PR #14, opened 2017-02-15
Move stories to be next to their components.

### LINKED ISSUE #1: Opinions on the various dunder directories.
I kinda like `__tests__` and `__mocks__` etc being next to the source files. I also think we could do the same with storybook stories, i.e. `__stories__` ?

alloy:
Closes #1.


orta:
👍

PR #15, opened 2017-02-15
Use styled-components for our styling needs.

alloy:


orta:
That is not a lot of code 👍

PR #17, opened 2017-02-21
[fuse-box] A simple port of storybook’s webpack support.

alloy:
This replaces the [webpack bundler](http://webpack.github.io) that [Storybook normally uses](https://github.com/storybooks/react-storybook/tree/master/src/server) by [FuseBox](http://fuse-box.org).

For our “TypeScript -> Babel with Relay transforms” pipeline It is _insanely_ much faster. Changes are reflected in the running UI near instantaneously.

I simply vendored some of the files that the regular webpack setup would generate (HTML), omitted things that seemed irrelevant to us (CSS modules, `<head>` customization). There may be issues left in this first pass, but it works well enough in my limited testing and is _so_ much faster to develop with, that I think we should start using it sooner rather than later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "license": "MIT",
   "scripts": {
     "test": "jest",
-    "storybook": "start-storybook -p 9001 -c .storybook",
+    "storybook": "node fuse.js",
     "deploy-storybook": "storybook-to-ghpages",
     "relay2ts": "relay2ts src/components/*.tsx --update",
     "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && yarn install && npm run dump-schema -- ../../data",
@@ -26,6 +26,7 @@
     "babel-preset-react": "^6.22.0",
     "babel-preset-stage-3": "^6.22.0",
     "babel-relay-plugin": "https://github.com/alloy/relay/releases/download/v0.9.3/babel-relay-plugin-0.9.3.tgz",
+    "fuse-box": "^1.3.118",
     "graphql-config-parser": "^1.2.1",
     "jest": "^18.1.0",
     "react-dom": "^15.4.2",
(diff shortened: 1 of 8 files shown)


src/__stories__/config.js
```diff
@@ -0,0 +1,8 @@
+const { configure } = require("@kadira/storybook")
+
+function loadStories() {
+  require("../components/__stories__/artwork")
+  require("../components/__stories__/artwork_grid")
+}
+
+configure(loadStories, module);
```
alloy: This file has to exist _inside_ the FuseBox `homeDir`, trying to recurse from outside of it with a relative path does not seem to make FuseBox happy.

package.json
```diff
@@ -35,6 +36,7 @@
     "typescript": "^2.1.5"
   },
   "dependencies": {
+    "asap": "https://github.com/alloy/asap.git",
```
alloy: This hacked version of `asap` (to return the correct browser specific build of the lib) is needed until [this FuseBox limitation](https://github.com/fuse-box/fuse-box/issues/258) has been removed.

package.json
```diff
@@ -8,7 +8,7 @@
   "license": "MIT",
   "scripts": {
     "test": "jest",
-    "storybook": "start-storybook -p 9001 -c .storybook",
+    "storybook": "node fuse.js",
```
orta: This should make it really easy to add a debugger to the storybook too.

src/__stories__/config.js
orta: 👍 

Could it be worth dynamically `require`ing anything that has the `__stories__` as it's a bit less intuitive on where the default is now?

orta:
Running this from a fresh clone gives this:

screen shot 2017-02-22 at 08 53 20

---

Looking into it after I make a tea

src/__stories__/config.js
alloy: Ah yes, I had tried globbing a while ago, but the issue turned out to be that this file actually only gets eval’ed in the browser, not in the Node.js process, and thus globbing doesn’t actually work/make sense.

However… there might be a way to do it by using the FuseBox in-browser API 🤔

alloy: Yeah, so, that works, but it’s now tied to FuseBox, which is not ideal for when we deploy the static version.

orta:
I was toying with getting the fuse process to run in a debugging node instance, but as @alloy pointed out in slack. It's not really needed. This was fixed by `rm -rf .fusebox` on the local server

PR #18, opened 2017-02-21
add artsy-icons and render minimal header

xtina-starr:
An imperfect react thing/header. This is the first stab at trying to get a tiny react component to render in the storybook. This will eventually be a reusable header for the login modal. It's missing some styling and there is probably a better way to set this up which is why I'm soliciting feedback 😄 

screen shot 2017-02-21 at 6 50 45 pm

Thanks for the help @l2succes!


src/components/icons.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
+import icons, { IconName } from '../assets/icons'
+import styled from 'styled-components'
+import colors from '../assets/colors'
+
+export type FontName = string
+
+interface IconProps extends React.Props<Icon>, React.HTMLAttributes<Icon> {
```
orta: interesting, I didn't know interfaces supported multiple inheritance

orta:
Sorry, @xtina-starr you probably have a bit of merge conflicts to look at, I'd recommend reading this PR to understand how things have moved - #17

src/components/icons.tsx
alloy: Good stuff this

.storybook/head.html
```diff
@@ -0,0 +1,12 @@
...
+
+<style>
+  @font-face {
+	font-family: 'artsy-icons';
+	src: url("/fonts/artsy-icons.eot?uo9ko");
+	src: url("/fonts/artsy-icons.eot?#iefixuo9ko") format('embedded-opentype'), url("/fonts/artsy-icons.woff2?uo9ko") format('woff2'), url("/fonts/artsy-icons.ttf?uo9ko") format('truetype'), url("/fonts/artsy-icons.woff?uo9ko") format('woff'), url("/fonts/artsy-icons.svg?uo9ko#artsy-icons") format('svg');
+	font-weight: normal;
+	font-style: normal;
+}
+</style>
```
alloy: Instead of this separate head file, you can just insert it into the `iframe.html` file inside the `.storybook` directory.

Although I do wonder what the better styled-components way is. Q stick search turned up this https://github.com/styled-components/styled-components/blob/3e5d63fd71ad06c37cfb5fe6bc8e77553ddc0fb2/docs/css-we-support.md, I’m not entirely sure in what component you’re supposed to do that, maybe any component that refers to the font and they deduce it before actually inserting into the `<head>`?

package.json
```diff
@@ -8,7 +8,7 @@
   "license": "MIT",
   "scripts": {
     "test": "jest",
-    "storybook": "start-storybook -p 9001 -c .storybook",
+    "storybook": "start-storybook -p 9001 -c .storybook -s ./.storybook/static",
```
alloy: Static assets should just go inside the root `assets` dir, which I had added to the `.gitignore` file so you’ll have to remove that.

src/assets/colors.tsx
```diff
@@ -0,0 +1,14 @@
...
+  graySemibold: "#666666",
+  grayBold: "#333333",
+  purpleLight: "#e2d2ff",
+  purpleRegular: "#6e1fff",
+  redRegular: "#f7625a",
+  yellowRegular: "#fdefd1",
+  yellowMedium: "#fce1a8",
+  yellowBold: "#f1af1b",
+  greenRegular: "#0eda83"
+}
```
alloy: I like this, but on the other hand it’s also nice to be able to just refer to the data in elan rather than having to keep it in sync. Do you have any thoughts on that?

src/components/icons.tsx
```diff
@@ -0,0 +1,27 @@
...
+        {icons[this.props.name]}
+      </span>
+    )
+  }
+}
+
+export default styled(Icon)`
+  font-family: ${props => props.font || 'artsy-icons'};
+  color: ${props => props.color || 'purple'}
+`;
```
alloy: Ah yeah, dynamic styling 🕺

src/components/login_header.tsx
```diff
@@ -0,0 +1,27 @@
...
+  line-height: 1.3;
+`
+
+const HeaderLogo = styled.div`
+  margin: 0 0 20px 0;
+  font-size: 55px;
+  line-height: 55px;
+`
+
+export default class LoginHeader extends React.Component<any, null> {
```
alloy: Seeing as this component doesn’t take any props, I wonder if it should be `React.Component<null, null>` or if that even works.

Also, a completely stateless component (a “pure” component) can also be represented as a simple function instead of a full class:

```js
export default function LoginHeader() {
  return (
    <div>
        <Header>
        <Icon name={"logotype"} color={"black"} />
      </Header>
    </div>
  )
}
```

alloy: That’s just an FYI btw, not a blocker.

alloy:
Nice work! :tada: :clap:

src/components/__stories__/login_header.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+import { storiesOf, action } from "@kadira/storybook"
+
+import LoginHeader from "../login_header"
+
+storiesOf("LoginHeader", LoginHeader)
+  .add("Simple login header", () => (
+    <LoginHeader/>
+  ))
```
alloy: Do you think there will be multiple header stories? If not, maybe we can make this a ‘static/simple’ components group?

.storybook/head.html
l2succes: Yeah would make sense to have a module for global css instead using this syntax

```javascript
import { injectGlobal } from 'styled-components'

injectGlobal`
    @font-face {
      font-family: 'Operator Mono';
      src: url('../fonts/Operator-Mono.ttf');
    }
`
```

and import it from any module that needs it. I'll set it up.

src/assets/colors.tsx
l2succes: We tried using elan as a data source instead but it's not in a format that very useful for reaction-force. When we have setup as an exported object, we get autocompletion and compiler help from typescript.

src/components/icons.tsx
l2succes: Yeah, it's very useful indeed to inherit from existing component props

src/components/login_header.tsx
l2succes: It's an initial pass for now, we haven't defined the props for that component yet.

.storybook/head.html
alloy: > and import it from any module that needs it

Ah yes, good thinking 👌

src/assets/colors.tsx
alloy: True true. In wonder how others are using this. If the elan data is only being consumed by JS apps, maybe we can just change the JSON file into an ES6 module like this?

/cc @craigspaeth

xtina-starr:
@l2succes I just fixed the merge conflict locally. Was working on adding the `injectGlobal` for the styled component. Do you just want pair on this and work on one machine?

fuse.js
```diff
@@ -22,4 +22,6 @@ const server = fuse.devServer(">__stories__/config.js +babel-runtime/* +componen
 
 // Serve Storybook bootstrap HTML
 const app = server.httpServer.app
+
+app.
```
alloy: Whoops

src/__stories__/config.js
```diff
@@ -3,6 +3,7 @@ const { configure } = require("@kadira/storybook")
 function loadStories() {
   require("../components/__stories__/artwork")
   require("../components/__stories__/artwork_grid")
+  require("../components/__stories__/login_header")
 }
 
-configure(loadStories, module);
+configure(loadStories, module);
```
alloy: Can take out that trailing semi-colon

xtina-starr:
It says there is a merge conflict but I'll fix that.

src/assets/colors.tsx
craigspaeth: They're in JSON because Stylus [supports importing that](https://github.com/artsy/elan/blob/master/components/lib/index.styl#L4), so it might take some re-jiggering to get ES6 modules to work with the rest of Elan.

FWIW I support some copy pasting and not trying to reuse Elan if it makes life easier. I think using React and Javascript as the lower common denominator to share UI among projects will be the sanest path forward and I'm happy to completely abbandon Stylus/Jade/Coffeescript/etc. to make marching towards that easier.

src/assets/fonts.tsx
```diff
@@ -0,0 +1,11 @@
+import { injectGlobal } from 'styled-components'
+
+injectGlobal`
+    @font-face {
+        font-family: 'artsy-icons';
```
craigspaeth: Since we're paving the future here, I'll throw my 👎  to icon fonts and 👍  to rendering SVGs directly for a couple reasons:

1. Icon fonts are a PITA to update b/c you often have to go through some [complicated build process](https://github.com/artsy/artsy-iconfont#managing-the-font) to turn the source files into fonts
2. You don't have to have the overhead of making mapping font character to css class name—just import SVG file basically
3. They involve un-intuitive properties like using `font-size` to change the size of them
4. Although a more rare use case, you don't have fine-grain control of the SVG like being able to change things like `stroke` or animate color `fill`

The nicest way I've found to do this is to use `fs.readFileSync` and shim `fs` in [the browser](https://github.com/substack/brfs), but there's [other](https://github.com/necinc/msvgc) [tools](http://fuse-box.org/#loader-api) that do similar things—the important part is that something takes SVG file in and puts JS out. You can even [do this manually](https://github.com/artsy/team-navigator/blob/master/app/views/sidebar/artsy-logo.svg.js) to avoid build/tools plugins.

xtina-starr: I switched the logotype font-face icon with the svg per @craigs suggestion just to see what that might look like. @l2succes @alloy are there any strong opinions about which implementation we should go with? I can switch this back to go ahead and get this merged and we can decide later...

alloy: @craigspaeth has good arguments against an icon font, the only thing I have to add is that iOS does not natively support SVG, but I’m sure there’s a way we can make that work.

alloy: @craigspaeth Can you move that to a new ticket?

l2succes: @alloy  I opened a ticket for the issue @craigspaeth mentioned here https://github.com/artsy/reaction-force/issues/24 
So we can address this in a separate PR.

xtina-starr: @alloy @l2succes any additional work needed for this particular PR?

l2succes: @xtina-starr the initial feedback has been addressed so let's merge it. We can deal with the icon-font in a subsequent PR.

xtina-starr: 👌

alloy:
LGTM!

I wonder if/how well the static storybook deployer is going to deal with these assets.

PR #21, opened 2017-02-23
@mzikherman - Artwork metadata

broskoski:
Ports some of the `artwork_brick` logic over here. 

screen shot 2017-02-28 at 6 44 09 pm


src/components/artwork_contact.tsx
```diff
@@ -0,0 +1,93 @@
...
+          opening_bid {
+            display
+          }
+          counts {
+            bidder_positions
+          }
+        }
+        partner(shallow: true) {
+          type
+          href
```
mzikherman: Looks like this field isn't actually being used in this component?

src/components/artwork_details.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  artistLine() {
+    const cultural_maker = this.props.artwork.cultural_maker
+    const artists = this.props.artwork.artists
+
+    if (cultural_maker) {
+      return (<h4>{cultural_maker}</h4>)
+    } else if (artists && artists.length) {
+      let artistsEl = []
+      for (var i=0; i < artists.length; i++) {
```
mzikherman: Can/should we be using `map`, `join`, etc?

src/components/artwork_details.tsx
```diff
@@ -0,0 +1,120 @@
...
+      } 
+      return (<h4>{artistsEl}</h4>)
+    }
+  }
+
+  titleLine() {
+    return (
+      <a href={this.props.artwork.href}>
+        <em>{this.props.artwork.title}</em>
+        { this.props.artwork.date ? `, ${this.props.artwork.date}` : null }
```
mzikherman: This might benefit from the && trick.

``` typescript
this.props.artwork.date && `, ${this.props.artwork.date}`
```

src/components/artwork_details.tsx
```diff
@@ -0,0 +1,120 @@
...
+      return (
+        <div>{artwork.sale_message}</div>
+      )
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        {this.artistLine()}
```
mzikherman: Should each of these actually be it's own React component? I don't know how deep boxes on boxes goes (or should go)...

src/components/artwork_details.tsx
```diff
@@ -0,0 +1,120 @@
...
+          is_live_open
+          is_open
+          is_closed
+        }
+      }
+    `
+  },
+})
+
+interface RelayProps {
```
mzikherman: What's this? (and should it be present in the `ArtworkContact` relay-ed component?)

mzikherman:
Looking tight!

src/components/artwork_details.tsx
orta: This should be fine, these functions end up mapping pretty much directly to a HTML output (e.g. elements from react, with. no lifecycle methods )

src/components/artwork_details.tsx
orta: It's our generated Relay interfaces, https://github.com/alloy/relay2ts

This gets referenced right at the top: `export class ArtworkContact extends React.Component<RelayProps, null> {` which says that the interface for `this.props` would follow this shape.

src/components/artwork_contact.tsx
broskoski: Good catch!

src/components/artwork_details.tsx
broskoski: In this case we actually want to know if the element is the last in the array so a for loop makes sense.

src/components/artwork_details.tsx
broskoski: Nice, thanks!

broskoski:
C'mon circle.

In any case, @mzikherman this is ready to go.

mzikherman:
screen shot 2017-03-01 at 11 31 40 am

:*(

broskoski:
I'm so sorry @mzikherman

src/components/__stories__/buttons.tsx
```diff
@@ -6,7 +6,7 @@ import Button, { ButtonState } from '../buttons/default'
 import GhostButton from '../buttons/ghost'
 import InvertedButton from '../buttons/inverted'
 import FacebookButton from '../buttons/facebook'
-import TwitterButton from '../buttons/twitter'
+// import TwitterButton from '../buttons/twitter'
```
mzikherman: How come commented out?

broskoski: Does not exist.

src/components/artwork/contact.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <TextLink href={artwork.href}>Enter Live Auction</TextLink>
+      )
+    } else if (artwork.sale.is_open) {
+      const sa = artwork.sale_artwork
+      const bids = sa.counts.bidder_positions
+      if (bids > 0) {
+        const s = bids > 1 ? 's' : ''
+        return (
+          <span>
+            {sa.highest_bid.display}(bid{s})
```
mzikherman: confused about this rendering, does it produce: `$15,000(bid)` or `$15,000(bids)`? Sorry if I'm mis-reading

broskoski: Ah good catch needs a space. The `s` variable handles whether or not its bid or bids.

src/components/buttons/twitter.tsx
```diff
@@ -1,21 +0,0 @@
-import * as React from 'react'
```
mzikherman: Intentional?

broskoski: Sigh, no.

src/components/artwork/contact.tsx
broskoski: Actually, you're right this is weird, will fix.

src/components/text_link.tsx
```diff
@@ -0,0 +1,33 @@
...
+    return (
+      <a href={this.props.href} className={this.props.className}>
+        {this.props.children}
+      </a>
+    )
+  }
+}
+
+TextLink.defaultProps = {
+    underline: false
```
mzikherman: I'm guessing there's a couple uses of this you do want the underline for? Like the 'Enter Live Auction', or 'Contact Gallery'? Not sure, but I guess the 'styling' can be done later.

src/components/text_link.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export default styled(TextLink)`
+  font-size: 15px;
+  line-height: 12px;
+  color: ${colors.grayBold};
+  text-decoration: ${props => props.underline ? "underline" : "none"}
```
broskoski: You can pass in `underline=true` and the underline will be styled correctly.

broskoski:
👍  ?

PR #23, opened 2017-02-23
Adds support for Chrome-based debugging.

orta:
[screen shot 2017-02-23 at 18 09 03](https://ortastuff.s3.amazonaws.com/vscode/debugger/chromedebugger.mov)
(click for the video)

This PR adds support for using a Chrome tab with inline debugging inside VS Code. This works by the [VS Code Chrome extension](https://github.com/Microsoft/vscode-chrome-debug) and some serious work inside the [VS Code](https://code.visualstudio.com/blogs/2016/02/23/introducing-chrome-debugger-for-vs-code) on supporting the chrome debugger protocol.

In terms of workflow, you can continue to use `yarn run storybook` to launch the server, then any time you want in VS Code you can hit the debugger, and it'll launch a chrome tab that hooks back into your editor. Then your inline breakpoints, REPL, and stack-traces etc will work. 

I've not dug too deep into how this will work with after some extensive hot-module reloading, but I'd expect it to be okay but not great. Sourcemaps are always accurate on the filesystem, so maybe VS Code can handle that at runtime.

`<edit>`Not blocked by https://github.com/fuse-box/fuse-box/issues/280 since my PR is shipped  `<edit/>`


broskoski:
Beautiful, can't wait to try it out.

fuse.js
```diff
@@ -11,7 +11,11 @@ storybookManager.bundle("> index.js")
 const fuse = fsbx.FuseBox.init({
   homeDir: "src",
   tsConfig: "./tsconfig.json",
-  outFile: "./assets/reaction-force.js",
+  outFile: "./reaction-force.js",
+    sourceMap: {
+      bundleReference: "./reaction-force.js.map",
+      outFile: "./reaction-force.js.map",
+    },
```
alloy: Why does this need to be outside the assets dir?

orta: ah yeah, was meant to make a note, this was due to all of the relative paths inside the source mapping setup, I couldn't find a way to say that they are relative to to `../`of the assets dir instead of `.`. 

Can give it another look at some point too. Might be something I've missed now there's a working version to compare against.

alloy: Yeah should probably have to change that at some point. I’m not even sure `assets` is the right place, but probably not in the root.

liuliangsir:
vscode 1.10.2 release support some features, such as Column breakpoints

image

image

alloy:
@liuliangsir That's pretty dope!

stephencweiss:
@orta any chance you could post the `launch.json` config you have to actually get the breakpoints to stop? Mine is logging appropriately in the integrated terminal, but it's not stopping at breakpoints:
```json
    {
      // Requires the extension Debugger for Chrome: https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome
      "type": "chrome",
      "request": "launch",
      "name": "Storybook Debug",
      "url": "http://localhost:9001",
      "sourceMaps": true,
      "webRoot": "${workspaceFolder}",
      "sourceMapPathOverrides": {
        "webpack:///*": "${webRoot}/*",
        "webpack:///./*": "${webRoot}/*",
        "webpack:///src/*": "${webRoot}/*",
        "webpack:///./~/*": "${webRoot}/node_modules/*"
      }
    },
```

(Yes, I realize this is ~2.5 years late, so any help is appreciated 👍 )

PR #25, opened 2017-02-24
Button component

l2succes:
This PR is meant to be a discussion on the component styling pattern itself (that's why I added the whole squad)

buttons-gif


src/assets/fonts.tsx
```diff
@@ -8,4 +8,26 @@ injectGlobal`
...
\ No newline at end of file
+`
+
+const primaryFontFamily = `'ITC Avant Garde Gothic W04', 'AvantGardeGothicITCW01D 731075', 'AvantGardeGothicITCW01Dm', 'Helvetica', 'sans-serif'`,
+
+export const primary = {
+    fontFamily: primaryFontFamily,
+    style: `
+        font-family: ${primaryFontFamily};
+        -webkit-font-smoothing: antialiased;
```
mzikherman: Do we need to do anything to include vendor-prefixed stuff, or do we get that for free with styled components? (or is that not even a thing anymore)?

src/components/button.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react'
+import * as Bootstrap from 'react-bootstrap'
+import styled from 'styled-components'
+import colors from '../assets/colors'
+
+debugger
+const buttonsStyles = require('../../externals/elan/components/buttons/index.styl')
+console.log(buttonsStyles)
```
xtina-starr: Left behind `console.log`

src/assets/fonts.tsx
l2succes: Yeah good point, styled-components is supposed to add the vendor prefixes for us so let me try to remove it. Hopefully it should work

src/assets/fonts.tsx
```diff
@@ -8,4 +8,26 @@ injectGlobal`
         font-weight: normal;
         font-style: normal;
     }
-`
\ No newline at end of file
+`
+
+const primaryFontFamily = `'ITC Avant Garde Gothic W04', 'AvantGardeGothicITCW01D 731075', 'AvantGardeGothicITCW01Dm', 'Helvetica', 'sans-serif'`,
+
+export const primary = {
```
xtina-starr: nice!!

src/assets/fonts.tsx
l2succes: Yup, we need it. the fonts look ugly without it and it's not being autoprefixed

.storybook/iframe.html
```diff
@@ -9,6 +9,7 @@
       }
     </script>
     <title>React Storybook</title>
+    <link type="text/css" rel="stylesheet" href="//fast.fonts.net/cssapi/f7f47a40-b25b-44ee-9f9c-cfdfc8bb2741.css">
```
mzikherman: This may not work in the deployed storybook b/c of the allowed domains for the fonts.

src/components/__stories__/buttons.tsx
```diff
@@ -0,0 +1,52 @@
...
+                    <Button>Button</Button>
+                    <Button disabled>Button</Button>
+                </div>
+                <div>
+                    <Button state={ButtonState.Success}>Button</Button>
+                    <Button state={ButtonState.Success} disabled>Button</Button>
+                </div>
+                <div>
+                    <Button state={ButtonState.Failure}>Button</Button>
+                    <Button state={ButtonState.Failure} disabled>Button</Button>
```
yuki24: What do you think of specifying a state directly instead of importing and using `ButtonState`?

```es6
<Button failure>
```

or

```es6
<FailureButton>
```

l2succes: Well the state of the button is most likely gonna change based on some external state update (relay, API response, user action, etc...) so it makes more sense to make it a property

src/assets/fonts.tsx
alloy: Hmm, I think it should, though, so we must be doing something wrong.

> (The CSS rules are automatically vendor prefixed, so you don't have to think about it!)

From https://github.com/styled-components/styled-components/blob/master/README.md#L50

.storybook/iframe.html
l2succes: Sure, we can update the fonts.net whitelist if we deploy to a non-listed host

src/assets/fonts.tsx
l2succes: screen shot 2017-02-24 at 5 27 41 pm

so it's working for the `transition` css property but not `font-smoothing` which is leading me to think, `font-smoothing` is not included in the list of properties that need to be prefixed. I'll double-check how `styled-components` does it

l2succes: Here's the module in `styled-components` that does it https://github.com/styled-components/styled-components/blob/c268b4f62192f22e99283d8f5a41803f489c8468/src/utils/autoprefix.js

Which is using [glamor/autoprefix.js](https://github.com/threepointone/glamor/blob/master/src/autoprefix.js)
I'll debug and see what's going on.

src/components/buttons/default.tsx
```diff
@@ -0,0 +1,85 @@
...
+    state?: ButtonState
+    block?: boolean
+}
+
+export class ButtonState extends String {
+    static Default: ButtonState = "default"
+    static Loading: ButtonState = "loading"
+    static Success: ButtonState = "success"
+    static Failure: ButtonState = "failure"
+}
```
alloy: Hah, this is some smart declaration 👌

However, I do wonder if having simply an enum of the various strings isn't better to grok?

alloy: Thinking about it more, with this declaration it would be possible for a user to cast any string as `ButtonState` and be a valid state type, so I do think it should be an enum.

src/components/buttons/default.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+class Button extends React.Component<ButtonProps, any> {
+    public static defaultProps: ButtonProps
```
alloy: I don't understand what this ends up being? Can you use a TS interface as an object and what keys/values does it contain?

src/components/buttons/default.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    render(): JSX.Element {
+        const newProps: any = {...this.props}
```
alloy: Is this possible with TS?

`const { state, ...newProps } = this.props`

I find that to be a little easier to read, not sure if very personal.

src/components/__stories__/buttons.tsx
alloy: Just out of interest, what does `<Button failure />` do? Does it set a prop by that name (failure) to `true`

alloy:
Besides the enum thing I like it 👍

src/components/buttons/default.tsx
orta: Yeah, enum is a better fit for this modelling, it can only be one of a specific set of options

package.json
```diff
@@ -36,8 +36,10 @@
     "typescript": "^2.1.5"
   },
   "dependencies": {
+    "@types/react-bootstrap": "^0.0.45",
     "asap": "https://github.com/alloy/asap.git",
     "react": "^15.4.2",
+    "react-bootstrap": "^0.30.7",
```
orta: This is a massive framework ( https://react-bootstrap.github.io/components.html ) was there discussion about whether this is worth it anywhere I can read?

yuki24: I'd prefer not to pull in any bootstrap-related libraries.

l2succes: Yes, agreed. didn't mean to include this in `package.json`, I'll remove it.

alloy: Ah yes, good catch, I forgot to comment on this 👌

src/components/__stories__/buttons.tsx
l2succes: @alloy yes basically it's equivalent to `<Button failure={true} />`

src/components/buttons/default.tsx
l2succes: Yes, that syntax is supported and I agree it's a bit more readable but I needed to "cast" props to type `any` to silence the compiler when we pass props down to `<button>` or `<a>`

src/components/__stories__/buttons.tsx
yuki24: The only thing I'd like to avoid is the need to import `{ ButtonState }`. Having a type check to specify a button state seems a bit overkill to me. Other than that, I'm :+1:

xtina-starr:
Is there any more feedback regarding this PR?? If not I'd like to more it to incorporate this fonts that was added to this for the Title component I'm working on. 

If this has the ok. I'm going to go ahead and merge...
cc: @alloy

src/assets/fonts.tsx
```diff
@@ -14,7 +14,7 @@ const primaryFontFamily = `'ITC Avant Garde Gothic W04', 'AvantGardeGothicITCW01
 
 export const primary = {
     fontFamily: primaryFontFamily,
-    style: `
+    style: css`
```
alloy: Is this `css` addition for the prefixing?

alloy:
@xtina-starr LGTM, go for it 👍

src/assets/fonts.tsx
l2succes: Yes exactly

PR #28, opened 2017-02-27
Fusebox update and social buttons

l2succes:


src/components/buttons/facebook.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import Button, { ButtonProps } from './default'
+import Icon, { IconProps } from '../icon'
```
alloy: IconProps is unused

src/components/buttons/twitter.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import Button, { ButtonProps } from './default'
+import Icon, { IconProps } from '../icon'
```
alloy: Same

src/components/buttons/facebook.tsx
l2succes: Good catch

alloy:
Dope, very fast indeed

PR #29, opened 2017-02-27
Grid system component

yuki24:
### Screenshot:

screen shot 2017-02-27 at 6 18 41 pm

### Interface

```jsx
const boxStyle = {
  backgroundColor: '#f8f8f8'
  height: '100px',
  lineHeight: '100px',
  textAlign: 'center'
}

<div style={{margin: '20px'}}>
  <Row>
    <Col><div style={boxStyle}>1</div></Col>
    <Col><div style={boxStyle}>2</div></Col>
    <Col><div style={boxStyle}>3</div></Col>
    <Col><div style={boxStyle}>4</div></Col>
    <Col><div style={boxStyle}>5</div></Col>
    <Col><div style={boxStyle}>6</div></Col>
  </Row>

  <Row>
    <Col><div style={boxStyle}>1</div></Col>
    <Col><div style={boxStyle}>2</div></Col>
    <Col><div style={boxStyle}>3</div></Col>
    <Col><div style={boxStyle}>4</div></Col>
  </Row>

  <Row>
    <Col><div style={boxStyle}>1</div></Col>
    <Col><div style={boxStyle}>2</div></Col>
    <Col><div style={boxStyle}>3</div></Col>
  </Row>

  <Row>
    <Col><div style={boxStyle}>1</div></Col>
    <Col><div style={boxStyle}>2</div></Col>
  </Row>

  <Row>
    <Col><div style={boxStyle}>1</div></Col>
  </Row>
</div>
```

### DIFF
--- a/src/__stories__/config.js
+++ b/src/__stories__/config.js
@@ -7,6 +7,7 @@ function loadStories() {
   require("../components/__stories__/buttons")
   require("../components/__stories__/title")
   require("../components/__stories__/input")
+  require("../components/__stories__/grid")
 }
 
 configure(loadStories, module)
--- a/src/components/__stories__/grid.tsx
+++ b/src/components/__stories__/grid.tsx
@@ -1,0 +1,49 @@
+import { action, storiesOf } from "@kadira/storybook"
+import * as React from "react"
+import * as Relay from "react-relay"
+
+import { Col, Row } from "../grid"
+
+const boxStyle = {
+  backgroundColor: "#f8f8f8"
+  height: "100px",
+  lineHeight: "100px",
+  textAlign: "center",
+}
+
+storiesOf("Grids", Row)
+    .add("Basic Usage", () => {
+        return (
+          <div style={ {margin: "20px"} }>
+            <Row>
+              <Col><div style={ boxStyle }>1</div></Col>
+              <Col><div style={ boxStyle }>2</div></Col>
+              <Col><div style={ boxStyle }>3</div></Col>
+              <Col><div style={ boxStyle }>4</div></Col>
+              <Col><div style={ boxStyle }>5</div></Col>
+              <Col><div style={ boxStyle }>6</div></Col>
+            </Row>
+
+            <Row>
+              <Col><div style={ boxStyle }>1</div></Col>
+              <Col><div style={ boxStyle }>2</div></Col>
+              <Col><div style={ boxStyle }>3</div></Col>
+              <Col><div style={ boxStyle }>4</div></Col>
+            </Row>
+
+            <Row>
+              <Col><div style={ boxStyle }>1</div></Col>
+              <Col><div style={ boxStyle }>2</div></Col>
+              <Col><div style={ boxStyle }>3</div></Col>
+            </Row>
+
+            <Row>
+              <Col><div style={ boxStyle }>1</div></Col>
+              <Col><div style={ boxStyle }>2</div></Col>
+            </Row>
+            <Row>
+              <Col><div style={ boxStyle }>1</div></Col>
+            </Row>
+          </div>
+        )
+    })
--- a/src/components/grid.tsx
+++ b/src/components/grid.tsx
@@ -1,0 +1,24 @@
+import * as React from "react"
+import styled from "styled-components"
+
+const gapSize = 20
+
+export const Row = styled.div`
+  display: flex;
+  flex: 0 1 auto;
+  flex-direction: row;
+  flex-wrap: wrap;
+  margin-right: -${ gapSize / 2 }px;
+  margin-left: -${ gapSize / 2 }px;
+  margin-bottom: ${ gapSize }px;
+`
+
+export const Col = styled.div`
+  flex: ${ props => props.size } 0 auto;
+  padding-right: ${ gapSize / 2 }px;
+  padding-left: ${ gapSize / 2 }px;
+`
+
+Col.defaultProps = {
+  size: 1,
+}


src/components/grids.tsx
```diff
@@ -0,0 +1,25 @@
...
+  margin-bottom: ${gapSize}px;
+`
+
+export const Col = styled.div`
+  box-sizing: border-box;
+  flex: 0 0 auto;
+  padding-right: ${gapSize / 2}px;
+  padding-left: ${gapSize / 2}px;
+  flex-basis: ${props => (100 / numberOfColumns) * (props.size)}%;
+  max-width: ${props => (100 / numberOfColumns) * (props.size)}%;
```
alloy: Rather than doing math here and having the user of the API having to define `props.size`, I wonder if we could just use `flex-grow` here? https://css-tricks.com/snippets/css/a-guide-to-flexbox/#article-header-id-10

The size prop can still exist, it would just default to 1 so the user doesn’t need to specify it if they want equal distribution.

alloy:
Looking good thus far! Nice and short ✨ 

Let’s rename the module to (singular) `grid`, though, as there is only 1.

src/components/grids.tsx
yuki24: It seems like the value of `flex-grow` is relative to other columns while the current grid system has the fixed scale of 12 in total. We could also just use `flex` as suggested here: https://facebook.github.io/react-native/docs/height-and-width.html, but I'm not sure what it'd look like on modern browsers.

I feel like we should experiment with `flex` or other flex-related properties before we get too far so we can make grid perfectly re-usable across multiple platforms.

What are the browser versions we are targeting? Also, is there any example iOS app that actually uses reaction-force that we can use for acceptance?

alloy: > It seems like the value of `flex-grow` is relative to other columns while the current grid system has the fixed scale of 12 in total.

I’m thinking that maybe less static numbers of columns is good for responsive pages (now that Force and microgravity have merged) as well as indeed share with React Native?

> I feel like we should experiment with `flex` or other flex-related properties before we get too far so we can make grid perfectly re-usable across multiple platforms.

I don’t think we need to engineer it too much right now, I think we should aim for the simplest solution that works well for what we need _right now_, so I’d suggest you take the time to make the best use of flexbox (and e.g. get rid of manual calculations).

> What are the browser versions we are targeting?

Good question. @craigspaeth can you say what historically has been the target?

> Also, is there any example iOS app that actually uses reaction-force that we can use for acceptance?

Nah, not yet. Don’t worry right now about these components working or not for iOS. We’ll cross that bridge when we come to it, just make sure to keep it simple so that it will be flexible to adjust to when we _do_ try to make it cross platform.

yuki24: It seems like `flex` is just the shorthand for `flex-grow`, `flex-shrink` and `flex-basis`, so I'm actually :+1: with just using `flex-grow` anywhere. I'll also remove the `max-width`. My only concern now is browser compatibility (and what I mean by browser is, obviously, IE11).

yuki24:
updated and rebased. Now `size: 1` is the default so you even don't have to specify anything if all the elements should have the same width.

There is one possible problem though. If you specify `size: 2` for example, the width of the column wouldn't equal to the sum of the two widths (see the screenshot and the code below). This is happening because the width of the longer one doesn't include padding values.

grid

```jsx
<Row>
  <Col><div style={boxStyle}>1</div></Col>
  <Col><div style={boxStyle}>2</div></Col>
  <Col><div style={boxStyle}>3</div></Col>
  <Col><div style={boxStyle}>4</div></Col>
</Row>

<Row>
  <Col size={2}><div style={boxStyle}>1</div></Col>
  <Col><div style={boxStyle}>2</div></Col>
  <Col><div style={boxStyle}>3</div></Col>
</Row>
```

l2succes:
@yuki24 I think It'd be worthwhile to implement a responsive grid using flexbox. For one we can then design components that render well in multiple screen sizes including mobile web and even react-native which also supports flexbox.

We can use this project as inspiration for desiging the grid.
http://roylee0704.github.io/react-flexbox-grid/


css-flexbox-small

yuki24:
@l2succes I actually picked up some of the code from flexbox-grid originally (without responsive part though).

yuki24:
Oooh, I totally misunderstood you @l2succes! Yes, we can totally pull in `react-flexbox-grid`.

What do you think @alloy?

alloy:
I generally prefer less dependencies.

We really only need a simple squares grid for now, we don’t even need them to be able to span multiple columns, so let’s keep it as simple as possible and is actually required.

grid

yuki24:
I see this grid on the top page of force.

screen shot 2017-03-02 at 2 06 20 pm

If even this was out of scope, why would we even need a style guide?

yuki24:
Also, how do our designers build out a page? Do they use a 12 grid system? If yes, we should align with them and implement the same type of grid system.

mzikherman:
Are we getting a little ahead of ourselves here? If the purpose of this work is a 'spike' in experimenting how we would build the loyalty page, which is relatively simple, I would not want to spend too much time on super generic stuff before we really understood _how_ and _if_ we are actually going to use them.

alloy:
@yuki24 Yeah the point here is to test out using React by building something as simple as the loyalty program registration, we do not need to build out more than we need for just that.

l2succes:
Yeah good point @mzikherman. we don't need a very flexible grid system since the layout for the loyalty page is pretty straightforward. @yuki24 Let's hold off on that and just get this basic grid implementation merged in.

yuki24:
@mzikherman just had a conversation with @l2succes on this - and I misunderstood the entire goal of this approach.

I think we can merge this in since the current implementation only uses `flex-grow`.

xtina-starr:
👍

alloy:
Sounds good 👍 I resolved the merge conflict for you, but the CI fail is legit.

xtina-starr:
What up with CI? It appears as if there is a lint error but it isn't showing what the error is.

yuki24:
rebased and fixed all the lint errors 👍

alloy:
Aight 👍

PR #30, opened 2017-02-27
Adds tslint and enforces rules

l2succes:
This PR adds tslint and tslint-react which is a linter for typescript and jsx in typescript files.

It also adds a pre-commit hook which will run `yarn run test` and `yarn run lint` before allowing a commit.

To fix linter issues, you can run `yarn lint -- --fix` which will fix a number of small errors (whitespace, etc...) but some errors must still be fixed manually.

It also adds a `yarn lint` as a command to run on Circle CI .

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,14 +1,19 @@
 {
-    // See https://go.microsoft.com/fwlink/?LinkId=733558
-    // for the documentation about the tasks.json format
-    "version": "0.1.0",
-    "command": "yarn",
-    "isShellCommand": true,
-    "showOutput": "never",
-    "suppressTaskName": true,
-    "isBackground": true,
-    "tasks": [{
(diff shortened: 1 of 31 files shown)


alloy:
Good stuff.

Did you take the majority of the rules from Emission's config? If not, maybe do a comparison.

We definitely need to add the react config to Emission too, that seems useful 👍

xtina-starr:
@l2succes @alloy yeah I was wondering where the rules came from as well. 

This definitely gives the project some structure and consistency. I like!

yuki24:
:shipit: 👍

xtina-starr:
Mergin'

l2succes:
@alloy Na I wrote it from scratch but comparing it to the one in emission, the configs are almost the same except for a few rules so we can combine them pretty easily.

PR #32, opened 2017-02-28
Title component

xtina-starr:
This adds the title component as well as updates the LoginHeader to use the new title component.  

**Title Component:**
screen shot 2017-02-28 at 10 18 25 am

**Updated LoginHeader Component:**
screen shot 2017-02-28 at 10 23 02 am

### DIFF
--- a/src/components/__stories__/login_header.tsx
+++ b/src/components/__stories__/login_header.tsx
@@ -2,8 +2,11 @@ import * as React from 'react'
 import { storiesOf, action } from "@kadira/storybook"
 
 import LoginHeader from "../login_header"
+import Title from "../title"
 
 storiesOf("LoginHeader", LoginHeader)
   .add("Simple login header", () => (
-    <LoginHeader title={"Welcome back, please log in to your account"}/>
+    <LoginHeader> 
+      <Title>Welcome back, please log in to your account.</Title>
+    </LoginHeader>
   ))
--- a/src/components/__stories__/login_header.tsx
+++ b/src/components/__stories__/login_header.tsx
@@ -5,5 +5,5 @@ import LoginHeader from "../login_header"
 
 storiesOf("LoginHeader", LoginHeader)
   .add("Simple login header", () => (
-    <LoginHeader/>
+    <LoginHeader title={"Welcome back, please log in to your account"}/>
   ))
--- a/src/components/icon.tsx
+++ b/src/components/icon.tsx
@@ -6,21 +6,17 @@ import '../assets/fonts'
 
 export type FontName = string
 
-export interface IconProps extends React.Props<Icon>, React.HTMLAttributes<Icon> {
+export interface IconProps extends React.HTMLProps<JSX.Element> {
   font?: FontName
   name: IconName
   color?: string
 }
 
-export class Icon extends React.Component<IconProps, null> {
-  render() {
-    return (
-      <span className={this.props.className}>
-        {icons[this.props.name]}
-      </span>
-    )
-  }
-}
+const Icon: React.SFC<IconProps> = props => (
+  <span className={props.className}>
+    {icons[props.name]}
+  </span>
+)
 
 export default styled(Icon)`
   font-family: ${props => props.font || 'artsy-icons'};
--- a/src/components/login_header.tsx
+++ b/src/components/login_header.tsx
@@ -8,7 +8,6 @@ const Header = styled.header`
   font-size: 26px;
   line-height: 1.3;
   text-align: center;
-  width: 400px;
 `
 
 export default class LoginHeader extends React.Component<any, null> {
--- a/src/components/login_header.tsx
+++ b/src/components/login_header.tsx
@@ -1,6 +1,6 @@
 import * as React from "react"
 import styled from "styled-components"
-import Icon from "./icons"
+import Icon from "./icon"
 import Title from "./title"
 
 const Header = styled.header`
--- a/src/components/login_header.tsx
+++ b/src/components/login_header.tsx
@@ -14,14 +14,12 @@ const Header = styled.header`
 export default class LoginHeader extends React.Component<any, null> {
   render() {
     return (
-      <div>
-        <Header>
-          <Icon name={"logotype"} color={"black"} />
-          <Title titleSize={'small'}>
-            {this.props.title}
-          </Title>
-        </Header>
-      </div>
+      <Header>
+        <Icon name="logotype" color="black" />
+        <Title titleSize="small">
+          {this.props.children}
+        </Title>
+      </Header>
     )
   }
 }
--- a/src/components/login_header.tsx
+++ b/src/components/login_header.tsx
@@ -1,12 +1,14 @@
 import * as React from "react"
 import styled from "styled-components"
 import Icon from "./icons"
+import Title from "./title"
 
 const Header = styled.header`
-  margin: 20px;
+  margin: 20px auto;
   font-size: 26px;
   line-height: 1.3;
   text-align: center;
+  width: 400px;
 `
 
 export default class LoginHeader extends React.Component<any, null> {
@@ -15,6 +17,9 @@ export default class LoginHeader extends React.Component<any, null> {
       <div>
         <Header>
           <Icon name={"logotype"} color={"black"} />
+          <Title titleSize={'small'}>
+            {this.props.title}
+          </Title>
         </Header>
       </div>
     )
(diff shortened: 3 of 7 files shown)


src/components/title.tsx
```diff
@@ -0,0 +1,35 @@
...
+const titleSizes = {
+  small: '25px',
+  medium: '30px',
+  large: '37px',
+  xlarge: '50px',
+  xxlarge: '72px'
+}
+
+const Title: React.SFC<TitleProps> = props => (
+  <div className={props.className}>
```
yuki24: Does React Native recognize heading tags like `h1`? Or should we just use `div` all the time?

xtina-starr:
@l2succes curious as to why you committed to this PR vs making change suggestions inline. I planned to return this today once I finished peer reviews. Is this PR perhaps blocking something?

l2succes:
@xtina-starr I actually did make change suggestions inline but while I was working on another component yesterday I figured out how to implement stateless components in typescript. 

Just thought I'd push the stateless component pattern since we talked about it when we were pairing it's not present anywhere else in the codebase.

Sorry didn't mean to disturb your flow, I can revert if you'd like.

xtina-starr:
@l2succes Oh no need to revert. Maybe in the future another branch could be created or a change request could be added.

yuki24:
Also, @l2succes's commit could be a separate PR, since it is mostly refactoring.

src/components/title.tsx
l2succes: It does have support for `h1-h6` tags or any html tags for that matter. We thought about using them for semantic reasons but decided to follow how titles are implemented in elan which uses `p` tags for all sizes. `div` would work just the same

alloy: Good stuff, was wondering about this too 👍 

@craigspaeth Do you have any opinion on what to use going forward?

src/components/login_header.tsx
```diff
@@ -1,22 +1,25 @@
...
+import Icon from "./icon"
+import Title from "./title"
 
 const Header = styled.header`
-  margin: 20px;
+  margin: 20px auto;
   font-size: 26px;
   line-height: 1.3;
   text-align: center;
+  width: 400px;
```
yuki24: I'm not sure if the Header component should have a fixed width.

src/components/title.tsx
alloy: @yuki24 RN does not use any HTML tag names at all, instead it uses capitalised tags like `View` and `ScrollView`.

src/components/icon.tsx
```diff
@@ -6,21 +6,17 @@ import '../assets/fonts'
...
-export class Icon extends React.Component<IconProps, null> {
-  render() {
-    return (
-      <span className={this.props.className}>
-        {icons[this.props.name]}
-      </span>
-    )
-  }
-}
+const Icon: React.SFC<IconProps> = props => (
```
alloy: `React.SFC` is the interface for stateless components?

src/components/__stories__/login_header.tsx
```diff
@@ -5,5 +5,5 @@ import LoginHeader from "../login_header"
 
 storiesOf("LoginHeader", LoginHeader)
   .add("Simple login header", () => (
-    <LoginHeader/>
+    <LoginHeader title={"Welcome back, please log in to your account"}/>
```
yuki24: I'm not sure what a `LoginHeader` with a title looks like, but could we possibly do something like below so we could easily add more elements to it?

```jsx
<LoginHeader>
  <Title>Welcome back, please log in to your account</Title>
</LoginHeader>
```

src/components/icon.tsx
l2succes: Yup, it's a shorthand for `React.StatelessComponent<IconProps>`

src/components/login_header.tsx
alloy: Good point. Unless it is guaranteed that a component is _always_ going to be a certain size, it should refrain from specifying any of its own metrics and leave that up to the user.

src/components/__stories__/login_header.tsx
broskoski: +

broskoski: Just to save you googling, props like this are accessed via `this.props.children` as opposed to `this.props.title`.

alloy:
> Also, @l2succes's commit could be a separate PR, since it is mostly refactoring.

Yup, that’s how we’ve done that in the past 👍

src/components/__stories__/login_header.tsx
l2succes: Yeah agreed. actually we should probably rename this component from `LoginHeader` to `ModalHeader` after we start working on the modal component

src/components/title.tsx
l2succes: @yuki24 oh yeah. what @alloy said. sorry, missed the "Native" part.

There's a project called [React Native Web](https://github.com/necolas/react-native-web) that brings react native components to the web but otherwise, React and React Native use different tags.

craigspaeth: I used to be into semantic markup, but nowadays I don't think it's worth spending any brain power on. That is to say, if one thinks there's a more appropriate tag for the situation then definitely go ahead and use it but don't spend more than half a second on it and feel free to just use a `div` and move on. As the web gets further away from its simple document viewer days the arguments about semantic markup become weaker, and any onsite SEO tag cleanup efforts I've been a part of at Artsy haven't resulted in measurable SEO payoff. (However, I'm not the expert and Christine and friends in the #seo channel might have better thoughts.)

That said, `h1`–`h6` _is_ a more approriate tag for headers than `p` or `div` so there's one reason to pick it, but if we wanted a cross platform "React Native Web" style component then certainly don't let semantic markup stop use from the wins of reusability.

xtina-starr:
thanks for all the great feedback on this PR. I think I've made the recommended changes. Let me know if I missed anything.

src/components/title.tsx
yuki24: I don't actually have strong opinions on `h1`-`h6` vs `div`. The `p` tag on the other hand may have some default CSS properties and would require us to add extra overrides. So as long as we use `h*` or `div`, I'm 100% 👍.

yuki24:
@craigspaeth @alloy @l2succes @xtina-starr I should've asked this earlier... but should we rebase or merge?

xtina-starr:
@yuki24 I think the preferred method is to rebase, but I just quickly resolved the merge conflict with the github ui since it was a tiny one and I'm a scary cat when it comes to having to rebase and use the `--force` command

yuki24:
> I just quickly resolved the merge conflict with the github ui

I didn't know we can do this on GitHub!

alloy:
> I'm a scary cat when it comes to having to rebase and use the `--force` command

We’ll have to work on that in the future!

But for now I’m ok, let’s get this in 👍

src/components/__stories__/login_header.tsx
xtina-starr: @broskoski I think I halfway grokked that because I'm accessing the text for the Title component in the header from `this.props.children`. Thanks for clearing it up. Yuki's suggestions makes sense as far as how it all comes together.

alloy: @xtina-starr Ahhh, hmm, then maybe I prematurely merged after all. Basically you should not need to use `Title` inside `LoginHeader` at all, just use the passed in children (`this.props.children`). i.e. in the example that @yuki24 originally posted, the children would be a list consisting of `<Title>Welcome back, please log in to your account</Title>`

alloy: …that is, [this line](https://github.com/artsy/reaction-force/pull/32/files#diff-bf1946d7c4ad3781c0372789ab3432d0R18) and [this line](https://github.com/artsy/reaction-force/pull/32/files#diff-bf1946d7c4ad3781c0372789ab3432d0R20) are not needed when you allow the user of `LoginHeader` to just pass in their own title(s).

alloy: If you look at the DOM you’ll see that there’s now 1 more wrapping div than you probably intended:

screen shot 2017-03-01 at 19 54 20

xtina-starr: Ahh, I see. I can address this in a separate PR.

xtina-starr: @alloy so in this cause it's on the user to construct the header component with a title component?

alloy: Yeah. I mean, if that makes any sense at all, not sure what else could go into the header, but if there's no reason to limit it to a Title then this would be the way you'd do that.

PR #33, opened 2017-03-01
Fix the storybook for the LoginHeader component

yuki24:
The storybook for the `LoginHeader` component is broken due to a path issue.

screen shot 2017-03-01 at 11 07 13 am


broskoski:
👍

PR #34, opened 2017-03-01
TextLink in ArtworkDetails should have a key

yuki24:
This removes the warning below:

screen shot 2017-03-01 at 3 58 27 pm

### DIFF
--- a/src/components/artwork/details.tsx
+++ b/src/components/artwork/details.tsx
@@ -19,7 +19,7 @@ export class ArtworkDetails extends React.Component<DetailsProps, null> {
     } else if (artists && artists.length) {
       let artistsEl = []
       for (let i = 0; i < artists.length; i++) {
-        artistsEl.push(<TextLink href={ artists[i].href }>{ artists[i].name }</TextLink>)
+        artistsEl.push(<TextLink href={ artists[i].href } key={ i }>{ artists[i].name }</TextLink>)
         if (i !== artists.length - 1) {
           artistsEl.push(<span>,</span>)
         }


broskoski:
Thanks @yuki24 !

src/components/artwork/details.tsx
```diff
@@ -19,7 +19,7 @@ export class ArtworkDetails extends React.Component<DetailsProps, null> {
     } else if (artists && artists.length) {
       let artistsEl = []
       for (let i = 0; i < artists.length; i++) {
-        artistsEl.push(<TextLink href={ artists[i].href }>{ artists[i].name }</TextLink>)
+        artistsEl.push(<TextLink href={ artists[i].href } key={ i }>{ artists[i].name }</TextLink>)
```
alloy: A simple index should not be used as the key, because we’d quickly run into duplicate keys on a full page and React keeps track (and re-uses) elements by this key.

In general I advice to use an object’s `__id` key, because it’s globally unique across all models _and_ Relay already automatically fetches this field even if you don’t explicitly include it in your query (thus adding it to the query does not add more data that’s to be fetched).

alloy: For more info see [the docs](https://facebook.github.io/react/docs/lists-and-keys.html#keys) and e.g. a blog post like [this](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318#.wvko033lh)

yuki24: 👍  I just sent a new pull request that addresses the key issue: https://github.com/artsy/reaction-force/pull/36/files

PR #35, opened 2017-03-01
Modal header

xtina-starr:
Renames LoginHeader to ModalHeader as mentioned in the [title component PR](https://github.com/artsy/reaction-force/pull/32#discussion_r103729954). This also addresses' @alloy's [comment](https://github.com/artsy/reaction-force/pull/32#discussion_r103759753) about not needing to create the title component inside of the header component. Finally, I didn't for see any case where the modal header would need to have a state so I refactor it into a stateless component.


src/__stories__/config.js
```diff
@@ -3,7 +3,7 @@ const { configure } = require("@kadira/storybook")
 function loadStories() {
   require("../components/__stories__/artwork")
   require("../components/__stories__/artwork_grid")
-  require("../components/__stories__/login_header")
+  require("../components/__stories__/modal_header")
```
yuki24: This is unrelated, but I wonder if there is any way to automatically load all the stories in `../components/__stories__/`.

alloy: I had looked into this, the issue is that this file gets evaluated in both Node and the browser and thus globbing a FS at that point isn’t that straightforward.

However, FuseBox does have an in-browser API that we can use to glob, so maybe we should conditionally use either the `node-glob` package or the FuseBox API.

yuki24: Understood. I'm ok with maintaining this manually for now.

broskoski:
LGTM2

yuki24:
LTGM3

broskoski:
Gonna close and reopen to see if the build success triggers.

PR #36, opened 2017-03-01
Refactor ArtworkDetails

yuki24:


src/components/artwork/details.tsx
```diff
@@ -9,22 +9,17 @@ export interface DetailsProps extends React.HTMLProps<ArtworkDetails> {
...
-      let artistsEl = []
-      for (let i = 0; i < artists.length; i++) {
-        artistsEl.push(<TextLink href={ artists[i].href } key={ i }>{ artists[i].name }</TextLink>)
-        if (i !== artists.length - 1) {
-          artistsEl.push(<span>,</span>)
-        }
-      }
-      return (<div><strong>{ artistsEl }</strong></div>)
+      let artistLine = artists
+        .map(artist => <TextLink href={ artist.href } key={ artist.name }>{ artist.name }</TextLink>)
```
yuki24: A fix for the key is coming

src/components/artwork/details.tsx
```diff
@@ -9,22 +9,17 @@ export interface DetailsProps extends React.HTMLProps<ArtworkDetails> {
...
+        .reduce((prev, curr) => [prev, ", ", curr])
```
broskoski: 👍

src/components/artwork/details.tsx
```diff
@@ -9,22 +9,17 @@ export interface DetailsProps extends React.HTMLProps<ArtworkDetails> {
...
-      let artistsEl = []
-      for (let i = 0; i < artists.length; i++) {
-        artistsEl.push(<TextLink href={ artists[i].href } key={ i }>{ artists[i].name }</TextLink>)
-        if (i !== artists.length - 1) {
-          artistsEl.push(<span>,</span>)
-        }
-      }
-      return (<div><strong>{ artistsEl }</strong></div>)
+      let artistLine = artists
+        .map(artist => <TextLink href={ artist.href } key={ artist.__id }>{ artist.name }</TextLink>)
```
broskoski: ✔️

yuki24:
Thanks!

src/components/artwork/details.tsx
alloy: Interesting that `artist.__id` is available at all, because it’s not being requested in the query and thus should be [masked](https://facebook.github.io/relay/docs/thinking-in-relay.html#data-masking) 🤔 

Can you add it to the [the fragment](https://github.com/artsy/reaction-force/blob/c5a4a0a5fbb9b5fb1c11e5428fcb56569c83cd60/src/components/artwork/details.tsx#L79-L82) anyways? Because this is fishy and could break in the future.

yuki24: Interesting. It might be just `undefined` but then React should complain about duplicate keys. Not sure why, but I'll add `__id` to the fragment.

PR #38, opened 2017-03-01
explicitly declare __id to makre sure it is always available

yuki24:
Following up on @alloy's comment: https://github.com/artsy/reaction-force/pull/36#discussion_r103796738


alloy:
Ace

PR #40, opened 2017-03-02
Input and TextArea components

l2succes:


src/components/helpers.tsx
```diff
@@ -0,0 +1,12 @@
+import { css } from "styled-components"
+
+export const block = (margin: number = 0) => {
```
alloy: This could do with some documentation. Use jsdoc so the IDE will nicely display it inline.

mzikherman:
Nice, this looks pretty good to me! Happy to merge whenever you're ready.

src/components/helpers.tsx
l2succes: Sounds good

src/components/input.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react"
+import styled from "styled-components"
+import colors from "../assets/colors"
+import * as fonts from "../assets/fonts"
+import { block } from "./helpers"
+
+interface InputProps extends React.HTMLProps<HTMLInputElement> {
```
xtina-starr: Just a thought but the interfaces for the `InputProps` & `TextAreaProps` look pretty similar, well identical at this point. I noticed there is another component for TextBox which I assume will also be similar. Is it worth extracting this out into a reusable Interface. I'm also curious what that implementation would look like. Maybe not something to address in this PR, but I did want to bring it up for dialogue purposes...

l2succes: Yeah I was thinking the same thing, didn't want to prematurely extract common styles but I think it makes sense in this case. I'll create a mixin for this.

l2succes: So I actually ended up addressing another issue from the one you described 😅  but it's related to shared properties between the Input and TextArea. In terms of the Interfaces, yes we can extract shared properties into a base Interface. But let's address it in another PR.

l2succes:
I addressed the feedback on the PR will go ahead and merge.

xtina-starr:
👌

src/components/helpers.tsx
```diff
@@ -0,0 +1,16 @@
...
+ */
+export const block = (margin: number = 0) => {
+    return props => {
+        if (props.block)
+            return css`
+                display: block;
+                width: calc(100% - ${margin}px);
+                margin: 10px auto;
+            `
+        }
```
damassi: @l2succes, @alloy - Question - can we settle and enforce spacing? This is the oldest argument out there, but I think it's important for consistency.  

From my own opinion, since so much of our existing codebase -- across Ruby, JS, CoffeeScript, Scala, Python (hokusai) and more -- are 2 spaces, can we keep it that way or at least open up for discussion?

alloy: I agree and discussed this in Slack, but seems it got merged maybe before/during I raised that.

I had just assumed that tslint was going to enforce 2 spaces as  default, I can’t believe [it’s still not in core](https://github.com/palantir/tslint/issues/581), however that ticket does point out [this external set of rules](https://github.com/buzinas/tslint-eslint-rules/blob/master/src/docs/rules/terIndentRule.md) that do support it, so we should just switch to that.

l2succes: @damassi Yes for sure, I'll add those rules and make sure we enforce two spaces. I honestly assumed vscode's default setting was two spaces but in any case, I'll have that rule included in a subsequent PR.

damassi:
Thanks y'all 👍 I'm admittedly a little OCD about 4 spaces because it's
like the wider the area I have to read, the less my peripheral "background"
mind can quietly take in from the surrounding code (which impacts focus).
It also seems like it wears the eyes out slightly more over time!

src/components/helpers.tsx
damassi: Thanks y'all 👍 I'm admittedly a little OCD about 4 spaces because it's
like the wider the area I have to read, the less my peripheral "background"
mind can quietly take in from the surrounding code (which impacts focus).
It also seems like it wears the eyes out slightly more over time!

PR #41, opened 2017-03-03
Make internal classes and interfaces private

yuki24:


src/components/icon.tsx
```diff
@@ -4,9 +4,9 @@ import colors from "../assets/colors"
 import "../assets/fonts"
 import icons, { IconName } from "../assets/icons"
 
-export type FontName = string
+type FontName = string
 
-export interface IconProps extends React.HTMLProps<JSX.Element> {
+interface IconProps extends React.HTMLProps<JSX.Element> {
```
l2succes: In some cases, we need to export a Props interface for a Component, to extend the component for instance. But we probably don't need to for IconProps.

yuki24: understood 👍

l2succes:
So actually as you can see in the screenshot below. We need to export `IconProps`

screen shot 2017-03-07 at 3 02 47 pm

PR #39, opened 2017-03-02
@yuki24 - port artwork column logic to ArtworkGrid

broskoski:
screenshot 2017-03-02 10 20 56

PR #45, opened 2017-03-04
Updated artwork columns

alloy:
This supersedes #39:

* It is rebased
* Moves data the grid needs into the Grid Relay container
* Fixes type and lint errors

### DIFF
--- a/src/components/artwork_grid.tsx
+++ b/src/components/artwork_grid.tsx
@@ -1,9 +1,9 @@
 import * as React from "react"
+import * as Relay from "react-relay"
 import styled from "styled-components"
 import Artwork from "./artwork/index"
 
-export interface GridProps extends React.HTMLProps<ArtworkGrid> {
-  artworks?: any,
+export interface GridProps extends RelayProps, React.HTMLProps<ArtworkGrid> {
   columnCount?: number,
   sectionMargin?: number,
   itemMargin?: number,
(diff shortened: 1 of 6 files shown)


alloy:
@yuki24 While in principle I agree we need tests for the logic in this component, I’m afraid that it’s hard to test without moving the logic into separate functions, at which time it starts to deviate a lot from the component in Emission. So I would prefer we do that once the dust has settled and we take a good look at where the overlap is between reaction-force components and Emission and extract/generalise at that time.

alloy:
@broskoski Be sure to read the commit message in 221f7b97abbe0e21000eb24a725f8a88ceca5060 about the use of `__id` instead of `id`.

src/components/artwork_grid.tsx
```diff
@@ -66,10 +66,8 @@ export class ArtworkGrid extends React.Component<GridProps, null> {
       for (let j = 0; j < sectionedArtworks[i].length; j++) {
         const artwork = sectionedArtworks[i][j]
         artworkComponents.push(
-          <Artwork
-            artwork={artwork}
-            key={"artwork-" + j + "-" + artwork.__id}
-          />)
+          <Artwork artwork={artwork as any} key={"artwork-" + j + "-" + artwork.__id} />,
```
broskoski: I'm curious about this from a stylistic standpoint. I find the former easier to read, is there a reason for making it one line?

broskoski:
LGTM, thanks for picking up the last bits!

src/components/artwork_grid.tsx
alloy: I don't remember what the final reason was for me to go with one line, but it was the result of a bunch of linter faffing.

I am completely fine with the multi-line approach. Feel free to reintroduce it whenever.

PR #46, opened 2017-03-06
add simple nav component

xtina-starr:
Adds the nav component.

screen shot 2017-03-05 at 10 47 45 pm


src/components/__stories__/simple_nav.tsx
```diff
@@ -0,0 +1,23 @@
...
+  position: relative;
+  top: -8px;
+  font-family: ${ fonts.primary.fontFamily };
+`
+
+storiesOf("Simple Nav", SimpleNav)
+  .add("Simple Nav Bar", () => (
+    <SimpleNav>
+      <LinkContainer>
+        <TextLink href="https://www.artsy.net">Back To Artsy</TextLink>
```
xtina-starr: I added the link here versus in the actual component due to feedback in the [Title Component](https://github.com/artsy/reaction-force/pull/32#discussion_r103759753). I thought it was a similar case.

I added a styled component in order to be able to style the link. I'm wondering if this is the correct approach...?

Also the href is a hard-coded value. I imagine when it's actually used it would be some sort of relative path or something.

xtina-starr:
I realize I need to rebase. I do that in AM.

src/components/simple_nav.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const NavIcon = styled.div`
+  border-right: 1px solid ${ colors.grayRegular };
+  display: inline-block;
+  font-size: 32px;
+  padding: 10px 5px 5px 5px;
+  margin-right: 10px;
+`
+
+const SimpleNav: React.SFC<any> = props => (
```
alloy: Is this `any` typing correct? Does it automatically cast it as something like `React.HTMLProps< HTMLDivElement>`?

src/components/__stories__/simple_nav.tsx
alloy: Aye 👍

alloy:
I just have the one question about the props typing, otherwise this looks good to me 👌

src/components/simple_nav.tsx
l2succes: No, it doesn't automatically cast to the correct props type. we should use `React.SFC<HTMLDivElement>` in this case @xtina-starr

src/components/icon.tsx
```diff
@@ -10,6 +10,7 @@ export interface IconProps extends React.HTMLProps<JSX.Element> {
   font?: FontName
```
l2succes: we should change `JSX.Element` to `HTMLSpanElement`

src/components/__stories__/simple_nav.tsx
l2succes: @xtina-starr I think the styling for the link should be encapsulated in a component. In terms of the `Link` component, I agree that it's a similar case to the Title Component. 

Also I would suggest naming the component `Nav`, we can decide on whether it's simple or complex based on the children components.

So the component could look like this.

```javascript
const Link = styled.a`
  position: relative;
  ...
`

<Nav>
  <Link href="https://www.artsy.net">Back To Artsy</Link>
</Nav>
```

src/components/simple_nav.tsx
yuki24: It seems like we should always specify a type so we can take advantage of TypeScript.

@l2succes What are your thoughts?

src/components/__stories__/simple_nav.tsx
xtina-starr: I considered abstracting these style rules into the link component. The `top: -8px` is what discouraged me. I thought it was pretty specific to this element. I thought it may make sense to keep the link component lightweight and leave the responsibility of determining how it's styled to the user. 

Maybe there is a happy medium and some things could be passed in as props. Thoughts @l2succes @alloy @yuki24 @broskoski?

src/components/simple_nav.tsx
xtina-starr: Ah, totally. Thanks for point this out.

src/components/__stories__/simple_nav.tsx
yuki24: I'm in agreement with @l2succes

yuki24: Speaking of `top: -8px`, we might be able to use [flex's `align-items`](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#article-header-id-7).

src/components/simple_nav.tsx
alloy: > It seems like we should always specify a type so we can take advantage of TypeScript.

Yup :+1:

src/components/__stories__/nav.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Nav from "../nav"
+import NavItem from "../nav_item"
+import TextLink from "../text_link"
+
+import * as fonts from "../../assets/fonts"
+
+storiesOf("Nav", Nav)
+  .add("Simple Nav Bar", () => (
+    <Nav>
+      <NavItem href="https://www.artsy.net" >Back To Artsy</NavItem>
```
l2succes: remove space before `>`

src/components/nav_item.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const StyledLink = styled.a`
+  position: relative;
+  align-self: center;
+  font-family: ${ fonts.primary.fontFamily };
+  font-size: 15px;
+  line-height: 12px;
+  text-decoration: none;
+  color: black;
```
l2succes: I think the text should be all caps also, right?

xtina-starr: Yep!

l2succes:
👍🏿  Looks good to me! Let's merge after resolving conflicts

src/components/nav.tsx
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+`
+
+const NavIcon = styled.div`
+  border-right: 1px solid ${ colors.grayRegular };
+  display: inline-block;
+  font-size: 32px;
+  padding: 10px 5px 5px 5px;
+  margin-right: 10px;
+`
```
yuki24: @l2succes @xtina-starr we just had a conversation offline, but since we are using `display: flex` already in `Nav`, I'm throwing out the diff I just made to kill magic numbers...

```diff
+const NavHeight = 54
+
 const Nav = styled.div`
   border-bottom: 1px solid ${ colors.grayRegular };
   display: flex;
+  box-sizing: border-box;
+  height: ${NavHeight}px;
 `

 const NavIcon = styled.div`
   border-right: 1px solid ${ colors.grayRegular };
-  display: inline-block;
+  display: flex;
   font-size: 32px;
-  padding: 10px 5px 5px 5px;
+  align-self: stretch;
+  align-items: center;
+  justify-content: center;
   margin-right: 10px;
+  box-sizing: border-box;
+  width: ${NavHeight}px;
 `
```

src/components/nav_item.tsx
```diff
@@ -0,0 +1,27 @@
...
+import styled from "styled-components"
+import * as fonts from "../assets/fonts"
+import TextLink from "./text_link"
+
+interface NavLinkProps extends React.HTMLProps<HTMLAnchorElement> {
+  href: string
+}
+
+const StyledLink = styled.a`
+  position: relative;
```
yuki24: I think we can remove this since `top: -8px` is gone now.

package.json
```diff
@@ -13,7 +13,7 @@
     "relay2ts": "relay2ts src/components/*.tsx --update",
     "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && yarn install && npm run dump-schema -- ../../data",
     "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
-    "lint": "tslint -c tslint.json --project tsconfig.json",
+    "lint": "tslint -c tslint.json --project tsconfig.json --fix",
```
alloy: I don’t think we should do this, it would also auto-fix fails on CI (but not commit them) and be pseudo-green.

Instead you can pass the flag when you need it: `yarn lint -- --fix`

alloy: (The `--` part is for yarn to know that whatever comes after it should be passed on to the commands.)

l2succes: Right that's a good point... let's remove the auto-fixing and just run it ad-hoc then.

yuki24: > I don’t think we should do this, it would also auto-fix fails on CI (but not commit them) and be pseudo-green.

What status does the `tslint` command return if auto-fix is on and there's a failure? Rubocop always returns 0 if auto-fix is on, but that might not be the case in tslint.

yuki24: confirmed it returns 0... 😱

alloy: Hah, that seems like an overly positive attitude of them :)

PR #47, opened 2017-03-06
Text component

yuki24:
screen shot 2017-03-06 at 4 10 00 pm

One thing that needs to be discussed is that [React native also provides `Text` component](https://facebook.github.io/react-native/docs/text.html) and having a component that has the same name may be confusing or could possibly cause a naming conflict.

We also might be able to refactor the `Title` component since `Title` is just a special version of `Text`.


yuki24:
For some reason the tslinter in CircleCI is smarter and finds more lint flaws while the one in my local machine is stupid and doesn't check anything.

l2succes:
@yuki24 you need to run `yarn type-check` which runs on CI. we should run it locally as well which by adding it to precommit hook script list in `package.json`

src/assets/colors.tsx
```diff
@@ -1,3 +1,19 @@
...
+  "#e2d2ff" |
+  "#6e1fff" |
+  "#f7625a" |
+  "#F1170D" |
+  "#fdefd1" |
+  "#fce1a8" |
+  "#f1af1b" |
+  "#0eda83" |
+  "#0aa462" |
+
```
yuki24: Not sure if we need this...

src/components/text.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react"
+import styled from "styled-components"
+import * as fonts from "../assets/fonts"
+
+type TextAligns = "start" | "center" | "end"
+type TextSizes = "small" | "large"
+
+interface TextProps extends React.HTMLProps<JSX.Element> {
```
alloy: Instead of `JSX.Element`, should this be`HTMLDivElement`?

src/components/text.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const textSizes = {
+  small: "17px",
+  large: "20px",
+}
+
+const Text: React.SFC<TextProps> = props => (
+  <div className={props.className}>
+    {props.children}
+  </div>
```
alloy: Out of curiosity, why did you choose a `div` instead of `p` ?

alloy:
Good point, but for now we don't need to worry about the details of a shared code base just yet.

src/components/text.tsx
yuki24: `p` has a default margin while `div` does not, but I don't really have strong opinions.

src/components/text.tsx
yuki24: 👍 

I guess it's time for me to learn all the element options...

src/components/text.tsx
alloy: I have a slight preference towards `p`, because I’d assume that assisted reading methods make use of these semantic differences. However, I have not looked into it, so maybe my assumption is incorrect?

src/components/text.tsx
alloy: If you’re using VSCode, you can command+click one of these element types and you’ll be taken to the definitions file where you will find all the others too.

src/components/text.tsx
l2succes: I agree with @alloy for having a preference for using `p` for encapsulating text, we can override the default margin as needed in the component itself

yuki24: sounds good, I'll change it to `p`.

Also, this is unrelated, but do we have a consistent set of margin/padding values?

src/components/text.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react"
+import styled from "styled-components"
+import * as fonts from "../assets/fonts"
+
+type TextAligns = "start" | "center" | "end"
+type TextSizes = "small" | "large"
+
+interface TextProps extends React.HTMLProps<HTMLDivElement> {
```
yuki24: I changed `div` to `p` but do we still need to use `HTMLDivElement`?

alloy: No you need to change that to the corresponding `p` type instead.

src/components/text.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react"
+import styled from "styled-components"
+import * as fonts from "../assets/fonts"
+
+type TextAligns = "start" | "center" | "end"
```
l2succes: the types should be you singular so `TextAlign` and `TextSize`

PR #48, opened 2017-03-06
Login page

l2succes:
screen shot 2017-03-07 at 3 45 16 pm


src/components/title.tsx
```diff
@@ -17,7 +17,7 @@ const titleSizes = {
 }
 
 const Title: React.SFC<TitleProps> = props => (
-  <div className={props.className}>
+  <div className={props.className} {...props}>
```
alloy: Might as well take out the `className` prop, it's going to be in the spread

mzikherman:
Screenshot!

mzikherman:
So, this can maybe go in this PR or a different one, but I guess this component should not assume anything about the paths that are used for the different authentication strategies, but allow those to be passed in via a parent?

Then, this can be used on Force (and most web node apps that use https://github.com/artsy/artsy-passport, with https://github.com/artsy/artsy-passport/blob/40dce9176f80c7829396495437236db7e70fa04e/README.md#and-maybe-a-signup-form showing a little example of how you just essentially point your facebook/twitter buttons and forms to the right `href`), where the parent passes that in?

It can then also be used in apps that roll their own authentication, and we wouldn't need to update this component if we ever change that.

Does that sound correct? If so, then I think the only thing we need with this component is allowing that info to be passed in, (the paths to wire up the buttons and form to). I am also just realizing that if we want to use this login form in a modal, we'd want that form submit to be client-side (the facebook/twitter stuff can't be b/c the oauth flow requires a redirect).

So, I think the form should also accept an `onSubmit` as well, which would basically need to make an ajax call with email/name/password to the sign up path, so that we can do something on success/error on that. Or...maybe we don't need that since we'll always just be passing in a redirect url to get redirected to after the server-side login...I'm actually not sure yet!

Either way, I think just letting it accept the paths w/ redirects for now, and waiting to see if we need any client-side form handling is probably sufficient. We do technically do some client-side stuff in the form like using [Mailcheck](https://www.google.com/search?q=mailcheck&oq=mailcheck&aqs=chrome..69i57j69i60l3.708j0j1&sourceid=chrome&ie=UTF-8) to help with some email validation/prompts, and in the context of the modal, it's nice to handle that client-side b/c any errors can just be meaningfully printed in the modal w/o any navigation, giving the user a chance to correct.

l2succes:
@matt yeah that makes sense, I'll add an interface for passing the form submit url and any other data needed for the form to work properly

mzikherman:
I realized I put a lot of stuff in my last comment.

Maybe some of that url stuff is sufficient for now, and we can pair on getting a login modal working? We can point it to staging, though I think we'll have CSRF issues with that.

I think trying to get the real login modal working will help us better figure out what API we want (from both login form and generic modal).

l2succes:
@mzikherman Ok I added a prop that includes a url and csrfToken to pass into the login form.

That's what the form looks like, still have to do some minor stylign and add the "forgot password" link but we can do afterwards.

screen shot 2017-03-07 at 4 24 55 pm

src/containers/login.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <Icon name="logotype" color="black" fontSize="30px" />
+        <Text textSize="large" align="center">
+          Welcome back, please log in <br /> to your account.
+        </Text>
+
+        <form action={form.url} method="POST" onSubmit={this.props.onSubmit}>
+          <StyledInput name="email" placeholder="Email" block />
+          <StyledInput name="password" placeholder="Password" type="password" block />
+
+          {form.csrfToken && <input type="hidden" name="_csrf" value={form.csrfToken} />}
```
mzikherman: cool!

src/containers/login.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+          <Button block>Log In</Button>
+          <div style={{textAlign: "center"}}>or</div>
+          <FacebookButton block />
```
mzikherman: These buttons will need an `href` too. (later)

src/containers/login.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <TwitterButton block />
+        </form>
+
+        <br />
+
+        <Text align="center">
+          <span>Don't have an account? </span>
+          <TextLink href={form.forgotPasswordUrl} underline>Sign Up</TextLink>
```
mzikherman: I think you meant `signUpUrl` for this one? (and you can throw the forgot password link in there, or even leave it out for now).

src/containers/login.tsx
l2succes: right good point

src/containers/login.tsx
l2succes: yeah, i'll just leave it out for now

src/components/title.tsx
```diff
@@ -16,11 +16,16 @@ const titleSizes = {
   xxlarge: "72px",
 }
 
-const Title: React.SFC<TitleProps> = props => (
-  <div className={props.className}>
-    {props.children}
-  </div>
-)
+const Title: React.SFC<TitleProps> = props => {
+  const newProps: TitleProps = {...props}
```
mzikherman: I def see this pattern in our components a lot, it sort of makes sense in order to allow passing options/props along to components.

mzikherman:
Looks great! :shipit:

PR #49, opened 2017-03-06
Add 'yarn type-check' to precommit hook

yuki24:


package.json
```diff
@@ -18,6 +18,7 @@
   },
   "precommit": [
     "lint",
+    "type-check",
     "test"
```
alloy: Nice. Didn’t know that the npm tooling provides a mechanism to automatically enable git hooks 👌

l2succes: It's a actually a node package called [pre-commit](https://www.npmjs.com/package/pre-commit), but yeah comes in handy

alloy: 👌

PR #52, opened 2017-03-08
@mzikherman - Artwork filter components

broskoski:
Doesn't do much but starting to wire up the components needed for a basic artwork filter component, starting with the dropdowns.

dropdown-component


src/components/__stories__/artwork_filter.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { FilterArtworksQueryConfig } from "../../relay/root_queries"
+
+import Dropdown from "../artwork_filter/dropdown"
+
+export class FilterArtworks extends React.Component<RelayProps, null> {
+  render() {
+    let dropdowns = []
+    this.props.filter_artworks.filter_artworks.aggregations.forEach((aggregation) => {
+      dropdowns.push(<Dropdown aggregation={aggregation}/>)
+    })
```
yuki24: This logic could be simplified using `map`:

```jsx
const dropdowns = this.props
  .filter_artworks
  .filter_artworks
  .aggregations.map(aggregation => <Dropdown aggregation={aggregation} key={someKey} />)
```

src/components/artwork_filter/dropdown.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    let buttonColor = "white"
+    let buttonTextColor = "black"
+    let navStyle = { display: "none" }
+
+    if (this.state.is_hovered) {
+      buttonColor = "black"
+      buttonTextColor = "white"
+      navStyle = { display: "block" }
+    }
```
yuki24: Isn't it possible to just use CSS instead of using `state`?

src/components/__stories__/artwork_filter.tsx
broskoski: 👍

src/components/artwork_filter/dropdown.tsx
broskoski: Good question, I couldn't figure out a non-complicated way of doing this with `styled-components`. If you can think of a more straightforward way to do it, lmk!

broskoski:
Added a basic demo of selecting an item for the dropdown (still need to represent that state in the dropdown button), and an extremely basic `TotalCount` component.

dropdown-component-2
screenshot 2017-03-09 12 44 26

src/components/artwork_filter/total_count.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface TotalCountProps extends RelayProps, React.HTMLProps<TotalCount> {
+  filter_artworks: any
+}
+
+export class TotalCount extends React.Component<TotalCountProps, null> {
+
+  render() {
+    const total = this.props.filter_artworks.counts.total
+    const s = total > 1 ? "s" : ""
```
alloy: Assuming `total` could be 0, this should use `!=` instead of `>`.

src/components/__stories__/artwork_filter.tsx
```diff
@@ -0,0 +1,145 @@
...
+import * as React from "react"
+import * as Relay from "react-relay"
+
+import { artsyNetworkLayer } from "../../relay/config"
+import { FilterArtworksQueryConfig } from "../../relay/root_queries"
+
+import Dropdown from "../artwork_filter/dropdown"
+import TotalCount from "../artwork_filter/total_count"
+
+export interface FilterArtworksDropdownState {
```
alloy: Why are you exporting from this stories file?

alloy:
Looking nice thus far! 👌

src/components/artwork_filter/dropdown.tsx
l2succes: @broskoski you can do it that way https://github.com/artsy/reaction-force/blob/master/src/components/buttons/default.tsx#L67

src/components/artwork_filter/dropdown.tsx
```diff
@@ -0,0 +1,145 @@
...
+import colors from "../../assets/colors"
+import { primary, secondary } from "../../assets/fonts"
+
+export interface DropdownProps extends RelayProps, React.HTMLProps<Dropdown> {
+  aggregation: any
+  onSelect?: any
+}
+
+export interface DropdownState {
+  is_hovered: boolean
```
l2succes: can we use camel casing for variable names

src/components/artwork_filter/dropdown.tsx
```diff
@@ -0,0 +1,145 @@
...
+      navStyle = { display: "block" }
+    }
+
+    return (
+      <div 
+        className={this.props.className}
+        onMouseEnter={() => this.toggleHover()} 
+        onMouseLeave={() => this.toggleHover()}
+      >
+        <Button style={{ backgroundColor: buttonColor, color: buttonTextColor }}>
```
l2succes: use a styled component to make background and text color configurable like

```javascript
const StyledButton = styled.button`
  backgroundColor: ${props => backgroundColor: props.backgroundColor};
  ...
`
```

src/components/artwork_filter/dropdown.tsx
broskoski: @l2succes eh, more complicated than that, child elements have to show/hide.

src/components/artwork_filter/dropdown.tsx
broskoski: Sure

src/components/__stories__/artwork_filter.tsx
broskoski: 🍝

alloy: Mmmmm

broskoski:
dropdown-component-3

So the `Dropdown` component is pretty much ready to go, I think it's probably good to merge, before this PR gets too large.

package.json
```diff
@@ -51,6 +51,7 @@
     "asap": "https://github.com/alloy/asap.git",
     "react": "^15.4.2",
     "react-relay": "https://github.com/alloy/relay/releases/download/v0.9.3/react-relay-0.9.3.tgz",
+    "react-url-query": "^1.1.1",
```
alloy: Where is this used?

broskoski: Ah, this is where I decided to cut myself off. Started down the path of making a component to manage url state and pass it down to filter components.

broskoski:
@l2succes @yuki24 @alloy are you guys married to having the linter and tests run on pre-commit? I find it kind of annoying (not being able to commit if my code doesn't pass lint).

mzikherman:
Ok I'm merging, looks awesome @broskoski

l2succes:
@broskoski I honestly think it helps keep a uniform codestyle throughout the project. Personally, I know i'd forget to fix them without it. Since it's gonna run on CI anyway might as well catch the problems before pushing to a remote branch and waiting for the build to fail. 

For most linter errors, you can just run `yarn lint -- --fix` and it will fix them automagically.
If there's a rule thats really a pain in the ass we can always remove or tweak it.

broskoski:
@l2succes oh yeah, def not arguing with a linter, more prohibiting a commit before fixing lints and passing tests. I suppose I'm more accustomed to getting code in and committed and then running the linters and tests.

l2succes:
@broskoski Oh word, yeah I mean we could move to a pre-push hook instead, that way it'll just run the linter and tests before pushing upstream. 

There's a node package called [pre-push](https://www.npmjs.com/package/pre-push) that works the same way as the `pre-commit` package we're currently using. You can swap them if you want.

src/components/__stories__/artwork_filter.tsx
```diff
@@ -0,0 +1,144 @@
...
+import * as React from "react"
+import * as Relay from "react-relay"
+
+import { artsyNetworkLayer } from "../../relay/config"
+import { FilterArtworksQueryConfig } from "../../relay/root_queries"
+
+import Dropdown from "../artwork_filter/dropdown"
+import TotalCount from "../artwork_filter/total_count"
+
+interface FilterArtworksDropdownState {
```
l2succes: @broskoski Is there a reason why you put `FilterArtworksDropdown` and its state interface in a story file?

broskoski: Just to show a basic example of how to handle state changes in the dropdowns

l2succes: Ah ok 👍🏾 cool

alloy:
Yeah I agree on moving to `pre-push` 👍

PR #53, opened 2017-03-15
@mzikherman - Artwork filter pt. 2

broskoski:
This time I am building a component that will wrap all of them together.

artwork-filter

This is still missing:
- [ ] Component to listen to url params
- [x] Sort dropdown
- [x] Headline component ("Small conceptual art under $100)")
- [x] For Sale checkbox


broskoski:
Added headline component
artwork-filter-headline

src/components/artwork_filter/headline.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { secondary } from "../../assets/fonts"
+import labelMap from "./param_map"
+
+import { compact } from "lodash"
+import styled from "styled-components"
+
+interface HeadlineProps extends React.HTMLProps<Headline> {
+  medium: any,
+  price_range: any,
+  dimension_range: any,
```
alloy: Can these all be `foo: string | null` or `foo?: string`?

alloy:
Off-topic, but just wanna double-check with you that the artwork brick needs a better loading state (grey background and correct height laid out immediately).

alloy:
Would it be possible to change the label of the selected price filter to remove ‘BETWEEN’ and just show the price range instead?

broskoski:
@alloy first point: yes! Second point: just replicating what we do on artsy.net currently. These filters are the least used across the site (only on the tag, gene and fair pages). I suppose it would be up to @briansw or @katarinabatina whether or not to change the way those filters behaved.

briansw:
@alloy you're just saying change it from "Between $1,000 and $5,000" to "$1,000–5,000" right? Seems like a good call considering the range gets truncated so severely. If it's a big time investment, it's probably not worth the effort since these pages don't drive many inquiries. 

$50–1,000
$1,000–5,000
$5,000–10,000
$10,000–50,000
$50,000+

alloy:
@briansw Yep, that’s exactly what I meant, but, just to be clear, _only_ for the selected state. In the dropdown I think it’s fine to prefix with ‘Between’. I hear you both on them being little, so let’s indeed skip it if it’s more than a few minutes of work.

briansw:
I'd say let's make them match in the dropdown and selected state if we end up doing this.

src/components/__stories__/artwork_filter.tsx
```diff
@@ -135,10 +136,23 @@ function FilterArtworksTotalCountExample() {
   )
 }
 
+function FilterArtworksExample() {
+  Relay.injectNetworkLayer(artsyNetworkLayer())
```
l2succes: We should move this to the storybook config file and only call it once.

src/components/artwork_filter/headline.tsx
```diff
@@ -0,0 +1,64 @@
...
+      return this.props.medium
+    }
+    return "Works"
+  }
+
+  priceRange() {
+    if (this.props.price_range && this.props.price_range !== "*") {
+      return this.props.price_range
+    }
+    return false
```
damassi: Why not 
```javascript
const { price_range } = this.props

if (price_range && price_range !== "*") {
  return price_range
}
```

I always think to myself if `this.props` is used more than once its best to destructure a value out, especially because you might grab more props later.

src/components/__stories__/artwork_filter.tsx
alloy: I think there was a reason to have to call it each time, but can’t recall from the top of my head.

src/components/artwork_filter/headline.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from "react"
+
+import { secondary } from "../../assets/fonts"
+import labelMap from "./param_map"
+
+import { compact } from "lodash"
+import styled from "styled-components"
+
+interface HeadlineProps extends React.HTMLProps<Headline> {
+  medium: any,
```
l2succes: Could you assign types to these properties instead of `any`? It'll make the typescript compiler a lot of more helpful with type checking and autocompletion

src/components/artwork_filter/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import Artworks from "../artwork_grid"
+import Dropdown from "./dropdown"
+import Headline from "./headline"
+import TotalCount from "./total_count"
+
+const PageSize = 10
+
+interface ArtworkFilterProps extends RelayProps, React.HTMLProps<ArtworkFilter> {
+  filter_artworks: any
```
l2succes: see above ^

alloy: This can simply be removed, because `ArtworkFilterProps` extends `RelayProps` which contains the actual `filter_artworks` typings.

src/components/artwork_filter/headline.tsx
broskoski: 👍

src/components/artwork_filter/for_sale_checkbox.tsx
```diff
@@ -0,0 +1,63 @@
...
+import colors from "../../assets/colors"
+import { primary } from "../../assets/fonts"
+
+interface ForSaleCheckboxProps extends React.HTMLProps<ForSaleCheckbox> {
+  onClick?: any
+}
+
+interface ForSaleState {
+  isChecked: boolean,
+}
```
alloy: FYI As long as they are not exported, you can just call these interfaces `Props` and `State`.

alloy:
Looking dope, @broskoski :bowtie: 👌

broskoski:
Here's how it's looking now:
artwork-filter-sort-forsale

In my rudimentary tests, I couldn't successfully capture changes to url params, but I suspect it's due to the React Storybook router. I think that part will need to be done outside of this context.

mzikherman:
ok I'm merging, I think this looks stellar! 🎉

alloy:
Looks great @broskoski! 👏

PR #61, opened 2017-03-28
Inquiry Artwork Brick

xtina-starr:
This creates a new inquiry artwork brick that has minimal metadata and includes the `isSelected` tap state.

new artwork brick


yuki24:
Overall looks good to me :+1:

Also, this is just an idea, but the other thing we could do, instead of using `state`, is to use an actual checkbox in the component and [the :checked CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:checked). That way we don't have to use any state and the "what artworks did you purchase?" form will be able to capture the checkbox values directly. Also, anyone who is not familiar with React will be able to understand how the form works since this is just a plain html form. What do you think @xtina-starr @alloy @mzikherman @broskoski?

src/components/__stories__/inquiry_artwork.tsx
```diff
@@ -0,0 +1,30 @@
+import { action, storiesOf } from "@kadira/storybook"
+import * as React from "react"
+import * as Relay from "react-relay"
+
+import Artwork from "../inquiry_artwork_brick/index"
```
alloy: FYI There’s no need to include the `index` filename, it is a special filename that by default is the entry point of the directory it resides in.

src/components/inquiry_artwork_brick/details.tsx
```diff
@@ -0,0 +1,102 @@
+import * as React from "react"
+import * as Relay from "react-relay"
+import styled from "styled-components"
+import colors from "../../assets/colors"
+import TextLink from "../text_link"
+
+export interface DetailsProps extends React.HTMLProps<ArtworkDetails> {
+  artwork: any
```
alloy: You should extend from `RelayProps` as well and drop this `any` typing to get the actual types listed below https://github.com/artsy/reaction-force/pull/61/files#diff-bb6edb8ba51d5c558582b32f2884d355R88. E.g.

```ts
interface DetailsProps extends RelayProps, React.HTMLProps<ArtworkDetails> {}
```

(PS: I left out the `export` because I don’t see this interface being used outside of this file.)

src/components/inquiry_artwork_brick/details.tsx
```diff
@@ -0,0 +1,102 @@
...
+`
+
+export class ArtworkDetails extends React.Component<DetailsProps, null> {
+  artistLine() {
+    const { cultural_maker, artists } = this.props.artwork
+
+    if (cultural_maker) {
+      return <div><strong>{cultural_maker}</strong></div>
+    } else if (artists && artists.length) {
+      let artistLine = artists
```
alloy: Use `const` instead of `let` if the variable is never going to change. I’m surprised the linter didn’t catch this 🤔

src/components/inquiry_artwork_brick/details.tsx
alloy: Once you do this, you’ll notice that when you access `this.props.artwork` you’ll start getting completion for the fields listed in the `RelayProps` interface.

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  position: absolute;
+  z-index: 10;
+  top: 0;
+  left: 0;
+  opacity: 0.8;
+  text-align: center;
+  line-height: 350px;
+`
+
+export class ArtworkState {
```
alloy: I don’t believe this is being used outside of this file, so `export` can be removed. (State details should probably _never_ be shared outside of a file, as they should be internal details.)

Additionally, I wonder if you really meant to use `class` here instead of `interface`? An interface is just a compile time thing that allows the compiler to check that the object conforms to a specific shape, whereas a class is an actual OOP class that can be instantiated, which we don’t do, as can be seen on line 53.

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  isSelected: boolean
+}
+
+export class Artwork extends React.Component<RelayProps, ArtworkState> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      isSelected: false,
+    }
+  }
```
alloy: Nit: missing newline

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    let selectedOverlay
+    if (this.state.isSelected) {
+      selectedOverlay = (
+        <SelectedArtworkOverlay onClick={this.onSelect.bind(this)} > 
+          <Icon name="check" color="white" />
+        </SelectedArtworkOverlay>
+      )
+    }
+    return (
+      <div style={{position: "relative"}} >
```
alloy: This should move to a styled component.

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    this.setState({
+      isSelected: !this.state.isSelected,
+    })
+  }
+
+  render() {
+    let selectedOverlay
+    if (this.state.isSelected) {
+      selectedOverlay = (
+        <SelectedArtworkOverlay onClick={this.onSelect.bind(this)} >
```
alloy: I think you can use `pointer-events: none;` to make this element ignore events and thus they would be handled by the `ImageOuterContainer`. This only works with IE >= 11, what versions are we targeting?

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+export default Relay.createContainer(Artwork, {
+  fragments: {
+    artwork: () => Relay.QL`
+      fragment on Artwork {
+        image {
+          url(version: "large")
+          aspect_ratio
```
alloy: This field doesn’t appear to actually be used in the component.

alloy:
Looking good thus far 👌

alloy:
> the other thing we could do […] is to use an actual checkbox in the component and the :checked CSS selector. That way we don't have to use any state and the […] form will be able to capture the checkbox values directly.

While I like the idea of using HTML as closely as possible, I wonder if making assumptions about the implementation details of child components jives well with the React way. Because the child components _could_ change away from checkboxes to something else and the parent component wouldn’t know and effectively break. By using an API contract, especially a TypeScript one, any such change is hidden and would break the build.

src/components/inquiry_artwork_brick/details.tsx
```diff
@@ -0,0 +1,102 @@
+import * as React from "react"
+import * as Relay from "react-relay"
+import styled from "styled-components"
+import colors from "../../assets/colors"
+import TextLink from "../text_link"
+
+export interface DetailsProps extends React.HTMLProps<ArtworkDetails> {
```
broskoski: As much as you can, try and extend the existing classes instead of duplicating them in a new folder. If any of that logic changes, we will now have to change it in two places.

src/components/inquiry_artwork_brick/details.tsx
xtina-starr: So does that meet that the DetailsProps object should now be empty? It seems like I lose access to some of the props on artwork by doing that.

src/components/inquiry_artwork_brick/index.tsx
xtina-starr: I don't quite understand this comment

src/components/inquiry_artwork_brick/details.tsx
alloy: By lose, do you mean the TS compiler doesn’t allow you to access them? If so, which props?

src/components/inquiry_artwork_brick/index.tsx
alloy: I assume the second paragraph? Let’s reboot it, why are you using `class` here instead of `interface`?

xtina-starr: Oh okay, yeah I should use interface here.

src/components/inquiry_artwork_brick/details.tsx
xtina-starr: I was missing some of the properties on the artist object. I figured this out and realized I needed to add them to the `RelayProps` interface. However, now it doesn't like the types passed to the `reduce` method called on the artists collection.

xtina-starr:
Does anyone have any more feedback for this PR?

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+import colors from "../../assets/colors"
+import Icon from "../icon"
+import ArtworkMetadata from "./metadata"
+
+const InquiryArtworkBrick = styled.div`
+  position: relative;
+`
+
+const ImageContainer = styled.div`
+  height: 350px;
```
l2succes: consider adding variable sizes based on the size of the screen (e.g. desktop vs mobile). We can do that in a separate PR when we add mobile support though.

For instance:

```
  height: ${props => props.size == Sizes.small ? 150 : 350 }
  ...
```

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  display: table-cell;
+  vertical-align: middle;
+  width: 100%;
+  height: 100%;
+  text-align: center;
+`
+
+const Image = styled.img`
+  max-width: 100%;
+  max-height: 350px;
```
l2succes: make width and height variables defined at the top and use them here

```
max-height: ${defaultHeight};
max-width: ${defaultWidth};
```

src/components/inquiry_artwork_brick/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  margin: auto;
+`
+
+const SelectedArtworkOverlay = styled.div`
+  background-color: ${colors.purpleRegular};
+  height: 350px;
```
l2succes: see above ^

src/components/inquiry_artwork_brick/index.tsx
xtina-starr: just added this suggestion.

src/components/inquiry_artwork_brick/index.tsx
xtina-starr: I agree that this can be added in a separate PR.

xtina-starr:
@l2succes I made the change you recommended. Is this good to merge?

l2succes:
Yup! Let's merge it.

xtina-starr:
👍

xtina-starr:
Do we check-in `dist/`??? @l2succes

l2succes:
@xtina-starr Na we're not supposed to check in `dist` anymore. I added it to `.gitignore` in my PR so no worries, i'll take care of it.

xtina-starr:
_phew!_

PR #69, opened 2017-04-04
[WIP] Inquiry artwork grid

xtina-starr:
I'm not sure if we want to create an entirely new artwork grid here, but I just tweaked the previous one to use the new inquiry artwork brick. I figured it was at a good point to get feedback.

artwork gird

### DIFF
--- a/src/components/artwork_grid.tsx
+++ b/src/components/artwork_grid.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import * as Relay from "react-relay"
 import styled from "styled-components"
-import Artwork from "./artwork/index"
+import Artwork from "./inquiry_artwork_brick"
 
 export interface GridProps extends RelayProps, React.HTMLProps<ArtworkGrid> {
   columnCount?: number,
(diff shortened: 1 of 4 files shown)


broskoski:
Yeah, definitely don't want entirely new `ArtworkGrid` here. The dumb fix would be to pass in a param that would specify which `Artwork` component to use.

Also, I still think the `InquiryArtwork` shouldn't be entirely new either, it should just extend the current one (or at the very least, use the current `ArtworkMetadata` component).

It _looks_ beautiful tho 👍

xtina-starr:
@broskoski the original artwork component and the new artwork component have pretty different style rules, no? Plus one has state and one doesn't.

broskoski:
@xtina-starr to some extent, but I think the fact that a lot of the code is duplicate should be an indication that some refactoring can be done to not repeat any code. If any of the artwork brick metadata rules change, now we have to update them in two places. Just food for thought!

xtina-starr:
@broskoski I'm not totally sure what it would look like to extend the original artwork, but extracting the metadata and using a prop or something to determine whether or not the `ArtworkContact` displays makes sense.

PR #73, opened 2017-04-05
Sets up isomorphic relay and styling for inquiries page

l2succes:


src/components/artwork/index.tsx
```diff
@@ -1,42 +1,70 @@
...
 `
 
-export class Artwork extends React.Component<RelayProps, null> {
+export interface ArtworkProps extends RelayProps {
+  size?: number
+  extended?: boolean
+  overlay?: JSX.Element
+}
+
+export interface ArtworkState {
```
alloy: Do the props and state really need to be exported?

.babelrc
```diff
@@ -5,7 +5,7 @@
     "react"
   ],
   "plugins": [
-    "react-relay"
+    "./data/schema"
```
alloy: Did the plugin not work?

alloy:
Looks great 👏

.babelrc
l2succes: It didn't work at first for me when I was trying different things but it does work so I'll put it back

src/components/artwork/index.tsx
l2succes: Yeah they're being used [here](https://github.com/artsy/reaction-force/pull/73/files/0a06da98fb41eb18b3d35a0cd712d7f6caf4fab6#diff-d9d55e0bf5232ce704f25858eb13380fR5)

l2succes:
Ok 🙏🏾, cool merging

alloy:
👌

src/components/artwork/index.tsx
```diff
@@ -1,42 +1,70 @@
...
-import * as Relay from "react-relay"
+
 import styled from "styled-components"
-import ArtworkMetadata from "./metadata"
+import ArtworkMetadata, { ArtworkMetadataProps } from "./metadata"
+import createContainer, { RelayProps } from "./relay"
+
+const ImageContainer = styled.div`
+  height: ${props => `${props.size}px`};
+  width: ${props => `${props.size}px`};
```
broskoski: :| this breaks the `ArtworkGrid`, I think we are going to have to make separate containers for non-grid artworks.

src/components/artwork/metadata.tsx
```diff
@@ -9,14 +9,19 @@ import ArtworkDetails from "./details"
 
 export interface ArtworkMetadataProps extends React.HTMLProps<ArtworkMetadata> {
   artwork: any
+  extended?: boolean
 }
 
 export class ArtworkMetadata extends React.Component<ArtworkMetadataProps, null> {
+  static defaultProps = {
+    extended: false,
```
mzikherman: I'm coming to this way late, but I'm taking a look at the artwork bricks, and this seems to be an issue.

Namely, this default means this component won't show the sale line or the `ArtworkContact` component, which is generally always a part of artwork bricks. It might even be _the_ most important part! You'd need to pass in 

The `Artwork/GridItem` and `Artwork` index components don't work with this- neither allows passing in the`extended` prop down to this component.

Sure enough, the Gene page artwork bricks aren't correct:

screen shot 2017-10-26 at 11 23 42 am

There should be a price and contact button underneath, like:

screen shot 2017-10-26 at 11 24 09 am

Do you recall if this was just fallout? Most likely all the base components should allow this to be passed thru (with a default of true), and then the Loyalty component can override this to false, if it needs to.

alloy: The default should definitely be `true` indeed. Also, do we need not extended components anywhere besides these loyalty program bricks? Otherwise maybe it can just be a similar but separate component that doesn’t include the metadata component?

alloy: Looking at https://github.com/artsy/reaction/blob/a486ed3651ee5e0a4ae5852c05d1ed3f6dd5cb6f/src/components/artwork/index.tsx I think even _more_ that separating bricks for loyalty and all others is the way to go. Normal bricks don’t need an overlay either, do they?

PR #88, opened 2017-04-10
[dev] Make HMR work on /inquiries.

alloy:
Need to not only export the Relay container, but also the component itself.

See https://github.com/alloy/relational-theory/commit/2cc900ea216f61c7a4a8096bb9177a648b135129

### DIFF
--- a/src/apps/loyalty/containers/inquiries/index.tsx
+++ b/src/apps/loyalty/containers/inquiries/index.tsx
@@ -28,7 +28,7 @@ export interface State {
   self_reported_purchases?: string
 }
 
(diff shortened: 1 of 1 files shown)


orta:
Is this something we should make  danger rule for in the future? In Emission there's some per-component based rules already

alloy:
Ooh, yeah, I like that 👌

l2succes:
@orta Would the rule just be export all React components by default?
I feel like it'd be too much work to check that component is wrapped in a Relay container somewhere

orta:
it's code that doesn't exist yet, we can do whatever we want 👍

PR #90, opened 2017-04-11
[Loyalty] updates graphQL schema and adds mutation for inquiries

l2succes:


src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -59,42 +65,76 @@ export class Inquiries extends React.Component<RelayProps, State> {
...
+              artwork={artwork as any}
+              onSelect={this.onArtworkSelected.bind(this, impulse_conversation_id)}
+            />
           </InquiryContainer>
         </Col>
       )
     })
   }
 
+  onArtworkSelected(conversationId: string, selected: boolean) {
```
mzikherman: I think @alloy was mentioning these can be typed a bit more in https://github.com/artsy/reaction-force/pull/80#discussion_r110269117

That's a todo

src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -59,42 +65,76 @@ export class Inquiries extends React.Component<RelayProps, State> {
...
+    const mutation = new UpdateConversationMutation({input: {
+      ids,
+      buyerOutcome: "PURCHASED",
+    }})
 
-    const mutation = new UpdateCollectorProfileMutation(this.state)
+    Relay.Store.commitUpdate(mutation, {
+      onFailure: this.onSubmitUpdatesFailed,
+      onSuccess: response => {
+        this.submitCollectorProfileUpdate()
```
mzikherman: mutations on mutations!

mzikherman:
LGTM! 🎉

src/apps/loyalty/containers/inquiries/index.tsx
l2succes: This is actually just a regular callback since I'm not forwarding the `React.FormEvent` in this case, but we add that to the Artwork component though. Good point

src/apps/loyalty/server/middlewares/user.ts
```diff
@@ -0,0 +1,21 @@
+import { NextFunction, Request, Response } from "express"
+import { fetchCollectorProfile } from "../gravity"
+
+export default (req: Request, res: Response, next: NextFunction) => {
```
mzikherman: This approach sort of gives me pause.

For one, we're using Backbone by attaching it to the user. We actually have https://github.com/artsy/artsy-passport/issues/104 and may hopefully be able to remove Backbone entirely very soon. Is there a way to just use plain JS objects? (and is `res.locals` the place of choice for it?)

Also, doing it this way loses the type information we have associated with the response object. From https://github.com/artsy/reaction-force/blob/2a364e2ebe1ed6d8c0eec63add9bfc41927d9a4b/src/apps/loyalty/server/enroll_loyalty_applicant.ts#L20, the return value is typed, but doing it this way loses that and we just get an `any`.

l2succes: Yeah, I actually had this dilemma as well. I was going to use `res.locals` but ended up resorting to the user model since I had doubts whether properties would get cleared on res.locals after each request. I'm looking into it now.

l2succes: I think it's fair to move to res.locals once we remove `artsy-passport`'s dependency on backbone. It works pretty well for now.

src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -59,42 +73,76 @@ export class Inquiries extends React.Component<RelayProps, State> {
...
-      if (!response.updateCollectorProfile.loyalty_applicant_at) {
-        console.log("Loyalty Applicant Not Saved") // tslint:disable-line:no-console
-      } else {
-        console.log("Success") // tslint:disable-line:no-console
+  submitCollectorProfileUpdate() {
+    const mutation = new UpdateCollectorProfileMutation(this.state)
+
+    Relay.Store.commitUpdate(mutation, {
+      onFailure: this.onSubmitUpdatesFailed,
+      onSuccess: response => {
```
mzikherman: kewl

src/apps/loyalty/containers/inquiries/mutations/update_conversation.ts
```diff
@@ -0,0 +1,47 @@
...
+export type BuyerOutcome =
+  "PURCHASED" |
+  "STILL_CONSIDERING" |
+  "HIGH_PRICE" |
+  "LOST_INTEREST" |
+  "WORK_UNAVAILABLE" |
+  "OTHER"
+
+interface MutationProps {
+  input: {
```
mzikherman: I guess it's weird that for the conversation mutation we didn't need this?

src/apps/loyalty/containers/inquiries/mutations/update_conversation.ts
```diff
@@ -0,0 +1,47 @@
...
+    return Relay.QL`
+      mutation {
+        updateConversation
+      }
+    `
+  }
+
+  getVariables() {
+    return {
+      ids: this.props.input.ids,
```
mzikherman: So without the `input` here and above it wasn't working? Strange...

src/apps/loyalty/server/middlewares/user.ts
mzikherman: yea, I hear you on that. I think that makes total sense 👍 , sorry for the noise.

src/apps/loyalty/server/middlewares/user.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (req.user && !res.locals.profile) {
```
mzikherman: i guess we can remove the `!res.locals.profile` check for now

mzikherman:
Feel free to merge when ready.

src/apps/loyalty/containers/3w_thank_you/index.tsx
```diff
@@ -2,12 +2,13 @@ import * as React from "react"
...
-import Text from "../../../../components/text"
-import TextLink from "../../../../components/text_link"
-import Title from "../../../../components/title"
+
+import Button from "components/buttons/inverted"
+import Nav from "components/nav"
+import NavItem from "components/nav_item"
+import Text from "components/text"
+import TextLink from "components/text_link"
+import Title from "components/title"
```
damassi: So much nicer :)

src/apps/loyalty/containers/acb_thank_you/index.tsx
```diff
@@ -2,12 +2,13 @@ import * as React from "react"
...
-import Text from "../../../../components/text"
-import TextLink from "../../../../components/text_link"
-import Title from "../../../../components/title"
+
+import Button from "components/buttons/inverted"
+import Nav from "components/nav"
+import NavItem from "components/nav_item"
+import Text from "components/text"
+import TextLink from "components/text_link"
+import Title from "components/title"
```
alloy: This might look even nicer if we’d align the `from` parts. Is that something that’s commonly done in JS?

src/apps/loyalty/containers/inquiries/index.tsx
alloy: 👍

src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -59,42 +73,76 @@ export class Inquiries extends React.Component<RelayProps, State> {
...
+        if (response.updateCollectorProfile.loyalty_applicant_at) {
+          window.location.pathname = "/loyalty/thank-you"
+        }
+      },
+    })
+  }
+
+  submitInquiriesUpdate() {
+    let ids: string[] = []
```
alloy: Should be `const`. We’re only modifying the contents of the array, not changing the array that this variable is referring to.

src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -59,42 +73,76 @@ export class Inquiries extends React.Component<RelayProps, State> {
...
+    const artworks = this.state.selected_artworks
+
+    for (let artworkId of Object.keys(artworks)) {
```
alloy: Should be `const`.

src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -59,42 +73,76 @@ export class Inquiries extends React.Component<RelayProps, State> {
...
+        this.submitCollectorProfileUpdate()
+      },
+    })
+  }
 
-    Relay.Store.commitUpdate(
-      mutation, {onFailure, onSuccess},
-    )
+  onSubmitUpdatesFailed(transaction) {
+    console.log(transaction.getError()) // tslint:disable-line:no-console
```
alloy: Nit: Use `console.error`

src/apps/loyalty/containers/inquiries/index.tsx
```diff
@@ -148,6 +199,7 @@ interface RelayProps {
       edges: Array<{
         node: {
           id: string | null,
+          impulse_conversation_id: string | null,
           artwork: Array<any | null> | null,
         } | null,
       } | null> | null,
```
alloy: All these `null`s……… 😢🐼

src/apps/loyalty/containers/inquiries/mutations/update_conversation.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type BuyerOutcome =
+  "PURCHASED" |
+  "STILL_CONSIDERING" |
+  "HIGH_PRICE" |
+  "LOST_INTEREST" |
+  "WORK_UNAVAILABLE" |
+  "OTHER"
+
+interface MutationProps {
```
alloy: Nit: Now that these mutations have moved out of the component file, these interfaces could just be called `Props` again.

src/apps/loyalty/server/index.tsx
```diff
@@ -62,64 +64,57 @@ app.get("/inquiries", (req, res) => {
...
+  }
+
+  IsomorphicRelay.prepareData({
+    Container: Inquiries,
+    queryConfig: new CurrentUserRoute(),
+  }, res.locals.networkLayer).then(
+    ({data, props}) => {
+      const html = renderToString(<IsomorphicRelay.Renderer {...props} />)
+      const styles = styleSheet.rules().map(rule => rule.cssText).join("\n")
+      res.locals.sharify.data.USER_DATA = req.user.toJSON()
```
alloy: Isn’t sharify already storing the `user` under something like `CURRENT_USER`?

src/apps/loyalty/server/middlewares/relay.ts
```diff
@@ -1,7 +1,7 @@
 import { NextFunction, Request, Response } from "express"
 import * as Relay from "react-relay"
 
-export const RelayMiddleware = (req: Request, res: Response, next: NextFunction) => {
+export default (req: Request, res: Response, next: NextFunction) => {
```
alloy: I would prefer that we use named function syntax where possible, because that aids in debugging (named functions get identified as such).

src/components/artwork/relay.ts
```diff
@@ -8,6 +8,7 @@ function createContainer<T, S>(ComponentClass: React.ComponentClass<T> | React.S
     fragments: {
       artwork: () => Relay.QL`
         fragment on Artwork {
+          id
```
alloy: What is this for?

package.json
```diff
@@ -80,6 +80,7 @@
     "numeral": "^2.0.4",
     "react": "^15.4.2",
     "react-relay": "https://github.com/alloy/relay/releases/download/v0.9.3/react-relay-0.9.3.tgz",
+    "react-styled-flexboxgrid": "^1.0.2",
```
alloy: Do our needs really go way beyond the grid we already had?

l2succes: We need a grid that supports the responsive layout. When rendering on a mobile display, we need to drop two columns. That is not supported by the grid we have and would take some time to implement. That grid component fits right in as it's built on top styled-components and the API is virtually the same except you can specify column widths depending on the screen size.

```javascript
import React from 'react'

import {Grid, Col, Row} from 'react-styled-flexboxgrid'

const App = props =>
  <Grid>
    <Row>
      <Col xs={6} md={3}>Hello, world!</Col>
    </Row>
  </Grid>
```

screen shot 2017-04-13 at 10 00 14 am

PR #99, opened 2017-04-14
@alloy => [dist] fixes paths for js bundles

l2succes:
Fixes #99.

### DIFF
--- a/src/apps/loyalty/server/template.tsx
+++ b/src/apps/loyalty/server/template.tsx
@@ -4,9 +4,10 @@ interface TemplateData {
   entrypoint: string
   bootstrapData?: string
   sharify?: string
+  baseURL: string
 }
 
-export default ({styles, html, entrypoint, bootstrapData, sharify}: TemplateData) => {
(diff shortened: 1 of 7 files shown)


src/apps/loyalty/server/template.tsx
```diff
@@ -20,7 +20,7 @@ export default ({styles, html, entrypoint, bootstrapData, sharify}: TemplateData
           <div id="app-container">${html}</div>
           ${sharify ? sharify : ""}
           ${bootstrapData ? "<script>" + bootstrapData + "</script>" : ""}
-          <script src="/bundles/commons.chunk.js"></script>
+          <script src="/loyalty/bundles/commons.chunk.js"></script>
```
alloy: Why not use `req.basseUrl` here as well?

alloy:
Linked it to #99 for ya.

src/apps/loyalty/server/template.tsx
l2succes: yeah ur right. laziness lol

PR #109, opened 2017-04-18
[Loyalty QA] center log in button text in safari

xtina-starr:
closes: https://github.com/artsy/reaction-force/issues/103

By default, the `flex-direction` property is set to `row` which aligns the flex-items along the main (x) axis. It looks like that is what's causing the "LOG IN" text to sit to the left. I think we want to style it to have`flex-direction: column` in this case. 

screen shot 2017-04-18 at 1 12 47 pm

### DIFF
--- a/src/apps/loyalty/containers/login/index.tsx
+++ b/src/apps/loyalty/containers/login/index.tsx
@@ -53,7 +53,7 @@ class Login extends React.Component<LoginProps, LoginState> {
 
           {form.csrfToken && <input type="hidden" name="_csrf" value={form.csrfToken} />}
 
-          <Button block>Log In</Button>
+          <Button block style={{flexDirection: "column"}}>Log In</Button>
           <div style={{textAlign: "center"}}>or</div>
           <FacebookButton href={form.facebookPath} block />
           <TwitterButton href={form.twitterPath} block />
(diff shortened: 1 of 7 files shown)


src/apps/loyalty/containers/login/index.tsx
```diff
@@ -53,7 +53,7 @@ class Login extends React.Component<LoginProps, LoginState> {
 
           {form.csrfToken && <input type="hidden" name="_csrf" value={form.csrfToken} />}
 
-          <Button block>Log In</Button>
+          <Button block style={{flexDirection: "column"}}>Log In</Button>
```
l2succes: can you add that style to the base button component instead?

xtina-starr: The facebook and twitter buttons need to be set to the default which is why I added it here but I can still move it to the base button component and set the `flex-direction` to row for the twitter and facebook button components

xtina-starr:
I anticipate my last commit will fail the build because of the snap shot. How do I go about updating that?

l2succes:
@xtina-starr run `yarn test -- -u` to update snapshots.

l2succes:
Looks good! We can merge after updating the snapshot tests

xtina-starr:
tests are 💚  gonna go ahead and merge.

PR #113, opened 2017-04-18
@mzikherman - Adds SaveButton component

broskoski:
This will be our first reaction-force component that makes a mutation!

It's not quite working yet (the mutation happens, but the component doesn't re-render), but wanted to get a WIP PR up in the meantime. 

There's one thing I haven't totally figured out how to approach – this component will have to have some knowledge of whether or not a user is logged in or not (and if not, pop up the login modal). We could probably pass user data down at the top level component (kind of like what @l2succes is doing to init the networkLayer), but I imagine there will be other places where this kind of (typically global) state is needed. I'd rather not have to query `me` on every page load!


src/components/artwork/grid_item.tsx
```diff
@@ -23,13 +24,18 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItem> {
...
   render() {
+    const { style, artwork } = this.props
     return (
-      <div style={this.props.style}>
-        <Placeholder style={{ paddingBottom: this.props.artwork.image.placeholder }}>
-          <Image src={this.props.artwork.image.url} />
+      <div style={style}>
+        <Placeholder style={{ paddingBottom:  artwork.image.placeholder }}>
+          <Image src={artwork.image.url} />
+          <SaveButton
```
broskoski: Every `ArtworkGridItem` will have a `SaveButton` just like on Force.

src/components/save.tsx
```diff
@@ -0,0 +1,128 @@
...
+    background-color: black;
+  }
+  &[data-saved='true'] {
+    background-color: ${colors.purpleRegular};
+    &:hover {
+      background-color: ${colors.redBold};
+    }
+  } 
+`
+class SaveArtworkMutation extends Relay.Mutation<Props, null> {
```
broskoski: Here's where the mutation config for saving an artwork.

alloy:
> We could probably pass user data down at the top level component, but I imagine there will be other places where this kind of (typically global) state is needed.

Maybe we should pass some/all of the `req.user` data into the React ‘context’ ?

https://facebook.github.io/react/docs/context.html

src/components/save.tsx
```diff
@@ -0,0 +1,128 @@
+import * as React from "react"
```
l2succes: Since the component will only be used inside of the artwork grid item component, could you move it to the `artwork` folder?

broskoski: Ya, good call.

broskoski:
So haven't actually successfully applied gotten this mutation to work yet, and the reason why is totally mysterious to me. Will comment inline.

src/components/save.tsx
```diff
@@ -0,0 +1,137 @@
...
+        is_saved: this.props.artwork.is_saved ? false : true,
+      },
+    }
+  }
+
+  getFatQuery() {
+    return Relay.QL`
+      fragment on SaveArtworkPayload {
+        artwork {
+          is_saved
```
broskoski: This is what we expect back from Metaphysics. However, all we currently get is the `clientMutationId` because relay is somehow translating this to:

```
mutation Save($input_0:SaveArtworkInput!) {
  saveArtwork(input:$input_0) {
    clientMutationId
  }
}
```
With the input set to:
screenshot 2017-04-19 18 23 08

The `input` is set correctly, Relay isn't setting the response payload correctly.

alloy:
Let’s pair on this when you get in. I’d love to also start getting myself familiar with mutations.

src/components/artwork/grid_item.tsx
```diff
@@ -45,6 +45,7 @@ export default Relay.createContainer(ArtworkGridItem, {
   fragments: {
     artwork: () => Relay.QL`
       fragment on Artwork {
+        id
```
alloy: What is this for?

src/components/__stories__/save.tsx
```diff
@@ -3,22 +3,23 @@ import * as React from "react"
...
 }
 
 storiesOf("Save Button", SaveButton)
-  .add("Save Button", () => {
+  .add("Save Button (logged in)", () => {
     return (
       <div style={{width: "200px"}}>
         <ArtworkExample
           artworkID="damon-zucconi-tetradic-edit-1"
+          user={{id: "547e891d7261691220dc0000", accessToken: "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI1NDdlODkxZDcyNjE2OTEyMjBkYzAwMDAiLCJzYWx0X2hhc2giOiI2OTg3ZjUzZmExOTQxZGUwM2E0YTA2NWNjZjA0YjBiZiIsInJvbGVzIjoidXNlcixhZG1pbiIsInBhcnRuZXJfaWRzIjpbXSwiZXhwIjoxNDk1OTg1MTU0LCJpYXQiOjE0OTA4MDExNTQsImF1ZCI6IjRlMzZlZmE0ZGI0ZTMyMDAwMTAwMDM1OSIsImlzcyI6IkdyYXZpdHkiLCJqdGkiOiI1OGRiZDIwMjhiMGMxNDU3OWQ1YTEzOTgifQ.NNxQnjIAX8DTLtkR8XwB8-FTrZNqjLU9Y_Sf20kviLc"}}
```
alloy: I don’t remember what we settled on. Are these real?

broskoski:
@mzikherman I'm gonna go ahead and merge this and handle the logout stuff in a separate PR, if that's cool with you.

PR #119, opened 2017-04-19
[dev] Verify node version

alloy:
Force on Heroku uses LTS, which currently is 6.10.2, so we should dev/test against that too.


orta:
👍 Yep - this looks good - internal context [from slack](https://artsy.slack.com/archives/C0GNAB0US/p1492628536943634)

PR #122, opened 2017-04-19
[dev] move longjohn to dependency

l2succes:


alloy:
Please always include summaries in PRs 🙏🏽

What is the reason to move this to runtime deps?

l2succes:
I moved it to runtime deps in an attempt to fix the runtime crash on force that I noticed after deploying `reaction-force` https://sentry.io/artsynet/force-staging/issues/256060860/ It looked like the module was missing and was needed at runtime
```
Error: Cannot find module 'longjohn'
```
That was before we noticed `dist` was missing from the package. So because `dist/index.js` was missing, node would load the `index.js` at the root of the package which contains our dev server setup.

l2succes:
@alloy The root of the problem is the version of `npm` we have running on `circle` is outdated and doesn't run `npm prepublishOnly` before publishing the package. So we'd end up with a package with no `dist` files 😩

A solution is to explicitly call `npm prepublishOnly` in the config. we could also upgrade npm on CI which I'm not sure how to do.

l2succes:
Anyway, we can move it back `devDependencies` as it's not needed at runtime

alloy:
Aha, I see, I see. Cool, yeah please move it back indeed 👍

alloy:
About updating npm on [CircleCI](https://circleci.com/docs/1.0/configuration/), try something like:

```yaml
deployment:
  pre:
    - npm install npm@latest -g
```

This should make sure we don’t upgrade it for each test run as well, I guess.

PR #126, opened 2017-04-20
Better IE11 and IE Edge support

yuki24:
Because of the opacity bug in IE11, we needed to give up on the nice transition, but in terms of functionality, it totally works fine now.

### DIFF
--- a/webpack/base.js
+++ b/webpack/base.js
@@ -5,11 +5,11 @@ module.exports = () => {
   return {
     entry: {
       "login": [
-        "./src/apps/loyalty/containers/login/browser",
+        "babel-polyfill", "./src/apps/loyalty/containers/login/browser",
       ],
       "inquiries": [
-        "./src/apps/loyalty/containers/inquiries/browser",
-      ]
+        "babel-polyfill", "./src/apps/loyalty/containers/inquiries/browser",
+      ],
     },
     module: {
       rules: [
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
   "dependencies": {
     "artsy-passport": "^1.6.1",
     "artsy-xapp": "^1.0.4",
+    "babel-polyfill": "^6.23.0",
     "backbone": "^1.3.3",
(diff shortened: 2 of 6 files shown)


webpack/base.js
```diff
@@ -5,11 +5,11 @@ module.exports = () => {
   return {
     entry: {
       "login": [
-        "./src/apps/loyalty/containers/login/browser",
+        "babel-polyfill", "./src/apps/loyalty/containers/login/browser",
       ],
       "inquiries": [
-        "./src/apps/loyalty/containers/inquiries/browser",
-      ]
+        "babel-polyfill", "./src/apps/loyalty/containers/inquiries/browser",
```
alloy: Do we really need all of babel-polyfill? Or will it tree shake and only import the actual needed polyfills?

alloy: And btw, out of curiosity, how does adding the polyfill actually fix the issue?

l2succes: +1 @yuki24 ^

yuki24: I forgot to mention that there was another issue where clicking wasn't working at all. There is no relation between the opacity bug and polyfill. The issue I saw seemed to come from inside React. We also use other functions like `Object.assign` that's not available on IE11, so we'd have to polyfill anyway.

yuki24: Our `Object.assign` call is invoked on nodejs, so that shouldn't be a problem. However I still the following error when running without polyfill:

screen shot 2017-04-21 at 11 43 18 am

alloy: Oh my bad, I now see there were 2 commits indeed. Did you by any chance verify if webpack is tree shaking most of the polyfill out? (i.e. only including required code in the bundles)

yuki24: I'm not actually sure whether webpack does that or not. I'll double-check that as well 👍

alloy: iirc webpack 2 should tree shake by default, but haven’t done any verification myself

alloy:
Sniping this

PR #128, opened 2017-04-21
Fixes CircleCI auto publish

l2succes:
The `npm prepublishOnly` hook is not being run on circle CI because the version of npm is outdated so it has to be triggered manually

https://github.com/travis-ci/travis-ci/issues/6953

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,4 +22,5 @@ deployment:
   npm:
     tag: /v[0-9]+(\.[0-9]+)*/
     commands:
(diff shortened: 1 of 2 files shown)


alloy:
Did you not have to update the `files` entry of `package.json` as discussed in Slack yesterday?

l2succes:
We don't have to, turns out the `.npmignore` file works fine but I'll make the switch

alloy:
LGTM 👌

PR #135, opened 2017-04-25
Login form

l2succes:
Adds proper headers for artsy-passport to send correct status codes on XHR requests


xtina-starr:
Solid catch! LGTM, I think we can merge when 💚

src/apps/loyalty/containers/login/index.tsx
```diff
@@ -92,11 +92,17 @@ class Login extends React.Component<LoginProps, LoginState> {
   }
 
   handleEmailChange (e) {
-    this.setState({email: e.target.value})
+    this.setState({
+      email: e.target.value,
+      error: "",
```
alloy: Why set `error` to a blank string rather than `null`?

xtina-starr: @alloy, I didn't think of that. Do you prefer `null` over an empty string but because it makes more sense syntax wise? Don't they both evaluate to falsey?

alloy: It is true that they both are falsey, but I think that `null` better reflects that difference and the _lack_ of an error.

PR #138, opened 2017-04-26
Make artwork grid infinite scroll based on element, not document

broskoski:


src/components/artwork_grid.tsx
```diff
@@ -26,7 +27,9 @@ export class ArtworkGrid extends React.Component<Props, State> {
   }
 
   maybeLoadMore() {
-    if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight / 2) {
+    const threshold = window.innerHeight + window.scrollY
+    const el = ReactDOM.findDOMNode(this)
+    if (threshold >= el.clientHeight + el.scrollTop) {
```
alloy: Can you either add a comment that describes exactly when more content will be loaded (without having to parse the logic of the implementation) or rename the method to be self-documenting?

PR #162, opened 2017-05-16
@mzikherman - Fillwidth component

broskoski:
After a few false starts I'm starting to understand the react / graphql version of replicating @craigspaeth's jQuery filldwidth plugin. This will definitely be less generic than that, the reason being that we can make use of the `aspect_ratio` attribute that gets returned on artwork images (which makes calculating a row a bit easier, especially not having to wait for images to load.

This doesn't really do much atm, just getting some code up here.


broskoski:
Hm, looks like `Dimensions` screws with the type-check...that will be fun to figure out.

broskoski:
Looking like this so far:
(6 artworks)
screenshot 2017-05-17 12 10 20

(10 artworks)
screenshot 2017-05-17 12 09 36

tsconfig.json
```diff
@@ -9,7 +9,7 @@
         "allowSyntheticDefaultImports": true,
         "baseUrl": "src",
         "outDir": "dist",
-        "noUnusedLocals": true
+        "noUnusedLocals": false
```
alloy: Why is this change needed?

broskoski: We already check for it in esLint and as far as I can tell there is no way to disable on a per-line basis (https://github.com/artsy/reaction-force/pull/162/files#diff-b7c8554bab28defe2bb4f7c32a1e2d33R68)

alloy: We don’t use eslint but tslint, which I iirc doesn’t catch e.g. unused imports.

src/components/artwork/fillwidth.tsx
```diff
@@ -0,0 +1,147 @@
...
+    })
+
+    // If the total width difference is less than 0, it is larger than the container
+    // so we need to scale down. If not, scale up.
+    let dir = this.widthDiff(dimensions) < 0 ? -1 : 1
+
+    // Keep looping until we get an acceptable width difference
+    const list = new Array(999)
+
+    for (let i of list) { // eslint-disable-line no-unused-vars
```
alloy: Is the intention here to just loop indefinitely until `break` is called? If so, something like the following will get rid of the unused array and unused variable:

```js
let proceed = true
while (proceed) {
  // ...
}
```

broskoski: 👍

src/components/artwork/fillwidth.tsx
```diff
@@ -63,9 +63,7 @@ export class Fillwidth extends React.Component<Props, null> {
     let dir = this.widthDiff(dimensions) < 0 ? -1 : 1
 
     // Keep looping until we get an acceptable width difference
-    const list = new Array(999)
-
-    for (let i of list) { // eslint-disable-line no-unused-vars
+    while (true) {
```
alloy: Even better! I wasn’t sure what the rules were for nested loops (i.e. the `for` inside this `while`) 👍

broskoski:
This is ready to go, what say y'all?

alloy:
Let’s 👍

PR #164, opened 2017-05-18
Copy .env file for webpack config for storybook

yuki24:
It seems like a `.env` file (used to inject `METAPHYSICS_ENDPOINT` in `.storybook/webpack.config.js`) is missing in CircleCI and we need to copy `.env.oss` to `.env`.


alloy:
🎉 🌮

PR #168, opened 2017-05-30
Update styled components to v2

alloy:
Best to read the commits (and their messages).

- [x] Enable stable class names via babel plugin
- [x] Make rehydration fully work, we’re currently still generating [different markup](https://www.diffchecker.com/r6dOkXlL) on the client-side.
      screen shot 2017-05-30 at 22 41 18


src/assets/icons.tsx
```diff
@@ -100,4 +49,6 @@ const IconNames: { [i: string]: string; } = {
   "search" : "\ue62c",
 }
 
+export type IconName = keyof typeof IconNames
```
alloy: Came across this file and remembered this one weird trick ✨

l2succes: 🔥🔥🔥 that's exactly the trick I was looking for

src/components/helpers.tsx
```diff
@@ -18,7 +18,11 @@ export const block = (margin: number = 0) => {
 
 const sizes = theme.flexboxgrid.breakpoints
 
-export const media: any = Object.keys(sizes).reduce((accumulator, label) => {
+type Media = {
+  [S in keyof typeof sizes]: typeof css
+}
```
alloy: So, what’s up with this, right?

* First we get a `union type` that consist of the keys of [the `sizes` object](https://github.com/artsy/reaction-force/blob/afa0a8d3811a10cea11f2c0affe2bd13436f1061/src/assets/theme.ts#L12-L15) with `keyof typeof sizes`, which results in something like: `type SizeKeys = "xs" | "sm" | "md" | "lg"`.
* Then we create a `mapped type` by iterating over the union elements with the `S in ...` syntax to produce the keys of the `media` object and we type the values to be the same of the `css` function, which is the function we wrap internally and pass arguments verbatim.

The result is:

```ts
type Media = {
    xs: ThemedCssFunction<any>
    sm: ThemedCssFunction<any>
    md: ThemedCssFunction<any>
    lg: ThemedCssFunction<any>
}
```

alloy:
Done. Also did a bunch of dependency cleaning/upgrading.

src/components/helpers.tsx
l2succes: that `keyof typeof` trick is very useful to use object keys as an union type.
thanks for the detailed explanation

src/typings/global.d.ts
```diff
@@ -34,3 +35,13 @@ declare module "express" {
     locals: ArtsyResponseLocals
   }
 }
+
+declare module "styled-components" {
```
l2succes: By the way, the `ServerStyleSheet` type definition has been merged. https://github.com/styled-components/styled-components/pull/839 so we can remove this

alloy:
Oh great! Will do now.

alloy:
Ok, going to merge this.

PR #169, opened 2017-05-31
Update reaction-force exports

broskoski:
As discussed in https://github.com/artsy/force/pull/1492 , some more exports to make it easier to initialize reaction-force components in force

### DIFF
--- a/src/components/index.tsx
+++ b/src/components/index.tsx
@@ -1,20 +1,48 @@
 import Artwork from "./artwork"
 import ArtworkFilter from "./artwork_filter"
-import Buttons from "./buttons/index"
-import Modal from "./modal/modal"
-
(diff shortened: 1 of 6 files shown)


broskoski:
Actually, hold on merging, gonna feature creep this PR and add some quick styling updates to the gene components

src/components/index.tsx
```diff
@@ -1,21 +1,49 @@
...
+import * as Artsy from "./artsy"
+
+interface InitOptions {
+  user: User,
+  component: any,
+  domID: string,
+  queryConfig: any,
+}
+
+export function init(options: InitOptions) {
```
broskoski: This function is what we will use to initialize reaction-force components on the client-side.

alloy:
LGTM 👍

PR #170, opened 2017-05-31
Import React and ReactDOM properly

broskoski:


broskoski:
It still compiles, but to `react_dom_1.default.render` instead of `ReactDOM.render`.

alloy:
Ah gotcha.

PR #171, opened 2017-05-31
Add gene filter url params

broskoski:
So in clicking around the reaction-force gene page and comparing to the one currently in production, I realized the url param -> filter state was missing. I tried to accomplish this in storybooks, but something about that router eats any URL params that are set, hence the entirely new app. 

Also, the bundling settings are specific to the loyal app right now, I started to change this over but wasn't familiar enough with what is going on to feel confident to change things around there. This Gene filter app is not going to be used in production, just a clean slate to test on.

In terms of the URL params, I have it setting them properly on dropdown change, but will have to do a little work to refactor those filter components to receive outside props initially.

screenshot 2017-05-31 18 30 35


src/components/index.tsx
```diff
@@ -19,8 +19,10 @@ import * as React from "react"
 import * as ReactDOM from "react-dom"
 import * as Relay from "react-relay"
 
+import { configureUrlQuery } from "react-url-query"
 import { artsyNetworkLayer } from "../relay/config"
 import * as Artsy from "./artsy"
+import history from "./history"
```
alloy: What is this needed for and does it work when rendering server-side?

broskoski: Getting and setting browser history. Nope, client-side only.

alloy: Right right, but it’s not entirely clear to me _why_ you need access to the history?

In any case, loading the module at the file level and the way the module is defined means it’s always going to be making that history object. Is there some way to only create the history object _when a component really needs it_ (lazily) and _only_ in the browser?

broskoski: Sure, we can only use it in the places where we will actually need it.

alloy: Ace 🕺

alloy:
Still a WIP, right? Let me know if you want/need assistance with webpack faffing.

broskoski:
Still a WIP, but not for long! I'm debating on whether or not just to remove the app after getting the URL params stuff working.

broskoski:
I have this working but I'd like to refactor a bit to make it nicer and add some tests, the logic is a little crazy ATM. 

One unfortunate drawback to the way relay takes in params is that we have to set them outside the component to get the initial fetch with the right params without a refetch after the component mounts. I _think_ we can get around this by setting the params in the queryConfig.

See: 
gene-url-filter

broskoski:
Just confirmed that passing in params outside of the component into the `QueryConfig` prevents a subsequent fetch:

```javascript
init({
  component: components.Gene,
  user: null,
  domID: "app-container",
  queryConfig: new GeneQueryConfig({
    geneID: "conceptual-art",
    dimension_range: getURLParameter("dimension_range"),
    price_range: getURLParameter("price_range"),
    medium: getURLParameter("medium"),
  }),
})
```

Gonna add a few tests here and move on!

broskoski:
Mergin!

alloy:
@broskoski I don’t fully understand the issue about updating a query config yet, can you elaborate here/slack?

PR #175, opened 2017-06-14
Adds fair booth component

broskoski:
Just getting some code up!

screenshot 2017-06-14 17 38 39

**TODO:**
- [x] Add follow button
- [x] Fix profile subquery (getting maximum call stack exceeded currently)
- [x] Add pagination

### DIFF
--- a/src/components/fair/booth.tsx
+++ b/src/components/fair/booth.tsx
@@ -1,0 +1,33 @@
+import * as React from "react"
+import * as Relay from "react-relay"
+
+interface Props extends RelayProps, React.HTMLProps<FairBooth> {
+  show: any
+}
+
+export class FairBooth extends React.Component<Props, null> {
+  render() {
+    const { show } = this.props
+    return (
+      <div>
+        {show.name}
+      </div>
+    )
+  }
+}
+
+export default Relay.createContainer(FairBooth, {
+  fragments: {
+    show: () => Relay.QL`
+      fragment on Show {
+        name
+      }
+    `,
+  },
+})
+
+interface RelayProps {
+  show: {
+    name: string | null,
+  } | any
+}
--- a/src/components/fair/booth.tsx
+++ b/src/components/fair/booth.tsx
@@ -6,18 +6,48 @@ import Artworks from "../artwork_grid"
 import Button from "../buttons/ghost"
 import Text from "../text"
 
+import * as fonts from "../../assets/fonts"
+
 const PageSize = 10
 
 interface Props extends RelayProps, React.HTMLProps<FairBooth> {
   show: any
-  onContactGallery?: () => any,
+  relay?: any,
+  onContactGallery?: (showId: number) => any,
 }
 
 const LocationText = styled(Text)`
   display: block
 `
 
-export class FairBooth extends React.Component<Props, null> {
+interface State {
+  loading: boolean,
+}
+
+export class FairBooth extends React.Component<Props, State> {
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      loading: false,
+    }
+  }
+
+  loadMoreArtworks() {
+    const hasMore = this.props.show.artworks.pageInfo.hasNextPage
+    if (!this.state.loading && hasMore) {
+      this.setState({ loading: true }, () => {
+        this.props.relay.setVariables({
+          artworksSize: this.props.relay.variables.artworksSize + PageSize,
+        }, readyState => {
+          if (readyState.done) {
+            this.setState({ loading: false })
+          }
+        })
+      })
+    }
+  }
+
   render() {
     const { show, onContactGallery } = this.props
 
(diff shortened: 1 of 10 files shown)


alloy:
Call stack blowing up might be similar/same to https://github.com/facebook/relay/issues/1740#issuecomment-308164130 ?

src/components/fair/booth.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import Artworks from "../artwork_grid"
+import Button from "../buttons/ghost"
+import Text from "../text"
+
+const PageSize = 10
+
+interface Props extends RelayProps, React.HTMLProps<FairBooth> {
+  show: any
+  onContactGallery?: () => any,
```
alloy: Should call this callback with the partner ID / whatever else is needed by Force’s modal.

broskoski:
@alloy I think this is ready to merge, gonna move on to the show feed after!

alloy:
Will review tonight!

src/components/fair/booth.tsx
```diff
@@ -0,0 +1,179 @@
...
+import Button from "../buttons/ghost"
+import FollowButton from "../follow"
+import Text from "../text"
+
+import * as fonts from "../../assets/fonts"
+
+const PageSize = 8
+
+interface Props extends RelayProps, React.HTMLProps<FairBooth> {
+  show: any
```
alloy: Out of curiosity, what does this do? Does TS use the definition from `RelayProps` because that one is more restricted than this `any` or does this `any` override the definition from `RelayProps`?

src/components/fair/booth.tsx
```diff
@@ -0,0 +1,179 @@
...
+      <LocationText textSize="small" textStyle="secondary">
+        {show.location.display}
+      </LocationText>
+    )
+
+    const artworksLeft =  show.counts.artworks - relay.variables.artworksSize
+
+    const loadMoreButton = (
+      <LoadMoreContainer>
+        <LoadMoreButton onClick={() => this.loadMoreArtworks()}>
```
alloy: This could be just `<LoadMoreButton onClick={this.loadMoreArtworks}>`, no need to create a new function on each render. Does require you to bind `loadMoreArtworks` to `this` in the `constructor`, though.

src/components/fair/booth.tsx
broskoski: Ah good point, I usually add the normal props first and then end up adding `RelayProps` in later, so this is just a result of forgetting to remove this one. Pretty sure it gets overridden though, to answer your question.

src/components/fair/booth.tsx
broskoski: True, for whatever reason I find this more intuitive but I can bind it (can't really imagine a performance difference, but that might be the case). 👍

src/components/fair/booth.tsx
```diff
@@ -0,0 +1,179 @@
...
+              Contact Gallery
+            </Button>
+          </div>
+        </Header>
+        <Artworks
+          artworks={show.artworks}
+          columnCount={4}
+          itemMargin={40}
+        />
+        {show.artworks && show.artworks.pageInfo.hasNextPage && !this.state.loading && loadMoreButton}
```
alloy: How about only ever defining `loadMoreButton` if all these conditions are met? Then you can just always interpolate `loadMoreButton` here, which might communicate the overall intent a bit better too.

src/components/fair/booth.tsx
```diff
@@ -0,0 +1,179 @@
...
+    super(props)
+    this.state = {
+      loading: false,
+    }
+  }
+
+  loadMoreArtworks() {
+    const hasMore = this.props.show.artworks.pageInfo.hasNextPage
+    if (!this.state.loading && hasMore) {
+      this.setState({ loading: true }, () => {
```
alloy: Nice use of `setState` callback 👌

src/components/fair/booth.tsx
```diff
@@ -0,0 +1,179 @@
...
+            <PartnerLine>
+              <PartnerText textSize="small" textStyle="primary">
+                {show.partner.name}
+              </PartnerText>
+              <FollowButton type="profile" profile={show.partner.profile}/>
+            </PartnerLine>
+            {showLocation}
+          </div>
+          <div>
+            <Button onClick={() => onContactGallery(show.id)}>
```
alloy: You’re passing a string (`show.id`) here to the callback, however [the callback’s signature](https://github.com/artsy/reaction-force/pull/175/files#diff-12ccaa558d908f4178256ea8b8b75bacR17) is defined to take a `number`.

This was probably not raised by the compiler because your `Props` interface overrides the `show` field typings, so that answers [this question](https://github.com/artsy/reaction-force/pull/175/files#r123081031), I guess.

src/components/follow.tsx
```diff
@@ -11,35 +11,38 @@ import * as Artsy from "../components/artsy"
 const SIZE = 32
 
 interface Props extends RelayProps, React.HTMLProps<FollowButton>, Artsy.ContextProps {
-  artist: any
+  artist?: any
+  profile?: any
```
alloy: Overriding here too?

src/components/fair/booth.tsx
broskoski: Ah yes true, good catch.

src/components/fair/booth.tsx
alloy: True. It’s more of a style thing that just ensures that things don’t add up too much somewhere in the future.

src/components/fair/booth.tsx
alloy: Yup 😄

alloy:
Few minor things, otherwise LGTM 👌 Merge on 💚

PR #177, opened 2017-06-20
Add show feed component for fair filter

broskoski:
Getting some code up here, running into two errors: 1. The dreaded `Maximum call stack size exceeded` again, randomly. 2. `RelayQueryPath.getQuery(): Cannot generate accurate query for path with connection` when trying to load more artworks in an individual booth component.

### DIFF
--- a/src/components/__stories__/fair_booth.tsx
+++ b/src/components/__stories__/fair_booth.tsx
@@ -3,9 +3,11 @@ import * as React from "react"
 import * as Relay from "react-relay"
 
 import FairBooth from "../fair/booth"
+import ShowFeed from "../fair/feed"
 
 import * as Artsy from "../../components/artsy"
 import { artsyNetworkLayer } from "../../relay/config"
+import FairQueryConfig from "../../relay/queries/fair"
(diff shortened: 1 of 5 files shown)


alloy:
> The dreaded Maximum call stack size exceeded again, randomly.

Remind me, what did you end up doing last time? Hot patch your local installation of the Relay package?

broskoski:
It actually stopped happening in the middle of trying to debug it, its very mysterious to me.

alloy:
Nonetheless, I wanna be sure we don’t deploy code in Force that’s dependent on a fix being in Relay without us adding that to our fork.

broskoski:
Nope, turned out to be two components both requesting an `id` (or at least, removing that made the error go away). Still figuring out the second issue.

broskoski:
This is now in a mergeable state!
fair-feed

orta:
@mzikherman this got merged with my prettier PR, it should still get a review

alloy:
@orta @mzikherman Not sure I follow. Has this PR been reviewed properly?

PR #181, opened 2017-06-21
[Dev] Convert to using prettier

orta:
## What does prettier do?

Prettier takes your code and reprints it from scratch by taking into account the line length.

For example, take the following code:

```js
foo(arg1, arg2, arg3, arg4);
```

It fits in a single line so it's going to stay as is. However, we've all run into this situation:

```js
foo(reallyLongArg(), omgSoManyParameters(), IShouldRefactorThis(), isThereSeriouslyAnotherOne());
```

Suddenly our previous format for calling function breaks down because this is too long. Prettier is going to do the painstaking work of reprinting it like that for you:

```js
foo(
  reallyLongArg(),
  omgSoManyParameters(),
  IShouldRefactorThis(),
  isThereSeriouslyAnotherOne()
);
```

Prettier enforces a consistent code **style** (i.e. code formatting that won't affect the AST) across your entire codebase because it disregards the original styling[\*](#styling-footnote) by parsing it away and re-printing the parsed AST with its own rules that take the maximum line length
into account, wrapping code when necessary.

---

> taken from https://github.com/prettier/prettier#prettier

## What does this PR do?

* Adds [prettier](https://github.com/prettier/prettier#prettier), you can run prettier against the entire project with `yarn prettier-project`
* Adds the prettier extension as a recommended extension
* Sets vscode to automatically format on save
* Adds dev hooks for running prettier on pre-commit
* Adds a compiler check on a pre-push
* Removes some TSLint rules `"ter-indent": [true, 2],` and `"jsx-wrap-multiline": [ true, "always" ],` as they conflicted with prettier output
* Added [tslint-config-prettier](https://github.com/alexjoverm/tslint-config-prettier) which disables any tslint rules 

Similar to https://github.com/danger/danger-js/pull/277, https://github.com/artsy/emission/pull/567 and https://github.com/artsy/prediction/pull/455.

/cc @broskoski  @l2succes @alloy @xtina-starr @eessex @mzikherman @yuki24


orta:
This PR  is blocked by https://github.com/artsy/reaction-force/pull/177 as it merges and builds on it

orta:
This _does not_ add it as a CI step. I still think the way prettier handles this is a bit confusing, so I'm not gonna add it until that's cleared up.

src/components/artwork/fillwidth_item.tsx
```diff
@@ -83,11 +80,11 @@ export default Relay.createContainer(StyledFillwidthItem, {
...
   artwork: {
-    href: string | null,
+    href: string | null
     image: {
-      placeholder: number | null,
-      url: string | null,
-      aspect_ratio: number | null,
-    } | null,
-  },
+      placeholder: number | null
```
damassi: Oh no! It seems like Prettier is stripping the commas off of interfaces

orta: Yeah, we had that discussion in the Emission PR, we're good with this 👍 

It doesn't change behaviour, Microsoft use semi-colons here.

orta:
lols @broskoski left today for reels this time, reassigning

yuki24:
LTGM👍

.vscode/settings.json
```diff
@@ -10,5 +10,14 @@
...
+    "editor.tabSize": 2,
+    "editor.formatOnSave": true,
+    "prettier.singleQuote": false,
+    "prettier.trailingComma": "es5",
+    "prettier.typescriptEnable": [
+        "typescript",
+        "typescriptreact"
+    ],
+    "prettier.semi": false,
+    "prettier.printWidth": 120
```
damassi: On https://github.com/prettier/prettier#options they emphatically recommend that users stick to `80` since Prettier (very unfortunately, IMO) has no concept of vertical space until it hits the width boundary. E.g.,

```javascript
  const {
    foo: {
      bar: {
        baz
      }
    },
    foo: {
      bar: {
        baz
      }
    },
    foo: {
      bar: {
        baz
      }
    },
    foo: {
      bar: {
        baz
      }
    },
  } = props
```
is collapsed into 
```javascript
  const { foo: { bar: { baz } }, foo: { bar: { baz } }, foo: { bar: { baz } }, foo: { bar: { baz } } } = props
```
(110 chars long)

Since we're dealing with React and propTypes where this sort of thing is common, this can get very unreadable very fast. I really want to argue keeping this value at 80 or max 90 if possible...

orta: Personally: I'd rather have awkward interfaces than awkward code reviews, 80 columns is so hard to write JS to, function definitions get really awkward across many lines as we have the types which make them much longer

damassi: Sounds good 👍  If anything ever gets out of hand its as easy as updating settings and rerunning over codebase.

orta: EZ - yeah 👍

PR #183, opened 2017-06-21
[Dev] Adds a getting started guide

orta:
Based on my discussions with @eessex on porting a JS component into RF.

@anandaroop please read this before we pair tomorrow 👍

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -1,0 +1,193 @@
+# How do I work in Reaction Force?
+
+We're going to take a component, and make it fit inside Reaction Force. This doc is based 
+on [this PR](https://github.com/artsy/reaction-force/pull/176).
+
+## Simple
+
+```jsx
+import React, { Component } from "react"
+
+class IconImageSet extends Component {
+  render() {
+    return (
+      <svg id="imageset" data-name="imageset" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 44 44">
+        <path d="M6760.67,551.72h-32v-32h32v32Zm-30-2h28v-28h-28v28Z" transform="translate(-6728.67 -507.72)" />
+        <polygon points="38 37 36 37 36 8 7 8 7 6 38 6 38 37" />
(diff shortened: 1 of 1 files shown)


broskoski:
Very nice, I can add some relay tips to this as well.

docs/getting_started.md
```diff
@@ -0,0 +1,191 @@
...
+    )
+  }
+}
+export default IconImageSet
+```
+
+This is a component which returns an SVG. To move this into Reaction force, you'll need to find a place to put it. 
+
+It should go somewhere inside `src/components/` - but it will need to be a `.tsx` file. So for now, let's imagine it is
+in `src/components/icons/icon_image_set.tsx`. Make the file and paste the contents in in.
```
damassi: Do you think we should enforce / encourage `components/CamelCase.tsx`? Since Relay 2 now requires it it seems like a good habit to get into.

alloy: One less ‘in’

alloy: I agree with @damassi that we’ll want that, but maybe we should start enforcing that once we update the whole code base and migrate this repo to Relay Modern?

docs/getting_started.md
```diff
@@ -0,0 +1,191 @@
...
+```ts
+interface IconProps {
+  width: number
+  height: number
+  color: string
+}
+
+interface IconState {
+  appliedColors: boolean
+}
```
alloy: We’ve been calling these interfaces simply ‘Props’ and ‘State’ usually.

docs/getting_started.md
```diff
@@ -0,0 +1,191 @@
...
+
+class IconImageSet extends Component<IconProps, IconState> {
+```
+
+If you don't have state at all, write `null` instead.
```
alloy: I’ve started using `void`, which I’ve picked up from various other sources.

orta:
Updated docs, happy to have @broskoski add some relay notes too ( might even be worth it's own doc?)

Agree re: relay modern and CamelCasing, but right now that'd look really inconsistent with the rest of the app - so I'm happy to wait on that one 👍

PR #184, opened 2017-06-23
[Dev] Adds Danger

orta:
This is going to be working off our [staging peril](https://dashboard.heroku.com/apps/artsy-peril) instance, instead of running on CI. So it might take a bit of setup.

You can see the peril settings here: https://github.com/artsy/artsy-danger/blob/master/peril.settings.json

If you want to understand how all this works, see:  https://github.com/danger/peril/blob/master/docs/setup_for_org.md


orta:
Going to self-merge this, and retain the fail from Peril on purpose - for archival reasons 👍

PR #185, opened 2017-06-28
Update to Artsy passport 2.x - Removes CoffeeScript

orta:
This is the result of https://github.com/artsy/artsy-passport/pull/112 https://github.com/artsy/artsy-passport/pull/113 and https://github.com/artsy/artsy-passport/pull/114

It removes the only CoffeeScript dependency in reaction-force, and thus kills the need for the language inside the app.

Interestingly in doing so, TypeScript raised a compiler error because the typings could then be inferred . However, as `options` is added [dynamically](https://github.com/artsy/artsy-passport/blob/master/lib/index.coffee#L18) then it isn't seen. So I any'd it. If people care enough I can add types to the lib _ooooorrrrr_ someone else can, and I'll pair with them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
   "dependencies": {
     "@storybook/addon-options": "^3.1.2",
     "@storybook/react": "^3.1.3",
-    "artsy-passport": "^1.6.1",
+    "artsy-passport": "^2.0.3",
     "artsy-xapp": "^1.0.4",
     "babel-polyfill": "^6.23.0",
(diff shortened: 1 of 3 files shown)


eessex:
Goodbye ☕️  🎉

alloy:
Excellent 👌 

>  If people care enough I can add types to the lib ooooorrrrr someone else can, and I'll pair with them.

Yeah, this would be a great #first-step issue for others to pick-up 👍

PR #186, opened 2017-06-28
[Publishing] Adds image component and font system

kanaabe:
Thanks @eessex for pairing on this & getting me started in reaction-force! Main thing I changed since the pair sesh is how the fonts work. I tried making it feel more similar to Force/Positron so whenever a font needs to be used, we can `${Fonts.garamond("s15")}` in the styled component.

This screenshot doesn't show the new font since it's not in my system yet, but the fallback is good old Arial. 

image


src/components/publishing/artwork.tsx
```diff
@@ -1,14 +1,14 @@
 import React, { Component } from "react"
 import styled from "styled-components"
 import TextLink from "../text_link"
+import Fonts from "./fonts"
 
 const TruncatedLine = styled.div`
+  ${Fonts.garamond("s15")}
   display: block;
```
eessex: this is great!

src/components/publishing/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+    color: #999;
+  }
+`
+
+const CaptionLink = styled.div`
+  ${Fonts.unica("s14")}
+  margin: 0;
+  border-bottom: 1px solid black;
+  cursor: pointer;
+  display: inline-block;
```
eessex: might be a good idea to add padding left here, so that long captions don't sit too close

src/components/publishing/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const CaptionContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  margin: 10px 0 10px 0;
+`
+
+const BlockImage = styled.img`
+  display: block;
+`
```
eessex: what does the display do here?

orta: `img` is typically an inline element, I think, so this tells it to "move to a new line"

src/components/publishing/fonts.tsx
```diff
@@ -0,0 +1,110 @@
...
+  'Unica77LLWebRegular',
+  'Arial',
+  'serif'
+`
+
+const unica = size => {
+  const evaluatedSize = textSizesForUnica[size]
+  const style = css`
+    font-family: ${unicaFontFamily};
+    -webkit-font-smoothing: antialiased;
```
alloy: Are you prefixing this yourself because you want it to _only_ apply to webkit? Otherwise styled-components should auto-prefix it.

src/components/publishing/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+class Image extends Component<any, null> {
```
alloy: Prefer `void` over `null` for stateless components.

src/components/publishing/artwork.tsx
alloy: Agreed ✨

src/components/publishing/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+  render() {
+    const { image } = this.props
+    return (
+      <div className="article-image">
+        <BlockImage src={image.url} width={"100%"} />
```
alloy: The curly brackets aren’t strictly needed when specifying a static string.

src/components/publishing/image.tsx
alloy: Btw, in the case of stateless components you can also choose for just [a function](https://facebook.github.io/react/docs/components-and-props.html#functional-and-class-components).

alloy:
No blockers from my side, looks great! 🚀

src/components/publishing/image.tsx
kanaabe: What he said 😺 . Without it the inline version gives us a ghost space of something like 3px

src/components/publishing/fonts.tsx
kanaabe: Hmm interesting point. I was mostly following https://github.com/artsy/reaction-force/blob/master/src/assets/fonts.tsx#L31 but it looks like the prefix doesn't get applied if I use `font-smooth: antialiased` or `font-smoothing: antialiased`.

kanaabe:
Thanks for the comments! I simplified the image component to a function (good call this will likely never have a state) and added a long caption image to the storybook. Will merge when 🍏

src/components/publishing/fonts.tsx
alloy: Oh yeah indeed, I remember asking @l2succes the same thing :D

PR #188, opened 2017-06-29
[WIP] Add <MasonryGrid /> component

damassi:
(WIP) Refactor of the `<ArtworkGrid />` component. I needed a react-based masonry component for some responsive work I'm doing in Force so I took the existing RF-based `<ArtworkGrid />` component and abstracted it a bit to accept dynamic content. Wanted to open the PR now so that I don't forget. 

**Usage Example:**

```javascript
render() {
   return (
     <MasonryGrid
        items={artworks}
        getAspectRatio={(artwork) => {
          return get(artwork, 'images.0.aspect_ratio')
        }}
        getDisplayComponent={(artwork) => {
          const artworkImage = get(artwork, 'images.0.image_url', '/images/missing_image.png')

          return (
            <img
              src={artworkImage}
              style={{
                width: '100%'
              }}
            />
          )
        }}
      />
  )
}
```

**TODO:**
- Port back over to formal TypeScript
- Update references 
- Add tests

### DIFF
--- a/src/components/masonry_grid.tsx
+++ b/src/components/masonry_grid.tsx
@@ -1,0 +1,137 @@
+import { times } from "lodash"
+import PropTypes from "prop-types"
+import * as React from "react"
+
+interface Props extends React.HTMLProps<MasonryGrid> {
+  columnCount?: number
+  columnMargin?: number
+  getDisplayComponent: (artwork: object) => any
+  getAspectRatio: (artwork: object) => number
+  items: Array<object>
+  rowMargin?: number
+}
+
+export default class MasonryGrid extends React.Component<Props, void> {
+  static propTypes = {
+    columnCount: PropTypes.number,
+    columnMargin: PropTypes.number,
+    getDisplayComponent: PropTypes.func.isRequired,
+    getAspectRatio: PropTypes.func.isRequired,
+    items: PropTypes.array.isRequired,
+    rowMargin: PropTypes.number,
+  }
+
+  static defaultProps = {
+    columnCount: 3,
+    columnMargin: 20,
+    rowMargin: 20,
+  }
+
+  createGrid() {
+    const { items, columnCount, getAspectRatio } = this.props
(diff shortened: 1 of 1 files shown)


alloy:
Love that API

damassi:
For the brief few moments I was in TypeScript land trying to appease Circle it really felt like the future. Reaction Force is nice! Excited to be working in it more in the coming months.

src/components/masonry_grid.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+export default class MasonryGrid extends React.Component<Props, void> {
+  static propTypes = {
+    columnCount: PropTypes.number,
+    columnMargin: PropTypes.number,
+    getDisplayComponent: PropTypes.func.isRequired,
+    getAspectRatio: PropTypes.func.isRequired,
+    items: PropTypes.array.isRequired,
+    rowMargin: PropTypes.number,
+  }
```
alloy: Ah yeah, runtime assertions will be required again for safety when integrating into a vanilla JS code base.

orta: good call, maybe worth a danger rule

damassi: `PropTypes` can be a bit grueling to type out on every component but time and again they're helpful catching stuff during dev / runtime. In Force they're enforced by eslint.

alloy: Not sure I really like loosing the TS type syntax in favour of this API, but it might be an option https://github.com/gcanti/prop-types-ts

orta: is there a way we can them outselves during the compilation cycle?

orta: need us some typescript plugins ;)I

damassi:
Actually, I'm going to close this. I just remembered a recent convo with @katarinabatina about how the asymmetrical "masonry" look is going to be phased out.

PR #192, opened 2017-07-06
@kanaabe => Publishing typography story

eessex:
- Adds a story/style-guide for typography
- Changes size to font-size in font declarations
- Uses new font-format for imageset
- Adds min-width to image 'view fullscreen' button to prevent line-breaks

screen shot 2017-07-06 at 11 13 10 am


kanaabe:
👍  Looks great! Heads up that in my other PR I intro'd a second arg to unica to configure font families so this story may change slightly. 

Would you mind adding a test for this story?

PR #199, opened 2017-07-13
@alloy => Image collection component + fillwidth utility

kanaabe:
WIP because I'm working thru a few opportunities to clean things up / add more types.

This extracts the math-y logic from the fillwidth component in `/components/artwork/fillwidth.tsx` and puts it into a util folder. The new `fillwidthDimensions` function does the same thing as `getDimensions` (returns an array of image sizes and ids given the target heights/widths/gutter) but now it also accepts other kinds of items like publishing images + artworks.

Then, this `fillwidthDimensions` gets used to create a new component called `ImageCollection` which is a mix of `Image` and `Artwork`

Image Collection fillwidth
image

Regular fillwidth stays the same
image


src/components/publishing/image_collection_item.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from "react"
+import styled from "styled-components"
+
+interface FillwidthItemProps extends React.HTMLProps<HTMLDivElement> {
+  width?: any
```
l2succes: you could set `width` `height` and `margin` to type `number` instead of `any`

src/utils/fillwidth.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const fillwidthDimensions = (items, containerWidth, gutter, targetHeight) => {
+  /**
+   * Scales an image object proportionally based on a direction (either -1 or 1)
+   * @param img a dimension object that references an artwork image
+   * @param dir the direction we need to scale an image, either -1 or 1
+   */
+  const resizeHeight = (img, dir) => {
+    img.width += img.width / img.height * dir
+    img.height += dir
```
l2succes: I would return a new object with the new sizes instead of mutating `img` directly.
so
```javascript
return {
  width: img.width + img.width / img.height * dir,
  height: img.height + dir
}
```

src/components/publishing/image_collection.tsx
```diff
@@ -0,0 +1,52 @@
...
+        height={imageSize.height}
+      >
+        {renderedImage}
+      </ImageCollectionItem>
+    )
+  })
+  return renderedImages
+}
+
+function ImageCollection(props) {
```
l2succes: I know it's a WIP but an interface for props here would be nice :)

src/components/publishing/image_collection_item.tsx
kanaabe: yess, sorry i have issues with letting go

src/components/publishing/image_collection.tsx
kanaabe: Definitely! coming

src/utils/fillwidth.tsx
kanaabe: Looking over this again (this was copied over from `artwork/fillwidth.tsx`) -- I _think_ it might be ok to mutate here. The fillwidth starts by creating a `dimensions` array with a best guess of width/height and the point of `resizeHeight` is for the images in `dimensions` to keep adjusting themselves until an acceptable difference is found in a while loop. If we don't mutate, we'd have to also reconstruct the `dimensions` array for each adjustment.  

```
  // Keep looping until we get an acceptable width difference
  while (true) {
    for (let img of dimensions) {
      resizeHeight(img, dir)
      if (widthDiff(dimensions) > 1) {
        break
      }
    }
    if (widthDiff(dimensions) > 1) {
      break
    }
  }
```

src/components/publishing/image_collection_item.tsx
alloy: Hahaha. To be fair, we did tell @kanaabe and @eessex that they don’t need to add typings immediately.

kanaabe:
@alloy this is ready to be reviewed now! 🙏

src/components/publishing/image_collection_item.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface FillwidthItemProps extends React.HTMLProps<HTMLDivElement> {
+  width?: number
+  height?: number
+  margin?: number
+  key: string
+}
+
+const FillwidthItem: React.SFC<FillwidthItemProps> = props => {
+  const { key, ...rest } = Object.assign({}, props)
+  return <div {...rest} key={key} />
```
alloy: I’m not quite sure I understand what’s happening here. Isn’t this all essentially equal to the following?

```js
return <div {...props} />
```

src/utils/fillwidth.tsx
alloy: I agree, `resizeHeight` is a function private to the scope of `fillwidthDimensions`, so it’s fair to make assumptions about being able to mutate the input.

alloy: I _am_ slightly surprised the linter didn’t complain about it, though.

alloy:
Another thing I’m surprised about is that the linter didn't complain about using `let` when `const` could be used 🤔

alloy:
Oh and a file that has no `JSX` can be named `.ts` instead of `.tsx`. I’m not sure if there’s any benefits in using one over the other, though, so I don’t feel strongly about it 🤷‍♂️

src/components/publishing/image_collection_item.tsx
kanaabe: Ah yes sorry I forgot to clean this up. I remember this was a sad attempt for me to figure out why I kept getting errors like:
```
Warning: Unknown prop `margin` on <div> tag. Remove this prop from the element.
```
even though I defined it in my prop definition.

What I didn't realize is that `<div>` itself isn't going to get the definition but props will be passed to it, so the way to write this instead is to not define a React.SFC at all, and just pass the proptypes to the styled component:
```
export default styled.div`       // previously styled(FillwidthItem)
 margin-right: ${(props: FillwidthItemProps) => props.margin + "px" || 0};
 width: ${(props: FillwidthItemProps) => props.width + "px" || "auto"};
 height: ${(props: FillwidthItemProps) => props.height + "px" || "auto"};
`
```
Does that make sense? Still kinda figuring out this stuff but this seems to work nicely.

alloy: Ah gotcha, yeah that makes sense 👍 

if you want typings it gets a bit hairy, something like the following should do it:

```ts
import styled, { StyledFunction } from "styled-components"

interface FillwidthItemProps {
  width?: number
  height?: number
  margin?: number
  key: string
}

/**
 * Need to type the function first so that props are typed correctly. Alas `styled.div<P>` doesn’t work at the moment
 * due TypeScript not (yet) supporting generic on template string functions.
 * 
 * See:
 * - https://github.com/styled-components/styled-components/issues/630 
 * - https://github.com/Microsoft/TypeScript/issues/11947
 */
const div: StyledFunction<FillwidthItemProps & React.HTMLProps<HTMLDivElement>> = styled.div

export default div`
  margin-right: ${props => props.margin + "px" || 0};
  width: ${props => props.width + "px" || "auto"};
  height: ${props => props.height + "px" || "auto"};
`
```

alloy: I guess the `key` prop can be dropped too? That should already be in `React.HTMLProps`.

alloy: Btw, if `props.margin` is not defined in `props.margin + "px" || 0`, the result would be `"undefinedpx"`, not `0`. I assume that’s not the intention, perhaps `props.margin && props.margin + "px" || 0` is?

alloy: …or `props.margin || 0 + "px"`

alloy:
Looks good 👌

PR #200, opened 2017-07-17
@alloy => Image set previews

eessex:
- Adds imagesetPreview component for new article layouts
- Renames old image-set preview to 'imagesetPreviewClassic'
- [minor] bumps up min-width for 'view fullscreen' link in images

screen shot 2017-07-17 at 1 42 09 pm


src/components/publishing/__test__/imageset_preview.test.tsx
```diff
@@ -1,10 +1,14 @@
...
-import { Images } from "./fixtures"
+import { ImageSetFull, ImageSetMini } from "./fixtures"
 
 import * as React from "react"
 import * as renderer from "react-test-renderer"
 
-it("renders properly", () => {
-  const imageset = renderer.create(<ImagesetPreview images={Images} />).toJSON()
+it("renders a full image-set properly", () => {
+  const imageset = renderer.create(<ImagesetPreview section={ImageSetFull} />).toJSON()
```
alloy: Is ‘imageset’ 1 word? I would _think_ it’s 2 words ‘image-set’ and thus the fixture name is more correct than the component name. What say you?

src/components/publishing/__test__/imageset_preview.test.tsx
```diff
@@ -1,10 +1,14 @@
...
+  expect(imageset).toMatchSnapshot()
+})
+it("renders a mini image-set properly", () => {
```
alloy: Oh, here’s my answer :D

src/components/publishing/imageset_preview.tsx
```diff
@@ -1,105 +1,131 @@
...
-      position: "absolute",
-      left: 8,
-      top: 14,
-    }
-    if (images.length > 9) {
-      length.left = 4
-    }
-
+    const { layout } = this.props.section
+    const image = layout === "full" ? <img src={this.getImageUrl()} width="100%" /> : ""
```
alloy: Do you really need an empty string if `layout === "mini`? If not, the code would be more portable to React Native if you’d use `null` instead. (React Native does not accept an empty string without explicit `Text` tag.)

src/components/publishing/imageset_preview.tsx
```diff
@@ -1,105 +1,131 @@
...
-  constructor(props) {
-    super(props)
-    this.state = {
-      visibleImages: this.getVisibleImages(this.props.images),
+  getImageUrl() {
+    const image = this.props.section.images[0]
+    const src = image.url ? image.url : image.image
+    return src
+  }
+  textSection(layout) {
```
alloy: Why does this method and `image` take `layout` as a param rather than pulling it out of the `props` themselves? For testability?

src/components/publishing/imageset_preview_classic.tsx
```diff
@@ -0,0 +1,106 @@
...
+  getVisibleImages(images) {
+    let widths = []
+    let hidden = 0
+    images.map((item, i) => {
+      const adjustedWidth = 150 * item.width / item.height
+      widths.push(adjustedWidth)
+      const total = widths.reduce((a, b) => a + b, 0)
+      const margins = widths.length * 10
+      if (total + margins + 50 > 560) {
+        hidden = hidden + 1
```
alloy: Would be nice to move all these magic numbers to the top of the file as self-describing constants.

src/components/publishing/imageset_preview_classic.tsx
```diff
@@ -0,0 +1,106 @@
...
+      }
+    })
+    return widths.length - hidden
+  }
+
+  renderImages(images) {
+    const items = images.slice(0, 4).map((item, i) => {
+      const src = item.image ? item.image : item.url
```
alloy: You could shorten this and line 66 into `const src = item.image || item.url || ""`

eessex:
Thanks for these comments, will make these changes now -- regarding the 'classic' naming, we will be deprecating the classic component in favor of the other in September, when we launch new article layouts, but it is currently used in production and we'll need it around until the switch flips. 

Also yes -- totally down to hear type-script specific suggestions!

alloy:
> regarding the 'classic' naming, we will be deprecating the classic component in favor of the other in September, when we launch new article layouts, but it is currently used in production and we'll need it around until the switch flips.

Gotcha, thanks for filling me in 👍

src/components/publishing/imageset_preview.tsx
```diff
@@ -1,106 +1,131 @@
...
 `
 
-class ImagesetPreview extends Component<any, any> {
-  constructor(props) {
-    super(props)
-    this.state = {
-      visibleImages: this.getVisibleImages(this.props.images),
+class ImageSetPreview extends Component<any, any> {
+  getImageUrl() {
+    const image = this.props.section.images[0]
```
alloy: Why does this component accept a list of images but only ever show the first?

eessex: Thought it would be easier to implement if passing the entire 'section' rather than extracting the title, layout and images as separate props.

src/components/publishing/imageset_preview.tsx
```diff
@@ -1,106 +1,131 @@
...
+    height: 98%;
+  }
 `
 
-class ImagesetPreview extends Component<any, any> {
-  constructor(props) {
-    super(props)
-    this.state = {
-      visibleImages: this.getVisibleImages(this.props.images),
+class ImageSetPreview extends Component<any, any> {
```
kanaabe: Not necessary but it would be good practice to get some typings here instead of `<any, any>`. Also, I think this doesn't have a `state` so probably good to pass `void`, so something like `<ImageSetPreviewProps, void>`

alloy: Typing the props would help the user of the component know how to use it more easily, especially with the `section.layout` prop, because it’s only 1 of 2 static strings and additionally they’d be told that the prop is required to specify. E.g.

```
interface Props extends {
  section: {
    layout: "mini" | "full"
    images: {
      url?: string
      image?: string
    }
  }
}

class ImageSetPreview extends Component<Props,null> {
  // ...
}
```

In the above example I only indicated with the second type parameter to `Component` that this class holds no state.

alloy: Note that if this component gets used directly from vanilla JS, then the only way to add safety is still to add runtime `PropTypes`.

alloy: Good call on the `void`. That is indeed what I had said before and what we have in the README, I think, but after recently discussing it with some people, the conclusion was that `null` is the most accurate, as that’s what the value _actually_ is at runtime.

src/components/publishing/imageset_preview.tsx
alloy: Ahhh, gotcha. Still not entirely clear why a section always has just 1 image, but passing in the whole thing makes sense to me.

eessex: There is an expanded view, yet to be created that will use all the images.

src/components/publishing/imageset_preview.tsx
```diff
@@ -1,106 +1,131 @@
...
+const TitleWrapper = styled.div`
   display: flex;
   flex-direction: column;
-  justify-content: center;
-  align-items: center;
-  border: 1px solid #e5e5e5;
-  text-align: center;
+  justify-content: space-between;
+  height: 100%;
+  &[data-layout='mini'] {
```
kanaabe: Like we talked about earlier today, it's definitely fine to use `data-layout` but if we're already using `this.props.layout` to make rendering decisions, it might make sense to let `TitleWrapper` accept the `layout` prop and use that to manipulate CSS. [Here](https://github.com/artsy/reaction/pull/199#discussion_r127554161) is the link to the helpful comment from @alloy.

src/components/publishing/imageset_preview.tsx
eessex: I'm having some issues getting and array of objects for the images prop -- when done like above it doesn't provide in array.  I tried wrapping it in an array but got some errors in the console (though the storybook is working and no errors in vscode):

When wrapping the image object in an array, 
`Type '{ section: { type: string; layout: string; title: string; images: ({ type: string; id: string; sl...' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<ImageSetPreview> & Readonly<{ children?: ReactNode...'.
  Type '{ section: { type: string; layout: string; title: string; images: ({ type: string; id: string; sl...' is not assignable to type 'Readonly<ImageSetPreviewProps>'.
    Types of property 'section' are incompatible.
      Type '{ type: string; layout: string; title: string; images: ({ type: string; id: string; slug: string;...' is not assignable to type '{ images: [{ name?: "string"; url?: "string"; image?: "string"; }]; layout: "full" | "mini"; titl...'.`

alloy: Oh my bad, I forgot to add the array specifier, sorry 😞 I’ve updated my example above.

With TS there’s 2 ways to describe an array. Either with `Array<T>` or with `T[]`. I personally prefer the former when the description of `T` spans multiple lines (like the above `images` example) and the latter when it’s a single line. E.g. an array of strings would be `string[]`.

eessex:
Turns out it was the `layout` field that was actually holding it up -- this is ready to go now, thanks for all the typescript help!

alloy:
Awesome!

This needs a rebase, though 👍

eessex:
Rebasing after merging conflicts earlier on has made this into a bit of a mess -- closing this PR in favor of a new one that i'll open momentarily.

PR #201, opened 2017-07-17
[doc] Use `null` for state instead of `void`.

alloy:
https://github.com/artsy/reaction/pull/200#discussion_r127812738

The actual value of `this.state` at runtime is `null`, so that’s more accurate.


kanaabe:
Nooice, thanks for the diligence 👍

PR #202, opened 2017-07-17
@eessex => [Publishing] Refactor Artwork and Image components and support "classic"

kanaabe:
Refactors the Artwork and Image components to be more Typescript-friendly and adds in support for the "classic" views.  

imageartwork

### DIFF
--- a/src/components/publishing/__stories__/publishing.tsx
+++ b/src/components/publishing/__stories__/publishing.tsx
@@ -17,12 +17,12 @@ storiesOf("Publishing", Artwork)
     return (
       <div>
         <div style={{ width: 800 }}>
-          <Artwork linked artwork={Artworks[0]} />
+          <Artwork artwork={Artworks[0]} />
         </div>
         <hr />
         <p>Multiple Artists: </p>
         <div style={{ width: 800 }}>
-          <Artwork linked artwork={Artworks[1]} />
+          <Artwork artwork={Artworks[1]} />
         </div>
         <hr />
         <p>Unlinked: </p>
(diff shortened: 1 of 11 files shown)


src/components/publishing/__stories__/publishing.tsx
```diff
@@ -8,15 +8,37 @@ import Image from "../image"
...
+        <hr />
+        <p>Multiple Artists: </p>
+        <div style={{ width: 800 }}>
+          <Artwork linked artwork={Artworks[1]} />
+        </div>
+        <hr />
+        <p>Unlinked: </p>
+        <div style={{ width: 800 }}>
+          <Artwork linked={false} artwork={Artworks[1]} />
+        </div>
```
eessex: Does 'linked' need to be explicitly declared because of typescript?

src/components/publishing/image_caption.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react"
+import styled, { StyledFunction } from "styled-components"
+import Fonts from "./fonts"
+import ViewFullscreen from "./view_fullscreen"
+
```
eessex: 👍 to extracting this

src/components/publishing/__stories__/publishing.tsx
kanaabe: nope! but that brings up a good point which is that our default should probably be `true` for `linked`.

kanaabe:
@eessex is this ok to merge? Changed the default behavior of `layout` to be `linked: true`

PR #204, opened 2017-07-21
@eessex => [Publishing] Responsive Header (Standard + Feature)

kanaabe:
`<Header/>` : Displays `<FeatureHeader/>` or `<StandardHeader/>`

FeatureHeader: Has `split`, `fullscreen`, and `text` layouts
StandardHeader: Has one layout
AuthorDate: Used in FeatureHeader and StandardHeader
Adds publishing breakpoints

header2


src/components/publishing/header/author_date.tsx
```diff
@@ -0,0 +1,80 @@
...
+      return str
+    } else {
+      return authors[0].name
+    }
+  } else {
+    return "Artsy Editorial"
+  }
+}
+
+const getDate = date => moment(date).tz("America/New_York").format("MMM D, YYYY h:mm a")
```
alloy: Why does this hardcode NY’s TZ instead of the user’s TZ?

alloy:
There are couple of more complex functions that probably have code paths that are not covered by the snapshot tests (e.g. the byline function), might be nice to have some unit tests for those. Other than that this looks good to me :ok_hand:

src/components/publishing/header/author_date.tsx
kanaabe: Yea.. that's apparently what editorial wants now. Fwiw, this only affects editorial articles, since only editorial can have feature/standard headers.

src/assets/theme.ts
```diff
@@ -15,4 +15,11 @@ export default {
       lg: 75, // em
     },
   },
+  publishing: {
+    breakpoints: {
+      sm: 600, // px
+      md: 900, // px
+      lg: 1250, // px
+    },
```
eessex: 👍

src/components/publishing/header/feature_header.tsx
```diff
@@ -0,0 +1,261 @@
...
+interface FeatureHeaderProps {
+  article?: any
+  size?: any
+}
+
+const FeatureHeader: React.SFC<FeatureHeaderProps> = props => {
+  const { article } = props
+  const { width } = props.size
+  const hero = article.hero_section
+  return (
```
eessex: Realized when starting to implement image-set in Positron that passing larger objects doesn't account for Backbone's attributes.  Hadn't considered that we will need to convert our articles to json before passing in. (this is maybe more of a problem for positron than force)

eessex:
This looks good to me when 🍏

src/components/publishing/header/author_date.tsx
alloy: Ah, is it “the local TZ of the place the article was published from”, but just hardcoded to by NY?

kanaabe: Yup!

alloy: Makes sense 👍

PR #206, opened 2017-07-24
@kanaabe => Imageset updates

eessex:
- Uses mini layout by default, and does not require layout prop to be passed in
- Does not require title prop, use alternate layout when title not included

screen shot 2017-07-24 at 1 23 42 pm

### DIFF
--- a/src/components/publishing/__test__/fixtures.tsx
+++ b/src/components/publishing/__test__/fixtures.tsx
@@ -93,8 +93,18 @@ export const ImageSetFull: ImageSetPreviewProps["section"] = {
 
 export const ImageSetMini: ImageSetPreviewProps["section"] = {
   type: "image_set",
-  layout: "mini",
   title: "The Work of Bruce M. Sherman",
+  images: Images,
+}
+
+export const ImageSetFullSansTitle: ImageSetPreviewProps["section"] = {
+  type: "image_set",
+  layout: "full",
+  images: Images,
+}
+
+export const ImageSetMiniSansTitle: ImageSetPreviewProps["section"] = {
+  type: "image_set",
   images: Images,
 }
 
--- a/src/components/publishing/imageset_preview.tsx
+++ b/src/components/publishing/imageset_preview.tsx
@@ -7,7 +7,7 @@ import IconImageSet from "./icons/icon_imageset"
 type Layout = "mini" | "full"
 
 interface DivLayoutProps {
-  layout: Layout
+  layout?: Layout
 }
 const div: StyledFunction<DivLayoutProps & React.HTMLProps<HTMLDivElement>> = styled.div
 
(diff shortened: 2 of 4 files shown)


src/components/publishing/__test__/fixtures.tsx
```diff
@@ -93,11 +93,21 @@ export const ImageSetFull: ImageSetPreviewProps["section"] = {
...
+export const ImageSetFullSansTitle: ImageSetPreviewProps["section"] = {
+  type: "image_set",
+  layout: "full",
+  images: Images,
+}
+
+export const ImageSetMiniSansTitle: ImageSetPreviewProps["section"] = {
+  type: "image_set",
+  images: Images,
+}
```
kanaabe: The pattern I've been using for supporting various fixtures is to create an `ImageSetPreview` array that contains the different types of image sets (ie: Image or HeroSection). This totally works fine too, but it might be confusing later on if we're using different patterns and maybe we should stick to one? What's the reason you separated it out for the ImageSet?

src/components/publishing/imageset_preview.tsx
```diff
@@ -112,31 +119,35 @@ class ImageSetPreview extends Component<Props, null> {
...
+    } else {
+      return (
+        <TitleWrapper layout={this.props.section.layout}>
+          <Title>{this.props.section.images.length} Images</Title>
+          <SubTitle>
+            <SubTitlePrompt>View Slideshow</SubTitlePrompt>
+          </SubTitle>
+        </TitleWrapper>
+      )
+    }
```
kanaabe: Noticing there's a lot of duplication here -- maybe we can extract the title check in a function like:
```        
        <TitleWrapper layout={this.props.section.layout}>
          {renderTitle()}
          <SubTitle>
            <SubTitlePrompt>View Slideshow</SubTitlePrompt>
            {renderSubtitleCount()}
          </SubTitle>
        </TitleWrapper>
```

src/components/publishing/__test__/fixtures.tsx
eessex: The ImageSetPreview sections need to be linked to the ImageSetPreviewProps, not sure how to do this inside an array without already declaring them as they are here and then nesting after

src/components/publishing/imageset_preview.tsx
```diff
@@ -112,31 +119,35 @@ class ImageSetPreview extends Component<Props, null> {
...
   }
   icon() {
     return (
```
kanaabe: Random comment not having to do with the diff: I saw some media queries in this component, but we should make use of our new `pMedia` helper [here](https://github.com/artsy/reaction/blob/master/src/components/helpers.tsx#L39)

src/components/publishing/__test__/fixtures.tsx
kanaabe: I see. It doesn't have to be now, but I think it's worth figuring that part out!

kanaabe:
Cool 👍  -- gonna merge! Whenever you implement the mobile versions, remember to use `pMedia`!

PR #207, opened 2017-07-26
@eessex => Patches/Bugs on ImageCollection and Artwork components

kanaabe:
- Adds responsiveness to `ImageCollection`
- Fixes bug with heights of `ImageCollectionItem`
- Fixes bug with missing info on `Artwork`


eessex:
Looks good!  Curious to play with the resize me library, maybe this could be useful for our pop-up text menus in Writer.

PR #208, opened 2017-07-27
@alloy => Upgrade node version

kanaabe:
Upgrades node version to LTS (we never upgraded since the vulnerability announcement).

Also, I'm seeing failing builds on CircleCI that's related to setting the default alias:

image

This is where it's actually failing:

image

Not sure why this hasn't been a problem until recently, but it's setting the default alias to node 8.2.0 (which happens to be the latest in https://circleci.com/docs/1.0/build-image-trusty/#nodejs) instead of the one specified in the .nvmrc and the one that actually gets installed. I'd like to see if this PR changes anything -- maybe Circle dropped support for this vulnerable version?


kanaabe:
Alright...it's happy after I `cat` the .nvmrc version to the circle.yml config.

image

alloy:
Nice! I assume this is the version that Force/L runs on too? If so, go ahead and merge 👍

kanaabe:
@alloy Good point...Force is configured to use 6.x.x and is currently on 6.11.1. Should be no problem I think?

orta:
Yep - that's fine @kanaabe

alloy:
@kanaabe Perfect, thanks 👌

PR #210, opened 2017-08-01
[Dev] Removes console logs we have no control over

orta:
End result:
screen shot 2017-08-01 at 09 50 16

---

Every test was giving the error `Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.` - why?

I added a throw to the see to see what causes the log

screen shot 2017-08-01 at 09 45 26

So what is the problem?

```
import { shallow } from "enzyme"
import * as renderer from "react-test-renderer"

jest.mock("isomorphic-fetch")
const fetch = require("isomorphic-fetch").default

import Button from "../../../../../components/buttons/inverted"
import Message from "../../../../../components/message"
import ForgotPassword from "../index"
```

the key is `_interopRequireWildcard` - in React it checks to see if you ever try to `get` the property `PropTypes` from the export. This happens in inside `import * as React ...`. 

This is an unintended side-effect, and IMO it's fine to suppress these warnings.


src/setup_jest.ts
```diff
@@ -0,0 +1,14 @@
...
+// Look into removing this on the next React update to 0.16.
+console.error = (message?: any, ...optionalParams: any[]) => {
+  if (
+    typeof message === "string" &&
+    (message.includes("Accessing PropTypes via the main React package") ||
+      message.includes("React.createClass is no longer supported."))
+  ) {
+    // NOOP
+  } else {
+    originalConsoleError(message)
```
damassi: Are you sure you want to overwrite native code? These warnings will disappear in time and _have_ to be gone by the time React 16 comes around. Also, I think the proper way to shim this to preserve existing functionality is

```js
const originalConsoleError = console.error.bind(console)

function error() {
  const message = arguments[0]

  if (
    typeof message === "string" &&
    (message.includes("Accessing PropTypes via the main React package") ||
      message.includes("React.createClass is no longer supported."))
  ) {
    // NOOP
  } else {
    originalConsoleError(...arguments)
  }
}

console.error = error
```

orta: I'm sure, yep, we don't update React / React Native often and it's tough to look at test results when every test has lines of deprecation warnings. Happy to switch to this version though 👍

orta:
This is still happening

screen shot 2017-10-10 at 8 37 01 am

orta:
Though @l2succes  - did https://github.com/artsy/reaction/pull/273 fix it?

alloy:
I’ll merge this as-is for now, @l2succes will continue on the React 16 upgrade after I’ve renamed all files, but in the meantime I’d like to have as little noise as possible while migrating to Relay Modern.

l2succes:
@orta The react 16 upgrade did fix it so we can remove it afterwards

PR #212, opened 2017-08-01
@eessex=> [Pub] Extract children from Header components

kanaabe:
Thanks for the nice pair sesh! 👭 

- Allow children to be passed to `StandardHeader` and `FeatureHeader` so Writer can use it with Draft! 
- Support missing deck in fullscreen headers

### DIFF
--- a/src/components/publishing/header/feature_header.tsx
+++ b/src/components/publishing/header/feature_header.tsx
@@ -71,10 +71,10 @@ const FeatureHeader: React.SFC<FeatureHeaderProps> = props => {
       <HeaderTextContainer>
         <HeaderText>
           <Vertical>{article.vertical.name}</Vertical>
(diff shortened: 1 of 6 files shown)


src/components/publishing/header/feature_header.tsx
```diff
@@ -211,11 +191,11 @@ const FeatureHeaderContainer = Div.extend`
       align-items: flex-start;
       flex-direction: column;
     }
-    ${SubHeaderText} {
+    .feature__deck {
       margin-bottom: 30px;
     }
     ${pMedia.sm`
-      ${Title} {
+      .feature__title {
```
kanaabe: Had to opt for classNames on styling the unknown children here 🤷‍♀️

eessex:
Hooray!  Looks awesome

Issue #187, opened 2017-06-29
[Storybooks] Static build is broken


alloy:
Probably [this issue](https://github.com/storybooks/storybook/issues/1203#issuecomment-311885419).

orta:
When Storybooks 3.2.0 comes out, I'll look into this and updating storybooks

joefraley:
what's the status of this? i'm still seeing errors during uglification:

```shell
Building storybook ...
Failed to build the storybook
static/preview...bundle.js from UglifyJs
Unexpected token: operator (>) [static/preview...bundle.js:...]
```

kanaabe:
After 3.2.0 it looks like regular builds are now failing too. Issue seems to be in the `node ./node_modules/.bin/build-storybook` step of `yarn deploy-storybook`. 

If I turn silent mode off [here](https://github.com/storybooks/storybook-deployer/blob/master/src/utils.js#L6), I see the bug is from chunkhash:
```
(node:31241) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.
Failed to build the storybook
chunk manager [entry]
static/[name].[chunkhash].bundle.js
Cannot use [chunkhash] for chunk in 'static/[name].[chunkhash].bundle.js' (use [hash] instead)
Error: Cannot use [chunkhash] for chunk in 'static/[name].[chunkhash].bundle.js' (use [hash] instead)
    at MainTemplate.replacePathVariables (/Users/kanaabe/Development/Artsy/reaction/node_modules/@storybook/react/node_modules/webpack/lib/TemplatedPathPlugin.js:64:9)
```

Next step for me is to figure out how to fix this, but i'm posting here in case this is obvious to anyone!

kanaabe:
The problem seems to be with using the  [HotModuleReplacementPlugin](https://github.com/artsy/reaction/blob/master/.storybook/webpack.config.js#L47) in our storybook webpack config. Removing this will build the storybook and make it deployable. @orta I know you were working on HMR a bit...do you know why we use this plugin? I thought storybook comes with HMR. Removing the plugin doesn't make HMR work, but maybe it's a different issue?

orta:
@kanaabe - make the config not do HMR when deploying 👍

That's legit

alloy:
Nice sleuthing @kanaabe 🙌

PR #214, opened 2017-08-09
@orta => Fix static build and deploy

kanaabe:
Only use the HMR Plugin in dev mode. I tested the deploy successfully here: https://artsy.github.io/reaction/

Closes https://github.com/artsy/reaction/issues/187


orta:
Looks legit to me 👍

PR #215, opened 2017-08-09
@orta => Fix deploy-storybook command related to NODE_ENV in Circle

kanaabe:
The deploy is still failing because we're copying the env.oss file to .env which contains `NODE_ENV=development`.

There are a few ways to solve this, but I just switched to using `process.env.NODE_ENV` and passed in the configuration on the `yarn deploy-storybook` command. 

image


orta:
Yes! Also makes sense 👍

PR #216, opened 2017-08-10
@kanaabe => Text component

eessex:
TextContainer component accepts either html or child component.  Adds text styles for all article layouts, can optionally take a 'postscript' option and render everything in italics. 

Still needs some mobile styles -- will add in a separate PR when confirmed by owen. 

blockquote overflow is not pictured in screenshots, but accounted for in the sectionContainer.

Classic:
screen shot 2017-08-10 at 3 30 34 pm

Standard:
screen shot 2017-08-10 at 3 31 39 pm

Feature:
screen shot 2017-08-10 at 3 31 26 pm


src/assets/theme.ts
```diff
@@ -17,7 +17,8 @@ export default {
   },
   publishing: {
     breakpoints: {
-      sm: 600, // px
+      xs: 600, // px
+      sm: 720, // px
```
kanaabe: 🍬

.vscode/cSpell.json
```diff
@@ -0,0 +1,17 @@
...
+    "words": [
+        "garamond"
+    ],
+    // flagWords - list of words to be always considered incorrect
+    // This is useful for offensive words and common spelling errors.
+    // For example "hte" should be "the"
+    "flagWords": [
+        "hte"
+    ]
+}
```
kanaabe: Intentional?

eessex: hm maybe not this one -- i wonder if vscode adds automatically?  adding garamond was intentional

eessex: will add this file to the gitignore

src/components/publishing/sections/sections.tsx
```diff
@@ -34,10 +36,11 @@ const Sections: React.SFC<SectionsProps> = props => {
 function renderSections(article) {
   const sections = article.sections
   const renderedSections = sections.map((section, i) => {
-    const child = getSection(section)
+    const child = getSection(section, article.layout)
+    const layout = section.body ? findBlockquote(section.body, article.layout) : section.layout
```
kanaabe: Hm this feels pretty specific to the text component -- would it be possible to move this logic there since section.body and article.layout is available in TextContainer?

eessex: would be possible, but we'd have to not set the width on the container in this file.

eessex: problem is blockquote is a different width than all the sections -- is bigger

src/components/publishing/sections/text.tsx
```diff
@@ -1,8 +1,132 @@
...
+    background-size: 1.25px 3px;
+    background-repeat: repeat-x;
+    background-position: bottom;
+  }
+  p, ul, ol {
+    ${props => (props.layout === "classic" ? Fonts.garamond("s20") : Fonts.garamond("s23"))};
+    padding-top: ${props => (props.layout === "classic" ? ".75em;" : "1em;")};
+    padding-bottom: ${props => (props.layout === "classic" ? ".75em;" : "1em;")};
+    margin: 0;
+    font-style: ${props => (props.postscript ? "italic;" : "inherit;")};
```
kanaabe: I think ideally we want to use the italic font here instead of font-style since we have it.

eessex: i thought we only had a weight argument for unica?

src/components/publishing/sections/sections.tsx
eessex: maybe would make more sense to set the layout on these text sections directly, so blockquote doesn't have to be identified on the client.

src/components/publishing/sections/text_container.tsx
```diff
@@ -0,0 +1,19 @@
...
+const TextContainer: React.SFC<TextContainerProps> = props => {
+  const { html, layout, postscript } = props
+  const child = html ? <div dangerouslySetInnerHTML={{ __html: html }} /> : props.children
+  return (
+    <Text layout={layout} postscript={postscript}>
+      {child}
+    </Text>
+  )
+}
+export default TextContainer
```
kanaabe: This is a bit of a nitpick, but could we name this `Text` and make the styled div either in the same file as this or separate and call it something like `StyledText`? Unless I'm missing a good reason not to.

I only say this because our components at the top-most level are consistently named like: `Embed`, `Video`, `Header`, etc. Not necessary but just a thought.

kanaabe:
Great work! Excited to see these in action. Maybe a screenshot would be nice?

eessex:
see screenshot above

src/components/publishing/sections/text.tsx
kanaabe: This should work: https://github.com/artsy/reaction/blob/master/src/components/publishing/fonts.tsx#L76

eessex: this text is in garamond -- no weight option i can see

src/components/publishing/sections/sections.tsx
kanaabe: Yeah that's right we want `SectionContainer` to always set the section's width. Definitely 👍  for setting the layout for blockquote at the section level in Writer. Seems much cleaner. Soo for now we can keep this!

src/components/publishing/sections/text.tsx
kanaabe: Ohh yes you're totally right. I kept thinking this was supposed to be unica. 👍

PR #219, opened 2017-08-15
@alloy => [Publishing] Use self-hosted webfont and styling QA

kanaabe:
- We no longer need the proxy middleware to support our font in dev mode to avoid CORS issues because we have a dedicated bucket on webfonts.artsy.net. This _should_ also fix our fonts on artsy.github.io/reaction. I've added all our icons to http://webfonts.artsy.net/all-webfonts.css as well which means we can remove the files from this project.
- Small styling QA on section margins and responsive


alloy:
So goooood! 👍

PR #221, opened 2017-08-17
@kanaabe => Accessible images, caption accepts children

eessex:
- Adds caption or artwork title as alt text for all body images and artworks
- Adds title as alt text for header images
- Updates Caption component to accept children and exports

### DIFF
--- a/src/components/publishing/sections/image.tsx
+++ b/src/components/publishing/sections/image.tsx
@@ -17,7 +17,8 @@ const Image: React.SFC<ImageProps> = props => {
   const { image, layout, width, height } = props
   return (
     <div className="article-image">
-      <BlockImage src={image.url} width={width} height={height} alt={image.caption.replace(/<[^>]*>/g, "")} />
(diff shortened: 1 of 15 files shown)


src/components/publishing/sections/image.tsx
```diff
@@ -17,7 +17,7 @@ const Image: React.SFC<ImageProps> = props => {
   const { image, layout, width, height } = props
   return (
     <div className="article-image">
-      <BlockImage src={image.url} width={width} height={height} />
+      <BlockImage src={image.url} width={width} height={height} alt={image.caption.replace(/<[^>]*>/g, "")} />
```
kanaabe: Maybe we can include a comment on what this regex does?

kanaabe:
Looks great!! Thanks for this thoughtful work :)

src/components/publishing/sections/image.tsx
eessex: Sure thing -- regex is for removing html tags from captions -- added a comment

PR #222, opened 2017-08-17
@kanaabe => Header tweaks

eessex:
Minor updates to standard header.  Should span full width (1250) of page (over sidebar).


kanaabe:
👍

PR #225, opened 2017-08-21
[Dev] Support showing CSS inside test snapshots

orta:
This makes PRs make more sense when looking at the results of tests 👍


alloy:
Groovy 👌

PR #228, opened 2017-08-23
@alloy => Caption editable updates

eessex:
Minor updates: 
- Adds our text-editor's generic placeholder class to caption component (not ideal to include a vendor class, but it saves a lot of css on the other side).

- Flips weighting in Caption for displaying children vs section html (will display a child if it has one)

### DIFF
--- a/src/components/publishing/sections/caption.tsx
+++ b/src/components/publishing/sections/caption.tsx
@@ -17,6 +17,7 @@ interface FigcaptionProps {
   layout: string
 }
 const div: StyledFunction<FigcaptionProps & React.HTMLProps<HTMLDivElement>> = styled.div
+// includes draft placeholder class for editable text in Writer
 const Figcaption = div`
   & > p, p, .public-DraftEditorPlaceholder-root {
     ${props => (props.layout === "classic" ? Fonts.garamond("s15") : Fonts.unica("s14", "medium"))}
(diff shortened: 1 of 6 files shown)


src/components/publishing/sections/caption.tsx
```diff
@@ -18,7 +18,7 @@ interface FigcaptionProps {
 }
 const div: StyledFunction<FigcaptionProps & React.HTMLProps<HTMLDivElement>> = styled.div
 const Figcaption = div`
-  & > p, p {
+  & > p, p, .public-DraftEditorPlaceholder-root {
```
alloy: It may be a good idea to add a comment that explains what this class is for so it doesn’t accidentally get removed in the future.

src/components/publishing/sections/caption.tsx
```diff
@@ -33,7 +33,7 @@ interface CaptionProps {
 
 const Caption: React.SFC<CaptionProps> = props => {
   const { layout, caption, viewFullscreen } = props
-  const child = caption ? <div dangerouslySetInnerHTML={{ __html: caption }} /> : props.children
+  const child = props.children ? props.children : <div dangerouslySetInnerHTML={{ __html: caption }} />
```
alloy: #minor maybe also add `children` to the destructuring on line 35?

alloy:
Looks good, just wondering if it would be beneficial to add a test for this logic?

eessex:
Thanks @alloy, these comments have been included in the most recent commit

alloy:
Perfect 👍

PR #230, opened 2017-08-29
@orta => Images send react children to caption

eessex:
Small update for Image component to accept children, and send them to the caption component.
Moves comment in image component out of render (was rendering)

### DIFF
--- a/src/components/publishing/sections/image.tsx
+++ b/src/components/publishing/sections/image.tsx
@@ -14,12 +14,15 @@ interface ImageProps extends React.HTMLProps<HTMLDivElement> {
 }
 
 const Image: React.SFC<ImageProps> = props => {
-  const { image, layout, width, height } = props
+  const { image, layout, width, height, children } = props
+  const child = children && children
+  // alt replace strips caption html
   return (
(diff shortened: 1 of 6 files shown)


src/components/publishing/sections/image.tsx
```diff
@@ -14,12 +14,15 @@ interface ImageProps extends React.HTMLProps<HTMLDivElement> {
 }
 
 const Image: React.SFC<ImageProps> = props => {
-  const { image, layout, width, height } = props
+  const { image, layout, width, height, children } = props
+  const child = children && children
```
orta: What does the `&& children` do here?

eessex: works similar to `children ? children : false`

orta: I could be wrong here, but I think then it'll be fine to just pass the children in directly - if not, then merge away 👍

eessex: The caption component defaults to showing children over a passed caption, so we do want to pass false here rather than children in this case.

PR #232, opened 2017-08-31
@alloy => FullscreenViewer QA

kanaabe:
- New ImageWrapper component handles artwork/image + expand icon overlay to enter the fullscreen viewer
- Focus slider on component update to ensure that key commands work without clicking/focusing on component.

image

### DIFF
--- a/src/components/publishing/article.tsx
+++ b/src/components/publishing/article.tsx
@@ -6,9 +6,13 @@ import FeatureLayout from "./layouts/feature_layout"
 import StandardLayout from "./layouts/standard_layout"
 import FullscreenViewer from "./sections/fullscreen_viewer/fullscreen_viewer"
 import Sections from "./sections/sections"
+import { Layout } from "./typings"
 
-interface ArticleProps {
-  article: any
+export interface ArticleProps {
+  article: {
+    layout: Layout
+    [x: string]: any
+  }
 }
 interface ArticleState {
   viewerIsOpen: boolean
--- a/src/components/publishing/article.tsx
+++ b/src/components/publishing/article.tsx
@@ -6,13 +6,10 @@ import FeatureLayout from "./layouts/feature_layout"
 import StandardLayout from "./layouts/standard_layout"
 import FullscreenViewer from "./sections/fullscreen_viewer/fullscreen_viewer"
 import Sections from "./sections/sections"
-import { Layout } from "./typings"
+import { ArticleData } from "./typings"
 
 export interface ArticleProps {
-  article: {
-    layout: Layout
-    [x: string]: any
-  }
+  article: ArticleData
 }
 interface ArticleState {
   viewerIsOpen: boolean
--- a/src/components/publishing/icon/expand.tsx
+++ b/src/components/publishing/icon/expand.tsx
@@ -1,0 +1,28 @@
+import React, { Component } from "react"
+
+class IconImageSet extends Component<any, null> {
+  render() {
+    return (
+      <svg viewBox="0 0 27 27" version="1.1" xmlns="http://www.w3.org/2000/svg">
+        <g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+          <g>
+            <circle fill="#000000" fillRule="nonzero" opacity="0.6" cx="13.9575195" cy="13.0588379" r="13" />
+            <g transform="translate(8.000000, 7.000000)" stroke="#FFFFFF">
+              <g>
+                <g transform="translate(0.382532, 0.382532)">
+                  <polyline points="4.78165571 0 11.1571966 0 11.1571966 6.37554094" />
+                  <polyline
+                    transform="translate(3.187769, 7.969426) rotate(-180.000000) translate(-3.187769, -7.969426) "
+                    points="-1.01881144e-06 4.78165537 6.37553992 4.78165537 6.37553992 11.1571963"
+                  />
+                  <path d="M10.360254,0.796942618 L0.796942618,10.360254" strokeLinecap="square" />
+                </g>
+              </g>
+            </g>
+          </g>
+        </g>
+      </svg>
+    )
+  }
+}
+export default IconImageSet
--- a/src/components/publishing/icon/expand.tsx
+++ b/src/components/publishing/icon/expand.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react"
 
-class IconImageSet extends Component<any, null> {
+class IconExpand extends Component<any, null> {
   render() {
     return (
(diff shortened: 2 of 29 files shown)


src/components/publishing/sections/fullscreen_viewer/__test__/__snapshots__/fullscreen_viewer.test.tsx.snap
```diff
@@ -140,7 +140,7 @@ exports[`renders properly 1`] = `
   height: 100vh;
   overflow: hidden;
   position: fixed;
-  z-index: 20;
+  z-index: 1070;
```
kanaabe: This _seems_ arbitrary, but it's actually the z-index we use in Force for modals: https://github.com/artsy/force/blob/master/desktop/components/stylus_lib/index.styl#L40. Maybe worth putting in a comment.

src/components/publishing/icon/expand.tsx
```diff
@@ -0,0 +1,28 @@
...
+                    transform="translate(3.187769, 7.969426) rotate(-180.000000) translate(-3.187769, -7.969426) "
+                    points="-1.01881144e-06 4.78165537 6.37553992 4.78165537 6.37553992 11.1571963"
+                  />
+                  <path d="M10.360254,0.796942618 L0.796942618,10.360254" strokeLinecap="square" />
+                </g>
+              </g>
+            </g>
+          </g>
+        </g>
+      </svg>
```
alloy: How do you make these SVG icons? Or does design output those for you?

src/components/publishing/sections/image_wrapper.tsx
```diff
@@ -0,0 +1,37 @@
...
+  alt?: string
+  index?: number
+}
+
+const ImageWrapper: React.SFC<ImageWrapperProps> = props => {
+  const { layout, index } = props
+  const viewFullscreen = layout !== "classic" ? <ViewFullscreen index={index} /> : false
+  const newProps: any = { ...props }
+  delete newProps.layout
+  delete newProps.index
```
alloy: You can also do:

```tsx
const { layout, index, ...blockImageProps } = props
```

src/components/publishing/sections/fullscreen_viewer/__test__/__snapshots__/fullscreen_viewer.test.tsx.snap
alloy: Maybe instead of a comment, this could go into a central file that’s all about styling constants and name the variable something like “modal z-index” so it’s self-documenting. I.e. like that file you linked to in Force.

src/components/publishing/sections/fullscreen_viewer/fullscreen_viewer.tsx
```diff
@@ -23,11 +23,20 @@ class FullscreenViewer extends React.Component<FullscreenViewerProps, Fullscreen
     onToggleCaption: PropTypes.func,
   }
 
+  private slider: any
```
alloy: Fancy 👌

src/components/publishing/sections/image_wrapper.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react"
+import styled from "styled-components"
+import ViewFullscreen from "./view_fullscreen"
+
+interface ImageWrapperProps extends React.HTMLProps<HTMLDivElement> {
+  src: any
+  layout?: string
```
alloy: Maybe type this as a union of the possible strings so it’s documented when people try to use it.

alloy: I see that `layout` is used everywhere, so in case you do decide to define a union, maybe it’s worth it to create a central typings file for this which you can import wherever this typing is used.

src/components/publishing/icon/expand.tsx
kanaabe: Ah design outputs these for us. They're generated via Sketch

src/components/publishing/sections/image_wrapper.tsx
kanaabe: oo yeah love that

src/components/publishing/icon/expand.tsx
alloy: Dope!

src/components/publishing/article.tsx
```diff
@@ -6,9 +6,13 @@ import FeatureLayout from "./layouts/feature_layout"
...
 import FullscreenViewer from "./sections/fullscreen_viewer/fullscreen_viewer"
 import Sections from "./sections/sections"
+import { Layout } from "./typings"
 
-interface ArticleProps {
-  article: any
+export interface ArticleProps {
+  article: {
+    layout: Layout
+    [x: string]: any
```
kanaabe: Just making note of this change to get our Layout type working across components. Not sure if there's a cleaner solution so lmk if you can think of one.

If you see the next file below in `fixtures/articles.ts`, we reuse this interface when defining the various types of articles. This makes testing a little more sane too since it gives proper names to articles instead of pulling them arbitrarily from an array. Soo the diff grew by a lot but it's mostly whitespace.

alloy: Is the interface for an ‘article’ always going to be the same? If so, then maybe you could move the interface declaration for it to the central typings too and import/use it here like so:

```ts
import { ArticleData } from "./typings"

interface ArticleProps {
  article: ArticleData
}
```

Then in the `fixtures/articles.ts` file you won’t need to do `ArticleProps["article"]` but just `ArticleProps`, which makes more sense.

I didn’t entirely follow what your question about a better way was about, does this answer it?

src/components/publishing/sections/fullscreen_viewer/__test__/__snapshots__/fullscreen_viewer.test.tsx.snap
kanaabe: I'm going to table this for now and eventually get real (typed) system for these. There's also the `faux-underline` CSS that could be nice to include.

src/components/publishing/article.tsx
kanaabe: oh nice! yeah that seems more flexible. I was moreso wondering about the `[x: string]: any` part since we needed to pull out layout to define it.

src/components/publishing/sections/fullscreen_viewer/__test__/__snapshots__/fullscreen_viewer.test.tsx.snap
alloy: 👍

PR #233, opened 2017-09-04
@alloy => [Publishing] Sidebar, share, related articles panel

kanaabe:
- Share component
- RelatedArticles component
- Sidebar (layout) component
- Adds gemini on-the-fly resizer to utils
- Fixes https link on embed
- Fix sizing on section width (fixes frozen pages after resize)

image

### DIFF
--- a/src/components/publishing/constants.ts
+++ b/src/components/publishing/constants.ts
@@ -1,8 +1,5 @@
 export const sizeMeRefreshRate = 500
 
-export const zIndex = (size: string) => {
-  const sizes = {
-    modal: 1070,
-  }
-  return sizes[size]
+export const zIndex = {
+  modal: 1070,
 }
--- a/src/components/publishing/constants.ts
+++ b/src/components/publishing/constants.ts
@@ -1,0 +1,8 @@
+export const sizeMeRefreshRate = 500
+
+export const zIndex = (size: string) => {
+  const sizes = {
+    modal: 1070,
+  }
+  return sizes[size]
+}
--- a/src/components/publishing/header/feature_header.tsx
+++ b/src/components/publishing/header/feature_header.tsx
@@ -3,6 +3,7 @@ import sizeMe from "react-sizeme"
 import styled from "styled-components"
 import { resize } from "../../../utils/resizer"
 import { pMedia } from "../../helpers"
+import { sizeMeRefreshRate } from "../constants"
 import Fonts from "../fonts"
 import AuthorDate from "./author_date"
 
@@ -243,7 +244,7 @@ const FeatureHeaderContainer = Div.extend`
 `
 
 const sizeMeOptions = {
-  refreshRate: 500,
+  refreshRate: sizeMeRefreshRate,
   refreshMode: "debounce",
 }
 
--- a/src/components/publishing/header/feature_header.tsx
+++ b/src/components/publishing/header/feature_header.tsx
@@ -243,7 +243,7 @@ const FeatureHeaderContainer = Div.extend`
 `
 
 const sizeMeOptions = {
-  refreshRate: 64,
+  refreshRate: 500,
   refreshMode: "debounce",
 }
 
--- a/src/components/publishing/header/feature_header.tsx
+++ b/src/components/publishing/header/feature_header.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import sizeMe from "react-sizeme"
 import styled from "styled-components"
+import { resize } from "../../../utils/resizer"
 import { pMedia } from "../../helpers"
 import Fonts from "../fonts"
 import AuthorDate from "./author_date"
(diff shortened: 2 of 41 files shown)


src/components/publishing/header/feature_header.tsx
```diff
@@ -240,7 +243,7 @@ const FeatureHeaderContainer = Div.extend`
 `
 
 const sizeMeOptions = {
-  refreshRate: 64,
+  refreshRate: 500,
```
alloy: What does this do in practice?

src/utils/resizer.ts
```diff
@@ -0,0 +1,56 @@
...
+  const config = {
+    resize_to: resizeTo,
+    src,
+    width,
+    height,
+    quality: quality || 95,
+  }
+
+  return [GEMINI_CLOUDFRONT_URL, qs.stringify(config)].join("?")
+}
```
alloy: Do you double resolution for retina displays? Or is that something that’s hard to figure out when doing SSR? If not, we should really have a ticket for that and figure it out sometime.

In [Emission we do it](https://github.com/artsy/emission/blob/3a920657e68dce5a9fc9a2a61affbe9c831144ed/src/lib/Components/OpaqueImageView.tsx#L88-L96), but it’s easy because we’ll know the exact laid out size and the screen pixel density.

src/components/publishing/sections/video.tsx
```diff
@@ -132,7 +129,7 @@ const PlayButton = styled.div`
 `
 
 const sizeMeOptions = {
-  refreshRate: 64,
+  refreshRate: 500,
```
alloy: Should this be a global constant?

src/components/publishing/header/feature_header.tsx
kanaabe: From their docs:
  // The maximum frequency, in milliseconds, at which size changes should be
  // recalculated when changes in your Component's rendered size are being
  // detected. This should not be set to lower than 16.

This is half a second now, instead of 64 milliseconds. We currently have an issue where the browser is occasionally freezing components. There are (I believe) 2 reasons why...the first is that I implemented react-sizeme a little naively. It's being used multiple times within an article on the individual component as opposed to at MOST once or twice on the article. This would be a larger-scale refactor.

The second is this factor, where because we have multiple components with recalculations every 64 milliseconds, the browser can't keep up. Changing it to half a second gives us a little more breathing room (and actually doesn't break the browser anymore). I think we still need that larger refactor but for now this helps. 

Sorry this should have definitely been explained as part of the PR! Lmk if u have questions.

src/components/publishing/sections/video.tsx
kanaabe: Yeah, maybe it's time we add that constants file - will add now!

src/utils/resizer.ts
kanaabe: We're not as precise with it -- for images that are cropped, we double exactly. But for everything else, we ask for 1200px resized width to sort of account for retina. There are definitely problems with this -- the max-width of images on the `StandardLayout` are 780px. For `FeatureLayout` they could be full-bleed. Also on mobile we use the same image size so 1200 is kind of an arbitrary in-between. This should be handled better.

For SSR in Force, we have an `isMobile` constant that is determined by the user-agent. We could use that for choosing sizes. Retina detection is a little harder with SSR. We could detect once and store it in a session cookie? 

I guess what we want here is a combo of knowing user-agent (for SSR Mobile) and some predetermined constants for various layouts. Unlike Gravity artworks, article images are just uploads to S3 (backed by Gemini/Cloudfront), and we don't have various pre-made images (i.e large, larger, medium) so either we add that kind of support in Positron (our team has reviewed this possibility a bunch and it would require quite a bit of work) or we can use predetermined constants.

src/components/publishing/header/feature_header.tsx
alloy: Not sure if it applies to the second issue you were referring to, but you may also want to look at [`shouldComponentUpdate`](https://facebook.github.io/react/docs/react-component.html#shouldcomponentupdate).

src/utils/resizer.ts
alloy: Interesting stuff to think about. We’re probably not the first ones, though, so probably should also do some research.

Re those predetermined constants; to be clear, we don’t use those in Emission, I think the future is only dynamic sizing (with appropriate caching).

alloy:
Thanks for the explanations! 👍

src/utils/resizer.ts
orta: To add a me too: I had this same SSR problem in team nav, couldn't really figure out a good answer

src/components/publishing/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const sizeMeRefreshRate = 500
+
+export const zIndex = (size: string) => {
+  const sizes = {
+    modal: 1070,
+  }
+  return sizes[size]
+}
```
alloy: Why is this a function rather than a single object?

kanaabe: Ah yeah you're right this is overkill

alloy:
Alright, this feels good to me 👍

PR #235, opened 2017-09-06
Update to webpack v3.5.x

l2succes:


orta:
wow - that breaks so many Peril things o.0

orta:
None of it makes sense

l2succes:
@orta Yeah... not sure what's going on here. Is it because we have a `danger/pr.js` instead of `danger/pr.ts`? Looks like it's looking for a `.ts` file

orta:
Ah yeah, I added that 16 days ago: https://github.com/artsy/reaction/commit/a9256fca93b98bca890ec7f24869e40a49ccca09

So it's not an issue 👍 - you can rebase to update if you want

PR #237, opened 2017-09-07
@eessex => [Publishing] Related Articles Canvas

kanaabe:
- Adds a RelatedArticlesCanvas
- Moves the AuthorDate component to the top-level /publishing dir
- New `related_articles` dir. For naming I went with Panel + Canvas, which maps similarly to the ad offerings.

Desktop:
image

Mobile:
image


src/components/publishing/related_articles/related_article_figure.tsx
```diff
@@ -0,0 +1,67 @@
...
+    </ArticleFigure>
+  )
+}
+
+const ImageTitle = styled.a`
+  display: flex;
+  flex-direction: column;
+  color: black;
+  text-decoration: none;
+  height: 270px;
```
eessex: It is possible that long titles could overflow this height?  Maybe min-height would make sense here -- not sure if the character count on thumbnail titles will cover it.

src/components/publishing/related_articles/related_articles_canvas.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const RelatedArticlesContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  max-width: 1250px;
+  margin: 30px auto 60px auto;
+  ${pMedia.lg`
+  margin: 30px 20px 60px 20px;
+  `}
```
eessex: * minor, indent missing in pMedia

eessex:
Some very minor comments, feel free to address or merge

kanaabe:
Addressed! Thanks. Will merge this after GH has its moment. Seems like they're having webhook delays.

eessex:
Ahh, is that why circle is not running?

Issue #211, opened 2017-08-01
HMR could be faster


orta:
I pulled out some stats on our webpack 

https://ortastuff.s3.amazonaws.com/data/webpack_stats.json - which you can download, and open with http://webpack.github.io/analyse/

---

I don't have too much internal understanding of webpack as my focus has mainly been on [metro bundler](https://github.com/facebook/metro-bundler/). 

We see most of the wait inside ` 92% chunk asset optimization`. So what can we learn from this?

We have 2 chunks, I expect this chunk (id: 1) is the [storybooks code](http://webpack.github.io/analyse/#chunk/1) 

![screen shot 2017-08-01 at 15 26 56](https://user-images.githubusercontent.com/49038/28843068-dec91e0c-76cd-11e7-9c45-b2b8ece9d56a.png)

We don't want this to be re-created on an incremental build.

This second one is our app, 

![screen shot 2017-08-01 at 15 31 42](https://user-images.githubusercontent.com/49038/28843270-87c4a1ca-76ce-11e7-841c-1740a06bb0eb.png)

You can find the modules in here: http://webpack.github.io/analyse/#module/81

![screen shot 2017-08-01 at 15 31 42](https://user-images.githubusercontent.com/49038/28843288-9979fc58-76ce-11e7-9466-230f32745d30.png)

--

When you look at the data that gets sent on a hot update, it's tiny:

![screen shot 2017-08-01 at 15 33 06](https://user-images.githubusercontent.com/49038/28843324-ba6f3e0a-76ce-11e7-9e5a-514066925dbf.png)

It's just taking a bunch of time to get there

orta:
I pushed a branch with what it took to generate this stuff:

```sh
git clone https://github.com/artsy/reaction-force.git
git checkout stats
yarn install
yarn storybooks
```

Then use `curl http://localhost:9001/stats.json > stats.json` to download the stats JSON.

orta:
Options for further digging:

* https://github.com/amireh/happypack

orta:
This HMR is what I care about: https://twitter.com/niieani/status/894630819192557568

orta:
https://medium.com/webpack/typescript-webpack-super-pursuit-mode-83cc568dea79

damassi:
@orta - this might provide some visual insight in terms of efficient organization: https://github.com/webpack/webpack/issues/690

orta:
[...]

PR #238, opened 2017-09-07
Bump Node v.8.4.0

damassi:
Following on [this comment](https://github.com/artsy/reaction/issues/211#issuecomment-327710945) it appears the latest Node version (8.4.0) provides significant performance gains during compilation. 

Matches current [setting in Force](https://github.com/artsy/force/blob/master/circle.yml#L3).


orta:
Gave artsy.net a browse now that 8.4.x is running it, seems fine so far 👍

PR #239, opened 2017-09-07
@kanaabe => More header updates

eessex:
- Header can now accept verticals as children
- Adds tests for headers with children
- Moves layout title styles into respective header components
- Passes children to header-components as named props
- Fixes errors in icons
- Exports article test fixtures


src/components/publishing/header/feature_header.tsx
```diff
@@ -64,7 +64,7 @@ function renderTextLayoutAsset(url, layout, title, children) {
       const image = <Image src={src} alt={alt} />
       return (
         <TextAsset>
-          {children[2]}
+          {children[3]}
```
kanaabe: More of a readability thing, but at the top of the render where we pull out children in the first place, it might be helpful to de-mystify what these children are by naming them before passing them into other places.

src/components/publishing/header/header.tsx
```diff
@@ -17,18 +17,20 @@ const Header: React.SFC<HeaderProps> = props => {
...
+        {props.children && props.children[3]}
       </FeatureHeader>
     )
   } else if (article.layout === "standard") {
     return (
       <StandardHeader article={article}>
+        {props.children && props.children[0]}
         <StandardTitle>
-          {props.children ? props.children : article.title}
+          {props.children ? props.children[1] : article.title}
```
kanaabe: I think here too, seeing a lot of repetition that could be avoided by deciding the content outside of building out JSX elements

eessex: Good call, just reworked this to pass the children as named props to each header

kanaabe:
Nice refactor! Feel free to merge when 🍏

PR #242, opened 2017-09-12
@alloy => Icon fullscreen, minor QA

eessex:
- Adds an Icon for fullscreen image UI
- Removes header class selectors & replaces with styled component references
- Sets lead_paragraph width to 100%

### DIFF
--- a/src/components/publishing/header/feature_header.tsx
+++ b/src/components/publishing/header/feature_header.tsx
@@ -215,7 +215,7 @@ const FeatureHeaderContainer = Div.extend`
   height: 100vh;
   &[data-type='text'] {
     height: auto;
-    .feature__title {
+    ${Title} {
       margin-bottom: 150px;
     }
     ${HeaderText} {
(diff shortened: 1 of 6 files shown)


src/components/publishing/header/feature_header.tsx
```diff
@@ -215,15 +215,15 @@ const FeatureHeaderContainer = Div.extend`
   height: 100vh;
   &[data-type='text'] {
     height: auto;
-    .feature__title {
+    ${Title} {
```
alloy: Were these referring to the same thing but previously in a roundabout way?

alloy:
LGTM 👍

src/components/publishing/header/feature_header.tsx
eessex: Yeah, they are the same div

alloy: Ah ok, so where was that classname coming from? Is that what styled-components generated?

eessex: Originally they came from the [header component](https://github.com/artsy/reaction/blob/master/src/components/publishing/header/header.tsx), the title class has already been removed.  The deck will probably be removed too in a future PR.

alloy: Gotcha, thanks 👍

PR #243, opened 2017-09-13
[Analytics] @kanaabe @mennenia: Add tracking to share buttons on new article components

craigspaeth:
**Don't merge yet**

This starts to add [react-tracking](https://github.com/NYTimes/react-tracking/tree/aaf9ed0b5f04b89ad5b0e6346e1877a1d16342ab) to our article components based on [Maxim's work](https://github.com/artsy/emission/pull/491). Aside from tests/cleanup, the one bigger remaining question in my mind is—do we want to implement Segment's Analytics.js `analytics.track()` calls in Reaction (sharing the segment key via environment variables) or delegate that out to the calling code in Force? My first instinct is to choose the former to better encapsulate that code in Reaction, but I'm probably overlooking caveats.

cc @orta @alloy


.vscode/settings.json
```diff
@@ -1,24 +1,21 @@
...
-    "prettier.trailingComma": "es5",
-    "prettier.typescriptEnable": [
-        "typescript",
-        "typescriptreact"
-    ],
-    "prettier.semi": false,
-    "prettier.printWidth": 120,
-    "typescript.tsdk": "./node_modules/typescript/lib"
-  }
+  "files.exclude": {
```
craigspaeth: Noticed these files using 4 spaces when we're using 2 everywhere else.

src/components/publishing/share.tsx
```diff
@@ -7,29 +8,48 @@ interface ShareProps extends React.HTMLProps<HTMLDivElement> {
...
-const Share: React.SFC<ShareProps> = props => {
-  const { url, title } = props
-  const encodedUrl = encodeURIComponent(url)
-  const getHref = type => {
+@track()
+class Share extends React.Component<ShareProps, null> {
+  constructor(props) {
+    super(props)
+    this.getHref = this.getHref.bind(this)
+    this.trackShare = this.trackShare.bind(this)
```
kanaabe: We can probably get rid of these by defining the functions with ES6 arrows: `trackShare = (e) => {}`

alloy:
Nice, more and more things are coming together 🙏🏽

orta:
In Emission, we pass in the analytics function, this means consumer apps can have their own custom functionality and keeps emission free of the analytics deps / EMV vars. 

You can use this to have a debugging analytics mode in reaction for example that shows a popup when you trigger an event (making the analytics team's life easier ) vs force/positron where it would actually send.

mennenia:
I wrote a whole preamble, and then Orta came and posted :D. 

Basically, the benefit of passing it through to the host app is that that application is in charge of where things get sent. In Eigen, we don't just use Segment actually, but also Adjust, and so just telling Eigen "this is an event, please send" means that it goes to all the right places (including our debug interceptions). 

Because this would be going into Reaction, events would stay cohesive with Force and Positron. If there's a specific framework we'd want to move towards, and unify all of our apps so that we send the same type of information consistently to specific platforms, we could move it into Reaction. I do think that's a separate project and something we could tackle further on as a second phase.

src/components/publishing/article.tsx
```diff
@@ -23,6 +24,8 @@ interface ArticleState {
   article: any
 }
 
+// TODO: Determine where to implement Segment calls
+// @track({}, { dispatch: data => console.log(data) })
```
mennenia: good ol' console.log! 

I pushed some new things to my PR yday, we use the Emission/Eigen bridge to send events now and it [looks like this](https://github.com/artsy/emission/pull/693/files#diff-bb6e7220fc0d9b5c39f6b5b40352d143R22).

src/components/publishing/share.tsx
mennenia: This should work, because react-tracking uses this also. Fun fact, in React Native / Typescript, this actually gives issues when decorating functions that use this approach. If anyone looked at [my PR](https://github.com/artsy/emission/pull/693) and noted the second "Important" point, it's specific to Emission's setup.

mennenia: Wait, forget what I just said. Test it, but it might also give issues (like I had in Emission) because obviously Reaction is also using Typescript.

mennenia:
In general, this is extremely exciting!! 🙌 👏

craigspaeth:
Thanks for the comments all! I'll update to inject the dispatching like Emission, test things out + add tests, and add some additional tracking to publishing components.

src/components/publishing/share.tsx
alloy: This feature (specifically ES7 class properties, not ES6 arrow functions) has only [very recently](https://github.com/tc39/proposal-class-fields/commit/668b51adc7b62a98f81f49b122302d3fbd9d6f06) moved to stage 3 of TC39, which is why it was dangerous to use (implementations could differ, debuggers may not be able to make sense of the emitted code, etc). Now that it has moved to stage 3, the TypeScript implementers will probably want to hear about cases where it doesn’t work well, so let’s verify on the next TS update and otherwise submit a bug report.

craigspaeth:
@mennenia @kanaabe @alloy I added an events library with a similar API to Emission's which is an [event emitter](https://nodejs.org/api/events.html) under the hood. This should be convenient for hooking into tracking calls at any level (example below). Lmk what you think, and if all's good we can merge and I'll move on to more tracking decorators for publishing components in a separate PR.

````javascript
import { Events } from "../../utils"

@track({}, { dispatch: Events.postEvent })
class Article extends React.Component {
}
````

````javascript
import { Events } from "@artsy/reaction-force/dist/utils"
import analytics from "@segment/analytics"

Events.onEvent((data) => {
  analytics.track(data)
})
````

mennenia:
@craigspaeth awesome! 

Does this capture all the data that you need? E.g. page view / source screen events etc? I'm not sure how Force handles these. I used to extrapolate the screen ("app" in Force?) from the class for example. If this catches that all, it looks so neat! 😄 And please ignore if this is something you'll be adding later on :).

alloy:
@mennenia @craigspaeth @kanaabe I made a PR against _this PR’s branch_ that adds both a vendored version of the generic `react-tracking` typings and specific ones tailored to Artsy and its analytics schema https://github.com/craigspaeth/reaction/pull/1.

craigspaeth:
@mennenia we can track page views via [lifecycle methods](https://github.com/NYTimes/react-tracking#when-to-use-optionsdispatchonmount), I guess other "screen" events we could also hook into browser APIs and use the `Events.postEvent` API directly like `document.addEventListener("DOMContentLoaded", () => Events.postEvent({ action: 'dom loaded' })`.

kanaabe:
@craigspaeth this looks good to go on 🌵

alloy:
Groovy 🕺

craigspaeth:
Got the 💚  margin' !

mennenia:
🎉

PR #245, opened 2017-09-16
[fillwidth] Fix a couple of infinite loop bugs.

### LINKED ISSUE #240: [fillwidth] Optimise calculations such that it doesn’t freeze the browser.
Currently, loading a page like https://www.artsy.net/gene/russia and resizing it a couple of times will freeze the browser, because [the fillwidth component](https://github.com/artsy/reaction/blob/master/src/utils/fillwidth.ts) is trying to recalculate the layout too often.

It would be ideal if we can leverage the browser’s layout engine, by for instance calculating the layout once and then translating the results into relative styling that leave further calculations up to the browser.

Alternatively, a first possible stopgap solution could be to have a single `sizeme` instance that measures the width of the window and passes that down to all the `fullwidth` components.

alloy:
Also adds a limit on the number of iterations the loop may make.

Fixes #240, at least the freezing aspect. This code still could do with optimising, some layouts take 100s of iterations to calculate.

### DIFF
--- a/src/utils/fillwidth.ts
+++ b/src/utils/fillwidth.ts
@@ -1,5 +1,7 @@
 import { reduce } from "lodash"
 
+const MAX_ITERATIONS = 1000
+
 const fillwidthDimensions = (items, containerWidth, gutter = 10, targetHeight) => {
   /**
    * Scales an image object proportionally based on a direction (either -1 or 1)
@@ -47,19 +49,24 @@ const fillwidthDimensions = (items, containerWidth, gutter = 10, targetHeight) =
     }
   })
 
-  // If the total width difference is less than 0, it is larger than the container
-  // so we need to scale down. If not, scale up.
-  let dir = widthDiff(dimensions) < 0 ? -1 : 1
+  // If the total width difference is too small or negative we need to scale down. If not, scale up.
+  let dir = widthDiff(dimensions) < 1 ? -1 : 1
 
   // Keep looping until we get an acceptable width difference
-  while (true) {
+  let count = 0
+  while (widthDiff(dimensions) <= 1) {
     for (let img of dimensions) {
       resizeHeight(img, dir)
       if (widthDiff(dimensions) > 1) {
         break
       }
     }
-    if (widthDiff(dimensions) > 1) {
+    // Seeing as there have been a couple of bugs in this code and there are no proper tests yet, let’s at least make
+    // sure to no longer run into infinite loops.
+    count++
+    if (count === MAX_ITERATIONS) {
+      const data = { items, containerWidth, gutter, targetHeight, dir, dimensions }
+      console.error(`Was unable to calculate a filling width for data: ${JSON.stringify(data)}`)
       break
     }
   }


src/utils/fillwidth.ts
```diff
@@ -47,19 +49,24 @@ const fillwidthDimensions = (items, containerWidth, gutter = 10, targetHeight) =
     }
   })
 
-  // If the total width difference is less than 0, it is larger than the container
-  // so we need to scale down. If not, scale up.
-  let dir = widthDiff(dimensions) < 0 ? -1 : 1
+  // If the total width difference is too small or negative we need to scale down. If not, scale up.
+  let dir = widthDiff(dimensions) < 1 ? -1 : 1
```
alloy: This was testing against `0`, which does make sense, but later on the difference is checked against `1`, presumably because we want to have _some_ margin. The problem with comparing different values is that we risk resizing in the wrong direction.

Seeing as it’s hard to conclude those details all right away (there are no tests that specify one or the other), I opted to just compare against `1` here as well.

src/utils/fillwidth.ts
```diff
@@ -47,19 +49,24 @@ const fillwidthDimensions = (items, containerWidth, gutter = 10, targetHeight) =
...
-  while (true) {
+  let count = 0
+  while (widthDiff(dimensions) <= 1) {
     for (let img of dimensions) {
       resizeHeight(img, dir)
       if (widthDiff(dimensions) > 1) {
         break
       }
     }
-    if (widthDiff(dimensions) > 1) {
```
alloy: Putting the first check _after_ resizing at least once, means that the containers could get resized too much. Instead we now check it at the start of the loop, possibly never resizing if the difference was already ok to begin with.

src/utils/fillwidth.ts
```diff
@@ -47,19 +49,24 @@ const fillwidthDimensions = (items, containerWidth, gutter = 10, targetHeight) =
...
+    // Seeing as there have been a couple of bugs in this code and there are no proper tests yet, let’s at least make
+    // sure to no longer run into infinite loops.
+    count++
+    if (count === MAX_ITERATIONS) {
+      const data = { items, containerWidth, gutter, targetHeight, dir, dimensions }
+      console.error(`Was unable to calculate a filling width for data: ${JSON.stringify(data)}`)
```
alloy: Just trying to be safe and never run into an infinite loop again. Ideally these cases wouldn’t only get logged to the console, but also reported to Sentry as an error, which should give us enough data to reproduce the bug.

@damassi @craigspaeth Do you know offhand if `console.error` logs are intercepted by Sentry and reported? They do do so with React Native, but haven’t verified with web JS yet.

src/utils/fillwidth.ts
```diff
@@ -1,5 +1,7 @@
 import { reduce } from "lodash"
 
+const MAX_ITERATIONS = 1000
```
alloy: E.g. the fullwidth story in the publishing section required 100s of iterations.

kanaabe:
Thanks @alloy these are great improvements. Sorry I didn't catch these before when I extracted it from its previous location. I'm not sure about the Sentry reporting so I'm fine with merging and finding out 👍

src/utils/fillwidth.ts
damassi: Yup, apparently it's captured as a breadcrumb, [but not explicitely caught](https://github.com/getsentry/raven-node/issues/121#issuecomment-266919986).

PR #247, opened 2017-09-21
@kanaabe => More Icons

eessex:
Adds a few more icons for Writer

### DIFF
--- a/src/components/publishing/icon/edit_embed.tsx
+++ b/src/components/publishing/icon/edit_embed.tsx
@@ -4,7 +4,7 @@ class EditEmbed extends Component<any, null> {
   render() {
     return (
       <svg
-        id="edit-embed"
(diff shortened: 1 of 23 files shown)


src/components/publishing/icon/edit_embed.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from "react"
+
+class EditEmbed extends Component<any, null> {
+  render() {
+    return (
+      <svg
+        id="edit-embed"
```
kanaabe: Do we want to use className instead here? I _think_ having multiple SVGs with the same id on the page is technically not allowed.

eessex: good call

kanaabe:
👍  Cool, this looks great!

PR #249, opened 2017-09-22
@eessex => [Publishing] Read more

kanaabe:
Closes https://github.com/artsy/publishing/issues/44

There are a few different things going on here which I'll explain in the comments.

readmore

### DIFF
--- a/src/components/publishing/__stories__/read_more.story.tsx
+++ b/src/components/publishing/__stories__/read_more.story.tsx
@@ -1,33 +1,13 @@
 import { storiesOf } from "@storybook/react"
-import { extend } from "lodash"
 import * as React from "react"
 import Article from "../article"
-import { StandardArticle } from "../fixtures/articles"
+import { ImageHeavyStandardArticle, ShortStandardArticle, StandardArticle } from "../fixtures/articles"
 import { RelatedCanvas, RelatedPanel } from "../fixtures/components"
 
 storiesOf("Publishing/Read More", module).add("Text Article", () => {
-  const article = extend({}, StandardArticle, {
-    sections: [
-      {
-        type: "text",
-        body:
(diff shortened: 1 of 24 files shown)


src/components/publishing/header/feature_header.tsx
```diff
@@ -288,7 +288,7 @@ const FeatureHeaderContainer = Div.extend`
 
 const sizeMeOptions = {
   refreshRate: sizeMeRefreshRate,
-  refreshMode: "debounce",
+  noPlaceholder: true,
```
kanaabe: This helps address 2 issues:
1. No more flashing images
2. Inserts actual component elements instead of an empty div means we can calculate Read More truncation accurately on the `componentDidMount` lifecycle method

src/components/publishing/read_more/read_more_wrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+      if (height) {
+        return height
+      } else {
+        this.props.hideButton()
+        return "100%"
+      }
+    } else {
+      return "100%"
+    }
+  }
```
kanaabe: This method is a little...long, but it's basically finding the jQuery-like offset after determining the best paragraph to truncate on. We use the lodash `find` method to be able to short circuit the method when our character limit is exceeded.

src/components/publishing/__stories__/read_more.story.tsx
```diff
@@ -0,0 +1,114 @@
+import { storiesOf } from "@storybook/react"
+import { extend } from "lodash"
+import * as React from "react"
+import Article from "../article"
+import { StandardArticle } from "../fixtures/articles"
+import { RelatedCanvas, RelatedPanel } from "../fixtures/components"
+
+storiesOf("Publishing/Read More", module).add("Text Article", () => {
+  const article = extend({}, StandardArticle, {
+    sections: [
```
eessex: Maybe it would make sense to move these articles into the fixtures file?  Seems like it would be useful to have more sample data options for testing in Positron and Force.

kanaabe: Yeah, moving into fixtures seems like a good idea! Will do.

eessex:
Other than my last comment, this looks great! Feel free to merge when ready.

PR #250, opened 2017-09-25
@eessex => Validation on <EmailSignup/>

kanaabe:
- Fixes a bug where the email was not being attached to `body`
- Adds simple client-side email validation based on regex


eessex:
Excellent! Merging.

PR #253, opened 2017-09-26
@alloy: [Onboarding] created foundation

xtina-starr:
This is a WIP for the new onboarding flow and collector intent question more specifically. The ablility to submit and send the data to gravity along with some styling clean up still needs to be implemented.

screen shot 2017-09-26 at 6 14 23 pm

### DIFF
--- a/src/components/__stories__/collector_intent_page.story.tsx
+++ b/src/components/__stories__/collector_intent_page.story.tsx
@@ -1,0 +1,87 @@
+import { storiesOf } from "@storybook/react"
+import * as React from "react"
+import styled from "styled-components"
+
+import * as fonts from "../../assets/fonts"
+import Button from "../buttons/ghost"
+import Icon from "../icon"
+import Title from "../title"
+
+const IconContainer = styled.div`
+width: 15px;
+height: 15px;
+background-color: black;
+display: none;
+border-radius: 50%;
+float: right;
+margin-right: 15px;
+`
+
+const Link = styled.a`
+  display: block;
+  font-size: 14px;
+  color: black;
+  text-decoration: none;
+  text-transform: uppercase;
+  font-family: ${fonts.primary.fontFamily};
+  padding: 30px 0 30px 15px;
+  border-top: 1px solid #e5e5e5;
+  &:hover {
+    background-color: #f8f8f8;
+  }
+  &:hover .collector-intent-checked {
+    display: inline;
+  }
+`
+
+storiesOf("Onboarding", module).add("Collector Intent", () => {
+  return (
+    <div>
+      <div style={{ textAlign: "center", marginBottom: "100px" }}>
+        <Title titleSize="xlarge">Get started on Artsy, what are you most interested in doing?</Title>
+        <Title titleSize="xlarge" style={{ color: "#999999" }}>Select all that apply</Title>
+      </div>
+      <div style={{ width: "450px", margin: "0 auto 100px" }}>
+        <Link href="#">
+          Buy Art & Design
+          <IconContainer className="collector-intent-checked">
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+        <Link href="#">
+          Sell Art & Design
+          <IconContainer className="collector-intent-checked">
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+        <Link href="#">
+          Research Art Prices
+          <IconContainer className="collector-intent-checked">
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+        <Link href="#">
+          Learn About Art
+          <IconContainer className="collector-intent-checked">
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+        <Link href="#">
+          Find Out About New Exhibitions
+          <IconContainer className="collector-intent-checked">
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+        <Link href="#" style={{ borderBottom: "1px solid #e5e5e5", marginBottom: "100px" }}>
+          Read Art Market News
+          <IconContainer className="collector-intent-checked">
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+      </div>
+      <div style={{ textAlign: "center" }}>
+        <Button>Next</Button>
+      </div>
+    </div>
+  )
+})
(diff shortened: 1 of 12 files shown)


src/components/onboarding/wizard.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from "react"
+import Button from "../buttons/default"
+
+interface Props {
+  stepComponents: Array<React.ComponentClass<any>>
```
alloy: Out of curiosity, have you considered using `children` for this?

src/components/onboarding/wizard.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  getCurrentStep(): JSX.Element | null {
+    const currentStep = this.state.currentStep
+    if (currentStep > this.props.stepComponents.length - 1) {
+      return null
+    }
+
+    const Step = this.props.stepComponents[currentStep]
+    return <Step onStateChange={this.onStepStateChanged.bind(this)} ref={step => (this.currentStepComponent = step)} />
+  }
```
alloy: Are the on-boarding steps actually always going to be sequential, or will there be conditional steps like we have in Eigen? E.g. iirc we only ask the user to follow genes if they didn’t follow any artists.

alloy:
The code is looking good, my main question is about those step transitions and if they really are always sequential.

Additionally, have you already tried to run this on top of the existing infra? If not, how are you approaching this design?

src/components/__stories__/collector_intent_page.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+  border-top: 1px solid #e5e5e5;
+  &:hover {
+    background-color: #f8f8f8;
+  }
+  &:hover .collector-intent-checked {
+    display: inline;
+  }
+`
+
+storiesOf("Onboarding", module).add("Collector Intent", () => {
```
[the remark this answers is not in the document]
l2succes: I don't think we need that

src/components/onboarding/selectable_link.tsx
```diff
@@ -0,0 +1,80 @@
...
+class SelectableLink extends React.Component<SelectableLinkProps, LinkState> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      selected: false,
+    }
+    this.onSelect = this.onSelect.bind(this)
+  }
+
+  onSelect(): any {
```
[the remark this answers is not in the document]
l2succes: We forgot to remove it while we trying to fix a TS error it's not needed

src/components/onboarding/wizard.tsx
l2succes: We're passing in an array of component classes, not instances so `children` wouldn't work in this case. The idea is to simplify the wizard API, all you need to pass is an array of classes where the props inherit from `StepProps`.

Example:

```typescript
 <Wizard stepComponents={[CollectorIntent, Artists]} />
```

The type should be more specific though `Array<React.ComponentClass<StepProps>>` and we should rename this property as the name is misleading.

src/components/onboarding/wizard.tsx
l2succes: I'm not sure that's a requirement for this version but if it's something we'd like to do in the future we should take it into account when implementing `nextStep`.

@xtina-starr Do you know if we should have the ability to skip some steps?

src/components/__stories__/collector_intent_page.story.tsx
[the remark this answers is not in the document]
xtina-starr: Yeah we definitely don't need to. I was modeling this after another storybook I saw and it wasn't completely related.

src/components/onboarding/selectable_link.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const state = !this.state.selected
+    this.setState({ selected: state })
+    if (this.props.onSelect) {
+      this.props.onSelect(state)
+    }
+  }
+
+  render(): JSX.Element {
```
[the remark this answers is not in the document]
xtina-starr: We don't :)

src/components/onboarding/wizard.tsx
xtina-starr: @alloy @l2succes the steps are sequential but similar to eigen in that if the user follows 4 or more artists they will skip the categories section. That's noted in the [specs](https://app.zeplin.io/project/59b2e6a22224266545e2aea2/screen/59b2e7a8cfddb7ce01bee3d5).

src/components/onboarding/wizard.tsx
alloy: Makes sense 👍

src/components/__stories__/collector_intent_page.story.tsx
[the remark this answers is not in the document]
alloy: Okeydokey. Still interesting why others then do pass it

src/components/onboarding/wizard.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const stepIndex = this.state.currentStep + 1
+    this.setState({ currentStep: stepIndex })
+  }
+
+  render() {
+    const step = this.getCurrentStep()
+    return (
+      <div>
+        {step}
+      </div>
```
alloy: Are these actually needed or were they previously only added because this was returning a list of components?

l2succes: I can't push to this branch as it's on @xtina-starr's fork so I'll create a separate PR but here's what the code actually looks like.

```typescript
  render() {
    const step = this.getCurrentStep()
    return (
      <div>
        {step}
        <Button disabled={!this.state.nextButtonEnabled} onClick={this.onNextButtonPressed.bind(this)}>
          Next
        </Button>
      </div>
    )
  }
```

l2succes:
Closing in favor of https://github.com/artsy/reaction/pull/258

l2succes:
Sorry! didn't notice the refactoring work y'all did 🙏🏿 
I'll merge the branches together.

xtina-starr:
@l2succes wondering what we might want to do to combine what you've done and the refactor implemented yesterday?

xtina-starr:
@l2succes you mentioned you did some styling on you branch. I finished up some styling as well. When you are back online it might be worth pulling this branch down to make sure all the styling items are covered and this should be good to merge.

I'm also thinking you could implement the submission of the data (via metaphysics) but that can be in a separate PR.

l2succes:
@xtina-starr Yes, styling fixes and cleaning up the collector intent step storybook. I'll go ahead and merge your PR with the refactoring and just apply the relevant changes on top of master.

Sounds good for the data submission in the next PR too.

PR #254, opened 2017-09-28
@eessex => [Publishing] Display panel

kanaabe:
Closes https://github.com/artsy/publishing/issues/38

display-panel

### DIFF
--- a/src/components/publishing/display/display_panel.tsx
+++ b/src/components/publishing/display/display_panel.tsx
@@ -1,4 +1,3 @@
-import * as _ from "lodash"
 import React from "react"
 import styled from "styled-components"
 import Colors from "../../../assets/colors"
@@ -14,7 +13,7 @@ const DisplayPanel: React.SFC<DisplayPanelProps> = props => {
   const { unit, campaign } = props
   return (
     <DisplayPanelContainer>
-      <Image src={unit.image_url} />
+      <Image src={crop(unit.image_url, { width: 680, height: 284 })} />
       <Headline>{unit.headline}</Headline>
       <Body dangerouslySetInnerHTML={{ __html: unit.body }} />
       <SponsoredBy>{`Sponsored by ${campaign.name}`}</SponsoredBy>
@@ -27,11 +26,13 @@ const DisplayPanelContainer = styled.div`
   flex-direction: column;
   border: 1px solid ${Colors.grayRegular};
   padding: 20px;
-  width: 360px;
+  max-width: 360px;
+  box-sizing: border-box;
 `
 const Image = styled.img`
   display: block;
   margin-bottom: 15px;
+  width: 100%;
   height: 142px;
   object-fit: cover;
 `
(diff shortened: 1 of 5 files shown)


src/components/publishing/fixtures/components.ts
```diff
@@ -79,6 +79,17 @@ export const Authors = [
   },
 ]
```
eessex: For the canvas we need the unit fields to be slightly different -- here is what I've been using, let me know if this looks ok to you?
https://github.com/eessex/reaction/blob/display-canvas/src/components/publishing/fixtures/components.ts#L86

kanaabe: This makes sense for the Direct layout but make sure it's flexible enough to support the other layouts -- seems like we'd want to rename this to `UnitCanvasDirect` and also include a `layout` field?

eessex: Yeah, I haven't looked fully into the other layouts yet.  We will need some kind of image-array too.  But they all are in the same database, so we should match the fields.  Specifically brought this up because of the callout link.  It is inline html in the body in this PR, but I think makes sense to separate it out for all units.

eessex: And yeah - layout field makes sense!

kanaabe: I see what you mean with the inline html. My assumption was that this link could exist anywhere in the body which is why it's currently inline. I'll ask Owen if this link always exists at the end and if it's meant to be the same kind of link in Canvas.

kanaabe: Ok so I confirmed with Owen that we should keep it inline since we anticipate clients wanting to have that flexibility. The canvas will be the same, so tag-wise, it will be similar to `postscript` where we have some basic functionality like `<p>` and `<a>`

eessex: Hm in the canvas I was told the entire block is a link this morning, so maybe the broken out link field is specifically for that layout.

src/components/publishing/display/display_panel.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+const DisplayPanel: React.SFC<DisplayPanelProps> = props => {
+  const { unit, campaign } = props
+  return (
+    <DisplayPanelContainer
+      imageUrl={crop(unit.image_url, { width: 680, height: 284 })}
+      hoverImageUrl={resize(unit.logo, { width: 680 })}
+    >
+      <Image />
```
eessex: To return to the link idea, would assume that this Image is meant to link out too

src/components/publishing/fixtures/components.ts
kanaabe: I think there's some confusion with the design - lets get together with Owen and discuss.

kanaabe:
Allllrighty, panel is now 🎁  @eessex

PR #255, opened 2017-09-29
@eessex => [Publishing] Feature Header QA

kanaabe:
Addresses most of the points in: https://github.com/artsy/publishing/issues/52
I also broke out the header tests to their own types -- I think adding some unit tests to the main header.tsx will be helpful as well. 

TODO
~~- [ ] Updates to share components~~ will address in separate PR
- [x] Handle header heights more gracefully when there are nav bars present
- [x] Better unit testing in the main header.tsx where we check children


src/components/publishing/header/__test__/classic_header.test.tsx
```diff
@@ -0,0 +1,44 @@
+import "jest-styled-components"
```
eessex: These tests are much better!

eessex:
This looks great from here, merging!

PR #258, opened 2017-10-04
[Onboarding] Wizard and Collector Intent step

l2succes:
continuation of https://github.com/artsy/reaction/pull/253. This is ready to merge, we can do the TODO items in subsequent PRs.

But just to recap here. we have a `Wizard` component which is a parent container for the onboarding flow. The `Step` component is a base component to encapsulate shared styling and logic between steps (e.g. title, subtitle, margins, next event, backend call, etc...). 

Did some styling and code cleanup work on the Wizard and CollectorIntent components.
It also includes some fixes for relay related compiler errors.

TODO:

- [x] add POST call to send collector intent and to update the user model in gravity
- [ ] integrate with /personalize app in force. (We'll need to render our root onboarding component using [stitch](https://github.com/artsy/stitch) in the [route handler](https://github.com/artsy/force/blob/8b9ee8ab9c38cbf679719edc0f88aaa27cb30b10/mobile/apps/personalize/routes.coffee) based on our A/B test flag.)

For the force integration, here's an example of what rendering a react root component in express looks like (https://github.com/artsy/stitch#expressjs-and-pug)


l2succes:
@xtina-starr Do you know what the collector intent endpoint and data format look like?
If you have a reference I can add it to finish up the collector intent step.

src/components/onboarding/steps/collector_intent/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ]
+
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      selectedOptions: {},
+      selectedCount: 0,
+    }
+  }
```
damassi: I always call this out, but there's no need to use constructors in TypeScript because ES7 class properties are supported out of the box:

```js
class Foo extends Component {
  state = {
    selectedOptions: {},
    selectedCount: 0,
  }

  ... 
}

File: src/components/onboarding/steps/collector_intent/option.tsx [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+  }
+`
+
+class Option extends React.Component<OptionProps, LinkState> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      selected: false,
+    }
+    this.onSelect = this.onSelect.bind(this)
```
damassi: No need to bind event methods in this way -- again, ES7 classProperties allow you to write:

```js
class Option extends React.Component<OptionProps, LinkState> {
  onSelect = (event) => {
    ...
  }
}
```
The fat arrow binds to the outer scope.

src/components/onboarding/steps/collector_intent/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    let selectedOptions = this.state.selectedOptions
+    selectedOptions[option] = selected
+
+    let count = 0
+    for (let key in selectedOptions) {
+      if (selectedOptions[key]) count++
+    }
+
+    this.setState({ selectedOptions, selectedCount: count })
+    this.props.onStateChange({ nextButtonEnabled: count > 0 })
```
damassi: Would be worth exploring Redux here, because I bet this will only scale outwards with time. Then there's no need to pass things from component to component because you can just 

```js
@connect(state => ({
  selectedOptions: state.onboarding.selectedOptions
})
class MyComponent extends Component {
   ... 
} 
```
and keep things loosely coupled. There is a lot of FUD around Redux's boilerplate, but @sweir27 built something similar to this recently and it worked really well. Might be worth pinging her for more information on the overall flow.

src/components/onboarding/steps/collector_intent/option.tsx
alloy: This was not stage 3 until very recently, before then it was still causing issues (such as not working well with method decorators), so it would be great if we can first verify that all the compilers have ironed out bugs since it reached stage 3.

damassi: This has been in TypeScript for a few years now, and you can verify it works by trying it in whatever compnoent :)

src/components/onboarding/steps/collector_intent/index.tsx
alloy: I agree that it’s likely that this state machine will grow in the future and this is one of the few situations where I think the amount of dealing with state would benefit from using redux, but I’m not sure we should add it right away, it’s probably straight forward for now. So I’d say we convert to redux either once it gets very complicated very soon or once we’ve delivered the initial version for the test and we start work on the fully featured and flexible on-boarding framework.

src/components/onboarding/steps/collector_intent/option.tsx
alloy: The situation I’m referring to was just a month ago, though, and class properties only reached stage 3 a month or so ago as well.

alloy: (See the ‘important’ section here https://github.com/artsy/emission/pull/693 /cc @mennenia)

orta: I am sad about this too

src/components/onboarding/steps/collector_intent/index.tsx
sweir27: FWIW I wouldn't wait tooooo long. I've found that it's not too bad to get set up and once you _are_ set up, it increases productivity/testability/etc. immensely!

src/components/onboarding/steps/collector_intent/option.tsx
damassi: Aw damn -- I see what you mean. We should keep an eye on this.

alloy: ;/

src/components/onboarding/steps/collector_intent/index.tsx
damassi: This is totally off the top of my head, but once you have the basic skeleton (which isn't much) your app will pretty much scale out infinitely: 

```javascript
// actions.js 

const actions = {
  NEXT_PAGE: 'NEXT_PAGE',
  PREVIOUS_PAGE: 'PREVIOUS_PAGE'
}

const nextPage = () => ({
  type: actions.NEXT_PAGE
})

const previousPage = () => ({
  type: actions.PREVIOUS_PAGE
})


// reducer.js 

const initialState = {
  page: 0,
  totalPages: 10
}

const onboardingReducer = (state = initialState, action) => {
  switch (action.type) {
    case actions.NEXT_PAGE: {
      const page = state.page + 1
      const nextButtonEnabled = page < state.totalPages 

      return u({
        page,
        nextButtonEnabled
      }, state)
    }

    case actions.PREVIOUS_PAGE: {
      const page = state.page - 1
      const nextButtonEnabled = page > 0

      return u({
        page,
        nextButtonEnabled
      }, state)
    }
}

const reducers = composeReducers({ 
  onboarding: onboardingReducer
})

const middleware = [
  createLogger({ 
    collapsed: true 
  })
]

const store = createStore(reducer, middleware) 

// App.js

const App = () => 
  <Provider store={store}>
    <Layout />
  </Provider>

// Layout.js

@connect(state => ({
  nextButtonEnabled: state.onboarding.nextButtonEnabled,
  page: state.onboarding.page
}), {
  nextPageAction: nextPage,
  previousPageAction: prevPage
})
class Layout extends Component {

  render () {
    const {
      nextButtonEnabled, 
      nextPageAction,
      previousPageAction,
      page
    } = this.props

    return (
      <div>
        <div>
          Hello, you are on step {page}
        </div>

        <NextButton 
          enabled={nextButtonEnabled} 
          onClick={nextPageAction}
        />

        <PreviousButton 
          onClick={previousPageAction}
        />
      </div>
    )
  }
}

// Further down in the tree, other components can just "pull" from store state via @connect
// which makes things very portable and light-weight, lending to more refined (and rapid)
// refactors later on. 

```

Like @sweir27 mentioned above, a little goes a very long way and untangling later (since you know it _will_ be scaling in complexity and scope) might only serve to discourage refactors. 

Anyways, just my 2 cents! Carry on 😄

yuki24: I didn't know about the `@connect` annotation, that's a game changer...

damassi: Yeah! pretty much any higher-order-component can be used as a decorator. For example, [this bit here](https://github.com/artsy/force/blob/master/desktop/apps/consign/components/sign_up/index.js#L110) can easily be rewritten like 

```javascript
@reduxForm({
  form: 'signUp', 
  validate
}),
@connect(
  mapStateToProps,
  mapDispatchToProps
)
export default class SignUp extends Component {
  ... 
}
```

While I'm here (sorry sorry...) another thing I want to emphasize is now nice Redux's [Logger middleware](https://github.com/evgenyrodionov/redux-logger) is to use during development. There's literally zero ambiguity about whats going on when and where and this immediately translates into increased development speed.

src/components/onboarding/steps/collector_intent/index.tsx
l2succes: Thanks for pointing that out @damassi. FWIW totally aware of ES7 class properties support in TS and maybe an unfounded concern of mine but I tend to sway towards staying consistent throughout to avoid confusion as there are components that need to initialize their state based on props in the project. I'm also a fan of less boilerplate so it's a good idea to just start using it.

alloy:
As [mentioned in Slack](https://artsy.slack.com/archives/C2TQ4PT8R/p1507640935000564), I’m closing this PR as it seems things have gone lost in an earlier merge. Afaik there are only [small actual changes](https://github.com/artsy/reaction/pull/253#issuecomment-334352057) needed, @l2succes will annotate those here so that @xtina-starr can recreate them on a new branch based off of master.

src/components/onboarding/steps/collector_intent/__stories__/collector_intent_page.story.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+storiesOf("Onboarding", module).add("Collector Intent", () => {
+  const currentUser = {
+    accessToken: "blah",
+    id: "my_id",
+    name: "Joe",
+  }
+  return (
+    <ContextProvider currentUser={currentUser}>
+      <CollectorIntent />
```
l2succes: @xtina-starr This part updates the story to use the `CollectorIntent` component so we should add that back to master.

PR #259, opened 2017-10-04
Add ability to set a custom color for the Title tag

yuki24:
In order to implement https://github.com/artsy/collector-experience/issues/505 we need to be able to change the color of a title.

screen shot 2017-10-04 at 15 18 35

### DIFF
--- a/src/components/title.tsx
+++ b/src/components/title.tsx
@@ -7,6 +7,7 @@ type TitleSize = "small" | "medium" | "large" | "xlarge" | "xxlarge"
 
 interface TitleProps extends React.HTMLProps<HTMLDivElement> {
   titleSize?: TitleSize
+  color?: string
 }
 
 const titleSizes = {
@@ -30,6 +31,7 @@ const Title: React.SFC<TitleProps> = props => {
 
 const StyledTitle = styled(Title)`
   font-size: ${props => titleSizes[props.titleSize]};
+  color: ${props => props.color};
   margin: 20px 0;
   ${fonts.secondary.style}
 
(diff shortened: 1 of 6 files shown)


src/components/title.tsx
```diff
@@ -40,6 +42,7 @@ const StyledTitle = styled(Title)`
 
 StyledTitle.defaultProps = {
   titleSize: "medium",
+  color: "currentcolor",
```
l2succes: the default `color` should be assigned to a valid css value. Like `black` or `#000`.

yuki24: It does seem like `currentcolor` is a valid css value, but at the same time, this is not how it's supposed to be used, according to [CSS Tricks](https://css-tricks.com/currentcolor/).

> you can use this value to indicate you want to use the value of color for other properties that accept a color value: borders, box shadows, outlines, or backgrounds.

How about `color: inherit;`? I'm a bit hesitant to assign a default color because some times I want to use a wrapper with a color for the entire box rather than setting a color to each element:

```jsx
<Text align="center" color={colors.graySemibold}>
  <Title>
    Contact Us
  </Title>

  <Text>
    Have questions? Get in touch:&nbsp;
    <TextLink href="#">youremail@example.com</TextLink>
  </Text>
</Text>
```

l2succes: Ah ok! I stand corrected then, never seen it before 😅. 
Right, `color: inherit` would give us the behavior you described. Let's go with that

l2succes:
@yuki24 Feel free to merge after updating that default value.

yuki24:
merging

PR #260, opened 2017-10-04
Add ability to set leftView to input

yuki24:
Similar to the `rightView` option for `Input`, but this adds the ability to assign a view to the left of the input. I need this in order to implement: https://app.zeplin.io/project/59b2e6a22224266545e2aea2/screen/59b2e7a8968341983ee82a9b

screen shot 2017-10-04 at 16 13 36

I also added two examples of an `Input` with `leftView`/`rightView` options.


l2succes:
👍

PR #263, opened 2017-10-04
@kanaabe => Display Canvas

eessex:
- Adds Display canvas component with layouts: Overlay, Standard (Image/Video), and Slideshow.
- Adds 'assets' and 'type' field to display unit model
- Adds font size Garamond s11
- Adds new breakpoint (1150), and changes `pMedia.lg` occurrences to xl

screen shot 2017-10-05 at 10 09 24 am
screen shot 2017-10-05 at 10 09 31 am
screen shot 2017-10-05 at 10 09 39 am
screen shot 2017-10-05 at 10 09 46 am
screen shot 2017-10-05 at 10 08 51 am
screen shot 2017-10-05 at 10 08 59 am
screen shot 2017-10-05 at 10 09 05 am
screen shot 2017-10-05 at 10 09 12 am

### DIFF
--- a/src/components/publishing/display/canvas/canvas_video.tsx
+++ b/src/components/publishing/display/canvas/canvas_video.tsx
@@ -4,9 +4,6 @@ import { pMedia } from "../../../helpers"
 
 interface VideoProps {
   src: any
-  size?: {
-    width: number
-  }
 }
 
 class DisplayCanvasVideo extends React.Component<VideoProps, any> {
--- a/src/components/publishing/display/canvas/container.tsx
+++ b/src/components/publishing/display/canvas/container.tsx
@@ -1,0 +1,141 @@
+import React from "react"
+import styled, { StyledFunction } from "styled-components"
+import { crop } from "../../../../utils/resizer"
+import { pMedia } from "../../../helpers"
+import DisplayCanvasText from "./text"
+
+interface DisplayCanvasContainerProps extends React.HTMLProps<HTMLDivElement> {
+  unit: any
+  disclaimer: any
+}
+
+interface DivProps extends React.HTMLProps<HTMLDivElement> {
+  backgroundUrl?: string
+}
+
+function isVideo(url) {
+  return url.includes("mp4")
+}
+
+function renderAsset(asset) {
+  if (isVideo(asset.url)) {
+    return (
+      <VideoContainer>
+        <Video src={asset.url} autoPlay controls={false} loop muted playsInline />
+      </VideoContainer>
+    )
+  } else {
+    return <Image src={crop(asset.url, { width: 1200, height: 760 })} />
+  }
+}
+
+const DisplayCanvasContainer: React.SFC<DisplayCanvasContainerProps> = props => {
+  const { unit, disclaimer } = props
+
+  if (unit.layout === "overlay") {
+    return (
+      <Canvas href={unit.link.url} target="_blank">
+        <Background backgroundUrl={unit.assets[0].url} />
+        <DisplayCanvasText unit={unit} />
+      </Canvas>
+    )
+  } else if (unit.layout === "slideshow") {
+    return (
+      <Canvas href={unit.link.url} target="_blank">
+        <DisplayCanvasText unit={unit} disclaimer={disclaimer} />
+      </Canvas>
+    )
+  } else {
+    return (
+      <Canvas href={unit.link.url} target="_blank">
+        {renderAsset(unit.assets[0])}
+        <StandardContainer>
+          <DisplayCanvasText unit={unit} disclaimer={disclaimer} />
+        </StandardContainer>
+      </Canvas>
+    )
+  }
+}
+
+const Div: StyledFunction<DivProps> = styled.div
+
+const Image = styled.img`
+  display: block;
+  width: 65%;
+  height: 100%;
+  max-height: 460px;
+  object-fit: cover;
+  ${pMedia.sm`
+    width: 100%;
+    max-width: 100%;
+    height: auto;
+    max-height: initial;
+    object-fit: contain;
+  `}
+`
+const Canvas = styled.a`
+  width: 100%;
+  height: 460px;
+  color: #000;
+  text-decoration: none;
+  display: flex;
+  flex-direction: row-reverse;
+  justify-content: center;
+  align-items: center;
+  position: relative;
+  ${pMedia.sm`
+    height: 400px;
+    flex-direction: column;
+    justify-content: flex-start;
+  `}
+`
+const StandardContainer = styled.div`
+  max-width: 35%;
+  height: 100%;
+  ${pMedia.sm`
+    max-width: 100%;
+    height: auto;
+  `}
+`
+const Background = Div`
+  background: black;
+  position: absolute;
+  left: 0;
+  right: 0;
+  top: 0;
+  bottom: 0;
+  z-index: -1;
+  &:before {
+    content: '';
+    background-image: url(${props => props.backgroundUrl && props.backgroundUrl});
+    position: absolute;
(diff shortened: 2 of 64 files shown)


src/components/publishing/display/canvas/slideshow.tsx
```diff
@@ -0,0 +1,217 @@
...
+      onClick={props.onClick}
+      isVisible={!props.isOnTitle}
+    >
+      <Icon name="chevron-left" color="black" fontSize="24px" />
+    </NavArrow>
+  )
+}
+
+const RightArrow = props => {
+  const onClick = props.currentSlide + 1 === props.slideCount ? props.onRightArrow : props.onClick
```
kanaabe: Can you explain the reasoning for rewriting the infinite scroll functionality in desktop view? I believe it has to do with the title card, but I'm not understanding why we can't use `infinite: true` anyway.

src/components/publishing/display/canvas/slideshow.tsx
```diff
@@ -0,0 +1,217 @@
...
+      },
+      responsive: [
+        {
+          breakpoint: 900,
+          settings: {
+            infinite: true,
+          },
+        },
+      ],
+    }
```
kanaabe: Nice!

src/components/publishing/display/canvas/slideshow.tsx
```diff
@@ -0,0 +1,217 @@
...
+const arrowDiv: StyledFunction<NavArrowProps> = styled.div
+const responsiveDiv: StyledFunction<ResponsiveProps> = styled.div
+const responsiveImage: StyledFunction<ResponsiveProps> = styled.img
+
+const NavArrow = arrowDiv`
+  display: flex;
+  align-items: center;
+  position: absolute;
+  height: 100%;
+  max-height: ${props => props.containerWidth * 0.65 * 0.59 + "px;"}
```
kanaabe: These values seem arbitrary to me -- what are they representing? Maybe we can extract some mystery by making them constants.

src/components/publishing/display/canvas/text.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+interface DivProps extends React.HTMLProps<HTMLDivElement> {
+  layout?: any
+  isSlideshowWithCaption?: boolean
+}
+
+const DisplayCanvasText: React.SFC<DisplayCanvasTextProps> = props => {
+  const { disclaimer, unit } = props
+  const isSlideshowWithCaption = unit.layout === "slideshow" && compact(map(unit.assets, "caption")).length
```
kanaabe: Would the text layout have a slideshow?

src/components/publishing/display/canvas/slideshow.tsx
eessex: VariableWidth (needed for the title card) creates some bugs when combined with infinite mode -- https://github.com/kenwheeler/slick/issues/1207

We don't need VariableWidth in mobile, so it is ok to use infinite there (also just realizing I forgot to remove variableWidth in the responsive section).

src/components/publishing/display/canvas/slideshow.tsx
eessex: A constant might be clearer here -- .65 is the max width of an image, .59 is the aspect ratio.

src/components/publishing/display/canvas/text.tsx
kanaabe: Hm actually I seem to be misunderstanding how this component gets used.

eessex: This is the text of the unit, not a text layout.  It is used in all layouts.

src/components/publishing/display/canvas/container.tsx
```diff
@@ -0,0 +1,149 @@
+import React from "react"
+import sizeMe from "react-sizeme"
+import styled, { StyledFunction } from "styled-components"
+import { crop } from "../../../../utils/resizer"
+import { pMedia } from "../../../helpers"
+import { sizeMeRefreshRate } from "../../constants"
+import Slideshow from "./slideshow"
+import CanvasText from "./text"
```
kanaabe: Ah I see now. Maybe this is more a preference thing, but if you are exporting the component as `DisplayCanvasText`, it might be helpful to use the same name when using the component in here as well. I'd also say I tend to make filenames consistent with the component name, so naming it something like `display_canvas_text.tsx` might be clearer.

src/components/publishing/display/canvas/slideshow.tsx
kanaabe: Ok yeah, I see it being used throughout the file so a constant makes sense here.

src/components/publishing/display/canvas/text.tsx
```diff
@@ -0,0 +1,112 @@
...
+  line-height: ${props => (props.layout === "overlay" ? "1.35em;" : "1.1em;")} 
+  ${props => pMedia.lg`
+    ${props.layout !== "overlay" && Fonts.garamond("s23")}
+  `}
+  ${props => pMedia.sm`
+    ${props.layout === "overlay"
+      ? `margin: 0 auto 10px auto;
+         padding: 20px;`
+      : `margin-bottom: 10px;
+         max-width: calc(100% - 40px);`}
```
kanaabe: This totally works, but just as a suggestion I've found that when styled-component prop logic gets wild, it helps to move logic into a function like, `getMobileHeadlineStyle(props)` which makes it a bit easier on the eye.

src/components/publishing/display/canvas/video.tsx
```diff
@@ -0,0 +1,94 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  z-index: 2;
+  cursor: pointer;
+  border: 0;
+  outline: 0;
+`
+
+export default DisplayCanvasVideo
```
kanaabe: Great! I would love to see us export small components like PlayButton which is used in the Video component as well. Doesn't have to be now but it would be good to consider!

src/components/publishing/display/canvas/index.tsx
```diff
@@ -0,0 +1,64 @@
+import React from "react"
+import styled, { StyledFunction } from "styled-components"
+import Colors from "../../../../assets/colors"
+import { pMedia } from "../../../helpers"
+import Fonts from "../../fonts"
+import Canvas from "./container"
```
kanaabe: Ya I think this is another instance where the naming is a little confusing. Canvas and container seem like two different things when they are not.

kanaabe:
Woohoo! Great job on this 👏  Thats a ton of functionality in one PR and beautiful screenshots. There are a few comments and a merge conflict so lmk when you're feeling this is 100% to you and I can merge/help resolve anything else.

orta:
Yep 👍 on a great PR to read

src/components/publishing/display/canvas/canvas_video.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+class DisplayCanvasVideo extends React.Component<VideoProps, any> {
+  constructor(props) {
+    super(props)
+    this.state = { isPlaying: false }
+  }
+
+  onPlayVideo = e => {
+    e.preventDefault()
+    e.target.paused ? e.target.play() : e.target.pause()
```
damassi: I would be really careful using ternaries in this way, especially without semicolons -- they're typically for variable assignment and can lead to strange side-effects if used otherwise. 

```js
if (e.target.paused) {
  e.target.play()
} else {
  e.target.pause()
}
```
In most eslint settings theres a rule that prevent that kind of use and enforce the above; we should probably turn it on in tslint.

PR #264, opened 2017-10-04
[WIP] Onboarding follow artists

xtina-starr:
For what it's worth this is probably throw away code but is an example of how we might want to approach fetching popular artists from metaphysics. @orta and I paired on [adding popular artists to metaphysics](https://github.com/artsy/metaphysics/pull/770), but it's not yet been merged so I can't see if this actually works yet. If it's still now merged by the time you start work tomorrow perhaps you could use fixture data to build out the UI.

If anything, I'm hoping this unblocks Yuki to be able to continue work on the Artist Follow view of the onboarding flow with this as a sudo example and the popular artists data in flight in metaphysics.

### DIFF
--- a/src/components/__stories__/collector_intent_page.story.tsx
+++ b/src/components/__stories__/collector_intent_page.story.tsx
@@ -1,0 +1,87 @@
+import { storiesOf } from "@storybook/react"
+import * as React from "react"
+import styled from "styled-components"
+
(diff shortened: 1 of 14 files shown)


alloy:
Closing this for now, as I _think_ it was mainly an example for @yuki24 and he already has the code to fetch popular artists in his PR.

Please follow-up if there was more code here that needs to go into master.

PR #265, opened 2017-10-05
Update Graphql schema

yuki24:
I'm not familiar with GraphQL, so please correct me if I'm wrong or missing. It seems like running `yarn sync-schema` and making a commit for the diff is necessary to use https://github.com/artsy/metaphysics/pull/770 in reaction.

Either way, the current schema is 3months old, so we should update the schema.


alloy:
Sure, updating the schema won’t hurt 👍

PR #266, opened 2017-10-05
@kanaabe: Added tracking for publishing components

craigspaeth:
This attempts to fill out the tracking calls for publishing components for https://github.com/artsy/publishing/issues/50

There were a handful of tracking calls that I couldn't find homes for in the Reaction components.

Specified in the [analytics spreadsheet](https://docs.google.com/spreadsheets/d/17oqFli15x1SmNBu_-fPZhFifMw2BNSPJe9VdOgs1aKs/edit#gid=1499052602):
1. `Sign up for editorial email` 
2. `Clicked partner logo` (Venice 360)
3. `Video play` (Venice 360)
4. `Video duration` (Venice 360)

Seems the last three are irrelevant right now, but #1 I'm curious if we need to re-implement that component?

These events I saw in our current analytics code but didn't find homes for them, nor did I see them in the analytics doc

- `Clicked primary partner logo` (Super article)
- `Clicked secondary partner logo`  (Super article)
- `Clicked partner cta link` (Super article)
- `Clicked partner cta link in footer blurb` (Super article)
- `Clicked Read More`
- `Sign up for gallery insights email`
- `Dismiss editorial signup`
- `Clicked Venice 2017 Banner link`

So this is probably incomplete, but it should be a good starting point and maybe I can sync up with @wrgoldstein on ensuring we have our bases covered.

### DIFF
--- a/src/components/publishing/article.tsx
+++ b/src/components/publishing/article.tsx
@@ -34,7 +34,7 @@ interface ArticleState {
   isTruncated: boolean
 }
 
-@track({}, { dispatch: data => Events.postEvent(data) })
+@track({ page: "Article" }, { dispatch: data => Events.postEvent(data) })
 class Article extends React.Component<ArticleProps, ArticleState> {
   static childContextTypes = {
     onViewFullscreen: PropTypes.func,
@@ -51,6 +51,18 @@ class Article extends React.Component<ArticleProps, ArticleState> {
       isTruncated: props.isTruncated || false,
     }
   }
+
+  @track((props, [e]) => ({
+    action: "Article Impression",
+    article_id: props.article.id,
+    destination_path: window.location.pathname,
+    // TODO: What do we put here? According to analytics there are the valid types
+    // [newsletter signup, toc, artist follow, image set, article callout, social, related article]
+    impression_type: "related_article",
+    context_type: "article_fixed",
+  }))
+  // tslint:disable-next-line:no-empty
+  componentDidMount() {}
 
   getChildContext() {
     return { onViewFullscreen: this.openViewer }
--- a/src/components/publishing/article.tsx
+++ b/src/components/publishing/article.tsx
@@ -52,18 +52,6 @@ class Article extends React.Component<ArticleProps, ArticleState> {
     }
   }
 
-  @track((props, [e]) => ({
-    action: "Article Impression",
-    article_id: props.article.id,
-    destination_path: window.location.pathname,
-    // TODO: What do we put here? According to analytics there are the valid types
-    // [newsletter signup, toc, artist follow, image set, article callout, social, related article]
-    impression_type: "related_article",
-    context_type: "article_fixed",
-  }))
-  // tslint:disable-next-line:no-empty
-  componentDidMount() {}
-
   getChildContext() {
     return { onViewFullscreen: this.openViewer }
   }
--- a/src/components/publishing/related_articles/related_article_figure.tsx
+++ b/src/components/publishing/related_articles/related_article_figure.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import styled from "styled-components"
 import { crop } from "../../../utils/resizer"
+import { track } from "../../../utils/track"
 import { pMedia } from "../../helpers"
 import Byline from "../byline/byline"
 import { articleHref } from "../constants"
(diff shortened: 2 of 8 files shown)


src/components/publishing/article.tsx
```diff
@@ -52,6 +52,18 @@ class Article extends React.Component<ArticleProps, ArticleState> {
     }
   }
 
+  @track((props, [e]) => ({
+    action: "Article Impression",
+    article_id: props.article.id,
+    destination_path: window.location.pathname,
+    // TODO: What do we put here? According to analytics there are the valid types
+    // [newsletter signup, toc, artist follow, image set, article callout, social, related article]
+    impression_type: "related_article",
```
craigspaeth: I'm not sure what to do about impression type here? This fires when the page loads essentially.

src/typings/react-tracking.d.ts
```diff
@@ -10,47 +10,47 @@ declare module "react-tracking" {
...
   type Falsy = false | null | undefined | ""
 
   interface Options<T> {
     /**
-       * By default, data tracking objects are pushed to `window.dataLayer[]`. This is a good default if you use Google
-       * Tag Manager. You can override this by passing in a dispatch function as a second parameter to the tracking
-       * decorator `{ dispatch: fn() }` on some top-level component high up in your app (typically some root-level
-       * component that wraps your entire app).
-       */
+     * By default, data tracking objects are pushed to `window.dataLayer[]`. This is a good default if you use Google
```
craigspaeth: Got a TSLint warning about lining these up so 🤷‍♂️

src/components/publishing/article.tsx
kanaabe: Ah so these should fire each time an interactive component comes into the viewport, but only once per component. By interactive component, I mean anything with a link, signup, or button (i.e video, image set, view fullscreen, etc). We use this in combination with click tracking to determine engagement on parts of an article. 

The tricky thing is knowing when the component is in view with React. We do it this crazy way with jQuery currently: https://github.com/artsy/force/blob/master/desktop/analytics/article_impressions.js

Either way it seems like there needs to be clarification of _which_ components we need impression/click tracking on. Feel free to remove this part for now and I can refine that list later with owen or will.

craigspaeth: Ahhh, that makes more sense now 😛  I'll remove this, and no worries—I'll sync up with Will/Owen on clarification.

src/components/publishing/related_articles/related_article_figure.tsx
```diff
@@ -16,17 +17,37 @@ interface RelatedArticleFigureProps extends React.HTMLProps<HTMLDivElement> {
...
+    super()
+    this.onClick = this.onClick.bind(this)
+  }
+
+  @track((props, [e]) => ({
+    action: "Clicked article impression",
+    article_id: props.article.id,
+    destination_path: e.currentTarget.attributes.href.value,
+    // TODO: What do we put here? According to analytics there are the valid types
+    // [newsletter signup, toc, artist follow, image set, article callout, social, related article]
```
kanaabe: I think we'll want to use a new one here, something like `related articles canvas` since this gets used by RelatedArticlesCanvas, but owen/will may have their own prefs.

src/components/publishing/sections/view_fullscreen.tsx
```diff
@@ -1,28 +1,39 @@
...
+    onViewFullscreen: PropTypes.func,
+  }
+
+  constructor(props) {
+    super()
+    this.onClick = this.onClick.bind(this)
+  }
+
+  @track({ action: "Clicked article impression" })
+  onClick(e) {
```
kanaabe: Nit pick, this could be `onClick = (e) => {}` to avoid that `.bind`

craigspaeth: For some reason I had trouble getting our static analysis tools to not complain using `onClick = () => {}`, but I'll give it another shot and see if I missed something/follow up with a better explanation why.

craigspaeth: Yeah when using decorators on class-level arrow functions I get `[ts] Unable to resolve signature of property decorator when called as an expression.` in VSCode and `Cannot read property 'configurable' of undefined
TypeError: Cannot read property 'configurable' of undefined` in Storybooks 😒 

There's probably a somewhat reasonable explanation for this with how `onClick = () =>` is assigning a field on the class vs. a method for the "prototype"... but 🤷‍♂️ these edge cases/complexity is what makes me resent all of the OO language feature 🍝  being added to Javascript /rant.

orta: Yeah, we've had the same issue in emission - just annoying

craigspaeth:
Alrighty, updated some things in line with your comments and I blocked out some time with @wrgoldstein tomorrow to generally go over tracking calls and figure out what else to add. I'll put up a separate PR after that with the rest of the tracking we come up with.

kanaabe:
👏  Woohoo, thanks!

src/components/publishing/sections/view_fullscreen.tsx
alloy: This is a matter of ES7 features being new still, static properties has only recently reached stage 3.

PR #269, opened 2017-10-09
[Onboarding] Budget screen views

mennenia:
Made with the 🎉 @xtina-starr 

Issue [here](https://github.com/artsy/collector-experience/issues/507). 

Currently looks like:

screen shot 2017-10-09 at 18 13 19

Still needs:
- [ ] Icon
- [ ] Potential refactor / cleanup? Some overlap code with `selectable_link.tsx`

We've just added a few reviewers just for some extra eyes 😄

### DIFF
--- a/src/components/onboarding/selectable_toggle.tsx
+++ b/src/components/onboarding/selectable_toggle.tsx
@@ -1,0 +1,71 @@
+import * as React from "react"
+import styled from "styled-components"
+
+import * as fonts from "../../assets/fonts"
+import Icon from "../icon"
+
+interface SelectableToggleProps {
+  href?: string
+  text?: string
+  onSelect: (selected: boolean) => void
+  selected: boolean
+}
+
+const IconContainer = styled.div`
+width: 15px;
+height: 15px;
+background-color: black;
+display: none;
+border-radius: 50%;
+float: right;
+margin-right: 15px;
+`
+
+const Link = styled.a`
+display: block;
+font-size: 14px;
+color: black;
+text-decoration: none;
+text-transform: uppercase;
+font-family: ${fonts.primary.fontFamily};
+padding: 30px 0 30px 15px;
+border-top: 1px solid #e5e5e5;
+&:hover {
+  background-color: #f8f8f8;
+}
+&:hover .collector-intent-checked {
+  display: inline;
+}
+& .collector-intent-checked.is-selected {
+  display: inline;
+}
+`
+
+class SelectableToggle extends React.Component<SelectableToggleProps, null> {
+  constructor(props) {
+    super(props)
+    this.onSelect = this.onSelect.bind(this)
+  }
+
+  onSelect() {
+    if (this.props.onSelect) {
+      this.props.onSelect(this.props.selected)
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <Link href={this.props.href} onClick={() => this.onSelect()}>
+          {this.props.text}
+
+          <IconContainer className={`collector-intent-checked ${this.props.selected ? "is-selected" : ""}`}>
+            <Icon name="check" color="white" fontSize="8px" />
+          </IconContainer>
+        </Link>
+      </div>
+    )
+  }
+}
+
+export default SelectableToggle
--- a/src/components/onboarding/steps/budget.tsx
+++ b/src/components/onboarding/steps/budget.tsx
@@ -1,0 +1,45 @@
+import * as React from "react"
+import styled from "styled-components"
+
+import SelectableLink from "../selectable_link"
+import { StepProps } from "../types"
+import { Layout } from "./layout"
+
+const OptionsContainer = styled.div`
+width: 450px;
+margin: 0 auto 100px;
+&:last-child {
+  border-bottom: 1px solid #e5e5e5;
+}
+`
+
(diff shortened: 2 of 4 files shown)


mennenia:
btw, @l2succes, I assume you implemented something responsive for Loyalty? We had a quick look through the project, but what's the pattern that you've used there? 
Thanks!

src/components/onboarding/selectable_toggle.tsx
```diff
@@ -0,0 +1,71 @@
...
+border-top: 1px solid #e5e5e5;
+&:hover {
+  background-color: #f8f8f8;
+}
+&:hover .collector-intent-checked {
+  display: inline;
+}
+& .collector-intent-checked.is-selected {
+  display: inline;
+}
```
alloy: It probably reads a bit nicer if all of the above styles are indented by 1 tabstop.

src/components/onboarding/selectable_toggle.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+class SelectableToggle extends React.Component<SelectableToggleProps, null> {
+  constructor(props) {
+    super(props)
+    this.onSelect = this.onSelect.bind(this)
+  }
+
+  onSelect() {
+    if (this.props.onSelect) {
+      this.props.onSelect(this.props.selected)
```
alloy: I’m not sure I understand this. I assume that the boolean being passed to the `onSelected` callback is supposed to reflect the new `selected` status? How does `this.props.selected` get updated after selecting an option?

src/components/onboarding/selectable_toggle.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <Link href={this.props.href} onClick={() => this.onSelect()}>
+          {this.props.text}
+
+          <IconContainer className={`collector-intent-checked ${this.props.selected ? "is-selected" : ""}`}>
```
alloy: You could use dynamic styling based on props instead of working with class names https://www.styled-components.com/docs/basics#attaching-additional-props

src/components/onboarding/steps/budget.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { StepProps } from "../types"
+import { Layout } from "./layout"
+
+const OptionsContainer = styled.div`
+width: 450px;
+margin: 0 auto 100px;
+&:last-child {
+  border-bottom: 1px solid #e5e5e5;
+}
+`
```
alloy: Same as before, could do with some indentation.

src/components/onboarding/steps/budget.tsx
```diff
@@ -0,0 +1,70 @@
...
+export default class Budget extends React.Component<StepProps, State> {
+  options = [
+    "UNDER $500",
+    "UNDER $2,500",
+    "UNDER $5,000",
+    "UNDER $10,000",
+    "UNDER $25,000",
+    "UNDER $50,000",
+    "NO BUDGET IN MIND",
+  ]
```
alloy: Nit: This is an instance variable, whereas this could just be a static constant outside the class.

src/components/onboarding/steps/budget.tsx
```diff
@@ -0,0 +1,70 @@
...
+const OptionsContainer = styled.div`
+width: 450px;
+margin: 0 auto 100px;
+&:last-child {
+  border-bottom: 1px solid #e5e5e5;
+}
+`
+
+interface State {
+  selection: string
```
alloy: Nit: It’s generally less brittle to use index based selections than comparing captions. It’s unlikely that we’ll do this any time soon, but for instance consider that we’d ever want to localise the copy, then this would need to be reworked to take multiple languages into account, whereas indices would be stable.

alloy:
Nice work and pairing 🎉 

Left some small comments, none are deal breakers.

src/components/onboarding/selectable_toggle.tsx
xtina-starr: @alloy the `onOptionSelected()` updates the selection state in the `BudgetComponent` and we are comparing whether the selection string matches the text of the clicked `SelectableToggle` to produce the `selected` prop. In that case, the one selected will be true and the other `SelectableToggle` components will be false. It seems like the BudgetComponent should be the one to know about which option was clicked and just let the SelectableToggle component know if it was selected or not in the form of a prop.

src/components/onboarding/steps/budget.tsx
xtina-starr: Not quite sure I understand what you mean here? Isn't the text being produced based on what the index of the options array is?

src/components/onboarding/steps/budget.tsx
xtina-starr: In the case of a static constant outside the class, what does that look like? Does it then change to be some sort of options prop that get's pass in via the constructor function?

src/components/onboarding/steps/budget.tsx
alloy: Yes it is _now_, but in general comparing indices should be favoured over comparing titles because those are UI level details, whereas indices are logic level details.

alloy:
Gonna merge this now as-is so that I can perform a path case change.

PR #271, opened 2017-10-09
[Publishing] @eessex => Another round of QA

kanaabe:
See commit messages for specific fixes

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,9 +125,9 @@
     "react-styled-flexboxgrid": "^2.0.0",
     "react-tracking": "^4.2.1",
     "react-url-query": "^1.1.4",
-    "request": "^2.81.0",
     "sharify": "^0.1.6",
     "styled-components": "^2.0.1",
+    "superagent": "^3.6.3",
     "uuid": "^3.0.1"
   },
   "graphql": {
--- a/src/components/publishing/read_more/read_more_wrapper.tsx
+++ b/src/components/publishing/read_more/read_more_wrapper.tsx
@@ -18,9 +18,8 @@ class ReadMoreWrapper extends React.Component<ReadMoreWrapperProps, ReadMoreWrap
       truncationHeight: "100%",
     }
   }
-
   componentDidMount() {
-    this.truncateArticle()
+    setTimeout(this.truncateArticle, 250)
     window.addEventListener("resize", this.truncateArticle)
   }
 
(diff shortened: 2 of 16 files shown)


src/components/publishing/read_more/read_more_wrapper.tsx
```diff
@@ -18,9 +18,8 @@ class ReadMoreWrapper extends React.Component<ReadMoreWrapperProps, ReadMoreWrap
       truncationHeight: "100%",
     }
   }
-
   componentDidMount() {
-    this.truncateArticle()
+    setTimeout(this.truncateArticle, 250)
```
kanaabe: This is temporary -- this `componentDidMount` is getting fired before `ImageCollection`s have had a chance to check its size (with react-sizeme) and adjust image heights. I'm happy to remove this since it's ugly but this seems to work for now.

package.json
```diff
@@ -125,9 +125,9 @@
     "react-styled-flexboxgrid": "^2.0.0",
     "react-tracking": "^4.2.1",
     "react-url-query": "^1.1.4",
-    "request": "^2.81.0",
     "sharify": "^0.1.6",
     "styled-components": "^2.0.1",
+    "superagent": "^3.6.3",
```
alloy: There’s already a (isomorphic) `fetch` dependency in Reaction, do these request/superagent dependencies offer anything that `fetch` doesn’t?

eessex:
This looks good for merge to me pending @alloy's question above.

alloy:
Oh we don’t need to address my question right away, the `request` dependency is already in here anyways so it’s something that can be done in a subsequent PR.

In that case I’ll go ahead and merge so I can get started on renaming files.

package.json
kanaabe: Well actually the `request` dependency is replaced with `superagent` in this PR. It wasn't being used anywhere in the codebase so I replaced it with the fetching lib that we use in Force/Positron to stay consistent. Do you see any issues with that @alloy?

alloy: Where possible I prefer to have less overall dependencies and it looks like in the future the `fetch` API might actually become standard in browsers and Node. Not a problem right now, just possibly more to maintain in the future.

damassi: As of now, the only place `fetch` missing is in [IE 11](https://caniuse.com/#search=Fetch) and this is often used as a universal polyfill: https://github.com/matthew-andrews/isomorphic-fetch

PR #272, opened 2017-10-10
Update prettier

orta:
I didn't run prettier across the whole project, because there are 8 open PRs and it'd likely force a bunch of merge conflicts.

The update moves the settings to be inside prettier, instead of inside vscode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,12 +6,7 @@
   "repository": "https://github.com/artsy/reaction-force.git",
   "author": "Eloy Durán <eloy.de.enige@gmail.com>",
   "license": "MIT",
-  "files": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -6,21 +6,23 @@
...
-    "assets",
-    "data",
-    "docs"
-  ],
+  "files": ["dist", "assets", "data", "docs"],
+  "engines": {
+    "node": "8.4.x",
+    "npm": "5.4.x",
+    "yarn": "1.x"
+  },
```
orta: I also added the same setup as force/emission - this is the same as our nvm file

PR #273, opened 2017-10-10
[Dev] upgrade to react 16

l2succes:
Depends on: https://github.com/artsy/force/pull/2053

upgrades react and enzyme to their latest versions


src/components/publishing/__test__/article_tracking.test.tsx
```diff
@@ -1,9 +1,15 @@
...
+import 'jest-styled-components';
+
+import { mount } from 'enzyme';
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import * as React from 'react';
+
+import Events from '../../../utils/events';
+import Article from '../article';
+import { StandardArticle } from '../fixtures/articles';
```
alloy: These do not look like they were appropriately formatted by prettier

package.json
```diff
@@ -73,16 +71,17 @@
...
-    "jest": "^20.0.4",
-    "jest-styled-components": "^4.4.1",
+    "jest": "^21.0.0",
+    "jest-styled-components": "^4.7.0",
     "lint-staged": "^4.0.0",
     "longjohn": "^0.2.12",
     "morgan": "^1.8.1",
-    "prettier": "^1.4.4",
+    "prettier": "^1.7.4",
+    "raf": "^3.4.0",
```
alloy: What is this dependency for? https://www.npmjs.com/package/raf

l2succes: Here's a detailed description of the issue https://github.com/facebook/jest/issues/4545

But after upgrading, I started seeing the following warning for every test 
`ReferenceError: requestAnimationFrame is not defined`.

this package is a requestAnimationFrame polyfill which we can configure jest to use to suppress the warnings. I think another option is to add the following lines to `setupJest.ts`. (Someone)[https://github.com/facebook/jest/issues/4545#issuecomment-332877412] mentioned that mocking `requestAnimationFrame` that way didn't fix the issue for them but I can try it in our setup anyway. If it works, we don't need that dependency, either way, it's a temporary fix until jest is updated with better react 16 support.

```typescript
global.requestAnimationFrame = function(callback) {
  setTimeout(callback, 0);
};
```

alloy: Gotcha, if you can’t get the simple version to work in 1 min then the temp dependency seems fine too 👍

damassi:
One thing to note -- this will have to be coordinated around updates to other codebases that use React, as [this PR](https://github.com/artsy/reaction/pull/302) now excludes React from the bundle, instead relying on the consuming client environment.

package.json
```diff
@@ -82,15 +90,21 @@
...
-    "react-test-renderer": "^15.6.1",
+    "react-dom": "^16.0.0",
+    "react-hot-loader": "^1.3.1",
+    "react-test-renderer": "^16.0.0",
+    "relay2ts": "^0.2.0",
+    "ts-babel-node": "alloy/ts-babel-node#artsy",
+    "ts-node": "^3.0.2",
+    "jest": "^21.1.1",
+    "jest-styled-components": "^4.9.0",
+    "merge-stream": "^1.0.1",
```
damassi: This is a lot of new deps; are these changes coming in from #master?

src/Components/Publishing/__test__/ArticleTracking.test.tsx
```diff
@@ -1,10 +1,16 @@
...
+import Adapter from "enzyme-adapter-react-16"
+import * as React from "react"
+
+import Events from "../../../utils/events"
+import { Article } from "../article"
 import { getArticleFullHref } from "../Constants"
-import { StandardArticle } from "../Fixtures/Articles"
+import { StandardArticle } from "../fixtures/articles"
+
+Enzyme.configure({ adapter: new Adapter() })
```
damassi: This should be configured in our Jest setup script, which gets loaded before each suite runs:  https://github.com/artsy/reaction/blob/master/src/setup_jest.js

damassi: Ah! just scrolled down 😄  maybe just remove? Why was this needed here?

package.json
l2succes: looks like a mistake from the merge. For sure some of these aren't needed or may be duplicates

src/Components/Publishing/__test__/ArticleTracking.test.tsx
l2succes: Yeah lol sorry, we don't need it here, it's already in the setup script

src/Components/Publishing/Display/__test__/DisplayPanel.test.js
```diff
@@ -65,190 +69,200 @@ describe('units', () => {
...
-      const spy = jest.spyOn(wrapper.props().tracking, 'trackEvent')
-      expect(spy).toHaveBeenCalledWith(expect.objectContaining({
-        action: 'Impression',
-        entity_type: 'display_ad',
-        unit_layout: 'panel'
-      }))
-    })
-
-    it('tracks video progress', () => {
+    // TODO: this never gets called because it gets fired before the test runs
```
l2succes: @damassi @kanaabe I ended up commenting that test because it was the only one left that wasn't passing. To fix it we'll have to reset [this map](https://github.com/artsy/reaction/blob/c07a2fb86e399097e0fc6eb638a3fb0984fb9fff/src/Components/Publishing/Display/track-impression.ts#L1) before the test runs

package.json
```diff
@@ -6,28 +6,31 @@
...
-    "assets",
-    "data",
-    "docs"
-  ],
+  "files": ["dist", "assets", "data", "docs"],
   "engines": {
     "node": "8.x.x",
     "npm": "5.4.x"
   },
+  "peerDependencies": {
```
l2succes: @damassi I moved react to `peerDependencies` here as well

src/Components/Publishing/Header/__test__/__snapshots__/FeatureHeader.test.tsx.snap
```diff
@@ -2902,8 +2907,9 @@ exports[`feature renders superArticle full header properly 1`] = `
 }
 
 .c0[data-type="split"] .c13 {
+  -webkit-box-flex: 1;
   -webkit-flex-grow: 1;
-  -ms-flex-grow: 1;
+  -ms-flex-positive: 1;
```
alloy: Any information about these changes? /cc @kana @eessex

src/Components/Publishing/RelatedArticles/RelatedArticleFigure.tsx
```diff
@@ -20,7 +20,7 @@ interface RelatedArticleFigureProps extends React.HTMLProps<HTMLDivElement> {
 @track()
 export class RelatedArticleFigure extends React.Component<RelatedArticleFigureProps, null> {
   constructor(props) {
-    super()
+    super(props)
```
alloy: Nice catch 👌

alloy:
I’m going to assume that there are good reasons for those flex box changes, but please keep it mind as versions are being updated and deployed.

src/Components/Publishing/Display/__test__/DisplayPanel.test.js
kanaabe: No worries @l2succes thanks for the TODO note.

src/Components/Publishing/Header/__test__/__snapshots__/FeatureHeader.test.tsx.snap
kanaabe: This looks correct according to https://msdn.microsoft.com/en-us/library/hh673531(v=vs.85).aspx. Maybe some dependency that got upgraded fixed the processor to spit out the right -ms prefix for `flex-grow`?

alloy: Probably yeah 👍

damassi:
Awesome, so we've now got four apps on React 16 :)

PR #277, opened 2017-10-10
@orta => Unlock yarn version

kanaabe:
Follow up from https://github.com/artsy/force/pull/1818. CircleCI doesn't support yarn@1.x yet so we can't install Reaction in Force atm. Following @damassi's suggestion, we probably don't need to specify this anyway. We might be missing something tho, so please lmk!

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,7 @@
   ],
   "engines": {
     "node": "8.4.x",
-    "npm": "5.4.x",
-    "yarn": "1.x"
+    "npm": "5.4.x"
   },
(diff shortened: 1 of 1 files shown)


orta:
👍 seems fine to me

alloy:
Another option would be to add yarn v1 as a dev dependency to Force and load it from `./node_modules/.bin/yarn` instead?

orta:
yarn supports that naturally, there's a config var for it, don't have time to find it thoug

PR #246, opened 2017-09-19
@alloy => [Publishing] Email signup

kanaabe:
Success:
emailsignupsuccess

Error:
emailsignuperror

Todo
~~- [ ] Fix babel-plugin-rewire error with TS~~
- [x] Error handling (pending design)
- [x] Thank you message (pending design)

PR #278, opened 2017-10-10
@alloy => Remove unnecessary storybook webpack config

kanaabe:
When I first implemented the email component (https://github.com/artsy/reaction/pull/246), I needed to add this bit of Webpack config to get the `request` package working in storybooks. Issue detailed here: https://github.com/request/request/issues/1529. Now that we're using `superagent`, this is no longer necessary.


alloy:
🔪:+1:

PR #280, opened 2017-10-11
Add storybook for animations

yuki24:
This adds a new storybook for animations.

Most popular animation libraries are based on [the `animation` property](https://developer.mozilla.org/en-US/docs/Web/CSS/animation) and [the `@keyframes` at-rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes). [animate.css](https://github.com/daneden/animate.css) has a good set of animations and provides low-level CSS definitions. I also looked at [react-animation](https://github.com/FormidableLabs/react-animations), but it's just a thin layer that encapsulates animate.css, so I just decided to take some animations from animate.css, make them compatible with styled-component, and build our own components. In the future, we may want to depend on animate.css, but for now we can maintain the animations since we only need a few at this point.

oct 11 2017 1-37 pm

### DIFF
--- a/src/Assets/Animations.tsx
+++ b/src/Assets/Animations.tsx
@@ -1,5 +1,6 @@
 import { css, keyframes } from "styled-components"
 
+// Fade effects
 export const fadeOut = css`
   animation: ${keyframes`
     from {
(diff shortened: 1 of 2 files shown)


yuki24:
closing as I wanted to add more flexibility (duration, delay, etc). I'll send another PR later.

alloy:
Nice!

> but for now we can maintain the animations since we only need a few at this point.

👍

PR #281, opened 2017-10-11
Fix the storybook for grid

yuki24:
It seems like there were some interface changes made to `Grid` and the storybook for Grid was broken.

screen shot 2017-10-11 at 14 04 06


orta:
👍 this should go in

PR #284, opened 2017-10-11
@kanaabe: Fix & refactor fullscreen slideshow 

craigspaeth:
This refactors the fullscreen slideshow component to move down the tree so we don't need to use React context (a feature React encourages not using). The ViewFullscreenLink now lives _next_ to the slideshow component so that the state toggling it visible is consolidated there instead of all the way at the top Article component.

This was an attempt at fixing...

image

of [this issue](https://github.com/artsy/publishing/issues/45). Lmk if I'm overlooking something.

### DIFF
--- a/src/Components/Publishing/Article.tsx
+++ b/src/Components/Publishing/Article.tsx
@@ -1,5 +1,4 @@
 import { cloneDeep, includes, map } from "lodash"
-import * as PropTypes from "prop-types"
 import * as React from "react"
 import styled, { StyledFunction } from "styled-components"
 import Events from "../../Utils/Events"
@@ -13,7 +12,6 @@ import ReadMore from "./ReadMore/ReadMoreButton"
 import ReadMoreWrapper from "./ReadMore/ReadMoreWrapper"
 import RelatedArticlesCanvas from "./RelatedArticles/RelatedArticlesCanvas"
 import RelatedArticlesPanel from "./RelatedArticles/RelatedArticlesPanel"
-import FullscreenViewer from "./Sections/FullscreenViewer/FullscreenViewer"
 import Sections from "./Sections/Sections"
 import { ArticleData } from "./Typings"
 
@@ -28,9 +26,6 @@ export interface ArticleProps {
 }
 
 interface ArticleState {
-  viewerIsOpen: boolean
-  slideIndex: number
-  fullscreenImages: any
   article: any
   isTruncated: boolean
 }
@@ -41,35 +36,13 @@ interface ArticleContainerProps {
 
 @track({ page: "Article" }, { dispatch: data => Events.postEvent(data) })
 class Article extends React.Component<ArticleProps, ArticleState> {
-  static childContextTypes = {
-    onViewFullscreen: PropTypes.func,
-  }
-
   constructor(props) {
     super(props)
-    const { fullscreenImages, article } = this.indexAndExtractImages()
+    const { article } = this.indexAndExtractImages()
     this.state = {
-      viewerIsOpen: false,
-      slideIndex: 0,
-      fullscreenImages,
       article,
       isTruncated: props.isTruncated || false,
     }
-  }
-
-  getChildContext() {
-    return { onViewFullscreen: this.openViewer }
-  }
-
-  openViewer = index => {
-    this.setState({
-      viewerIsOpen: true,
-      slideIndex: index,
-    })
-  }
-
-  closeViewer = () => {
-    this.setState({ viewerIsOpen: false })
   }
 
   removeTruncation = () => {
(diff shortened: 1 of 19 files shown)


src/Components/Publishing/Sections/ImageCollection.tsx
```diff
@@ -48,11 +48,25 @@ function renderImages(images, dimensions, gutter, sectionLayout, width) {
     let renderedImage
     if (image.type === "image") {
       renderedImage = (
-        <Image image={image} sectionLayout={sectionLayout} width={imageSize.width} height={imageSize.height} />
+        <Image
+          images={images}
```
craigspaeth: Instead of passing a callback through context we're passing the `images` and `slideIndex` property from the ImageCollection components through props.

src/Components/Publishing/Article.tsx
```diff
@@ -94,62 +65,58 @@ class Article extends React.Component<ArticleProps, ArticleState> {
...
+              {emailSignup}
+              {relatedArticlePanel}
+            </Sidebar>
+          </StandardLayout>
+          {relatedArticleCanvas}
+        </ReadMoreWrapper>
+        {this.state.isTruncated ? <ReadMore onClick={this.removeTruncation} /> : false}
+      </div>
+    )
+  }
```
kanaabe: Been meaning to separate these out - thanks!

src/Components/Publishing/Article.tsx
```diff
@@ -94,62 +65,58 @@ class Article extends React.Component<ArticleProps, ArticleState> {
...
+
+  render() {
+    const { marginTop } = this.props
+    const { article } = this.state
+    return (
+      <ArticleContainer marginTop={marginTop}>
+        {article.layout === "feature" ? this.renderFeatureArticle() : this.renderStandardArticle()}
```
kanaabe: Reminds me of this comment by @damassi: https://github.com/artsy/reaction/pull/263#discussion_r143336391

I never looked into why or when this can cause issues though. Any experience with that?

src/Components/Publishing/Sections/ViewFullscreen.tsx
```diff
@@ -3,52 +3,63 @@ import * as React from "react"
...
+      <div>
+        <ViewFullscreenLink onClick={this.openViewer}>
+          <IconExpand />
+        </ViewFullscreenLink>
+        <FullscreenViewer
+          slideIndex={this.props.slideIndex}
+          onClose={this.closeViewer}
+          show={this.state.viewerIsOpen}
+          images={this.props.images}
+        />
```
kanaabe: Doesn't this bloat the page _a lot_?

src/Components/Publishing/Article.tsx
damassi: In JSX this is common since it returns a value (renderable JSX)! But here's an illustration of a naked ternary (non-assignment) that's dangerous: 

```javascript
const isOpen = true
isOpen ? foo() : orSomeTrutyValue()
additionalCodeToExecute()
```
When run, `additionalCodeToExecute` will move up to the `else` conditional since there's no semi denoting a break. Most linters guard against this type of thing.

craigspaeth: Ah, I see what Chris is saying there—in that context you linked there are rare circumstances where using it in that imperitive way (e.g. not assigned to a variable or used within an expression as is the case here) combined with JS's automatic semicolon insertion (b/c we don't use mandatory semis) could cause an issue.

That said, it should be totally fine to use the terery in JSX expressions like this.

kanaabe:
Hmm I'm not sure if this solution will work because I think this renders a slideshow for _every_ image. That `images` field will contain every image in the article, not just the ones from that collection or set. Lmk if i'm missing something tho!

src/Components/Publishing/Article.tsx
kanaabe: OoOo ic. Thanks for the explanation!

src/Components/Publishing/Sections/ViewFullscreen.tsx
craigspaeth: IMO it's probably a premature optimization to consider the HTML bloat and something to be addressed _if_ there are performance problems after trying the most straightforward implementation. That said, it won't actually bloat the HTML size because if `show={false}` then it [won't render any HTML and instead return null](https://github.com/artsy/reaction/blob/master/src/Components/Publishing/Sections/FullscreenViewer/FullscreenViewer.tsx#L82).

craigspaeth:
@kanaabe The `images` field is coming [from the collection component](https://github.com/artsy/reaction/pull/284#discussion_r144118527) (and from trying it out doesn't seem to come from the whole article). Maybe I'm missing something?

kanaabe:
@craigspaeth Ah I see, you're right that `images` are just the ones from the collection. However this feature requires that we are able to carousel through _all_ images in the article (images in an image_set or image_collection) and not just the ones in that section -- does that make sense?

craigspaeth:
🤦‍♂️ realized I went down the wrong path trying to keep this close to the image collection since it's all images. Going to close and try fixing the current approach.

PR #285, opened 2017-10-11
[Onboarding] refactor ItemLink into separate component

xtina-starr:
**What we did:**
- Extract `ItemLink` into its own component. 
- Moved relay calls into corresponding components (components that will use that data themselves).
- Started looking at the input field (there's a stray console.log there)

**Why we did it:**
Primarily to contain the relay calls for specific data within the component that will use that data. Overall in React, we don't want to pass relay props from components that won't be using the data themselves. The reason why, is that if this property gets removed by the parent / further up in the tree, the child components will no longer have that data / lose the access to that variable, and shiz will break. As such, it more robust for components to be self sufficient. 

We started with `ItemLink`, because we ran into this issue for the whole list (`SelectableItemList`), because we were originally going to work on the search functionality, and would need to switch up the data in that list (or create a new component), to split up the content of popular artists vs your search results. We paired with @alloy, and down the rabbit hole we went. `ItemLink` was a good place to start refactoring. 

**What next:**
@yuki24 — not sure how much impact this will have on the work you've been doing, apologies if it caused disruption. Would you like to base off your future work / branch off this PR? 

Also, not sure what you'll be working on next. We (@mennenia + I) were hoping to pair tomorrow on getting that user input and then fetching the artist search results from MP. This would probably involve moving up the `Input` field, and there being two child lists, one dedicated to fetching the popular artists data, and the other to consume the search query and make those calls and populate a list. 

Structurally, it would look a little like:

```
Artists (step)
    Input
    List (toggle)
        Popular 
        SearchResults
```

What are your thoughts? Maxim will be up in the morning, so you should be able to sync over Qs.


xtina-starr:
@alloy do you think this is mergeable?

src/Components/Onboarding/Steps/Artists/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Artists
+ * -> Input
+ *    List
+ *    -> Popular
+ *       Search
+ */
+
```
alloy: This can be removed, it’s already in your excellent PR write-up, feel free to merge either way, up to you 👍

yuki24:
The Follow Artist step and the Follow Category step have the exact same styles and interactions. How are we going to avoid having duplicates?

alloy:
You can move the shared code out again and leave a item link wrapper component here that’s responsible for getting the data and render the item link component. We only moved it here to simplify the mental model while explaining how React and Relay work together.

yuki24:
> You can move the shared code out again and leave a item link wrapper component

Could you be more specific? I'm not sure how that is going to be different from the existing ones.

yuki24:
On a side note, I added the `ItemLink` component just to experiment with CSS animations in React. We can just inline that component in the container since we'll have to re-consider the interface for animation stuff.

alloy:
Let’s talk about/pair on sharing the component with the categories page once you’ve figured out how to best implement the animations.

yuki24:
I'll merge this anyway since there are some better patterns implemented in this PR.

PR #286, opened 2017-10-11
@kanaabe: Refactored article component to fix fullscreen slideshow

craigspaeth:
The issue with the slideshow was simply using `childContextTypes` in the child component vs. `contextTypes` but I also decided to loop in the refactors from the last PR.

I can't find a way to use the simulator to test the mobile tap issue you encountered because of the un-pollyfilled advanced browser features Reaction uses. Maybe @alloy or @orta has an idea of how to do this? Barring a way to locally test this should allow us to test on GH pages atleast.

image

### DIFF
--- a/src/Components/Publishing/Article.tsx
+++ b/src/Components/Publishing/Article.tsx
@@ -97,59 +97,61 @@ class Article extends React.Component<ArticleProps, ArticleState> {
     return { fullscreenImages, article }
   }
 
-  render() {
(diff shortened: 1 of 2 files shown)


alloy:
@craigspaeth Looks like these web pack bundles are [missing babel-polyfill](https://github.com/storybooks/storybook/issues/213#issuecomment-322011646), but that [should be fixed](https://github.com/storybooks/storybook/pull/1668) in newer storybook versions, so try upgrading to the latest version?

craigspaeth:
I'll follow up PR with the polyfill stuff. Thanks Eloy + Kana!

PR #291, opened 2017-10-12
@kanaabe => [Minor] Don't require hero.url to render FeatureHeader

eessex:
Related to https://github.com/artsy/publishing/issues/83

If there is no hero.url, pass a string rather than undefined. Allows a new feature article with empty heroSection to render in Writer.


kanaabe:
Noice!

PR #292, opened 2017-10-12
@alloy => [Publishing] Kill default exports in publishing components

kanaabe:
There's something strange going on with Prettier on styled components. It's adding commas to places where we use template literals...still haven't looked into it but it all looks good on Storybooks and tests.

Looks like ^ was happening with my other machine but not my home one so it must be a local issue.


src/Components/Publishing/Byline/AuthorDate.tsx
```diff
@@ -82,11 +94,10 @@ const StyledAuthor = Text.extend`
...
   }
-  ${props => adjustForCondensed(props.layout)}
-  ${pMedia.sm`
+  ${props => adjustForCondensed(props.layout)} ${pMedia.sm`
     &:before {
       min-width: 8px;
       min-height: 8px;
     }
-  `}
+  `};
```
kanaabe: It's a lot of odd behavior like this ^

alloy:
Woah, epic! :clap:

orta:
very cool 👍

alloy:
Oh no, looks like I never submitted my review comment... Will do so when I get back to my laptop.

src/Components/Publishing/index.tsx
```diff
@@ -39,7 +39,7 @@ import {
   StandardArticle,
 } from "./Fixtures/Articles"
 
-export default {
+export const components = {
```
alloy: Do you really want this to be named `components`? You’d have to use it like so:

```js
import { components } from "@artsy/reaction/dist/Components/Publishing"
const { Article } = components
```

You can use this ES6 syntax to just export the individual items:

```js
export { Article } from "./Article"
```

And then the import looks like:

```js
import { Article } from "@artsy/reaction/dist/Components/Publishing"
```

alloy:
The PR looks great, feel free to address/ignore my comment about the exports from the index file and merge 👍

src/Components/Publishing/index.tsx
kanaabe: Ah yeah the latter is definitely what we want. Good call!

PR #295, opened 2017-10-17
Unlock Node 8 minor versions

damassi:
Removes required `8.4.x` in favor of `8.x.x` in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "docs"
   ],
   "engines": {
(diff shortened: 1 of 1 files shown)


damassi:
I noticed that there's an `.nvmrc` file that's locked to 8.4.0, but I believe that only applies to running Reaction, not installing from another project.

alloy:
👍

PR #296, opened 2017-10-17
@kanaabe => [Bugfix]: Ad Unit video playback

damassi:
Fixes https://github.com/artsy/publishing/issues/96

This PR checks to see what target is underneath the click target and if a `CanvasVideo` prevent default behavior. 

**TODO:** 
Expand the click target to cover video if the video has already been played and playback is complete.


damassi:
Updated!

PR #298, opened 2017-10-18
Remove duplicate HMR loader

damassi:
Fixes issue where after file change the process locks up with thousands of console errors leading to slow iteration time: 

screen shot 2017-10-18 at 11 59 44 am

Hot reloading is already in the environment via `webpack/development.js`. 

Note that this warning now appears: 

screen shot 2017-10-18 at 12 15 15 pm

But its only a few and doesn't impact performance. (Don't have time right now to debug further.)

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -5,22 +5,24 @@ const sharify = require("./sharify")
 
 const webpack = require("webpack")
 const webpackMerge = require("webpack-merge")
-const { CheckerPlugin } = require("awesome-typescript-loader")
+
+const {
+  CheckerPlugin
+} = require("awesome-typescript-loader")
 
 /**
  * Write out a file that stubs the data that’s normally shared with the client through the `sharify` module. This file
  * is then replaced in the product of webpack where normally the actual `sharify` module would be loaded.
  */
-const { METAPHYSICS_ENDPOINT } = env.config().parsed
-const sharifyPath = sharify({ METAPHYSICS_ENDPOINT })
+const {
+  METAPHYSICS_ENDPOINT
+} = env.config().parsed
 
-/**
- * Only use HMR plugin in dev mode
- */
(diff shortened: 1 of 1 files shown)


.storybook/webpack.config.js
```diff
@@ -5,22 +5,24 @@ const sharify = require("./sharify")
 
 const webpack = require("webpack")
 const webpackMerge = require("webpack-merge")
-const { CheckerPlugin } = require("awesome-typescript-loader")
+
+const {
+  CheckerPlugin
+} = require("awesome-typescript-loader")
```
damassi: For some reason VS code is rewriting these vars on file save and I can't figure out how to stop it -- I don't have prettier installed!

orta: might be auto eslint

damassi: screen shot 2017-10-18 at 12 21 28 pm

orta: it has the option to hook in to the vscode formatter, which may not be turned off with that option

damassi: Ah found it -- its in the checked-in `.vscode` folder. `editor.formatOnSave: true`

alloy: So is this the way it should be now?

Also, we specify prettier as a dev dependency, so I don’t quite follow this thread, because you _should_ have it installed. Is this option not related to prettier?

damassi: So i wasn't aware of the checked-in project-level settings for VSCode when I posted the above comment, and then later found it -- so yeah, formatting is working exactly as it should; the format changes in the diff are due to preject settings configured to apply prettier on save.

alloy: 👍

alloy:
Dope, thanks for the cleanup ✨

damassi:
Np! In use it now feels much more light-weight

PR #302, opened 2017-10-20
@alloy => Exclude certain common packages from bundle 

damassi:
Noticed that we were getting some strange errors (and poor stack traces) in Force and realized that we were loading two versions of React -- one from force and one from Reaction. Webpack docs: https://webpack.js.org/configuration/externals/

**Exclusions from build:**
- `react`
- `react-dom`
- `styled-components`

Also, removed `formatOnSave` from `.vscode/settings.json` since we already have Prettier setup on git commit and we shouldn't have to adjust the way we code and look at code during dev...

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,6 +11,6 @@
     "dist/": true
   },
   "editor.tabSize": 2,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
   "typescript.tsdk": "./node_modules/typescript/lib"
 }
--- a/src/Components/Publishing/Byline/__test__/__snapshots__/AuthorDate.test.tsx.snap
+++ b/src/Components/Publishing/Byline/__test__/__snapshots__/AuthorDate.test.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`Author/Date renders a single author 1`] = `
 .c0 {
-  font-family: Unica77LLWebMedium ,      'Arial',      'serif';
+  font-family: Unica77LLWebMedium , 'Arial', 'serif';
   -webkit-font-smoothing: antialiased;
   font-size: 16px;
   line-height: 1.1em;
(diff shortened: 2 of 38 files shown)


package.json
```diff
@@ -137,10 +138,13 @@
...
     "react-url-query": "^1.1.4",
     "sharify": "^0.1.6",
-    "styled-components": "^2.0.1",
     "superagent": "^3.6.3",
     "uuid": "^3.0.1"
   },
+  "peerDependencies": {
+    "react": "^15.5.4",
+    "styled-components": "^2.0.1"
+  },
```
alloy: Why are these peer dependencies? I would think that they should be normal dependencies, we need the host to always include them, it’s not a suggestion.

.vscode/settings.json
```diff
@@ -11,6 +11,6 @@
     "dist/": true
   },
   "editor.tabSize": 2,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
alloy: In what way do you have to change the way you write code with this enabled?

damassi: Prettier makes a lot of assumptions about ternaries, destructuring and spacing, and its kinda buggy at times too.

package.json
damassi: Ah you're right -- I've been staring at this screen for too long; updated.

damassi:
@alloy: Can you help me debug these type / update issues? It seems like they have to do with styled-components being bumped automatically via semver to `styled-components@2.2.1`. Version `2.0.1` seems to work fine. Not sure why `master` doesn't update while this branch does.

.vscode/settings.json
alloy: But if you don’t disable the commit hook prettier still rewrites at commit time, so then this issue would remain. I would prefer we keep note of actual bugs and report/fix them instead.

alloy:
Fixes here https://github.com/damassi/reaction/pull/1

.vscode/settings.json
orta: Yep, agreed that this option shouldn't be changed

damassi: 👍 will do.

damassi:
@alloy - the latest commit was a snapshot update. There were a lot of little white-space changes. I image its related to Prettier?

src/Components/Publishing/Byline/__test__/__snapshots__/AuthorDate.test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`Author/Date renders a single author 1`] = `
 .c0 {
-  font-family: Unica77LLWebMedium ,      'Arial',      'serif';
+  font-family: Unica77LLWebMedium , 'Arial', 'serif';
```
alloy: Hah, interesting glitch previously

alloy:
It’s only i snapshot files, right? I think it has to do with the styled-components update, someone decided to ‘fix’ printed whitespace.

alloy:
Looking good 👍

damassi:
Yup, only whitespace.

damassi:
Thanks for your help!

alloy:
Ah yes, didn’t re-check that. Please do 👍

PR #304, opened 2017-10-22
@eessex => Refactor out JS based mobile detection for CSS 

damassi:
Addresses https://github.com/artsy/publishing/issues/103. 

See https://github.com/ctrlplusb/react-sizeme#server-side-rendering for details on why `react-sizeme` can't be used in SSR apps. Instead, replaced with [react-responsive-decorator](https://github.com/damassi/react-responsive-decorator) which injects CSS mediaQueries rather than polling / dimension checking. 

#### Remove `react-sizeme` package:
- [x] FeatureHeader
- [ ] CanvasContainer
- [ ] ImageCollection
- [ ] Video

#### Swap JS for CSS
- [ ] FeatureHeader
- [ ] CanvasContainer
- [ ] ImageCollection
- [ ] Video

#### Update `isMobile` in Force and pass to responsive component
- [x] `isMobile: res.locals.sd.IS_MOBILE`

#### Jump issues 
- [x] Background images on fullscreen headers slightly jump as they load https://cl.ly/003Y1s1L283t
- [x] Fullscreen headers w/ video jump / reload after approx 2 seconds https://cl.ly/003Y1s1L283t

#### Slow loading:
- [x] Header takes 2-3 seconds to load, begins with grey background and title text only. (not sure if there is much we can really do here?) Was called out specifically on these stories: https://www.artsy.net/article2/artsy-editorial-the-100-most-expensive-artists
https://www.artsy.net/article2/bryce-dessner-the-future-of-art-according-to-bryce-dessner
https://www.artsy.net/article2/artsy-editorial-how-kerry-james-marshall-became-a-superhero-for-chicago-s-housing-projects

### DIFF
--- a/src/Components/Publishing/Article.tsx
+++ b/src/Components/Publishing/Article.tsx
@@ -23,6 +23,7 @@ export interface ArticleProps {
   article: ArticleData
   relatedArticlesForPanel?: any
   relatedArticlesForCanvas?: any
+  isMobile?: boolean
   isTruncated?: boolean
   emailSignupUrl?: string
   headerHeight?: string
@@ -50,6 +51,10 @@ interface ArticleContainerProps {
 export class Article extends React.Component<ArticleProps, ArticleState> {
   static childContextTypes = {
     onViewFullscreen: PropTypes.func,
+  }
+
+  static defaultProps = {
+    isMobile: false
   }
 
   constructor(props) {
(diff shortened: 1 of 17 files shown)


src/Components/Publishing/Header/FeatureHeader.tsx
```diff
@@ -190,74 +229,80 @@ const TextAsset = styled.div`
...
+        width: 50vw;
     }
     ${SubHeader} {
-      align-items: flex-start;
+        align - items: flex-start;
       flex-direction: column;
     }
     ${Deck} {
-      margin-bottom: 30px;
+        margin - bottom: 30px;
```
damassi: @alloy, @orta - this is what I mean by bugs in Prettier; all of the formatting was blown up and now its introducing literal errors in our styles. Not sure why this is happening but would like to disable so as not to slow us down in this sprint.

orta: Prettier can probably have it's inline CSS styling disabled somehow

alloy: Looks like various issues around this were introduced in v1.5.0 https://github.com/prettier/prettier/issues/2291, https://github.com/prettier/prettier/issues/2623, https://github.com/prettier/prettier/issues/2350

I haven’t found any issue yet around nested rules, so that that’s a ticket that definitely needs to get filed with them.

I wonder if in the interim there’s another fix we can apply. E.g. the places were the linebreak after interpolated CSS gets removed can be fixed by adding an extra semicolon after the interpolation:

```diff
 const Vertical = styled.div`
-  ${Fonts.unica("s16", "medium")}
+  ${Fonts.unica("s16", "medium")};
   margin-bottom: 10px;
   ${breakpoint.sm`
-    ${Fonts.unica("s14", "medium")}
+    ${Fonts.unica("s14", "medium")};
-  `}
+  `};
 `
```

alloy: Hmm, this is weird, I can’t actually reproduce this specific breakage on save that you highlight @damassi, only the ones that I could fix by adding semicolons. Any chance there are other vscode extensions at play here?

alloy: For examples see [this playground snippet](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEaIFs4AiAlgG6YC82MAngDZwAmAdA6QAYA6UmmAJMBxAwAFkTSYARgCc4AQwDWaQQF8uPSLQhSkmGQwDcXTlC6QoGTBBFwpxMpQx1GLdmr7Ac+O8swDuPTFxZKQBzIigAWgkrPB0AJgAGAAcAD0N-VSg2EAAaEAgkmCJ0ZFBZDGQAM1laNDg86VkweTgYAGUkpvCQ5BgpAFd6kAYIMCqauryAKzQUgCEpJpb22XwAGXC4cdqhiH6YJP247cmQTqk6qWQQCVkJOmhcs6lwmAB1IgYRZAAOBLy6rgiL0BkM0N16ABFfpWLZIao7PIwO4fL7CZBxJGLIi0boAYTwQWuUGgWzy-TqABU7mgTnBlMogA).

alloy: @orta From those tickets I linked to it seems like you can only disable prettier completely (with disable comments).

orta: Hrm, good point (you can only disable it [inside vscode](https://github.com/prettier/prettier-vscode/#prettierstylelintintegration-default-false---css-scss-and-less-only) ) in which case it'd mysteriously get clobbered on commit.

damassi: @alloy, @orta - yeah, i'm not too sure what the deal is and def don't have time to debug right now... 

In the meantime, yet another issue I just caught; check out this screen cap. On save its munging the spacing: 

prettier

So (so) annoying... On commit this tool can do what it will, but while I'm writing code I'm pretty sure this stuff shouldn't be taking away from a developers focus. I feel kinda strongly about this. If a developer wants Prettier and TSLint-fix to reformat their code while they're typing they can very easily opt in via their personal editor settings, but it shouldn't be forced on us.

orta: I don't get why it's so buggy in Reaction? It's been perfect in Emission and Danger

orta: The last time people were having problems with this it was because they had other extensions installed which were breaking it on save

damassi: 👍 Just spoke to @alloy and I'm gonna dig into my editor and check to see if there's a conflict somewhere, which seems likely. Thanks y'all

alloy: Let us know if you find it, I’m so dumbfounded otherwise.

damassi: Found the culprit: https://marketplace.visualstudio.com/items?itemName=mgmcdermott.vscode-language-babel. Disabled and all ok!

orta: Interesting, I wonder if it's a language scoping issue ( e.g. the template string css is classed as something else )

alloy: Glad you found it!

Weird that the README says it’s _only_ syntax highlighting, but changes how code is formatted on save, though 🤔

src/Components/Publishing/Article.tsx
```diff
@@ -105,61 +110,82 @@ export class Article extends React.Component<ArticleProps, ArticleState> {
...
+    const { headerHeight, isMobile } = this.props
     const { article } = this.state
+
     return (
       <div>
-        <Header article={article} height={headerHeight} />
+        <Header
+          article={article}
+          height={headerHeight}
+          isMobile={isMobile}
```
kanaabe: 🙌

damassi:
(Will have to return to the other items in checklist at later point)

PR #307, opened 2017-10-24
Add 'xlarge: 26px' for to the Text component

yuki24:
We have [a requirement](https://github.com/artsy/collector-experience/issues/609#issuecomment-338811236) where we need to show texts in the size of 26px.

screen shot 2017-10-24 at 11 39 52 am

### DIFF
--- a/src/Components/Text.tsx
+++ b/src/Components/Text.tsx
@@ -3,7 +3,7 @@ import styled from "styled-components"
 import * as fonts from "../Assets/Fonts"
 
 type TextAlign = "start" | "center" | "end"
-type TextSize = "small" | "medium" | "large"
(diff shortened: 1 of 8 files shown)


src/Components/Text.tsx
```diff
@@ -22,6 +22,7 @@ const textSizesForPrimaryStyle = {
 }
 
 const textSizesForSecondaryStyle = {
+  xlarge: "26px",
```
yuki24: This is the only actual change. The rest is just snapshot updates.

mennenia:
🎉 

Just waiting for CI :)

yuki24:
👍

PR #308, opened 2017-10-24
Add the gray dark color #999999 to the Colors

yuki24:
Another requirement of https://github.com/artsy/collector-experience/issues/609, we need this color for the introduction of the fair and the date for our public viewing.


yuki24:
I think it's pretty safe to merge this, so why not...

PR #310, opened 2017-10-24
Text component should accept all props

yuki24:
Right now the `Text` component only passes the `className` prop down to the `p` tag. This makes it difficult to apply a style workaround and requires us to update the upstream `@artsy/reaction-force`(this repo), which has more steps than just changing it in e.g. Force. This PR changes the component to pass down any props to the `p` tag so we'll be able to have more control without having to change upstream.

### DIFF
--- a/src/Components/Text.tsx
+++ b/src/Components/Text.tsx
@@ -39,10 +39,11 @@ const textStyleNameToCss = {
   secondary: fonts.secondary.style,
 }
 
-const RawText: React.SFC<TextProps> = props =>
-  <p className={props.className}>
-    {props.children}
-  </p>
+const RawText: React.SFC<TextProps> = (props: TextProps) => {
+  const { textSize, textStyle, align, color, ...remainderProps } = props
+
+  return <p {...remainderProps}>{props.children}</p>
+}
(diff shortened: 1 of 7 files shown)


src/Components/Text.tsx
```diff
@@ -39,10 +39,15 @@ const textStyleNameToCss = {
...
-const RawText: React.SFC<TextProps> = props =>
-  <p className={props.className}>
-    {props.children}
-  </p>
+const RawText: React.SFC<TextProps> = (props: TextProps) => {
+  const newProps: TextProps = { ...props }
+  delete newProps.textSize
+  delete newProps.textStyle
+  delete newProps.align
+  delete newProps.color
```
alloy: Rather than creating a dupe and then deleting, you can leave them out and only create a new object with the remainder:

```js
const { textSize, textStyle, align, color, ...remainderProps } = props
```

yuki24:
updated!

alloy:
I’ll merge, seeing as @l2succes already approved.

alloy:
@yuki24 you can do so yourself next time btw

PR #311, opened 2017-10-24
@damassi => RelatedArticlesCanvas for non-super feature articles  

kanaabe:
- Addresses the 'Further Reading' part of https://github.com/artsy/publishing/issues/129
- makes marginTop a string - consistent with other margin-like props
- adds isSuper prop which gets passed through from Force


src/Components/Publishing/Article.tsx
```diff
@@ -110,72 +112,69 @@ export class Article extends React.Component<ArticleProps, ArticleState> {
...
-      />
-      : false
-
-    const relatedArticleCanvas = relatedArticlesForCanvas ?
-      <RelatedArticlesCanvas
-        articles={relatedArticlesForCanvas}
-        vertical={article.vertical}
-      />
-      : false
+    const relatedArticlePanel = relatedArticlesForPanel ? (
```
damassi: Don't need the ternaries here -- can do (preferably in the render function so ones eyes don't have to skip around and things can be more 1:1)  
```javascript
{ relatedArticlesForPanel && 
  <RelatedArticlesPanel 
     label={"Related Stories"} 
     articles={relatedArticlesForPanel} 
   /> }

File: src/Components/Publishing/Article.tsx [hunk 2]
```diff
@@ -110,72 +112,69 @@ export class Article extends React.Component<ArticleProps, ArticleState> {
...
+      <RelatedArticlesPanel label={"Related Stories"} articles={relatedArticlesForPanel} />
+    ) : (
+      false
+    )
+
+    const relatedArticleCanvas = relatedArticlesForCanvas ? (
+      <RelatedArticlesCanvas articles={relatedArticlesForCanvas} vertical={article.vertical} />
+    ) : (
+      false
+    )
```
damassi: Same here

src/Components/Publishing/Article.tsx
```diff
@@ -213,7 +212,7 @@ export class Article extends React.Component<ArticleProps, ArticleState> {
 const ArticleDiv: StyledFunction<ArticleContainerProps & React.HTMLProps<HTMLDivElement>> = styled.div
 
 const ArticleContainer = ArticleDiv`
-  margin-top: ${props => (props.marginTop ? props.marginTop + "px" : "50px")};
+  margin-top: ${props => (props.marginTop ? props.marginTop : "50px")};
```
damassi: Don't need a ternary here -- can be `marginTop || '50px'`

src/Components/Publishing/Article.tsx
kanaabe: Oo weird, this diff is from autosave, not new code. I'll change anyway, thanks!

src/Components/Publishing/Article.tsx
```diff
@@ -110,72 +112,69 @@ export class Article extends React.Component<ArticleProps, ArticleState> {
...
+    )
 
     return (
       <div>
         <ReadMoreWrapper isTruncated={this.state.isTruncated} hideButton={this.removeTruncation}>
-          <Header
-            article={article}
-            isMobile={isMobile}
-          />
+          <Header article={article} isMobile={isMobile} />
```
damassi: This is debatable as a stylistic convention, but one that I'm passionate about, so... I would like to argue that as things scale out, grow larger and change over time, getting in the habit of orienting your components vertically 

```
<Header 
  article={article} 
  isMobile={isMobile} 
/>
```
vs
```
<Header article={article} isMobile={isMobile} />
```

- Saves stress on your eyes: reading code is not like reading a book where the eye needs to explicitly focus on each word; generally a programmer has an _idea_ or _intention_ in their head as they're coding, and rather than "read" its more of a gaze as things work in parallel. When code is vertically oriented that gaze can take in more information immediately (almost like a ideogram vs a sentence), which not only makes coding cumulatively faster, it also 
- Makes things more organized: alphabetization, should one choose to use it and get in the habit of applying it everywhere (people really should!), is much easier to parse when its top to bottom rather than left to right, and adding a new prop later on is much more orderly when being able to insert in in alphabetical order, like an index. Everything should be like an index...  
- Vertical components also read better. Take this 
```javascript
<SomeTitleComponent>{someTitle}</SomeTitleComponent>
```
One has to move their eyes to the middle of the line and pass through a handful of characters before they get to the variable in question. When stacked vertically like
```javascript
<SomeTitleComponent>
  {someTitle}
</SomeTitleComponent>
```
the var `someTitle jumps right out, because the syntax signals `{` and `}` and surrounding whitesspace allow the eye to ignore the container and get right to it. 

And what about scaling out? Say `<SomeTitleComponent />` needs a few more children. If its compressed together in one line you have to break it out onto individual lines anyway before more content can be added in. Again, tiny cumulative changes that increase overall dev time but can be mitigated by following vertical orientation. 

- One last point: tabs and whitespace can also be used in the "gaze", meaning, when you have code that follows a vertical orientation, its easy for the background mind to recognize that given a form like the below:
```
<Component
  variable={foo}
  anotherVariable={bar}
/> 
```
The first line (no whitespace) is the token `Component` and then the following (where whitespace begins) are variables and other misc. It makes it easier to scan over and find stuff while ignoring other things. This principle scales out well along many different axis and increasingly becomes clearer over time as a codebase grows larger. 

Sorry for the novel.... and if you totally disagree then please ignore! Just wanted to provide my thinking here for why i've been systematically expanding all code I come across.

cc @eessex

kanaabe: Sorry I didn't mean to change any of your work! I blame the autosave in my VS Code settings. What extensions are enabled for you? I added the ones related to this project and it automatically makes components of certain length into a single-line.

damassi: LOL, ran into some autosave issues yesterday too.... its so annoying I want to 😢

PR #316, opened 2017-10-25
@kanaabe => Display spacing

eessex:
Closes https://github.com/artsy/publishing/issues/130

- Matches widths to article width of 1250 max size
- Removes right margin for last item in related articles canvas


kanaabe:
👍  but ❌ !

PR #317, opened 2017-10-25
@alloy => Small QA + TS Error fixes

kanaabe:
The small QA includes
- https://github.com/artsy/publishing/issues/141
- https://github.com/artsy/publishing/issues/137
- Fullscreen viewer caption spacing


src/Components/Artwork/index.tsx
```diff
@@ -1,15 +1,26 @@
...
+}
+
+interface SizeImgProps extends React.HTMLProps<HTMLImageElement> {
+  size?: number
+}
+
+const SizeDiv: StyledFunction<SizeDivProps> = styled.div
+const SizeImg: StyledFunction<SizeImgProps> = styled.img
+
+const Container = SizeDiv`
```
kanaabe: I wasn't really sure why `size` needed to be passed down to the styled divs/imgs below since it didn't _look_ like they were being used anywhere so I'm happy to remove them if you'd like!

src/Components/Publishing/Constants.ts
```diff
@@ -6,4 +6,6 @@ export const zIndex = {
 
 export const articleHref = slug => `/article/${slug}`
 
+export const articleFullHref = slug => `https://www.artsy.net/article/${slug}`
```
kanaabe: I think ideally we want to replace this with a configurable `baseUrl` from somewhere like `sharify` in Force but there's no nice way to pass this down from the main <Article/> component without having to go through a long tree or use `context`. Yet another reason to move to a better mgmt container like Redux!

src/Components/Publishing/__stories__/RelatedArticles.story.tsx
```diff
@@ -1,21 +1,17 @@
...
-
 storiesOf("Publishing/Related Articles", module)
   .add("Related Articles Panel", () => {
     return <RelatedArticlesPanel articles={RelatedPanel} />
   })
   .add("Related Articles Canvas", () => {
     return (
       <RelatedArticlesCanvas
-        articles={StandardArticle.relatedArticlesCanvas}
+        articles={RelatedCanvas}
```
kanaabe: Looks like there was some leftover testing code here and in `/Fixtures/Article.ts`

src/Components/Artwork/index.tsx
alloy: I guess it may have been used for dynamic styling previously? I don’t see why it’s currently needed either, so 🔪

src/Components/InquiryArtwork/index.tsx
```diff
@@ -21,8 +25,8 @@ const OverlayBackground = styled.div`
   justify-content: center;
   align-items: center;
 `
-
-const Circle = styled.div`
+const Div: StyledFunction<CircleProps> = styled.div
+const Circle = Div`
```
alloy: So I guess the `props` are no longer typed as `any` by default in this version of styled-components? :/

src/Components/Publishing/Constants.ts
alloy: Could this be based on `window.location` instead?

alloy: If you do ever want to add some context, you can do so here if it’s something that could be useful data for many components https://github.com/artsy/reaction/blob/master/src/Components/Artsy.tsx

src/Components/InquiryArtwork/index.tsx
kanaabe: Yeah I think that may have been around for awhile. You mean being able to use custom props without typing them, right?

src/Components/Publishing/Header/FeatureHeader.tsx
```diff
@@ -23,6 +23,11 @@ interface FeatureHeaderProps {
   vertical?: any
 }
 
+interface DivProps {
+  height?: string
+  src?: string
+}
```
alloy: This isn’t really correct. It’s now being applied to all those styled divs, but they really only refer to 2 instances and e.g. `FeatureHeaderContainer` really needs the `height` prop, it’s not optional. Is there no way to specify the specific typing when you `extend` the div?

src/Components/Publishing/Constants.ts
kanaabe: Oh cool, I'll take a deeper look at this context work. I think `window.location` could become a pain later on if we decide to do SSR-only stuff like Google AMP.

src/Components/Publishing/Sections/FullscreenViewer/Slide.tsx
```diff
@@ -1,20 +1,24 @@
...
 import * as React from "react"
-import styled from "styled-components"
+import styled, { StyledFunction } from "styled-components"
 import { resize } from "../../../../Utils/resizer"
 import { pMedia } from "../../../Helpers"
 import { Fonts } from "../../Fonts"
 import { Caption } from "./Caption"
 
+interface ImageProps {
+  src?: string
```
alloy: Is this really optional? I don’t actually know, but assume that `background: url()` would be illegal?

src/Components/Publishing/Constants.ts
alloy: Gotcha, makes sense 👍

src/Components/InquiryArtwork/index.tsx
alloy: Yeah

alloy:
Other than those comments this looks good to me, feel free to merge 👍

src/Components/Publishing/Header/FeatureHeader.tsx
```diff
@@ -189,12 +194,12 @@ function renderSuperArticleLogos(superArticle, isMobile) {
 }
 
 // Styles
-
 const Div = styled.div`
   width: 100%;
   height: 100%;
   box-sizing: border-box;
 `
+const DivWithProps: StyledFunction<DivProps> = Div.extend
```
kanaabe: @alloy good call! It works well with this -- `Div.extend`

PR #321, opened 2017-10-26
@craigspaeth => "clicked partner logo" event for super article

eessex:
Adds a tracking call to FeatureHeader component, watches for clicks on a SuperArticle partner link.

Closes https://github.com/artsy/publishing/issues/134

Tracking information was taken from existing Force call here: https://github.com/artsy/force/blob/master/desktop/analytics/articles.js#L48


craigspaeth:
Pretty straightforward, LGTM! Thanks!

PR #323, opened 2017-10-26
Make Events API truly global

craigspaeth:
This adds the `Events` API to the `window` object which is kinda ugly in TS, but luckily not something we commonly need to do. The reason for it here is that we want a truly global singleton object that is shared across asset packages and between Reaction & Force—otherwise, the Reaction components instantiate one object and Force instantiates a different one and they don't communicate. This is a pattern we use in other truly global event emitters like our [analytics hooks](https://github.com/artsy/force/blob/master/desktop/lib/analytics_hooks.coffee#L10).

### DIFF
--- a/src/Utils/Events.ts
+++ b/src/Utils/Events.ts
@@ -1,6 +1,14 @@
 import EventEmitter from "events"
 
-const emitter = new EventEmitter()
+/* tslint:disable:no-namespace */
+declare global {
+  interface Window { __reactionEventsEventEmitter: any; }
+}
+/* tslint:disable:no-namespace */
+
+const emitter = typeof window !== 'undefined'
+  ? window.__reactionEventsEventEmitter || (window.__reactionEventsEventEmitter = new EventEmitter())
+  : new EventEmitter()
 const postEvent = data => emitter.emit("postEvent", data)
 const onEvent = callback => emitter.on("postEvent", callback)
 


src/Utils/Events.ts
```diff
@@ -1,7 +1,15 @@
 import EventEmitter from "events"
```
eessex: Where does this events api come from?

eessex:
Conceptually this makes sense to me but I'm not so familiar with these API's -- would be good to touch base on how these are shared in more detail.

src/Utils/Events.ts
craigspaeth: It's from Node's standard library and pollyfilled (via browserify/webpack) for browsers.

craigspaeth:
@eessex An event emitter is kinda like a person with a megaphone that listens to people's whispers and decides what to shout back out to the crowd.

```javascript
emitter.on("whisper hello", (data) => {
  emitter.emit(`Hey everyone, ${data.personsName} says hi!`)
})
emitter.emit("whisper hello", { name: "Eve" }) // Emits "Hey everyone, Eve says hi!"
```

It's a useful way to decouple pieces of code so that you don't have to have objects/modules call methods/functions directly on one another.

This pattern also maps nicely to say a chat room or HTTP server where might have many different incoming requests that you want to filter and respond to. It's also similar to [RabbitMQ](https://www.rabbitmq.com/) which is a platform-wide implementation we use to [publish and subscribe](https://www.wikiwand.com/en/Publish%E2%80%93subscribe_pattern) to many different events from services e.g. "Gravity `emit`s that a new artwork was published and Pulse `.on('new artwork')` sends a "new works for you" email.

This seems like a natural fit for bridging tracking between Reaction and other repos because it allows Reaction to emit tracking events from its components without having to call code in the other apps—and other apps can subscribe to the tracking events they want and decide what to do with them (e.g. Force can send all the events to Segment and Positron can choose to ignore most of them).

eessex:
Awesome, thanks for this info!

PR #324, opened 2017-10-27
@eessex => Cleaner responsive StandardLayout

kanaabe:
Alright, this feels so much better. See screenshots below. Once @owendodd approves of the layout, I'll reassign it for code review.

Couple things to point out:

1. Breakpoint is changed to 1080 for `lg`
   - This number makes more sense with the layouts and a 40px margin. 680px (sections) + 120 (3*40px margins) + 280 (minimum sidebar width) = 1080
   - I saw this only affected the Display Canvas so I'll make sure that looks good too **(WIP)** 
2. No grid..yet. I think the layout could benefit from a Grid layout but it also worked nicely by adding a parent with a strict margin and letting the child div do auto margins. 
3. Right after 1080 I hide the sidebar because it will no longer fit - I'm thinking we either come up with a condensed sidebar, or just opt for centering everything. Centering everything kinda looks strange though with the header being so far to the left.

<br><br>

standard-text-layout
standard-image-layout

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ dist
 .vscode/cSpell.json
 yarn-error.log
 npm-debug.log
+package-lock.json
--- a/src/Assets/Theme.ts
+++ b/src/Assets/Theme.ts
@@ -20,7 +20,7 @@ export default {
       xs: 600, // px
       sm: 720, // px
       md: 900, // px
-      lg: 1150, // px
+      lg: 1080, // px
       xl: 1250, // px
     },
   },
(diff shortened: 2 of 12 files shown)


owendodd:
@kanaabe This looks great to me. Seems like we could do a condensed sidebar in the future but I think just centering the content once the sidebar disappears makes the most sense rn. 

Nice work!

eessex:
This looks great to me!  One q, do all the feature layouts look ok with the changed breakpoint sizes?

kanaabe:
Yep, I double checked visually, but code-wise `lg` is only used in Sidebar (standard) and DisplayCanvas.

PR #327, opened 2017-10-27
@kanaabe => Qa mobile

eessex:
Addresses points in https://github.com/artsy/publishing/issues/165, https://github.com/artsy/publishing/issues/166, and https://github.com/artsy/publishing/issues/168

Looked into adding a slider to the related articles canvas, but it was slow going -- this is addressed in a separate ticket for next sprint, so leaving slider implementation as a TODO for then. 

For the current moment added some margin changes to the canvas to fix up margin spacing/overflow scrolling, as well as the scroll property recommended by @damassi here: https://github.com/artsy/publishing/issues/126


src/Components/Publishing/Article.tsx
```diff
@@ -222,6 +226,7 @@ const ArticleDiv: StyledFunction<ArticleContainerProps & React.HTMLProps<HTMLDiv
 
 const ArticleContainer = ArticleDiv`
   margin-top: ${props => props.marginTop || "50px"};
+  ${props => props.viewerIsOpen && "position: fixed;"}
```
kanaabe: I gave this a try in storybooks and InfiniteScroll in Force but it looks like the scroll position gets reset to the top of the article after exiting. You might be better off targeting `body` + using `overflow` to disable scroll which is how we do it for modals in [Force](https://github.com/artsy/force/blob/master/desktop/components/scrollbar/index.styl#L1-L2).

src/Components/Publishing/RelatedArticles/RelatedArticlesCanvas.tsx
```diff
@@ -77,7 +87,17 @@ const ArticlesWrapper = styled.div`
...
+      margin: 0 10px;
+      &:first-child {
+        margin-left: 20px;
+      }
+      &:last-child {
+        border-right: 20px solid white;
+      }
+    `}
   }
+  ${props => props.isMobile && "-webkit-overflow-scrolling: touch;"}
```
kanaabe: `styled-components` _should_ be able to auto-prefix this but I've also seen [issues](https://github.com/artsy/reaction/pull/186#pullrequestreview-47033980) with it in the past. Curious if this one is also the case.

src/Components/Publishing/Article.tsx
eessex: Hm, do you know how to target out to the body without using jquery?

src/Components/Publishing/RelatedArticles/RelatedArticlesCanvas.tsx
eessex: Looks like this property only exists in webkit (is for ios scrolling), so I believe this is the only way to declare.

eessex:
Closing in favor of https://github.com/artsy/reaction/pull/330

PR #332, opened 2017-10-30
@eessex => Instant article email module

kanaabe:
Spoke with @owendodd -- we'll use this as a fallback for ads when there is no display to show (under 100% SOV). IA policy requires all ads to return a creative asset as per: https://developers.facebook.com/docs/instant-articles/policy so this is the workaround.

This email signup matches the [sailthru hosted page](http://link.artsy.net/join/sign-up-editorial-facebook) that is currently used in Instant Articles

image

Also moves email-related components into its own dir.


src/Components/Publishing/Email/__test__/InstantArticleEmailSignup.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const state = viewer.state()
+    expect(state.message).toEqual("Invalid Email... Please try again")
+    expect(state.error).toBe(true)
+
+    jest.runAllTimers()
+    const postTimeoutState = viewer.state()
+    expect(postTimeoutState.disabled).toBe(false)
+    expect(postTimeoutState.error).toBe(false)
+    expect(postTimeoutState.message).toEqual("")
+  })
```
eessex: awesome tests

eessex:
Looks great!

PR #335, opened 2017-10-30
@damassi => Fix snapshots in master

damassi:
Update snaps that for some reason were invalid in master


eessex:
I'm having problems with snapshots now as well -- they pass locally, but fail on circle.  Are these the same tests you had issues with?
https://circleci.com/gh/artsy/reaction/1502

eessex:
Ah found your note in GH about removing all snapshots, this did it for me

PR #336, opened 2017-10-31
@eessex => Guard against images in Video component

damassi:
This PR adds a guard against image urls being placed in iframe video. 

In addition:
- https://github.com/artsy/publishing/issues/132
- https://github.com/artsy/publishing/issues/170
- https://github.com/artsy/publishing/issues/171
- https://github.com/artsy/publishing/issues/169


src/Components/Publishing/RelatedArticles/RelatedArticlesCanvas.tsx
```diff
@@ -47,7 +48,7 @@ export const RelatedArticlesCanvas: React.SFC<RelatedArticlesCanvasProps> = prop
             })}
           </ArticlesWrapper>
         </RelatedArticlesContainer>
-      </div>
+      </StandardLayoutParent>
```
damassi: @kanaabe - i wrapped a few problematic areas in this new wrapper. At some point we could probably refactor the layout so that this is built into the overall container instead of needing to be applied to independent elements.

src/Components/Publishing/Sections/FullscreenViewer/FullscreenViewer.tsx
```diff
@@ -68,7 +68,7 @@ export class FullscreenViewer extends React.Component<FullscreenViewerProps, Ful
   render() {
     const sliderSettings = {
       dots: false,
-      infinite: true,
+      infinite: false,
```
damassi: Per offline discussion with Owen -- the scroll-to-beginning-from-end didn't match rest of site

src/Components/Publishing/__stories__/Article.story.tsx
```diff
@@ -59,6 +60,10 @@ ads.forEach(mediaType => {
   })
 })
 
+story.add("Basic Article", () => {
+  return <Article article={BasicArticle} />
+})
+
```
eessex: Basic is a sub-category of feature article, but not an article type - we don't need this here if it is already included in the headers story.

src/Components/Publishing/Sections/Video.tsx
```diff
@@ -99,6 +102,11 @@ class VideoComponent extends React.Component<VideoProps, VideoState> {
 }
 
 // Utils
+export function isValidVideoUrl(url: string) {
+  const urlExtension = url && url.split('.').pop()
+  const isValid = BLACKLIST.every(bad => urlExtension !== bad)
+  return isValid
+}
```
eessex: 👍

src/Components/Publishing/RelatedArticles/RelatedArticlesCanvas.tsx
eessex: Because features need to be full screen width I think we will still need to wrap the display/related units to give them a margin.  Maybe we could use a more generic name for the standard element padding?

src/Components/Publishing/Header/FeatureHeader.tsx
```diff
@@ -60,16 +61,24 @@ function renderTextLayoutAsset(url, layout, title, imageChild) {
...
-      return (
-        <TextAsset>
-          {imageChild}
-          {url.length && image}
-        </TextAsset>
-      )
+
+      // When toggling between layout states, guard 'Basic' video headers stored
+      // in temporary `hero_section` auto-save state. TODO: might want to revist.
+      if (!isValidVideoUrl(url)) {
```
eessex: I think mongo validation should do this already, but we should probably check on the positron side and make sure fields are not being saved if they aren't used.

damassi: I'm certain that it wasn't guarding against images / other urls, but will follow up in Positron

src/Components/Publishing/RelatedArticles/RelatedArticlesCanvas.tsx
damassi: Will revisit in another PR

PR #337, opened 2017-10-31
@damassi => Update react-lines-ellipses; add `yarn test:watch`

damassi:
Updates package away from my fork and adds `yarn test:watch` with additional param `--runInBand` so that we can do `.only` on `describe` and `it` blocks.


orta:
👍 makes sense

orta:
also makes tests easier to run the debugger on

PR #338, opened 2017-10-31
Added tracking to read more

craigspaeth:
Part of https://github.com/artsy/publishing/issues/109
Adds a tracking call to the read more button so we can hook into it for Parsely.


damassi:
LGTM 👍

PR #339, opened 2017-11-01
@damassi => Use authors field and set better empty state

kanaabe:
- Gives us a better empty state for twitter handles and author bios.
- Converts all the things to `authors` now that the backfill is complete

image


damassi:
Simple enough! 👍

PR #341, opened 2017-11-01
@craigspaeth => Mobile Panel video, allow empty links

eessex:
- Use `cover_image_url` instead of `assets` for video cover images
- Adds mobile styles for Panel video (cover image has play button, plays on click)
- Moves video caret/controls into reusable component
- Allows for unit.link to be empty

I added a tracking event for the video play with `action: "Click"` -- is there a way to make this accommodate both click and mouseenter, or should we make separate events?

panel-video

### DIFF
--- a/src/Components/Publishing/Display/DisplayPanel.tsx
+++ b/src/Components/Publishing/Display/DisplayPanel.tsx
@@ -80,7 +80,7 @@ export class DisplayPanel extends React.Component<DisplayPanelProps, null> {
   render() {
     const { unit, campaign } = this.props
     const url = get(unit.assets, '0.url', '')
-    const cover = get(unit.assets, '1.url', '')
+    const cover = unit.cover_image_url || ''
     const imageUrl = crop(url, { width: 680, height: 284 })
     const hoverImageUrl = resize(unit.logo, { width: 680 })
     const coverUrl = crop(cover, { width: 680, height: 284 })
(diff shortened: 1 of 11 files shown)


src/Components/Publishing/Display/DisplayPanel.tsx
```diff
@@ -47,70 +60,89 @@ export class DisplayPanel extends React.Component<DisplayPanelProps, null> {
...
+    const { link } = this.props.unit
     e.preventDefault()
-    window.open(this.props.unit.link.url, '_blank')
+    if (link) {
+      window.open(link.url, '_blank')
+    }
   }
 
-  onMouseEnter = () => {
+  @track(once((props) => ({
```
craigspaeth: `once` is probably not necessary here as I believe we'll actually want to track multiple plays.

src/Components/Publishing/Display/DisplayPanel.tsx
```diff
@@ -47,70 +60,89 @@ export class DisplayPanel extends React.Component<DisplayPanelProps, null> {
...
   renderVideo = (url) => {
+    const { isMobile } = this.props
     return (
-      <VideoContainer>
-        <VideoCover />
+      <VideoContainer onClick={isMobile && this.playVideo}>
+        {!this.state.isPlaying &&
+          <VideoCover>
+            {isMobile &&
+              <VideoControls mini />
```
craigspaeth: 👍

craigspaeth:
We should make the mouseenter and click tracking separate—we can always join them into the same thing via SQL/Looker later but conceptually they are different things. Minus the comment about `once` the code LGTM! Feel free to merge when you address that.

eessex:
@craigspaeth I separated the events for mobile and desktop plays, they are distinguished by the labelling of their actions: "Click" and "MouseEnter"

PR #342, opened 2017-11-01
@eessex => Publishing canvas ad height

damassi:
Addresses https://github.com/artsy/publishing/issues/185

### DIFF
--- a/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
@@ -115,13 +115,13 @@ export class CanvasSlideshow extends React.Component<CanvasSlideshowProps, any> 
       nextArrow: (
         <RightArrow
           containerWidth={containerWidth}
-          onChangeSlide={slide => this.onChangeSlide(slide)}
+          onChangeSlide={this.onChangeSlide}
         />
       ),
       prevArrow: (
         <LeftArrow
           containerWidth={containerWidth}
-          onChangeSlide={slide => this.onChangeSlide(slide)}
+          onChangeSlide={this.onChangeSlide}
           isOnTitle={this.state.isOnTitle}
         />
       ),
--- a/src/Components/Publishing/__stories__/Article.story.tsx
+++ b/src/Components/Publishing/__stories__/Article.story.tsx
@@ -1,5 +1,6 @@
 import { storiesOf } from "@storybook/react"
 import React from "react"
+import styled from 'styled-components'
 import _ from "underscore"
 import { Article } from "../Article"
 import {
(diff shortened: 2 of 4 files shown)


src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
```diff
@@ -89,7 +89,7 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
       const slideshowProps = { unit, campaign, disclaimer, containerWidth: size.width }
 
       return (
-        <CanvasSlideshow {...slideshowProps}>
+        <CanvasSlideshow {...slideshowProps} renderTitleCard>
```
damassi: This opens up the space for the mobile slideshow stuff to show but it's still hidden for reasons I can't quite figure out.

src/Components/Publishing/__stories__/Article.story.tsx
```diff
@@ -70,6 +71,52 @@ ads.forEach(mediaType => {
...
+        display={Display('image')}
+        relatedArticlesForPanel={RelatedPanel}
+        relatedArticlesForCanvas={RelatedCanvas}
+        emailSignupUrl="#"
+      />
+    </div>
+  )
+})
+
+const Break = styled.div`
```
damassi: Grabbed this from force code so we would have a better testbed for seeing how infinite scroll is displayed

src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
```diff
@@ -15,6 +15,7 @@ interface CanvasSlideshowProps {
   containerWidth: number
   disclaimer: any
   unit: any
+  renderTitleCard?: boolean
```
eessex: it is not clear to me where the renderTitleCard prop is being used

damassi: It's being used here: https://github.com/damassi/reaction/blob/master/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx#L151

eessex: Does it not get redefined here? https://github.com/damassi/reaction/blob/master/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx#L143

eessex: but i think it is redefined just above https://github.com/damassi/reaction/blob/master/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx#L144

damassi: Ah, you are correct -- will fix.

kanaabe:
Nice!

PR #343, opened 2017-11-01
@damassi => More consistent margins

eessex:
- Moves layout for margins in sidebar out of individual components
- Moves layout for margins in footer components mostly out
- Matches margins for canvas/related to article body

### DIFF
--- a/src/Components/Publishing/Article.tsx
+++ b/src/Components/Publishing/Article.tsx
@@ -2,8 +2,10 @@ import { cloneDeep, includes, map, omit } from "lodash"
 import PropTypes from "prop-types"
 import React from "react"
 import styled, { StyledFunction } from "styled-components"
+import Colors from "../../Assets/Colors"
 import Events from "../../Utils/Events"
 import track from "../../Utils/track"
+import { pMedia } from "../Helpers"
 import { DisplayCanvas } from "./Display/Canvas"
 import { DisplayPanel } from "./Display/DisplayPanel"
 import { EmailPanel } from "./Email/EmailPanel"
@@ -31,7 +33,7 @@ export interface ArticleProps {
   display?: {
     name: string
     panel: object
-    canvas: object
+    canvas: any
   }
 }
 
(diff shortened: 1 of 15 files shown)


damassi:
👍

src/Components/Publishing/Article.tsx
```diff
@@ -185,31 +194,43 @@ export class Article extends React.Component<ArticleProps, ArticleState> {
...
+
         {this.props.display && (
           <div>
-            <DisplayCanvasBreak />
-            <DisplayCanvas unit={this.props.display.canvas} campaign={campaign} />
+            <LineBreak />
+            {displayOverflows
+              ? <FooterContainerOverflow>
+                <DisplayCanvas unit={this.props.display.canvas} campaign={campaign} />
+              </FooterContainerOverflow>
```
damassi: Prettier + JSX is so broken

orta: Isn't this more that the code itself is pretty verbose for using the tertiary operator?  This could be more readable by choosing a container earlier in the function and using that instead of duplicating the code inside the container, e.g.

```js
const displayOverflows = this.props.display && this.props.display.canvas.layout === "slideshow"
const FooterWrapper = displaysOverflow ? FooterContainerOverflow : FooterContainer

[...]
<FooterWrapper>
   <DisplayCanvas unit={this.props.display.canvas} campaign={campaign} />
</FooterWrapper>
```

I can't really think of an optimal way to format code like this?

damassi: The code could indeed be structured in many different ways, but this is more of a comment on Prettier breaking in a JSX context due to confusion about indention.

damassi: That said, I like this pattern (`const FooterWrapper = displaysOverflow ? FooterContainerOverflow : FooterContainer`) a lot. I feel like a lot of devs aren't aware that components can be reassigned like this for readability.

eessex: Yes into that!  Will keep in mind for the future.

PR #349, opened 2017-11-03
@damassi => Fix mobile ad

kanaabe:
Fixes features on mobile


damassi:
Ty 👍  This makes me feel wayy better.

PR #354, opened 2017-11-03
@craigspaeth => QA / Headers margins

eessex:
- Ups breakpoint to 1280 -- macbook screens are ~ 1260, meaning they loaded pages with no margins.
- Adds relative positioning to feature header and cleans up some overlap/margin problems
- Adds top margin to basic headers without embeds


craigspaeth:
LGTM!

PR #356, opened 2017-11-03
@kanaabe => Fix CanvasDisplayVideo click-out

damassi:
Address https://github.com/artsy/publishing/issues/206

### DIFF
--- a/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
@@ -36,8 +36,9 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
     campaign_name: props.campaign.name,
     unit_layout: unitLayout(props)
   })))
-  // tslint:disable-next-line:no-empty
-  componentDidMount() { }
+  componentDidMount() {
+    // noop
+  }
 
   // TODO: Ensure that full element can be clicked on video complete
   // Prevent links from blocking video playback.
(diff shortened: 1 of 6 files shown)


src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
```diff
@@ -50,7 +51,14 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
   }))
   openLink(e) {
     e.preventDefault()
-    if (!e.target.className.includes('CanvasVideo')) {
+    const videoClasses = [
+      'VideoControls__PlayButton',
+      'VideoControls__PlayButtonCaret',
+      'CanvasVideo__video'
+    ]
+    const isVideoClickArea = videoClasses.some(c => e.target.className.includes(c))
```
damassi: Within the current context, I'm not sure of a better way to do this.

kanaabe: Hmm could z-index or absolute positioning work here?

damassi: The problem is that there is an outer `a` tag that then contains `n` kinds of elements (cover, bottom text area, video, caret, etc), in a dynamic way -- without blacklisting specific elements with classes that can then be filtered against there's no way to accurately know whats being clicked.

kanaabe:
Alrighty, LGTM!

PR #357, opened 2017-11-03
@damassi => Text feature children display

eessex:
Looks like there was some code related to basic header that doesn't need to be here (we are already checking for layout to be text, it can't be basic too). This was preventing the image upload from displaying in the text layout in writer.


damassi:
👍

PR #360, opened 2017-11-04
@damassi => Fix ad injection target, bugfixes

damassi:
Addresses
- https://github.com/artsy/publishing/issues/199
- https://github.com/artsy/publishing/issues/192#issuecomment-341931308

Also
- Set a new constant `GLOBAL_IMAGE_QUALITY = 80%` (down from 95) which cuts average page download size by almost 1mb. We should probably go lower for mobile. 
- Fixes alignment issue with logos (cc @eessex, does this ring any bells?)

#### Before 

before-logos

#### After 

after-logos


src/Components/Publishing/Sections/ImageCollection.tsx
```diff
@@ -20,7 +20,7 @@ interface ImageCollectionProps {
   }
 }
 
-class ImageCollectionComponent extends React.Component<ImageCollectionProps, null> {
+class ImageCollectionComponent extends React.PureComponent<ImageCollectionProps, null> {
```
damassi: @kanaabe, @eessex - [`PureComponent`](https://reactjs.org/docs/react-api.html#reactpurecomponent) is how I fixed the repetitive image fade-in issue. Since this content will only be set on initial render no need to allow re-renders (unless data changes). 

If you're ever running into perf issues this really helps, but I would consider it a premature optimization to be avoided until necessary or your component data-structure is entirely flat without nesting (as `PureComponent` only does shallow comparisons).

src/Components/Publishing/Sections/Sections.tsx
```diff
@@ -72,7 +68,7 @@ export class Sections extends React.Component<SectionsProps, any> {
   }
 
   mountDisplayToMarker() {
-    const displayMountPoint = document.getElementById(`#${MOBILE_DISPLAY_INJECT_ID}`)
+    const displayMountPoint = document.getElementById(MOBILE_DISPLAY_INJECT_ID)
```
damassi: oops

src/Components/__stories__/CollectorIntentPage.story.tsx
```diff
@@ -34,7 +34,7 @@ const Link = styled.a`
   }
 `
 
-storiesOf("Onboarding").add("Collector Intent", () => {
+storiesOf("Onboarding", module).add("Collector Intent", () => {
```
damassi: Fixes some console warnings

damassi:
Going to self-merge and continue testing (since no one is around), cc @kanaabe

src/Components/Publishing/Sections/ImageCollection.tsx
kanaabe: 👍 Very nice - I was wondering about the repetitive image fade, thanks!

PR #361, opened 2017-11-05
@kanaabe => Complete Display Ad interactions on mobile

damassi:
This PR refactors our display unit to further account for mobile behaviors. Some states were missing such as image toggles on mobile (as it only accounted for css `:hover` events on desktop). Additionally tightened up the behavior pipeline so that things funnel through the component more systematically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,7 +153,7 @@
     "transform": {
       ".(ts|tsx)": "typescript-babel-jest"
     },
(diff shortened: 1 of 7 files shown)


damassi:
@kanaabe, @eessex - going to do one more self-merge (sorry! last one, just trying to get myself a day off tomorrow) and then I think everything required will be covered for launch. Will release to staging now.

PR #364, opened 2017-11-06
@damassi => Fix SSR slideshow bug

damassi:
One of the more frustrating bug sequences I've encountered:

1. When testing the new Publishing layout in staging, I would notice that the view would occasionally break
1. Digging in, I noticed that it would only break when the Display ad was image-based, not video 
1. That led me to conclude that the immediate right sidebar link was the culprit -- and more specifically the new mobile-based injection code
1. I pulled that all apart, compiled and restarted innumerable times
1. This led me to uncover an issue related to infinite scroll Display injection, so i guess that 1/2 day's worth of time was worth it 😓 
1. Realized however that the issue has nothing to do with the sidebar Display, although the state of the sidebar Display is indicative of the problem -- if image, it breaks, if video, all good
1. Proceeded to comment out JSX node after JSX node to track down execution; yay to binary search
1. Found it was related to the Canvas Display and learned...
1. That there's always a canvas display if there's a sidebar display, and the canvas display state is tied to the sidebar display state -- if video on side, standard on bottom; if standard on side, slideshow on bottom
1. Continued commenting out nodes -- all the while having to compile Reaction, stop / start force, and then refresh page until proper combo appeared -- until it occurred to me that this was a race related to SSR and all I had to do was defer rendering of slideshow until on client
1. Bug fixed, chris ☠️

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "tslint-react": "^2.4.0",
     "typescript": "^2.3.4",
     "typescript-babel-jest": "^1.0.3",
+    "typescript-styled-plugin": "^0.1.2",
     "uglify-js": "^2.8.22",
     "uglifyjs-webpack-plugin": "^0.4.2",
     "webpack": "^3.5.6",
--- a/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
@@ -19,13 +19,21 @@ interface CanvasContainerProps {
   unit?: any
 }
 
+interface State {
+  isMounted: boolean
+}
+
 @track()
-export class CanvasContainerComponent extends React.Component<CanvasContainerProps, null> {
+export class CanvasContainerComponent extends React.Component<CanvasContainerProps, State> {
   static defaultProps = {
     size: { width: 1250 }
   }
 
   public canvasVideoHandlers: any
+
+  state = {
+    isMounted: false
+  }
 
   constructor(props) {
     super(props)
@@ -39,7 +47,9 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
     unit_layout: unitLayout(props)
   })))
   componentDidMount() {
-    // noop
+    this.setState({
+      isMounted: true
+    })
   }
 
   // TODO: Ensure that full element can be clicked on video complete
@@ -83,6 +93,12 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
       target: "_blank",
       containerWidth: size.width,
       layout
+    }
+
+    // TODO: For whatever the slideshow leads to a race condition when rendering
+    // app in Force. Defer execution and everything works fine.
+    if (!this.state.isMounted) {
+      return <div />
     }
 
     // Overlay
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -105,6 +105,7 @@
     "tslint-react": "^2.4.0",
     "typescript": "^2.3.4",
     "typescript-babel-jest": "^1.0.3",
+    "typescript-styled-plugin": "^0.1.2",
```
damassi: See https://github.com/Microsoft/typescript-styled-plugin. Provides intellisense and syntax-checking, but doesn't work on `StyledFunction` interface nodes (damn)

src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
```diff
@@ -85,6 +95,12 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
       layout
     }
 
+    // TODO: For whatever the slideshow leads to a race condition when rendering
+    // app in Force. Defer execution and everything works fine.
+    if (!this.state.isMounted) {
+      return <div />
```
damassi: ~11 hours, give or take

package.json
orta: Ah, nice - we'd been looking at adding this to emission, all our CSS recommendations are coming through from VS Code side rather than the language. I think this plugin is probably more reliable 👍

src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
orta: Are there ways to potentially _potentially_ highlight these edge cases in danger?

package.json
orta: I looked into the `StyledFunction` thing - it's because it's just looking at [every tagged string](https://github.com/Microsoft/typescript-styled-plugin/blob/master/src/index.ts#L28-L29) and seeing if the generator is [one of these](https://github.com/Microsoft/typescript-styled-plugin/blob/master/src/configuration.ts#L11)

Looks like it's not interpreting types itself, so the `x = StyledFunction(Y)` won't know x`` is a thing

damassi: Ah yeah, thats what i figured :/

src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
damassi: In this instance not necessarily, as there's no reason _why_ this would be an issue. I think a lot of these things could resolve if the publishing code followed more of a single responsibility principle, which would be a good refactor once the deadline is passed. There are a lot of instances where one component does many things internally, including accepting dynamic `{children}` from unknown places, so its hard to track down where one things begins and another ends.

damassi: @orta - In terms of potental Danger plugins though -- yeah! libraries that aren't supported by SSR should be flagged or explicitely warned against. [`react-sizeme`](https://github.com/ctrlplusb/react-sizeme#server-side-rendering) is one of them, which was used here, and [`react-slick`](https://github.com/akiran/react-slick/issues/274) is another. This issue will always be present when dealing with SSR code. I think the solution is more awareness, as its easy to build something in storybooks (no SSR) and have it break in force.  Which leads me to think... it would be nice if there was an SSR-aware version of storybooks.

PR #366, opened 2017-11-06
Add video playback tracking to display video component

craigspaeth:
This follows the tracking events we did for [Venice 360](https://github.com/artsy/force/blob/master/desktop/apps/editorial_features/components/venice_2017/client/video.coffee#L117-L145) for tracking 3 & 10 second intervals as well as percent complete. It also fixes a small bug where clicking the video play button would count as a clickthrough tracking event.

### DIFF
--- a/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
@@ -54,24 +54,17 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
 
   // TODO: Ensure that full element can be clicked on video complete
   // Prevent links from blocking video playback.
-  @track((props, [e]) => ({
+  @track((props, [e]) => !isVideoClickArea(e) && {
     action: "Click",
     label: "Display ad clickthrough",
     entity_type: "display_ad",
     campaign_name: props.campaign.name,
     unit_layout: unitLayout(props)
-  }))
+  })
   openLink(e) {
     e.preventDefault()
 
-    const videoClasses = [
-      'PlayButton',
-      'PlayButton__PlayButtonCaret',
-      'CanvasVideo__video'
-    ]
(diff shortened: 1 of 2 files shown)


src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
```diff
@@ -54,24 +54,17 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
 
   // TODO: Ensure that full element can be clicked on video complete
   // Prevent links from blocking video playback.
-  @track((props, [e]) => ({
+  @track((props, [e]) => !isVideoClickArea(e) && {
```
kanaabe: Nice secondary use for the video click area method

src/Components/Publishing/Display/Canvas/CanvasVideo.tsx
```diff
@@ -9,10 +9,31 @@ interface VideoProps {
...
+    this.props.tracking.trackEvent({
+      action: "Video duration",
+      label: "Display ad video duration",
+      percent_complete: percentComplete,
+      campaign_name: this.props.campaign.name,
+      unit_layout: "canvas_standard"
+    })
+  })
+
+  trackSeconds = memoize((secondsComplete) => {
```
craigspaeth: `memoize` is used here to prevent multiple tracking calls for the same second interval—e.g. `timeupdate` events will fire for unpredictable 3.1, 3.2, 3.5 intervals b/c of how the HTML5 video API works, so we `Math.floor` that time and use memoize as a way to prevent this function being called multiple times with the same floored time.

kanaabe: Very cool

kanaabe:
Not necessary for now (and i'm happy to do them to get more familiar), but these could probably use a good set of unit tests.

PR #369, opened 2017-11-06
@kanaabe => Add cover image support to Canvas video

damassi:
Adds cover image support to canvas video. 

**NOTE!** There is no field in positron for uploading cover images in this way, so pulls from display ad. 

screen shot 2017-11-06 at 3 38 50 pm

### DIFF
--- a/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
@@ -75,7 +75,7 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
 
   render() {
     const { campaign, disclaimer, size, unit } = this.props
-    const { assets, layout, link: { url } } = unit
+    const { assets, cover_image_url, layout, link: { url } } = unit
     const isOverlay = layout === 'overlay'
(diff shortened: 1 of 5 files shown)


src/Components/Publishing/Display/Canvas/CanvasContainer.tsx
```diff
@@ -75,7 +75,7 @@ export class CanvasContainerComponent extends React.Component<CanvasContainerPro
 
   render() {
     const { campaign, disclaimer, size, unit } = this.props
-    const { assets, layout, link: { url } } = unit
+    const { assets, cover_image_url, layout, link: { url } } = unit
```
kanaabe: Don't we need to pull this in from the panel?

damassi: Ahh I see: `this.props.display.canvas` is what `unit` resolves to further up the chain

kanaabe: Nevermind, discussed irl!

damassi: As we talked about offline, gonna patch this at the force query level

PR #370, opened 2017-11-07
@craigspaeth => Add video tracking to <DisplayPanel />

damassi:
Follow up to: https://github.com/artsy/reaction/pull/366

### DIFF
--- a/src/Components/Publishing/Display/DisplayPanel.tsx
+++ b/src/Components/Publishing/Display/DisplayPanel.tsx
@@ -1,4 +1,4 @@
-import { get, once } from "lodash"
+import { get, memoize, once } from "lodash"
 import React, { Component, HTMLProps  } from "react"
 import styled, { StyledFunction } from "styled-components"
 import Colors from "../../../Assets/Colors"
@@ -12,6 +12,7 @@ interface Props extends React.HTMLProps<HTMLDivElement> {
   campaign: any
   isMobile?: boolean
   unit: any
+  tracking?: any
 }
 
 interface State {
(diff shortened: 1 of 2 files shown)


src/Components/Publishing/Display/DisplayPanel.tsx
```diff
@@ -48,6 +49,55 @@ export class DisplayPanel extends Component<Props, State> {
...
+    }
+  }
+
+  componentDidUpdate() {
+    if (this.video) {
+      this.video.addEventListener("timeupdate", this.trackProgress)
+    }
+  }
+
+  // TODO: This could be shared with <CanvasVideo />
```
craigspaeth: 👍  thanks for this comment.

src/Components/Publishing/Display/DisplayPanel.tsx
```diff
@@ -48,6 +49,55 @@ export class DisplayPanel extends Component<Props, State> {
...
+    this.props.tracking.trackEvent({
+      action: "Video duration",
+      label: "Display ad video duration",
+      percent_complete: percentComplete,
+      campaign_name: this.props.campaign.name,
+      unit_layout: "panel"
+    })
+  })
+
+  trackSeconds = memoize((secondsComplete) => {
```
kanaabe: @damassi I'm curious why this memoize is considered an optimization? If the arguments are always different, would the cached method ever be used twice?

craigspaeth: This is a trick I used in the canvas component to avoid accidentally double tracking the same 3 or 10 second intervals. Notice how [we round up](https://github.com/artsy/reaction/pull/370/files#diff-cbef750817163c85f91a3113eb1ca059R66) above—this could cause 3:05, 3:20, 3:35, etc. to all track 3 seconds.

kanaabe: Ohh ok that makes sense with the rounding up. Wow that's cool! So on the first call, memoize applies the function and caches the result, then on subsequent calls it just returns the cached result? Verrrrry nifty!!

PR #371, opened 2017-11-07
@craigspaeth => Fix slideshow area on mobile

damassi:
Addresses https://github.com/artsy/publishing/issues/184

The margins aren't totally perfect (meaning, the spacing between the end of the ad and the next article inserted after infinite scroll) but we can return to it later as its hardly noticable: 

margin2

### DIFF
--- a/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
@@ -143,14 +143,16 @@ export class CanvasSlideshow extends React.Component<CanvasSlideshowProps, any> 
     const renderTitleCard = containerWidth < 900
 
     return (
-      <SliderContainer containerWidth={containerWidth}>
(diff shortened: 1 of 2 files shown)


src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
```diff
@@ -143,14 +143,16 @@ export class CanvasSlideshow extends React.Component<CanvasSlideshowProps, any>
...
-          {this.renderSlides()}
-        </Slider>
+      <div>
+        <SliderContainer containerWidth={containerWidth}>
+          <Slider {...sliderSettings} ref={slider => (this.slider = slider)}>
+            {this.renderSlides()}
+          </Slider>
+        </SliderContainer>
 
         {renderTitleCard &&
```
craigspaeth: Oh I see—this was moved out of the container. 👍

craigspaeth:
Thanks!

PR #373, opened 2017-11-08
@kanaabe => [Minor] Video cover classname

eessex:
Adds a class to the video cover image so it can be targeted for css.


eessex:
Gonna merge since is trivial and hope to include in force PR.

PR #375, opened 2017-11-08
Stop multiple impression tracking calls

craigspaeth:
Yikes, what a rabbit hole. This implements a `trackImpression` decorator that stores a cache of the previous impressions to prevent multiples being fired. This avoids the React lifecycle pitfalls and the decorator hides some ugliness and feels a bit more in the style of React Tracking. [Context from Slack](https://artsy.slack.com/archives/C3YTM6BTK/p1510171534000387)

PS. I thought I disliked decorators before—now after writing one... 😵 ...so complicated


craigspaeth:
👍  for sure. Your DisplayPanel tests did add some coverage to this but the caching behavior isn't covered. I'll try to include that in my next PR, but didn't want to hold this patch up any longer.

PR #376, opened 2017-11-09
[WIP] Gene View Component

xtina-starr:
I wanted to start to get some eyes on this. This is a WIP for the [Onboarding Gene Component](https://github.com/artsy/collector-experience/issues/506) using the relay modern approach for data fetching. 

Included in this PR:
- Rendering the suggested genes list from metaphysics
- Using the `match_gene` schema to search for genes to follow
- Returning _No Results Found_ when there are no search results

The work that is to be completed is creating the mutation to actually follow genes and replace with a related gene. That is outside of the scope of https://github.com/artsy/collector-experience/issues/506 and ventures into handling the data submission. If the all the UI work for both artist follow and gene follow component is completed it might be worth starting that.

**Default suggested genes list:**
screen shot 2017-11-09 at 12 00 06 pm

**Gene Search Results list:**
screen shot 2017-11-09 at 12 00 46 pm

**No Search Results:**
screen shot 2017-11-09 at 12 04 30 pm


src/Components/Input.tsx
```diff
@@ -9,37 +9,34 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
 }
 
 interface InputState {
   borderClasses: string
 }
 
 const StyledInput = styled.input`
-  ${borderedInput}
-  ${block(24)}
+  ${borderedInput} ${block(24)};
```
alloy: Add a semi-colon after the interpolated value so prettier doesn’t move these both to a single line.

src/Components/Input.tsx
```diff
@@ -9,37 +9,34 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
 `
 
 const BorderlessInput = styled.input`
   border: 0;
-  ${fonts.secondary.style}
-  font-size: 17px;
+  ${fonts.secondary.style} font-size: 17px;
```
alloy: Ditto.

src/Components/Input.tsx
```diff
@@ -9,37 +9,34 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
   outline: none;
   flex: 1;
 `
 
 const StyledDiv = styled.div`
-  ${borderedInput}
-  border: 0;
+  ${borderedInput} border: 0;
```
alloy: Ditto.

src/Components/Input.tsx
```diff
@@ -9,37 +9,34 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
   padding: 12px;
   margin-right: 0;
   display: flex;
   position: relative;
 
   & .border-container {
     z-index: -1;
-    ${border}
-    position: absolute;
+    ${border} position: absolute;
```
alloy: Ditto.

src/Components/Input.tsx
```diff
@@ -75,28 +73,55 @@ class Input extends React.Component<InputProps, InputState> {
   }
 
   render() {
-    if (this.props.rightView) {
+    if (this.props.leftView && this.props.rightView) {
+      const newProps: any = { ...this.props }
+      delete newProps.className
```
alloy: Rather than duplicating all of the props and then deleting, you can also omit:

```js
const { className, ...newProps } = this.props
```

src/Components/Input.tsx
```diff
@@ -75,28 +73,55 @@ class Input extends React.Component<InputProps, InputState> {
...
+            onFocus={this.onFocus.bind(this)}
+            onBlur={this.onBlur.bind(this)}
+            value={this.props.value}
+          />
+          {this.props.rightView}
+        </StyledDiv>
+      )
+    } else if (this.props.rightView) {
       const newProps: any = { ...this.props }
       delete newProps.className
```
alloy: Ditto.

src/Components/Input.tsx
```diff
@@ -75,28 +73,55 @@ class Input extends React.Component<InputProps, InputState> {
...
-
-        return (
-          <StyledDiv>
-            <div className={this.state.borderClasses}/>
-            {this.props.leftView}
-            <BorderlessInput {...newProps} onFocus={this.onFocus.bind(this)} onBlur={this.onBlur.bind(this)}/>
-          </StyledDiv>
-        )
+      const newProps: any = { ...this.props }
+      delete newProps.className
```
alloy: Ditto.

src/Components/Onboarding/ItemLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+  fadeOut: boolean
+  item_id: string
+}
+
+interface Props extends React.HTMLProps<HTMLAnchorElement> {
+  item?: any
+  id: string
+  _id: string
+  name: string
+  image_url
```
alloy: This property is missing a type. Interestingly that seems to default to `any`, was that the intent?

src/Components/Onboarding/ItemLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export default class ItemLink extends React.Component<Props, State> {
+  state = {
+    fadeIn: false,
+    fadeOut: false,
+    item_id: this.props.id,
```
alloy: This doesn’t really work, does it? If your initial state isn’t static, then you should set it from a `constructor` override.

src/Components/Onboarding/ItemLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  render() {
+    return (
+      <Link
+        onClick={this.onClick.bind(this)}
+        fadeIn={this.state.fadeIn}
+        fadeOut={this.state.fadeOut}
+        item_id={this.props.id}
+      >
+        <Col>{<Avatar src={this.props.image_url ? this.props.image_url : ""} width={50} height={50} />}</Col>
```
alloy: What falsy values could `image_url` hold? If it’s `null`, `undefined`, or `""` then I wonder if you can’t just use those? If you really do need to default to an empty string, then a shorter version would be to do `this.props.image_url || ""`

src/Components/Onboarding/ItemLink.tsx
xtina-starr: I meant to type this as a string.

src/Components/Onboarding/ItemLink.tsx
xtina-starr: No I think this is left over code from @yuki24 first implemented the artist component.

src/Components/Onboarding/ItemLink.tsx
xtina-starr: If my memory serves me correctly, I think the `image_url` can be `null`. I think I left this as an empty string because I thought there should be some sort of image place holder because that's the way search in force handles this. I planned to go back and add the image placeholder. Might be a good time to do that now 😄

src/Components/Input.tsx
xtina-starr: I inherited this lol. I only intended to add `value` to the input props and prettier did its thing.

src/Components/Input.tsx
xtina-starr: So in the above code `className` get restructured in its own thing and everything else is on `...newProps`?

alloy: Aye 👍

src/Components/Input.tsx
alloy: Yeah I know :( It’s a PITA with prettier 1.5.x

src/Components/Onboarding/ItemLink.tsx
alloy: 👍

src/Components/Onboarding/ItemLink.tsx
yuki24: I was so sure I did use this syntax at some point (without `this.props.id` for sure) but looking at git commits it seems like I actually used [`constructor`](https://github.com/artsy/reaction/pull/268/files#diff-617784f1b8c8a101065da7cc0270070cR66) instead, which later was re-written to [not use it](https://github.com/artsy/reaction/pull/285/files#diff-17dd94b4bd0e5bcc37ff8e8aad9e2027R49).

src/Components/Input.tsx
xtina-starr: When I try this method of accessing the `...newProps` I'm getting a type error: 

Type '{ onFocus: any; onBlur: any; value: string | number | string[]; children?: ReactNode; error?: boo...' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ThemedOuterStyledProps<ChangeTargetHTMLP...'.

alloy: Oh I guess that’s because `newProps` is no longer cast as `any`. Try:

```ts
const { className, ...newProps }: any = this.props
```

yuki24: doesn't `any` drop all the type goodness? I can't explain why but [dropping `ref`](https://github.com/artsy/reaction/blob/77a496f/src/Components/CircleIcon.tsx#L22) may make the editor happy:

```ts
const { ref, ...reminderProps } = props
```

alloy: Yes it would, good call 👍

xtina-starr: Sweet!

src/Components/Onboarding/ItemLink.tsx
```diff
@@ -78,7 +78,15 @@ export default class ItemLink extends React.Component<Props, State> {
         fadeOut={this.state.fadeOut}
         item_id={this.props.id}
       >
-        <Col>{<Avatar src={this.props.image_url ? this.props.image_url : ""} width={50} height={50} />}</Col>
+        <Col>
+          {
+            <Avatar
+              src={this.props.image_url ? this.props.image_url : "https://www.artsy.net/images/icon-70.png"}
```
xtina-starr: Adding an external link for now, but the plan is to create a lower level component that display a default image when the `image_url` isn't available.

src/Components/Onboarding/Steps/Artists/ItemLink.tsx
```diff
@@ -102,7 +102,17 @@ class ItemLink extends React.Component<Props, State> {
         artist_id={this.props.artist.id}
       >
         <Col>
-          {<Avatar src={this.props.artist.image ? this.props.artist.image.cropped.url : ""} width={50} height={50} />}
+          {
+            <Avatar
+              src={
+                this.props.artist.image
+                  ? this.props.artist.image.cropped.url
+                  : "https://www.artsy.net/images/icon-70.png"
```
xtina-starr: Same here.

xtina-starr:
@mzikherman @alloy I think handling the image thing was the last bit of feedback for this PR. Is there any additional feedback? If not, would we then merge this into the relay-modern branch?

PR #380, opened 2017-11-10
@eessex => Partner components

kanaabe:
- Pulls in [these two components](https://github.com/artsy/force/tree/master/desktop/apps/editorial_features/components/gucci/components/partner) written by @eessex and Typescriptifies them
- Add tracking to these components (**WIP WAITING ON** the rest of info from: https://waffle.io/artsy/publishing/cards/5a032d9aaf310e0030bca7be)
- Adds `IconPlus`

image
image

### DIFF
--- a/src/Components/Publishing/Icon/IconPlus.tsx
+++ b/src/Components/Publishing/Icon/IconPlus.tsx
@@ -11,7 +11,7 @@ export class IconPlus extends Component<any, null> {
         xmlns="http://www.w3.org/2000/svg"
         style={{margin: this.props.margin}}
       >
-        <g stroke="none" stroke-width="2" fill="none" fill-rule="evenodd">
+        <g stroke="none" strokeWidth="2" fill="none" fillRule="evenodd">
           <g transform="translate(-101.000000, -18.000000)" fill="#000000">
             <g transform="translate(101.000000, 18.000000)">
               <rect x="0" y="7" width="15" height="1" />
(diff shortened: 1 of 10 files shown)


src/Components/Publishing/Icon/IconPlus.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from "react"
+import styled from "styled-components"
+import { pMedia } from "../../Helpers"
+
+export class IconPlus extends Component<any, null> {
+  render() {
```
eessex: Would be good to include a color prop here -- on superArticles it is rendered in white

src/Components/Publishing/Partner/PartnerInline.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  constructor(props) {
+    super(props)
+    this.onPartnerClick = this.onPartnerClick.bind(this)
+  }
+
+  onPartnerClick(event) {
+    this.props.tracking.trackEvent({
```
eessex: Looking at the share tracking now, what is the props.tracking object and where does it come from?

src/Components/Publishing/Icon/IconPlus.tsx
kanaabe: Noted, I'll add this in another time since it's not required for this project.

src/Components/Publishing/Partner/PartnerInline.tsx
kanaabe: Comes from the track decorator!

PR #381, opened 2017-11-10
@kanaabe => Share tracking accepts props

eessex:
Updates Share's tracking event to accept props.

### DIFF
--- a/src/Components/Publishing/Byline/Share.tsx
+++ b/src/Components/Publishing/Byline/Share.tsx
@@ -1,20 +1,27 @@
 import React from "react"
 import styled from "styled-components"
-import { track } from "../../../Utils/track"
+import Events from "../../../Utils/Events"
+import track from "../../../Utils/track"
 import { pMedia } from "../../Helpers"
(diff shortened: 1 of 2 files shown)


src/Components/Publishing/Byline/Share.tsx
```diff
@@ -1,20 +1,27 @@
...
-export class Share extends React.Component<ShareProps, null> {
+@track((props) => {
+  return props.trackingData ? props.trackingData : {}
+}, {
+  dispatch: data => Events.postEvent(data)
+})
+export class Share extends React.Component<Props, null> {
   static defaultProps = {
     color: "black",
   }
```
kanaabe: Does this work without setting tracking in defaultProps?

eessex: I believe so, this is what I see in storybooks - 
screen shot 2017-11-10 at 5 52 37 pm

eessex: Added a story for the custom props as well, looks to be working - - 
screen shot 2017-11-12 at 1 37 46 pm

PR #384, opened 2017-11-13
@kanaabe => Update share labels

eessex:
Per [input from @wrgoldstein](https://github.com/artsy/positron/issues/1416), updates labels on 'share' tracking data

### DIFF
--- a/src/Components/Publishing/Byline/Share.tsx
+++ b/src/Components/Publishing/Byline/Share.tsx
@@ -37,8 +37,9 @@ export class Share extends React.Component<Props, null> {
     window.open(e.currentTarget.attributes.href.value, "Share", "width = 600,height = 300")
 
     this.props.tracking.trackEvent({
-      action: "Article share",
-      service: (() => {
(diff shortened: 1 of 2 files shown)


kanaabe:
🙌  for consistency

src/Components/Publishing/Byline/Share.tsx
```diff
@@ -37,8 +37,9 @@ export class Share extends React.Component<Props, null> {
     window.open(e.currentTarget.attributes.href.value, "Share", "width = 600,height = 300")
 
     this.props.tracking.trackEvent({
```
alloy: Normally the idea is that you would have a method that does some work unrelated to analytics (in this case that appears to be to open the shared link?) and then annotate the method with a track function. Or is there a reason I’m overlooking why that wasn’t possible here?

eessex: We are using a slightly different method here to allow for custom tracking params, for when the share component is not nested inside of an article.

PR #386, opened 2017-11-14
@kanaabe => Author updates

eessex:
Related to https://github.com/artsy/publishing/issues/244
- Changes font-family/size for author bio, makes twitter handle match size.
- Bumps image size down to 60px


kanaabe:
Easy peasy 👍

PR #390, opened 2017-11-17
@damassi => Use props to set drop-caps

eessex:
Rather than using html classes for drop-caps, the parent container (in this case sections, in Positron, sectionContainer) determines whether the section is the first text in a feature article.

Related to https://github.com/artsy/positron/pull/1451

### DIFF
--- a/src/Components/Publishing/Sections/StyledText.tsx
+++ b/src/Components/Publishing/Sections/StyledText.tsx
@@ -5,6 +5,7 @@ import { Fonts } from "../Fonts"
 import { Layout } from "../Typings"
 
 interface StyledTextProps {
+  isContentStart: boolean
   layout: Layout
   postscript?: Boolean
 }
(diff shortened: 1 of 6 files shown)


src/Components/Publishing/Sections/StyledText.tsx
```diff
@@ -111,13 +112,15 @@ export const StyledText = div`
...
     word-break: break-word;
   }
-  .content-start {
-    ${Fonts.unica("s67", "medium")}
-    float: left;
-    line-height: .5em;
-    margin-right: 10px;
-    margin-top: .298em;
-    text-transform: uppercase;
+  p:first-child:first-letter {
```
damassi: Wasn't aware of this selector, and it checks out here https://caniuse.com/#feat=css-first-letter 👍

src/Components/Publishing/Sections/StyledText.tsx
```diff
@@ -111,13 +112,15 @@ export const StyledText = div`
...
+    ${props => props.isContentStart && Fonts.unica("s67", "medium")}
+    ${props => props.isContentStart && `
+      float: left;
+      line-height: .5em;
+      margin-right: 10px;
+      margin-top: .298em;
+      text-transform: uppercase;
+  `}
```
damassi: Could collapse these two into one conditional: 

```
${props => props.isContentStart && `
  Fonts.unica("s67", "medium")
  float: left;
  ...
`}

PR #392, opened 2017-11-20
@kanaabe => Minor QA

eessex:
- Fixes problem in drop-caps, conditional wasn't rendering font-family
- Adds pointer cursor to video covers
- Adds max-width to basic headers


src/Components/Publishing/Header/BasicHeader.tsx
```diff
@@ -218,7 +218,9 @@ const Title = styled.div`
   line-height: 1.1em;
   letter-spacing: -0.035em;
   margin-bottom: 27px;
-
+  max-width: 1250px;
+  margin-left: auto;
+  margin-right: auto;
```
kanaabe: Minor, but this could consolidated into: `margin: 0 auto 27px auto;`

PR #394, opened 2017-11-22
[Onboarding] add data submission for Budget question

xtina-starr:
This adds the data submission for the [budget question](https://github.com/artsy/collector-experience/issues/507).

### DIFF
--- a/src/Components/Onboarding/Steps/Budget.tsx
+++ b/src/Components/Onboarding/Steps/Budget.tsx
@@ -16,7 +16,7 @@ const OptionsContainer = styled.div`
 `
 
 interface State {
-  selection: number
+  selection: number | null
 }
 
 class Budget extends React.Component<StepProps & ContextProps, State> {
@@ -30,12 +30,8 @@ class Budget extends React.Component<StepProps & ContextProps, State> {
     "NO BUDGET IN MIND": 1000000000000,
   }
 
-  constructor(props) {
-    super(props)
-
-    this.state = {
-      selection: null,
-    }
+  state = {
+    selection: null,
   }
 
   onOptionSelected = (index: number) => {
--- a/src/Components/Onboarding/Steps/Budget.tsx
+++ b/src/Components/Onboarding/Steps/Budget.tsx
@@ -1,70 +1,90 @@
 import * as React from "react"
+import { commitMutation, graphql } from "react-relay"
 import styled from "styled-components"
 
+import { ContextConsumer, ContextProps } from "../../Artsy"
 import SelectableToggle from "../SelectableToggle"
 import { StepProps } from "../Types"
 import { Layout } from "./Layout"
 
 const OptionsContainer = styled.div`
-width: 450px;
-margin: 0 auto 100px;
-&:last-child {
-  border-bottom: 1px solid #e5e5e5;
-}
+  width: 450px;
+  margin: 0 auto 100px;
+  &:last-child {
+    border-bottom: 1px solid #e5e5e5;
+  }
 `
 
 interface State {
-  selection: string
+  selection: number
 }
 
-export default class Budget extends React.Component<StepProps, State> {
-  options = [
-    "UNDER $500",
-    "UNDER $2,500",
-    "UNDER $5,000",
-    "UNDER $10,000",
-    "UNDER $25,000",
-    "UNDER $50,000",
-    "NO BUDGET IN MIND",
-  ]
+class Budget extends React.Component<StepProps & ContextProps, State> {
+  options = {
+    "UNDER $500": 500,
+    "UNDER $2,500": 2500,
+    "UNDER $5,000": 5000,
+    "UNDER $10,000": 10000,
+    "UNDER $25,000": 25000,
+    "UNDER $50,000": 50000,
+    "NO BUDGET IN MIND": 1000000000000,
+  }
 
   constructor(props) {
     super(props)
 
     this.state = {
-      selection: "",
+      selection: null,
     }
   }
 
   onOptionSelected = (index: number) => {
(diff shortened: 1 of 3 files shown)


src/Components/Onboarding/Steps/Budget.tsx
```diff
@@ -1,70 +1,90 @@
...
 
     this.state = {
-      selection: "",
+      selection: null,
     }
   }
 
   onOptionSelected = (index: number) => {
-    let selection = { selection: this.options[index] }
+    let selection = { selection: Object.values(this.options)[index] }
```
xtina-starr: The typescript compiler complains that `values` doesn't exist on type `ObjectConstructor`, but the [property values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) indeed exists. Any idea why typescript might not like it?

xtina-starr:
Also I actually paired with @yuki24 on this then went ahead and finished up the second part of this work on my own, but forgot to change the author in `duet-commit`. Sorry Yuks! I'm heading offline now but will try and resolve and amend the commit when I'm back.

src/Components/Onboarding/Steps/Budget.tsx
alloy: Welcome to the joy of the many JS specifications! :tada:

If you look at [the ‘specifications’ section of the MDN docs](https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Object/values#Specifications) for `Object.values()`, you will notice that it only exists in the ‘ES2017’ spec, not the ‘ES2015’ one.

‘Fun’ fact is that the JS specification version names such as ‘ES6’ are deprecated and replaced by a yearly scheme. ‘ES6’ _is_ ‘ES2015’.

So looking at our `tsconfig.json` file, we were targeting a standard library `lib` for version ‘ES6’/’ES2015’, whereas we should target ‘ES2015’ through ‘ES2017’, after which TS will know about these new APIs being available:

```diff
diff --git a/tsconfig.json b/tsconfig.json
index 329aaf4..e750a8e 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,8 @@
     ],
     "rootDir": "src",
     "sourceMap": true,
-    "target": "es6"
+    "target": "es2015",
+    "lib": ["dom", "es2015", "es2016", "es2017"]
   },
   "exclude": ["node_modules", "dist", "data", "externals", "danger"]
 }
```

You can apply this and add it to your PR.

src/Components/Onboarding/Steps/Budget.tsx
```diff
@@ -1,70 +1,90 @@
...
+    "UNDER $2,500": 2500,
+    "UNDER $5,000": 5000,
+    "UNDER $10,000": 10000,
+    "UNDER $25,000": 25000,
+    "UNDER $50,000": 50000,
+    "NO BUDGET IN MIND": 1000000000000,
+  }
 
   constructor(props) {
     super(props)
```
alloy: Side-note, unfortunately when you override the `constructor`, you need to re-specify the typing of `props` otherwise those get lost completely.

```ts
type Props = StepProps & ContextProps

class Budget extends React.Component<Props, State> {
  constructor(props: Props) {
    super(props)
    this.state = {
      selection: null
    }
  }

  // ...
}
```

Or another thing you can do, if you do nothing in your override besides setting **static** initial state, is to use the syntax you used to define the `options` instance variable, thus not needing to override the constructor at all:

```ts
class Budget extends React.Component<Props, State> {
  state = {
    selection: null
  }

  // ...
}
```

src/Components/Onboarding/Steps/Budget.tsx
```diff
@@ -1,70 +1,90 @@
...
+  width: 450px;
+  margin: 0 auto 100px;
+  &:last-child {
+    border-bottom: 1px solid #e5e5e5;
+  }
 `
 
 interface State {
-  selection: string
+  selection: number
```
alloy: You set the initial state to `null`, so this should really be `number | null`.

alloy:
Looking pretty good, just a few small comments 👍

src/Components/Onboarding/Steps/Budget.tsx
orta: Sidenote: This doesn't make guarantees about those functions existing at runtime though ( It's not going to add them , just declare they exist in the types) so if this is running on older web clients, the function may fail

alloy: Yes definitely, the assumption here is that there would be a polyfill for those features for older browsers (such as provided by Babel).

src/Components/Onboarding/Steps/Budget.tsx
xtina-starr: In this case would it be preferred to do both. Define the types and remove the constructor and define state as a static variable or is this an either or type of thing? I remember you pointing out the constructor thing out when pairing with Maxim and I awhile back. If I remember correctly removing the constructor is the preferred method. I'm just not 100% sure why.

src/Components/Onboarding/Steps/Budget.tsx
xtina-starr: Ah yes, this makes sense! Good catch 👌

src/Components/Onboarding/Steps/Budget.tsx
alloy: Yeah I would choose removing the constructor. Not because it’s inherently better than the other, jut because it leads to less code.

xtina-starr:
@alloy almost forgot about this guy. Do you have any more feedback?

alloy:
Nope 👍

PR #396, opened 2017-11-27
@kanaabe => Series: ArticleCard

eessex:
ArticleCard component shows a condensed article view, and displays video length and caret if the article has a video. 

Also adds the ability to pass a color as props to the Byline component. 

Video article:
screen shot 2017-11-27 at 3 34 06 pm

Non-video article:
screen shot 2017-11-27 at 3 34 16 pm


src/Components/Publishing/Series/ArticleCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+  bottom: 20px;
+  border-top: 30px solid transparent;
+  border-bottom: 30px solid transparent;
+  border-left: 40px solid ${props => props.color};
+  ${props => pMedia.md`
+    border-top: 22px solid transparent;
+    border-bottom: 22px solid transparent;
+    border-left: 30px solid ${props.color};
+  `}
+`
```
kanaabe: Ah the play button has been implemented a few times in Reaction now - might want to consider extracting this out into a reusable styled div or as an SVG?

eessex: We already have a component for video play, but left this one as css because it has no 'onPlay' triggering.

eessex: SVG is probably a good idea, though the width here is a little different from our normal caret -- will check w owen if we can standardize.

PR #398, opened 2017-11-28
@kanaabe => Video Series page

eessex:
Series page (except header menu)

screen shot 2017-11-30 at 10 46 39 am


src/Components/Publishing/Fixtures/Articles.ts
```diff
@@ -669,15 +669,29 @@ export const MissingVerticalStandardArticle = extend({}, StandardArticle, {
   vertical: null,
 })
 
+export const VideoArticle = extend({}, StandardArticle, {
```
kanaabe: This is minor, but it seems a little odd to extend another article layout in this instance since it means `VideoArticle` also has `sections` which shouldn't be true. I would prefer to create a new one or make a secondary Generic article with just the basics.

src/Components/Publishing/Fixtures/Articles.ts
```diff
@@ -669,15 +669,29 @@ export const MissingVerticalStandardArticle = extend({}, StandardArticle, {
...
+})
+
 export const SeriesArticle = extend({}, StandardArticle, {
   layout: "series",
   title: "The Future of Art",
+  about: "<p>Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Curabitur blandit tempus porttitor. Donec ullamcorper nulla non metus auctor fringilla. Donec ullamcorper nulla non metus auctor fringilla. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Maecenas faucibus mollis interdum. Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p><p>Duis mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit. Curabitur blandit tempus porttitor. Sed posuere consectetur est at lobortis. Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Nullam quis risus eget urna mollis ornare vel eu leo.</p><p>Donec sed odio dui. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Nullam quis risus eget urna mollis ornare vel eu leo. Nulla vitae elit libero, a pharetra augue.</p>",
+  related_articles: [
+    VideoArticle,
+    FeatureArticle
+  ]
```
kanaabe: Nice, I like `related_articles` as the relation. It will work well with other article layouts.

I'm not too sure with the naming for `about` though. Maybe it should be `about_the_film`. It makes me wonder -- if we're planning on having something like `media` to represent the "content" of the video article, how can we do the same for a series? Especially if the "content" of a series can expand in the future. (I'm still not convinced that series should be a layout, but I need to sit on that one)

src/Components/Publishing/Series/Components/ArticleCard.tsx
```diff
@@ -42,7 +42,7 @@ export class ArticleCard extends Component<Props, null> {
 
         <ImageContainer>
           <Image src={crop(article.thumbnail_image, { width: 680, height: 450 })} />
-          {video &&
+          {media &&
             <VideoPlay><IconPlayCaret color={color} /></VideoPlay>
           }
         </ImageContainer>
```
kanaabe: We've been putting our subcomponents in the main entity's directory, like with:

`Components/Publishing/Display/Canvas/CanvasText.tsx`
and
`Components/Publishing/Header/ClassicHeader.tsx`

It's not a big deal, but I'm wondering why you want to nest here with an additional `Components` directory? It looks like there are only 3 subcomponents and they all have good naming.

src/Components/Publishing/Series/Components/SeriesAbout.tsx
```diff
@@ -0,0 +1,112 @@
...
+  color?: string,
+  series?: any
+}
+
+export class SeriesAbout extends Component<Props, null> {
+  public static defaultProps: Partial<Props>
+
+  render () {
+    const { color, series } = this.props
+    const { about, sponsor } = series
```
kanaabe: This `series` is the same thing as `article`, right? Do you think that could be confusing? I think if we end up modeling Series as an entirely different model, this would make sense but in this implementation we should probably call it what it is. I'm thinking of it as a different article _layout_ here. But I admit there's still some ambiguity overall on how we want to approach this. Maybe we can leave this as is for now and loop back sometime tomorrow on a full plan for modeling Series.

src/Components/Publishing/Series/Components/SeriesAbout.tsx
```diff
@@ -0,0 +1,112 @@
...
+        display: block;
+      }
+    }
+    &--last {
+      width: 60%;
+      .PartnerBlock {
+        display: none;
+      }
+    }
+  }
```
kanaabe: Shouldn't we be able to do this the `styled-components` way without using `classNames`? Unless I'm missing something?

src/Components/Publishing/Series/Series.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  .ArticleCard {
+    margin-bottom: 60px;
+  }
+
+  .SeriesAbout {
+    padding-top: 60px;
+  }
```
kanaabe: Same here, we should be able to reference and update with: 

```
${SeriesAbout} {
   padding-top: 60px;
}
```
to avoid usage of className.

src/Components/Publishing/Series/__test__/SeriesTitle.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+it("Renders partner block for a sponsored series", () => {
+  const component = mount(<SeriesTitle series={SeriesArticleSponsored} />)
+  expect(component.find('.PartnerBlock').length).toBe(1)
+})
+
+it("Does not render partner block for an unsponsored series", () => {
+  const component = mount(<SeriesTitle series={SeriesArticle} />)
+  expect(component.find('.PartnerBlock').length).toBe(0)
+})
```
kanaabe: Appreciate the inclusion of unit tests! ❤️

src/Components/Publishing/Series/Series.tsx
eessex: Doesn't that only work for targeting a styled component?  These are references to react components nested in the series.  Didn't want to include the outside padding in the component itself because it could potentially change in the future based on context.

kanaabe: Yes, it's for targeting a styled component. I'm saying we could replace the usage of `className="SeriesAbout"` and make an exportable styled component called `<SeriesAbout>` to reference it like above. Orr we could even pass in margin as a prop?

src/Components/Publishing/Fixtures/Articles.ts
```diff
@@ -669,15 +669,48 @@ export const MissingVerticalStandardArticle = extend({}, StandardArticle, {
...
 
-export const VideoArticle = extend({}, StandardArticle, {
-  layout: "video",
-  video: {
-    url: '',
-    length: '03:12',
-  },
+export const SeriesArticleSponsored = extend({}, SeriesArticle, {
+ sponsor: {
+    partner_logo: 'https://artsy-media-uploads.s3.amazonaws.com/eZqsrnodcIlyyJzmRqLm4A%2FBombay_Sapphire_logo.png',
```
kanaabe: We can update this after merging since we've got more updates coming, but noting that this is now `partner_dark_logo`

src/Components/Publishing/Series/ArticleCard.tsx
```diff
@@ -17,38 +17,70 @@ interface Props {
...
+    }
+  }
+
+  openLink = (e) => {
+    e.preventDefault()
+
+    if (!this.isUnpublishedMedia()) {
+      window.open(e.currentTarget.attributes.href.value)
+    }
+  }
```
kanaabe: Ooh, is this because empty `href=""` will jump to the top of the page or something?

kanaabe:
Nice! This looks good to go now. 🥂

src/Components/Publishing/Series/ArticleCard.tsx
eessex: Yes exactly

PR #399, opened 2017-11-29
@eessex => Editorial Nav

kanaabe:
Regular:
image

Transparent:
image

Mobile:
image


src/Components/Publishing/Fixtures/Articles.ts
```diff
@@ -591,6 +591,12 @@ export const FeatureArticle: ArticleData = {
   slug: "artsy-editorial-path-winning-art-prize",
 }
 
+export const SponsoredArticle = extend({}, FeatureArticle, {
+  sponsor: {
+    white_logo: "https://artsy-media-uploads.s3.amazonaws.com/F4RVgsSXv3q9Nrt59P8gbA%2Ffullscreen.png"
```
kanaabe: I'm kinda feeling like we have this opportunity to redo partner logos now and maybe this format is a little clearer than `partner_logo_fullscreen`. Thoughts? Maybe `light_logo` and `dark_logo` is even more flexible...

eessex: Yeah light_logo/dark_logo makes sense to me -- are there places where a brand has used a different logo for a header placement versus footer for layout reasons as opposed to color?  For example, some brands have an alternative logo for a small placement versus large.

eessex:
It looks like 'artsy editorial' is not centered in the wide views, both with and without the logos -- would assume it should be centered.

kanaabe:
Yeah, they're spaced evenly between each other to avoid collision, but you're right, it makes sense to float the left/right and always center the title. Will double check w/owen and update!

src/Components/Publishing/Fixtures/Articles.ts
kanaabe: Mm not that I can recall -- we've had:

`partner_fullscreen_header_logo`: to account for dark bgs
`partner_logo`: for navs with white background
`secondary_partner_logo`: used only once afaik here:
image

kanaabe:
Oo just got the expanded mobile nav on this one so marking it WIP

kanaabe:
Added the mobile dropdown without the user logged in/out section (this will come later). The mobile menu lives in a directory one level up from `Publishing` in case we want to implement the general nav in React at some point.

image

image

src/Components/Publishing/Icon/IconPlus.tsx
```diff
@@ -33,8 +31,10 @@ export const IconPlus: React.SFC<Props> = props => {
 const StyledSvg = styled.svg`
   width: 15px;
   height: 15px;
+  margin: 0 20px;
```
eessex: Margins seem contextual -- wouldn't it make more sense to do this outside the svg or add a prop?

src/Components/Publishing/Nav/__test__/Nav.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+    expect(nav).toMatchSnapshot()
+  })
+  it("includes a partner logo in a sponsored nav", () => {
+    const nav = mount(<Nav sponsor={SponsoredArticle.sponsor} />)
+    expect(nav.html()).toMatch("https://artsy-media-uploads.s3.amazonaws.com/F4RVgsSXv3q9Nrt59P8gbA%2Ffullscreen.png")
+  })
+  it("renders a custom title", () => {
+    const nav = mount(<Nav title="Custom Title" />)
+    expect(nav.text()).toMatch("Custom Title")
+  })
```
eessex: Would make sense to add a test for the open/close action

src/Components/Publishing/Icon/IconPlus.tsx
kanaabe: Good catch!

kanaabe:
Addressed your comments -- going to merge this pup!

PR #400, opened 2017-11-30
Fix relay modern mutation

yuki24:
* adds a patch from https://github.com/artsy/reaction/commit/20ddcf2b20ea9b17accbbec23285c8544d88427b.patch
 * fixes the function to execute the follow artist mutation
 * replaces `any` with appropriate types


alloy:
Looks good!

PR #401, opened 2017-12-01
@eessex => Support missing logos in the DisplayPanel ad

kanaabe:
Addresses https://github.com/artsy/publishing/issues/258


eessex:
Looks good 👍

PR #402, opened 2017-12-01
User can see a related artist after searching and following an artist

yuki24:
Now the user can click an artist and see a related artist. 

 * moved the `#onArtistFollowed` function up to the `ArtistSearchResultsContent` and `PopularArtistsContent` to be able to update different stores for different relay queries. Not sure about the interface yet but I think we should be able to figure out and make it more mature as we add more logic to it (e.g. Follow Gene)
 * Animation is not implemented yet, but could be addressed as a separate task
 * The `SelectableItemContainer` component could be inlined since all it does is just to iterate over each item on `this.props.artists` (proof of concept: https://github.com/yuki24/reaction/commit/bd2bdd312c39340ed8014fe299de8b2c7e4a9b58)

## To-dos

 * [x] Rebase once https://github.com/artsy/reaction/pull/403 is merged
 * [x] Use TypeScript's `interface` to expose the API for `updater`

### DIFF
--- a/src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
+++ b/src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
@@ -1,18 +1,33 @@
 import * as React from "react"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 
+import { RecordSourceSelectorProxy, SelectorData } from "relay-runtime"
 import { ContextConsumer, ContextProps } from "../../../Artsy"
 import SelectableItemContainer from "./SelectableItemContainer"
 
 export interface RelayProps {
+  term: string
   viewer: {
     match_artist: any[]
   }
 }
 
 class ArtistSearchResultsContent extends React.Component<RelayProps, null> {
+  onArtistFollowed(artistId: string, store: RecordSourceSelectorProxy, data: SelectorData): void {
+    const suggestedArtist = store.get(data.followArtist.artist.related.suggested.edges[0].node.__id)
+
+    const popularArtistsRootField = store.get("client:root:viewer")
+    const popularArtists = popularArtistsRootField.getLinkedRecords("match_artist", { term: this.props.term })
+    const updatedPopularArtists = popularArtists.map(popularArtist =>
+      popularArtist.getDataID() === artistId ? suggestedArtist : popularArtist)
+
+    popularArtistsRootField.setLinkedRecords(updatedPopularArtists, "match_artist", { term: this.props.term })
+  }
+
   render() {
-    return <SelectableItemContainer artists={this.props.viewer.match_artist} />
+    return <SelectableItemContainer
+              artists={this.props.viewer.match_artist}
+              onArtistFollowed={this.onArtistFollowed.bind(this)} />
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
```diff
@@ -1,18 +1,32 @@
...
 
 export interface RelayProps {
+  term: string
   viewer: {
     match_artist: any[]
   }
 }
 
-class ArtistSearchResultsContent extends React.Component<RelayProps, null> {
+class ArtistSearchResultsContent extends React.Component<RelayProps, null> implements ItemLinkUpdater {
```
alloy: Oooh, very interesting use of `implements`!

However, it’s more idiomatic in JS (and our codebases) to pass just a function rather than an object that has a specific method.

src/Components/Onboarding/Steps/Artists/ItemLink.tsx
```diff
@@ -84,8 +78,8 @@ class ItemLink extends React.Component<Props, null> {
           unfollow: false,
         },
       },
-      updater: storeUpdater,
-      // optimisticUpdater: storeUpdater,
+      updater: (store: RecordSourceSelectorProxy, data: SelectorData) =>
+        this.props.itemLinkUpdater.onArtistFollowed(artist, store, data),
```
alloy: …following up on the `implements` comment, this should probably just be `this.props.onArtistFollowed(artist.__id, store, data)`

alloy:
Looks good! Just some small idiomatic JS comments

src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
yuki24: One nice thing about `implements Interface` is you can let the editor generate a method stub (at least in IntelliJ) and you don't have to deal with `.bind(this)`.

src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
```diff
@@ -1,18 +1,31 @@
...
+    const popularArtists = popularArtistsRootField.getLinkedRecords("match_artist", { term: this.props.term })
+    const updatedPopularArtists = popularArtists.map(popularArtist =>
+      popularArtist.getDataID() === artistId ? suggestedArtist : popularArtist)
+
+    popularArtistsRootField.setLinkedRecords(updatedPopularArtists, "match_artist", { term: this.props.term })
+  }
+
   render() {
-    return <SelectableItemContainer artists={this.props.viewer.match_artist} />
+    return <SelectableItemContainer artists={this.props.viewer.match_artist} onArtistFollowed={this.onArtistFollowed} />
```
alloy: I suspect that `this.props.term` in `onArtistFollowed` will not work because this is an unbounded function.

yuki24:
I'm still fixing my local but I'll test this once it's fixed.

yuki24:
confirmed this is working fine. In staging there is still some missing data, so follow the steps below for QA:

### Follow an artist without searching

 1. Select **Artist Selector** in the Story Book menu
 2. Select **JEFF KOONS**
 3. JEFF KOONS will be replaced with a related artist

### Search and follow

 1. Select **Artist Selector** in the Story Book menu
 2. Search for **Jonas Wood**
 3. Click **Jonas Wood**
 3. **Jonas Wood** will be replaced with a related artist

We have a separate story for artists that don't have any related artists: https://github.com/artsy/collector-experience/issues/811

alloy:
Nice! 👌

PR #403, opened 2017-12-01
Remove unused state in the Artist component

yuki24:
This removes unused state I added a few months ago. We are working on the Follow Artist component with a different approach (https://github.com/artsy/reaction/pull/402), and this has been just mysterious code that prevented us from refactoring. Let's just delete it and come back once we finish the Follow Artist and Follow Gene components.


xtina-starr:
This looks good to me. Merging!

yuki24:
Thanks!

PR #406, opened 2017-12-04
Use yarn 1.3.2 to install dependencies in CircleCI

yuki24:
This should fix the `tslint` failure occurring in CircleCI.


alloy:
Ah so we only use v1.3.2 to intall the deps? Yeah I guess there’s no real reason to use it for all commands, seeing as it’s really just the install phase that fails.

PR #408, opened 2017-12-05
@eessex => Video Player

kanaabe:
Oof this one turned out much larger of a PR than anticipated. Happy to go over the code together if that is more helpful. I have one more UI thing with the scrubber, but functionally, this all works!

Demo:
video-player


src/Components/Publishing/Icon/IconVideoFullscreen.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from "react"
+
+export class IconVideoFullscreen extends Component<any, null> {
```
damassi: I _think_ this could be different in reaction (as compared to emission), but if state isn't being used `null` can be omitted

src/Components/Publishing/VideoPlayer/MuteUnmute.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { IconVideoUnmute } from "../Icon/IconVideoUnmute"
+
+interface Props extends React.HTMLProps<HTMLSpanElement> {
+  isMuted: boolean
+  toggleMute: () => void
+}
+
+export class MuteUnmute extends Component<Props, null> {
+  shouldComponentUpdate(nextProps) {
+    return this.props.isMuted !== nextProps.isMuted
```
damassi: Since the props being passed in are simple and can be compared shallowly, can use `React.PureComponent` instead of checking against `shouldComponentUpdate`.

src/Components/Publishing/VideoPlayer/PlayPause.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface Props extends React.HTMLProps<HTMLSpanElement> {
+  isPlaying: boolean
+  togglePlay: () => void
+}
+
+export class PlayPause extends Component<Props, null> {
+  shouldComponentUpdate(nextProps) {
+    return this.props.isPlaying !== nextProps.isPlaying
+  }
```
damassi: Same here

src/Components/Publishing/VideoPlayer/Scrubber.tsx
```diff
@@ -0,0 +1,155 @@
...
+}
+
+export class Scrubber extends Component<Props, State> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      isScrubbing: false,
+      isPlayingOnMouseDown: false
+    }
+  }
```
damassi: Can move state out of constructor and eliminate whole block

src/Components/Publishing/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      isMuted: false,
+      isPlaying: false,
+      currentTime: 0,
+      duration: 0
+    }
+  }
```
damassi: No need for constructor

src/Components/Publishing/VideoPlayer/__test__/FullscreenHelpers.test.js
```diff
@@ -0,0 +1,136 @@
...
+      enumerable: true,
+      value: true,
+      writable: true
+    })
+    Object.defineProperty(document, 'fullscreenEnabled', {
+      configurable: true,
+      enumerable: true,
+      value: true,
+      writable: true
+    })
```
damassi: @kanaabe - curious why this was needed?

src/Components/Publishing/Icon/IconVideoMute.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from "react"
+
+export class IconVideoMute extends Component<any, null> {
```
eessex: Might make sense to include color prop on all icons, especially if you're thinking to package up.

src/Components/Publishing/VideoPlayer/Timestamp.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import styled from "styled-components"
+import { Fonts } from "../Fonts"
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  duration: number
+  currentTime: number
+}
+
+const formatTime = (time) => {
```
eessex: We will need to format the time in the ArticleCard component too if we aren't saving duration as a string -- maybe a good idea to move this into constants.

kanaabe: Ah yeah, the player doesn't actually use the duration from the article. It picks up the exact duration here once video metadata is loaded: https://github.com/artsy/reaction/pull/408/files#diff-fb979d99f63804b1dd7cc80ddab6a9a0R45

I'll likely keep ^ to keep the player generic but good call that for ArticleCard we'll want to make the Positron side be able to do the same for article saving. Good idea to move this to a util to anticipate that. 👍

src/Components/Publishing/Fixtures/Components.ts
```diff
@@ -351,6 +351,16 @@ export const ImageSetMiniSansTitle: ImageSetPreviewProps["section"] = {
   images: Images,
 }
 
+export const Media = [
+  {
+    title: "Trevor Paglan",
+    url: "http://files.artsy.net/videos/scenic_mono_3.mp4",
+    duration: 560948,
```
eessex: Here is another instance of the duration -- was using strings here but we can update articleCard to read this.

src/Components/Publishing/VideoPlayer/__test__/FullscreenHelpers.test.js
kanaabe: Ooh yeah I believe this was an artifact of when this test was in typescript. We can definitely just use `document.fullscreenElement = true` now! Good catch ⚾️

src/Components/Publishing/VideoPlayer/MuteUnmute.tsx
kanaabe: Oooo nice!

src/Components/Publishing/Icon/IconVideoFullscreen.tsx
kanaabe: Yeah I should actually just change all the icons to be SFC. They're easier to export/style outside of the component that way too. Like with: https://github.com/artsy/reaction/pull/408/files#diff-f5accae7b0f443e61496b5c152e113aaR1

Allows you to do:
```
import { IconVideoPause } from "../../Icons/IconVideoPause"

const IconContainer = styled.div`
   ${IconVideoPause} {
      margin: 20px;
   }
`
```

eessex: 👍

kanaabe:
Alright, comments addressed! Please have a final look and let me know. I changed ArticleCard to have a number duration instead of a string using the `formatTime` method.

PR #409, opened 2017-12-06
@kanaabe => Series components accept children

eessex:
In anticipation of editing new article types, adds support for rendering children to Series components.

### DIFF
--- a/src/Components/Publishing/Series/ArticleCard.tsx
+++ b/src/Components/Publishing/Series/ArticleCard.tsx
@@ -1,4 +1,4 @@
-import React, { Component  } from "react"
+import React, { Component } from "react"
 import styled, { StyledFunction } from "styled-components"
 import { crop } from "../../../Utils/resizer"
 import { pMedia } from "../../Helpers"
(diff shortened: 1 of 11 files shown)


src/Components/Publishing/Series/ArticleCard.tsx
```diff
@@ -62,10 +63,14 @@ export class ArticleCard extends Component<Props, null> {
...
 
   render () {
-    const { article, color, series } = this.props
+    const { article, children, color, series } = this.props
     const { media } = article
     const isUnpublishedMedia = this.isUnpublishedMedia()
 
+    const editTitle = children && children[0]
+    const editDescription = children && children[1]
+    const editImage = children && children[2]
```
kanaabe: Curious if you had further thoughts about this pattern. This works, but I can see how having to know the order of the children could become more of a burden as others try to build an editable  `ArticleCard`.

What if these components were passed down as optional props instead of looking at the children? Then, this parent component can decide to render it or not based on those props. That way we can be more compliant with TS and have autocomplete instead of having to go to this specific line of code to implement the component.

eessex: Sure will try this way instead

PR #410, opened 2017-12-06
@alloy => Don't show dropcaps unless feature, icon colors

eessex:
Two minor fixes -- 

- In StyledText, check if feature before showing drop-caps styles
- In IconDrag, add fill color two two shapes that were missed

### DIFF
--- a/src/Components/Publishing/Sections/StyledText.tsx
+++ b/src/Components/Publishing/Sections/StyledText.tsx
@@ -113,8 +113,10 @@ export const StyledText = div`
     word-break: break-word;
   }
   p:first-child:first-letter {
-    ${props => props.isContentStart && Fonts.unica("s67", "medium")}
-    ${props => props.isContentStart && `
+    ${props => props.isContentStart && props.layout === "feature" &&
+      Fonts.unica("s67", "medium")
+    }
+    ${props => props.isContentStart && props.layout === "feature" && `
       float: left;
       line-height: .5em;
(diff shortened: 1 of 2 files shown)


src/Components/Publishing/Sections/StyledText.tsx
```diff
@@ -113,8 +113,10 @@ export const StyledText = div`
     word-break: break-word;
   }
   p:first-child:first-letter {
-    ${props => props.isContentStart && Fonts.unica("s67", "medium")}
-    ${props => props.isContentStart && `
+    ${props => props.isContentStart && props.layout === "feature" &&
+      Fonts.unica("s67", "medium")
+    }
+    ${props => props.isContentStart && props.layout === "feature" && `
```
alloy: This seems to use the exact same conditionals as above. Can you combine the two?

eessex: For some reason the font declaration does not like being combined with the multiline styles --  tried a few ways, for example-- 
`${props => props.isContentStart && props.layout === "feature" &&
      Fonts.unica("s67", "medium") + `
      float: left...`
This rendered the styles as strings, but not the fonts.  Is there a way I'm missing?

alloy: Hmm, no not really alas. This is not blocking imo btw, so feel free to merge at any time 👍 

@l2succes Do you happen to know more about this with styled-components?

PR #412, opened 2017-12-11
@kanaabe => More edit changes for series/article card

eessex:
Minor changes for editing series/article card in writer:
- ArticleCard does not link out if editable fields are present
- ArticleCard accepts an editable date child
- Export SeriesContainer styled component


kanaabe:
Looks good!

PR #414, opened 2017-12-12
@eessex => Use grid in SeriesAbout

kanaabe:
One thing I noticed while building out the video landing pages was that Series components don't use `react-styled-flexboxgrid` so it causes a little inconsistency when importing them into the video pages. I think if we're committed to using grid systems, we should start with these pages since they're new. This converts the SeriesAbout component to use `react-styled-flexboxgrid`. We'll have to get to regular article pages another time!

Note that this uses the breakpoints and sizes of the regular grid layout, not our publishing ones.

Some screenshots:
image
image

### DIFF
--- a/src/Apps/Loyalty/Containers/3wThankYou/__tests__/__snapshots__/index.ts.snap
+++ b/src/Apps/Loyalty/Containers/3wThankYou/__tests__/__snapshots__/index.ts.snap
@@ -30,7 +30,7 @@ exports[`3-way Handshake Thank You renders the snapshot 1`] = `
     className="file__TitleSection-s1ds4mkl-1 kLJIlb"
   >
     <div
-      className="file__StyledTitle-s1hq9j8q-0 dyBmig"
+      className="file__StyledTitle-s1hq9j8q-0 cbQDSR"
       color="inherit"
     >
       Thank you, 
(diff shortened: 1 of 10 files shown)


eessex:
This works for me, though I thought we did not commit to the grid because it didn't match our publishing breakpoints?  Happy to go in this direction if grid is the plan.

kanaabe:
I think the idea was to start using them for new things, like we did with the basic header layout. Also these new video/series layouts don’t really need the ones specially designed for standard/feature with pixel values since we really only make use of the mobile/desktop switch. I kind of do feel strongly about getting a better grid system in our layouts but if you have additional concerns/caveats about them, lmk! 

When we swap out standard/feature for grid, we should probably establish new em/rem based breakpoints that could also work with video/series. That way, swapping out video/series will be a simple matter of adding a `ThemeProvider` to our layouts.

damassi:
Note that its possible to configure grid breakpoints: https://github.com/LoicMahieu/react-styled-flexboxgrid#configuration

kanaabe:
Yep thanks! I tried applying pixel values as a theme to ThemeProvider but it wasn’t too happy. I think this lib wants em/rem to work properly. I just spoke to owen about this too — I think we can figure out a system using em/rem that can be applied to all the layouts 👍🏼

PR #415, opened 2017-12-13
@eessex => Video Landing

kanaabe:
Landing page for the video article type. This doesn't include the analytics or editable props but this PR was getting too long anyway! I will follow up with the part 2. 

ezgif-3-558a31f078


src/Components/Publishing/Fixtures/Articles.ts
```diff
@@ -685,19 +685,33 @@ export const VideoArticle = {
   published_at: "2017-07-28T20:38:05.709Z",
   description: "The elegant spiral of the Nautilus shell, the sinuous pattern of the banks of a river, or the swirling vortex street of clouds - patterns exist.",
   media: {
-    url: "https://artsy-media-uploads.s3.amazonaws.com/z9w_n6UxxoZ_u1lzt4vwrw%2FHero+Loop+02.mp4",
+    title: "Trevor Paglan",
```
eessex: Looks like this info is repeated below and in the components file, maybe you could import this data from there?

src/Components/Publishing/Video/VideoLayout.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    color="white"
+                    series={seriesArticle}
+                  />
+                </Col>
+              </MaxRow>
+            )
+          })
+        }
+        <MaxRow>
+          {seriesArticle &&
```
eessex: Since there's nothing else in this last `<MaxRow>` they could nest inside the conditional to avoid rendering empty.

src/Components/Publishing/Constants.ts
```diff
@@ -25,6 +25,16 @@ export const getArticleHref = slug => `/article/${slug}`
  */
 export const getArticleFullHref = slug => `https://www.artsy.net/article/${slug}`
 
+/**
+ * Relative path to editorial entity
+ */
+export const getEditorialHref = (type, slug) => `/${type}/${slug}`
+
```
eessex: Nice, these will be useful!

PR #420, opened 2017-12-14
@kanaabe => Minor Series updates

eessex:
All in SeriesLayout:
- Uses relatedArticles as prop, rather than related_article_ids
- Adds Nav
- Adds backgroundColor prop and defaults to black
- Adds support to render background images


kanaabe:
Nice! Straightforward and well tested

PR #421, opened 2017-12-15
@eessex => Editable VideoCover and VideoAbout components

kanaabe:
Editable Cover and About sections:

image

image


src/Components/Publishing/Video/VideoInfoBlock.tsx
```diff
@@ -5,40 +5,44 @@ import { formatTime } from "../Constants"
...
 }
 
 export class VideoInfoBlock extends Component<Props, null> {
   render() {
     const {
+      editTitle,
       media,
-      seriesTitle,
+      tag,
+      title
```
eessex: Tag is a bit of a confusing name here because it is not literally a tag, although that is something that exists on articles. Maybe SuperTitle or SubTitle?

kanaabe: Good suggestion, thanks!

PR #423, opened 2017-12-15
@kanaabe => Nav fixes on scroll

eessex:
Adds waypoint to Nav component to fix the menu on scroll.  This is enabled by default but can be removed by setting props.canFix to false. 

Also adds the Nav to series pages.


kanaabe:
Solid!

PR #425, opened 2017-12-15
Exclude ids from gene default list

xtina-starr:
closes: https://github.com/artsy/collector-experience/issues/841


src/Components/Onboarding/Steps/Genes/SuggestedGenes.tsx
```diff
@@ -25,23 +25,33 @@ interface Props extends React.HTMLProps<HTMLAnchorElement>, RelayProps {
...
 class SuggestedGenesContent extends React.Component<Props, null> {
-  followedGene(geneId: string) {
+  private excludedGeneIds: Set<string>
+
+  constructor(props: RelayProps, context: any) {
+    super(props, context)
+    this.excludedGeneIds = new Set(this.props.suggested_genes.map(item => item._id))
+  }
+
+  followedGene(followedGene: any) {
```
alloy: Rather than typing this as `any`, how about moving the gene specific props out of `RelayProps` into a separate `Gene` interface and use that here?

PR #426, opened 2017-12-15
[WIP] Exclude ids from artist default list

xtina-starr:
This is our work in progress refactor. It does not yet have the implementation to exclude ids.

cc: @yuki24


yuki24:
I'll close and re-send this PR since the `relay-modern` branch has been merged into `master`.

PR #427, opened 2017-12-17
@kanaabe => Use 100vw for mobile embeds

kanaabe:
Prevents misconfigured iframe content from overflowing the div on mobile. This isn't necessary, but it helps quickly solve an issue with mobile embeds on a recent article: https://www.artsy.net/article/artsy-editorial-20-influential-artists-2017

Additionally renames a folder `__tests__` to `__test__` for consistency.


eessex:
It would be great to add a rule to enforce naming conventions for the test folder, as they are a bit different in every repo.

orta:
Aye, I'd love to move them all to `_tests` - don't think we need the double unders

@eessex make an RFC for discussion ( e.g. like https://github.com/artsy/emission/issues/861 )

PR #440, opened 2017-12-19
@kanaabe => Text component can accept color/style links

eessex:
Adds a color prop to Text and StyledText component, to accommodate styling of links color and underline.


kanaabe:
Nice, thank you!

PR #442, opened 2017-12-19
@kanaabe => Series Background Images

eessex:
To make this a bit easier on the writer side, moved the background color/image off of the series container and into its own component.  

If there is an image, it shows a fixed bg image with gradient.  Otherwise it gives a fixed bg color.


kanaabe:
Ah yeah, this is nice 👍

kanaabe:
Btw I have a working branch on Force with a Reaction bump, lmk once you bump the version again for this one and I'm happy to include it in the Force branch!

eessex:
Sounds good, will bump when the master build finishes

PR #443, opened 2017-12-19
@eessex => Update partner logos to use new fields

kanaabe:
This should do it for partner logos!

Nav now using `partner_condensed_logo` properly:
image

Titles and about sections all use `partner_light_logo` properly:
image


kanaabe:
Note that all these logos are based on fixtures and will not actually be this ugly blue on black! cc @orta 😛

PR #449, opened 2017-12-20
@kanaabe => iOS overlapping text bug

eessex:
Was able to replicate this bug in browserstack, but not xCode's simulators -- maybe we should get an account... the free trial is only 30min.

- Changes articleCard's image container height to 'inherit'


kanaabe:
Yes, lets definitely get an account! I forgot where it was left off but I remember @damassi was interested in getting one for the team. Maybe it exists already?

PR #452, opened 2017-12-21
[Dep] Move React to peerDeps 

damassi:
I'm upgrading Force to React 16 right now and ran into an issue where `yarn list react` shows multiple copies, which was leading to client errors; one copy installed at the root and one in `@artsy/reaction-force`. Moving to peerDeps and testing locally by pointing at this branch in Force's `package.json` and navigating to http://localhost:5000/gene/collective-history confirms a fix. 

https://github.com/artsy/force/pull/2053 is blocked in the meantime.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@artsy/reaction-force",
-  "version": "0.16.13",
+  "version": "0.16.14",
   "description": "Force’s React Components",
   "main": "dist/index.js",
   "repository": "https://github.com/artsy/reaction.git",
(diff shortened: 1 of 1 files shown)


orta:
Makes sense to me

l2succes:
So I'm currently working to get this PR ✅ https://github.com/artsy/reaction/pull/273

@damassi How is this going to work when reaction is upgraded to react 16 and force is still on react 15 if it's a peer dep?

damassi:
@l2succes - I'm updating Force now

orta:
This probably shouldn’t be permanent- ideally reaction force’s version of react should be canonical for downstream - but it gets hard when you can have many instances of a dep in the tree

l2succes:
FYI, I also move react to peerDeps in the react 16 upgrade PR (https://github.com/artsy/reaction/pull/273) so no need to merge that one

alloy:
Going to close this now that #273 has been merged.

PR #454, opened 2017-12-21
[storybook] Actually load env so external vars get merged with file.

alloy:


orta:
Does anyone ever run this manually? It could leak a real user's access token 

Could be worth adding an exception if it's not the demo user we expect?

alloy:
I personally don’t really worry about this, nobody runs this manually afaik. When CI got stuck in the past we just had an outdated deployed storybook.

PR #455, opened 2017-12-21
@kanaabe => Add ImageCollection subdirectory to /Sections

eessex:
Just a little bit of cleanup... the `/Sections` directory was getting a bit hard to navigate, so I've moved everything related to imageCollection into a subdirectory, similar to how positron is setup.  
There are a few other components that could use the same treatment but will save for separate PRs.

Also updated the Image component to use an editCaption prop rather than children.  Will follow up with a positron PR to use this.


kanaabe:
Hmm I'm not totally convinced that some of these components should be in `ImageCollection`. For example, `ArtworkCaption` gets used in the `FullscreenViewer` component, but it has to get it through the `ImageCollection` section's directory. The same goes for `ImageWrapper` -- this component isn't limited to use in an `ImageCollection` and although it's not used in another place atm, we did try to use it in `ImageSet` a while ago.

It seems like in this case we should move the more generic components into someplace outside of `Sections` entirely. Thoughts?

kanaabe:
Just noticed this is still open, thoughts on this one @eessex?

eessex:
Still thinking we need more order here, but will revisit at another time. Closing for now.

PR #457, opened 2017-12-22
[Toolchain] Add gulp-watch for recompiles

damassi:
This works hand-in-hand with our new `watchModules` option in `@artsy/express-reloadable` -- when making changes to Reaction while working in Force, updates are now automatically reflected without having to rerun commands.


damassi:
Don't want to block you while you're in this publishing crunch so going to merge!

alloy:
Can I get a bit more context on what triggered the need for this? Especially considering we may drop gulp.

damassi:
@alloy - Yup. I'm sure there are other ways to watch things, but the issue was the need to work concurrently between Reaction and Force for cross-browser, real-world testing (meaning, testing in force rather than testing in reaction). Before adding the watch, we'd always have to use the command line to recompile things, but now with `yarn watch` when changes are made to sources the `dist` folder is updated immediately and a reload in force reflects the change (due to `watchModules` option in `express-reloadable`. It speeds up developer experience since one can start the command and forget about it.

alloy:
Gotcha 👍 

Once/when/if we add TS emitting support to relay-compiler, then we no longer need that Babel part of the Gulpfile and we only needs `tsc`, which has its own watch mode which, I presume, will have incremental builds and be faster.

damassi:
Yeah I was wondering about that! Like if it would be possible to tap into tsc's incremental builds and pipe them into dist rather than having to do a full recompile on source change. 

I think all of this will become increasingly less important once we get Babel 7 into everything, but still something to think about for a little while.

PR #458, opened 2017-12-22
@damassi => IE Fixes for Video Page

kanaabe:
Fixes IE issues -- About sections without widths, -ms prefix for scrubber, etc. Looks like Prettier had a go with these files, I don't know remember turning that back on but 🤷‍♀️ I'll note the actual changes inline.

### DIFF
--- a/src/Components/Publishing/Nav/Nav.tsx
+++ b/src/Components/Publishing/Nav/Nav.tsx
@@ -24,14 +24,14 @@ interface DivProps {
 export class NavComponent extends React.Component<Props, State> {
   static defaultProps = {
     canFix: true,
-    transparent: false
+    transparent: false,
   }
 
   state = {
-    isFixed: false
+    isFixed: false,
   }
 
-  setPosition = (isFixed) => {
+  setPosition = isFixed => {
     const { canFix } = this.props
     const currentPosition = this.state.isFixed
 
@@ -41,39 +41,21 @@ export class NavComponent extends React.Component<Props, State> {
   }
 
   render() {
-    const {
-      sponsor,
-      className,
-      canFix,
-      transparent,
-      title
-    } = this.props
+    const { sponsor, className, canFix, transparent, title } = this.props
     const { isFixed } = this.state
 
     return (
       <div>
-        <NavContainer
-          transparent={!isFixed && transparent}
-          className={className}
-          isFixed={canFix && isFixed}
-        >
+        <NavContainer transparent={!isFixed && transparent} className={className} isFixed={canFix && isFixed}>
           <PartnerInline
             url={sponsor && sponsor.partner_logo_link}
             logo={sponsor && sponsor.partner_condensed_logo}
             color="white"
             margin="0 10px"
           />
-          <Title>
-            {title
-              ? title
-              : <a href="/magazine">Artsy Editorial</a>
-            }
-          </Title>
+          <Title>{title ? title : <a href="/magazine">Artsy Editorial</a>}</Title>
         </NavContainer>
-        <Waypoint
-          onEnter={() => this.setPosition(false)}
-          onLeave={() => this.setPosition(true)}
-        />
+        <Waypoint onEnter={() => this.setPosition(false)} onLeave={() => this.setPosition(true)} />
       </div>
     )
   }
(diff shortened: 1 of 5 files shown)


src/Components/Publishing/Nav/Nav.tsx
```diff
@@ -41,39 +41,21 @@ export class NavComponent extends React.Component<Props, State> {
...
+    const { sponsor, className, canFix, transparent, title } = this.props
     const { isFixed } = this.state
 
     return (
       <div>
-        <NavContainer
-          transparent={!isFixed && transparent}
-          className={className}
-          isFixed={canFix && isFixed}
-        >
```
damassi: 😢 bye bye readable code

damassi: 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢

src/Components/Publishing/Nav/Nav.tsx
```diff
@@ -41,39 +41,21 @@ export class NavComponent extends React.Component<Props, State> {
...
-              ? title
-              : <a href="/magazine">Artsy Editorial</a>
-            }
-          </Title>
+          <Title>{title ? title : <a href="/magazine">Artsy Editorial</a>}</Title>
         </NavContainer>
-        <Waypoint
-          onEnter={() => this.setPosition(false)}
-          onLeave={() => this.setPosition(true)}
-        />
```
damassi: Bye 😢

src/Components/Publishing/Video/VideoAbout.tsx
```diff
@@ -19,46 +19,31 @@ interface TitleProps {
...
     return (
       <VideoAboutContainer>
         <Col xs={12} sm={4}>
-          <Title color={color}>
-            Credits
-          </Title>
-          {editCredits
-            ? <Text layout="standard">{editCredits}</Text>
-            : <Text layout="standard" html={media.credits} />
-          }
```
damassi: Bye 😢

src/Components/Publishing/Video/VideoAbout.tsx
```diff
@@ -19,46 +19,31 @@ interface TitleProps {
...
-            About the Film
-          </Title>
-          {editDescription
-            ? <Text color={color} layout='standard'>{editDescription}</Text>
-            : <Text color={color} layout="standard" html={media.description} />
-          }
-          <ShareDate
-            color={color}
-            article={article}
-          />
```
damassi: I'm forking prettier

src/Components/Publishing/Nav/Nav.tsx
orta: I don't really get this, the new code looks better to me. 

It's short enough to fit on a single line, so it becomes more readable by being read left to right?

kanaabe: Yeah I guess it’s just personal preference - i like Prettier for everything except being able to do multiline components like this. We do this consistently in the Publishing directory and in Positron so I think we’re just used to it. But I can see how this is troublesome within the repo for half of us to use it and the other half not if we want to share components for instance!

orta: That feels like the answer is to get prettier installed and set up the same in Positron too 👍

PR #459, opened 2017-12-22
@damassi => Fixes IE for BMW

kanaabe:
Sorry, I opened the wrong branch. This is the one that fixes stuff for IE.


kanaabe:
Yeah maybe I should turn off and redo these. I'm finding it difficult to read too. 1 sec

damassi:
But for the record, one more time straight from the Source: 

screen shot 2017-12-22 at 1 26 13 pm

If we're going to be forced to use Prettier I'm going to go to my grave fighting for and complaining about 120 print width. It blows code out, make things unreadable, and isn't idiomatic.

PR #463, opened 2017-12-26
[WIP] Update to relay 1.4.1

alloy:
This uses my [updated Relay custom Node ID patches](https://github.com/facebook/relay/compare/master...alloy:custom-node-id-redone), which brings in the latest Relay. This seems to work well, but I’d like someone else to take this for a good spin.


alloy:
Reaction now has latest Relay.

PR #467, opened 2018-01-04
@kanaabe => Image editing tweaks

eessex:
Some changes to images and artworks to allow better control in Positron--

- Use `editCaption` prop rather than children to render caption fields
- Allow children in both artwork and image (for remove buttons and potentially other things, renders without styles or container)
- Allow image captions to be empty
- Use linked prop in images to determine if `ViewFullscreen` button is displayed
- Adds unit tests for images and artworks

  This will affect https://github.com/artsy/reaction/pull/455/files, but I can update that PR once this is merged.

### DIFF
--- a/src/Components/Publishing/Sections/Image.tsx
+++ b/src/Components/Publishing/Sections/Image.tsx
@@ -6,8 +6,10 @@ import { Caption } from "./Caption"
 import { ImageWrapper } from "./ImageWrapper"
 
 interface ImageProps extends React.HTMLProps<HTMLDivElement> {
+  editCaption?: any
   image?: any
   layout?: Layout
+  linked?: boolean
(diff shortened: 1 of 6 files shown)


src/Components/Publishing/Sections/Image.tsx
```diff
@@ -38,17 +43,19 @@ export const Image: React.SFC<ImageProps> = props => {
       />
 
       <Caption
-        caption={image.caption}
+        caption={caption}
         layout={layout}
         sectionLayout={sectionLayout}
       >
-        {children}
+        {editCaption && editCaption()}
```
kanaabe: 👍

kanaabe:
LGTM 👍

PR #469, opened 2018-01-04
[WIP] Add animation transition on follow

yuki24:
This PR adds a `<ReplaceTrnsition>` component. finishes https://github.com/artsy/collector-experience/issues/843

**Jan 22 2018: updated the screenshot**

 * Right now, people are moving forward to the `react-transition-group` version 2, which seems to cover many issues. Ideally, we should be using this version.
 * In the master branch, there's already a [`<ReplaceTransition>` component](https://github.com/reactjs/react-transition-group/blob/7a64114/src/ReplaceTransition.js) in `react-transition-group`. However, it requires [exactly two elements](https://github.com/reactjs/react-transition-group/blob/7a641148849fc3aa0aa0acb673da9bc3cc35a7a0/src/ReplaceTransition.js#L10) shown and mounted simultaneously. Unfortunately, this is not the case in our list components such as `<PopularArtists>`. I also tried adding an empty `<div>` just to confirm the interface, but it just transitioned to an empty `<div>` (of course...) or animation wasn't applied at all.
 * There's also another library [`react-css-transition-replace`](https://github.com/marnusw/react-css-transition-replace). While this library works out of the box, it still depends on the old `react-transition-group` v1.

In other words, there's no ideal solution out there at this point. For now, I just sticked to the `react-css-transition-replace` so I could send this PR and show you what's in the branch. There's [another WIP branch](https://github.com/yuki24/reaction/tree/animation-without-dependencies) I'm working on for implementing the same animation but without dependencies. I'm hoping that we'll be able to:

 * Understand how the entire `<ReactCSSTransitionReplace>` component works by reading and converting raw javascript to TypeScript
 * Upgrade the `react-transition-group` to v2 without braking the interface (`<ReplaceTransition>`  takes a single component, fades in/out after `render()`)
 * Propose the entire idea and implementation back to the `react-transition-group`

However, this has been taking a bit longer than it should. I'd like to hear from you before looking into this too deeply. This PR is meant to be a conversation starter for how we are going to implement/maintain animation components rather than just merging immediately. Let me know what you think!

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "numeral": "^2.0.4",
     "prop-types": "^15.5.10",
     "qs": "^6.5.0",
+    "react-css-transition-replace": "^3.0.2",
     "react-lines-ellipsis": "^0.8.0",
     "react-markdown": "^2.5.0",
     "react-relay": "https://github.com/alloy/relay/releases/download/v1.3.0-artsy/react-relay-1.3.0-artsy.1.tgz",
--- a/src/Components/Animation/ReplaceTransition.tsx
+++ b/src/Components/Animation/ReplaceTransition.tsx
@@ -1,0 +1,31 @@
+import * as React from "react"
+import ReactCSSTransitionReplace from 'react-css-transition-replace'
+import { injectGlobal } from "styled-components"
+
+injectGlobal`
+  .fade-wait-leave {
+    opacity: 1;
+  }
+  .fade-wait-leave.fade-wait-leave-active {
+    opacity: 0;
+    transition: opacity .4s ease-in;
+  }
+
+  .fade-wait-enter {
+    opacity: 0;
+  }
+  .fade-wait-enter.fade-wait-enter-active {
+    opacity: 1;
+    /* Delay the enter animation until the leave completes */
+    transition: opacity .4s ease-in .6s;
+  }
+
+  .fade-wait-height {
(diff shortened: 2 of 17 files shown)


xtina-starr:
Unless anyone can point out a viable alternative for implementing animations, I'm okay with adding this dependency using v1 and revisiting this to support the functionality we want (`<ReactCSSTransitionReplace>`) in v2 as a separate issue.

@orta, @l2succes, @mzikherman, and @alloy do you all have thoughts?

alloy:
I agree with @xtina-starr to finish this PR first (and get a version into staging), after/while QA happens you can then dive deeper into the state of animations.

Btw, I noticed that this PR does not strictly implement the transition [that was designed](https://github.com/artsy/collector-experience/issues/843#issuecomment-352081723). Has there been offline discussions and is this intentional?

xtina-starr:
@alloy thanks for pointing that out. @Yuki we can take a look at that together if you'd like.

yuki24:
You are right the proposed animation looks a bit different. I think we may be able to use the `optimisticUpdater` to show the loading state and just let the current `<>` animate as it should. I should be able to update this PR tomorrow afternoon.

katarinabatina:
works for me @yuki24 !

yuki24:
Ok merging!

PR #478, opened 2018-01-11
[Dep] Upgrade TypeScript to 2.6.2 for React <Fragment /> support 

damassi:
Follow up after noticing https://github.com/prettier/prettier/pull/3331

### DIFF
--- a/src/Components/Input.tsx
+++ b/src/Components/Input.tsx
@@ -124,7 +124,7 @@ class Input extends React.Component<InputProps, InputState> {
       )
     }
 
-    return <StyledInput {...this.props} />
+    return <StyledInput {...this.props as any} />
   }
 }
 
--- a/src/Utils/Responsive.tsx
+++ b/src/Utils/Responsive.tsx
@@ -2,9 +2,14 @@ import { each } from 'lodash'
 import React from 'react'
 import theme from '../Assets/Theme'
 
-// TODO: IProps
 // TODO: Docs
 // TODO: Tests
+
+interface Props {
+  children?: any
+  media: any
+  mobileBreakpoint: any
+}
 
 interface State {
   isMobile: boolean
@@ -12,7 +17,7 @@ interface State {
 
 const MOBILE_BREAKPOINT = 600
 
-class ResponsiveWrapper extends React.Component<any, State> {
+class ResponsiveWrapper extends React.Component<Props, State> {
 
   static defaultProps = {
     initialState: {
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "tslint": "^4.5.0",
     "tslint-config-prettier": "^1.1.0",
     "tslint-react": "^2.4.0",
-    "typescript": "^2.3.4",
+    "typescript": "^2.6.2",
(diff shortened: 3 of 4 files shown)


src/Utils/Responsive.tsx
```diff
@@ -2,17 +2,22 @@ import { each } from 'lodash'
...
 
-// TODO: IProps
 // TODO: Docs
 // TODO: Tests
 
+interface Props {
+  children?: any
+  media: any
+  mobileBreakpoint: any
+}
```
damassi: This will be cleaned up in a different PR

kanaabe:
Oo nice, fragment support! Do you like <Fragment> or <> ?

damassi:
I prefer `<Fragment>` as its more explicit, and also, for now, lots of tools don't know how to recognize `<>`.

src/Components/Input.tsx
```diff
@@ -124,7 +124,7 @@ class Input extends React.Component<InputProps, InputState> {
       )
     }
 
-    return <StyledInput {...this.props} />
+    return <StyledInput {...this.props as any} />
```
alloy: Why is this needed?

damassi: I'm not too sure and now I feel bad for not investigating. One second...

damassi: Yeah this is a weird one thats non-obvious unfortunately:

```
src/Components/Input.tsx(127,25): error TS2322: Type '{ children?: ReactNode; error?: boolean; block?: boolean; leftView?: Element; rightView?: Element...' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<any, ComponentState>> & Readonly<{ child...'.
  Type '{ children?: ReactNode; error?: boolean; block?: boolean; leftView?: Element; rightView?: Element...' is not assignable to type 'IntrinsicClassAttributes<Component<any, ComponentState>>'.
    Types of property 'ref' are incompatible.
      Type 'Ref<HTMLInputElement>' is not assignable to type 'Ref<Component<any, ComponentState>>'.
        Type '(instance: HTMLInputElement) => any' is not assignable to type 'Ref<Component<any, ComponentState>>'.
          Type '(instance: HTMLInputElement) => any' is not assignable to type '(instance: Component<any, ComponentState>) => any'.
            Types of parameters 'instance' and 'instance' are incompatible.
              Type 'Component<any, ComponentState>' is not assignable to type 'HTMLInputElement'.
                Property 'accept' is missing in type 'Component<any, ComponentState>'.

127     return <StyledInput {...this.props} />
```

I'm not sure why its saying that `accept` is missing as `styled-components`'s interface says that its optional, and `StyledInput` is just an instance of `styled.input`:

```ts
interface InputHTMLAttributes<T> extends HTMLAttributes<T> {
  accept?: string;
  ...
}
```

alloy: Interesting! 🤔

Feel free to merge now if this is blocking you, otherwise I’ll take a look too see if I can understand it more when I get back home.

damassi: No rush, but am definitely curious to know what's up; this only started appearing with the upgrade and _seems_ to be valid

PR #479, opened 2018-01-11
Create Relay-based <RailSlider /> component 

damassi:
**First Draft. This is being put on hold due to Relay-SSR issues that need to be solved in Force as well as other work related to Eigen 4.0.1**

While starting a new "Buy Now" rail feature for [this PR](https://github.com/artsy/force/pull/2072) I figured it would be a good time to build a rail slider on top of Relay that would replace existing backbone-based Homepage rails and make it flexible enough to accept different kinds of blocks. We can then share that rail in various places where need be. 

fillwidth


.vscode/settings.json
```diff
@@ -12,5 +12,9 @@
   },
   "editor.tabSize": 2,
   "editor.formatOnSave": true,
+  "[json]": {
+    "editor.formatOnSave": false
```
damassi: Dont let prettier et al rewrite our configs

package.json
```diff
@@ -21,26 +21,26 @@
...
+    "prettier-write": "yarn run prettier --write",
+    "relay": "env GRAPHQL_NO_NAME_WARNING=true relay-compiler --src ./src --schema data/schema.graphql --extensions ts tsx",
+    "start": "yarn storybook",
+    "storybook": "node verify-node-version.js && concurrently --kill-others 'yarn relay --watch' 'env GRAPHQL_NO_NAME_WARNING=true start-storybook -p 9001'",
+    "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
+    "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && yarn install && npm run dump-schema -- ../../data",
+    "sync-schema:localhost": "cd ../metaphysics && yarn dump-schema -- ../reaction/data",
+    "test": "node verify-node-version.js && jest",
+    "test:watch": "jest --watch --runInBand",
+    "type-check": "tsc --noEmit --pretty",
```
damassi: alpha

.vscode/settings.json
alloy: 👍

src/Components/RailSlider/Arrow.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const SliderArrow = styled.div`
+  display: flex;
+  align-items: center;
+  position: absolute;
+  height: 100vh;
+  top: 0;
+  box-sizing: border-box;
+  ${(props: Props) => (props.direction === "left" ? "left: 0px;" : "")}
+  ${(props: Props) => (props.direction === "right" ? "right: 0px;" : "")}
```
alloy: If you search Reaction for StyledFunction, you’ll find a way to write these components without needing to type props for each function.

src/Components/RailSlider/Slider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  slidesToShow: 1,
+}
+
+interface Props {
+  artist: string,
+  initialSlide?: number
+}
+
+export const Slider = createFragmentContainer(
+  (props: Props) => {
```
alloy: Hah, nice, never really considered inlining. Makes sense for such a short component 👍

src/Components/RailSlider/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  )
+})
+
+export const RailSlider: SFC = () => {
+  return (
+    <ContextProvider>
+      <SliderContainer artistID='andy-warhol' />
+    </ContextProvider>
+  )
+}
```
alloy: At some point the query renderer should probably move outside of this component, so that there can be a single query renderer for the home page with all components composed onto that.

src/Components/Input.tsx
```diff
@@ -124,7 +124,7 @@ class Input extends React.Component<InputProps, InputState> {
       )
     }
 
-    return <StyledInput {...this.props} />
+    return <StyledInput {...this.props as any} />
```
alloy: This change was also part of your TS bump PR, was that maybe unintentional?

src/Components/RailSlider/Arrow.tsx
damassi: I was wondering about this, and the approach above is deliberate. Is there any benefit in going the other way? IMO `props: Props` is much less noisy and more succinct then using `StyledFunction` as it takes away two lines of code and the need for an intermediate variable.

src/Components/RailSlider/index.tsx
damassi: I see, like wrap _all_ of the rails in a container. For sure -- once this is built out and things are working I'll be sure to loop back.

src/Components/Input.tsx
damassi: Yup, I rebased on top for the time being and have a note at the top of the PR.

src/Components/RailSlider/Arrow.tsx
alloy: The benefit is that the call sites where this component is used gets code completion and is checked.

Some people have made more succinct suggestions here since mine, maybe one of these can work for us? https://github.com/styled-components/styled-components/issues/630

src/Components/Input.tsx
alloy: Ah kk

src/Components/RailSlider/index.tsx
alloy: Aye. In general that should be the default pattern (make all components composable inside other Relay trees).

Although in the case of the home page rails we may not end up doing that :) Remember that faffing with the rails in Emission? That’s to defer queries, we should see if replacing that with separate query renderers is feasible and simpler.

src/Components/RailSlider/Arrow.tsx
damassi: Everything still seems to typecheck fine? 

screen shot 2018-01-10 at 11 27 32 pm

damassi: But yeah, looking through the responses yours still seems to be the clearest out of the bunch.

alloy: Oh that’s interesting! So either styled-components or TypeScript have made improvements.

We should revisit it then. Maybe something like this works as well nowadays which means you only need to type once?

```ts
const SliderArrow = styled.div<Props>`
  ...
`
```

yordis: @alloy did you figure SC with TS out?

damassi: > Maybe something like this works as well nowadays which means you only need to type once?

Unfortunately not, but maybe this makes sense to you? 

screen shot 2018-01-12 at 11 30 03 am

src/Components/Auctions/RailSlider/Slider.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Title style={{ marginLeft: "10%" }} titleSize="large">
+          Lots by Followed Artists
+        </Title>
+
+        <Slick {...SETTINGS}>
+          {props.sale.artworks.map((artwork, key) => {
+            const { image: { aspect_ratio } } = artwork
+
+            return (
+              <div key={key}>
```
damassi: An outer div is needed for `react-slick`

src/Components/Auctions/RailSlider/Slider.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </Slick>
+      </Container>
+    )
+  },
+  graphql`
+    fragment Slider_sale on Sale {
+      artworks {
+        image {
+          aspect_ratio
+          height
```
alloy: Is `height` unused?

src/Utils/metaphysics.ts
```diff
@@ -1,3 +1,4 @@
+import "isomorphic-fetch"
```
damassi: This was the first issue that popped up when starting down the SSR path but was an easy fix

damassi:
Note for later: When attempting to mount in Force, this error is being thrown: 

screen shot 2018-01-16 at 9 27 38 pm

It seems as though there is a race condition somewhere in`styled-components`, which is applying `ref='component'`, before an `instance._owner` is set. This is mediated in various ways by wrapping styles like: 

```javascript
const StyledFillwidthItem = styled.div`

... 

<StyledFillwidthItem>
 ...
</StyledFillwidthItem>
```

instead of 

```javascript 
const StyledFillwidthItem = styled(FillwidthItem)`
```

The problem however is still inconclusive.

damassi:
> [WIP], are those supposed to be squashed?

Yup! thanks for the reminder; squashing now...

src/Components/Auctions/RailSlider/Arrow.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const SliderArrow = styled.div`
+  display: flex;
+  align-items: center;
+  position: absolute;
+  height: 50vh;
+  top: 0;
+  box-sizing: border-box;
+  ${(props: Props) => (props.direction === "left" ? "left: -120px;" : "")};
+  ${(props: Props) => (props.direction === "right" ? "right: -120px;" : "")};
```
damassi: (Note to self: need to fix these hacky negative margins.)

src/Components/Auctions/RailSlider/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      }}
+    />
+  )
+})
+
+export const RailSlider: SFC = () => {
+  return (
+    <ContextProvider>
+      <SliderContainer saleID="phillips-evening-and-day-editions-3" />
+    </ContextProvider>
```
alloy: For follow-up work: A reason that this can’t really be inside this component is that the host (Force) is supposed to configure this context with e.g. the current user.

src/Components/Auctions/RailSlider/Arrow.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { SFC } from "react"
+import styled from "styled-components"
+import Icon from "../../Icon"
+
+interface Props {
+  direction: "left" | "right"
+  onClick?: () => void
+}
+
+export const Arrow: SFC<Props> = ({ direction, onClick }) => {
```
tal: Out of curiosity

a) why dont you have these things as default exports?
b) why do the pattern of `const Name = () => {}` instead of `function Name() {}`

alloy: Good questions!

a) We’ve been moving away from default exports for various reasons, this page summarises all of them https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html
b) We use it to be able to cast the function appropriately, afaik if you use the `function` keyword you can only type the individual parameters and return value

tal: Whats the reason for not doing `function Arrow({direction, onClick}: Props) {}`?

alloy: Oh yeah, I hadn’t considered that to be good enough, so no reason. What type of tooling are you referring to?

tal: Mostly symbol identification in editors (for jump to symbol) and stuff like git lens.

alloy: Nice, didn’t know about git lens!

@damassi Do you have any thoughts on function definition style?

damassi: In terms of information that is available to the compiler, I think function literals are less precise:

screen shot 2018-01-24 at 12 40 39 pm

Compared to 

screen shot 2018-01-24 at 12 44 15 pm

(JSX.Element vs React.StatelessComponent)

That said, I'm personally in the function literal camp as I find it more pleasing to read and less noisy, as well as providing sufficient safety and intellisense assuming the argument is typed. I also like that function literals are hoisted to the top before execution so that a function declared at the bottom of the file can be used in another function located in the middle without erring out; `const Foo = () =>`-style declarations have to be used in the order in which they are declared and do not have free access to the entire scope.

tal: Heh some people would say the lack of hoisting is a feature. It is worth saying that `SFC<P>` provides more type safety. if you look at the definition there's much more to it. The question is how these functions are passed around, usually they're used 100% w/in JSX so it doesn't actually matter too much.

This is the ultimate bike shed. My personal opinion is let anyone do it either way, whatever they individually like more.

damassi: > This is the ultimate bike shed

😄

orta: FWIW: In react native, until _very_ recently a function literal `const component = () => <View/>` would not get hot module reloaded. So that's been avoided (or [hijacked](https://github.com/artsy/emission/blob/e106feabfbf4663023e0c7010177527841884b23/src/lib/Components/Consignments/Components/ArtworkConsignmentTodo.tsx#L124-L126)) from the mobile side.

PR #486, opened 2018-01-17
@eessex => Artwork credits + restyle artwork captions

kanaabe:
Adds support for credits in artworks and restyles captions a bit. Addresses https://github.com/artsy/publishing/issues/306

image
image

### DIFF
--- a/src/Components/Publishing/Fixtures/Components.ts
+++ b/src/Components/Publishing/Fixtures/Components.ts
@@ -16,6 +16,7 @@ export const Artworks = [
     },
     width: 1152,
     height: 826,
+    credit: "Courtesy of Gary Nader"
   },
   {
     type: "artwork",
@@ -40,6 +41,7 @@ export const Artworks = [
     ],
     width: 1152,
     height: 826,
+    credit: "Courtesy of Gary Nader"
   },
   {
     type: "artwork",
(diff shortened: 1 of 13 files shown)


src/Components/Publishing/Sections/ArtworkCaption.tsx
```diff
@@ -189,27 +212,34 @@ export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
...
-              {this.renderArtists()}
-            </span>
+          <ArtistName>
+            {this.renderArtists()}
+          </ArtistName>
 
-            {this.renderTitleDatePartner()}
-          </Truncator>
+          <div>
+            <Truncator>
```
eessex: Seems like Truncator could be broken out into its own component.

kanaabe: Mmm I think it is?

kanaabe:
Realized we don't have any unit tests for `ArtworkCaption` so I'd like to include that in this PR as well --

src/Components/Publishing/Sections/ArtworkCaption.tsx
eessex: Own file rather

src/Components/Publishing/Fixtures/Components.ts
```diff
@@ -5,7 +5,7 @@ export const Artworks = [
     type: "artwork",
     id: "589a6291275b2410d1beb6a5",
     slug: "fernando-botero-nude-on-the-beach",
-    date: "2000",
+    date: "",
     title: "Nude on the Beach",
```
eessex: Curious why you wanted to remove the date here?

kanaabe: The fixture is meant to represent an artwork with missing info

PR #489, opened 2018-01-18
@kanaabe => Prefix draft selector

eessex:
One more small change to the classes for Draft -- prefix selector with `div[data-block=true]` so it does not apply to div's inside an `h2` or other kind of block.


kanaabe:
Kinda makes me think though, do we want Draft-level styles in Reaction when there's no Draft in Reaction? Seems like it can be hard to track down. I can see how colocating the styles is nice though. Wonder what's most ideal here.

eessex:
Yeah it is not my favorite solution -- which is why I went the other direction first with overwriting the block type.  Taking these styles out of reaction would mean needing to duplicate the styles entirely in Positron though, so while ugly this still seems more efficient.

eessex:
Another option would be going back to the first solution and manually overriding the error messages... also hacky

PR #491, opened 2018-01-19
[Prettier] Update column setting to default

damassi:
This updates Prettier column settings to the default (80) per the [RFC outcome](https://github.com/artsy/potential/issues/127#issuecomment-358997987). In favor of an incremental approach (while acknowledging existing PRs in the wild), it does _not_ reformat the codebase. Existing code will remain the same until touched at which point our git precommit hook (or VSCode editor settings) will rewrite.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,7 @@
     "**/node_modules": true,
     "dist/": true
   },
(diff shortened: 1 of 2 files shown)


orta:
👍

PR #493, opened 2018-01-19
@yuki24 => [Onboarding] Throttled search

kanaabe:
Addresses https://github.com/artsy/collector-experience/issues/847 by decoupling text input from query + throttles the query state change.

Noticed there weren't any tests in this project yet so I was going to add one but got a little stuck on an error with `ContextProvider` and mocking the Relay side. I'll follow up with tests but didn't want to hold back on the PR.

This uses 500 millisecond throttle which was a little nicer than 1 second which felt pretty slow.
throttle-search

### DIFF
--- a/src/Components/Onboarding/Steps/Artists/index.tsx
+++ b/src/Components/Onboarding/Steps/Artists/index.tsx
@@ -1,3 +1,4 @@
+import { throttle } from "lodash"
 import React from "react"
 import styled from "styled-components"
 
@@ -23,11 +24,13 @@ const OnboardingSearchBox = styled.div`
 interface State {
   inputText: string
   followCount: number
+  inputTextQuery: string
 }
 
 export default class Artists extends React.Component<StepProps, State> {
   state = {
     inputText: "",
+    inputTextQuery: "",
     followCount: 0,
   }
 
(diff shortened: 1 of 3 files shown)


src/Components/Onboarding/Steps/Artists/index.tsx
```diff
@@ -22,20 +23,36 @@ const OnboardingSearchBox = styled.div`
...
+    inputTextQuery: "",
   }
 
-  // onInputChange = e => {
-  //   this.props.onStateChange({ nextButtonEnabled: true })
-  // }
+  componentDidMount() {
+    this.throttledTextChange = throttle(this.throttledTextChange, 500, {
+      leading: true,
+    })
```
yuki24: This is a bit of a brain teaser. We could possibly use the [`lodash-decorators`](https://www.npmjs.com/package/lodash-decorators) to make it more declarative: 

```ts
@Bind()
@Debounce(500, { leading: true }) 
throttledTextChange(inputText) {
  this.setState({ inputTextQuery: inputText })
}
```

kanaabe: Ooh, this is really cool. I don't have a preference. Should I go for this?

PR #494, opened 2018-01-23
@kanaabe => [Publishing] Fix missing reduce start value

damassi:
Just noticed a weird issue in this Webpack PR where it was erroring out due to a missing start value in a `reduce` fn. I double checked the rest of the projects so ignore other changes as Prettier. Odd that this never came up before.

### DIFF
--- a/src/Components/Publishing/Sections/ArtworkCaption.tsx
+++ b/src/Components/Publishing/Sections/ArtworkCaption.tsx
@@ -22,22 +22,15 @@ interface StyledArtworkCaptionProps {
 }
 
 export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
-  joinParts(children, delimiter = ', ') {
-    const joined = _.compact(children)
-      .reduce((prev, curr) => {
-        return [
-          prev,
-          delimiter,
-          curr
-        ]
-      })
+  joinParts(children, delimiter = ", ") {
+    const joined = _.compact(children).reduce((prev, curr) => {
+      return [prev, delimiter, curr]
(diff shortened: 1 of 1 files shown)


src/Components/Publishing/Sections/ArtworkCaption.tsx
```diff
@@ -22,22 +22,15 @@ interface StyledArtworkCaptionProps {
...
-        return [
-          prev,
-          delimiter,
-          curr
-        ]
-      })
+  joinParts(children, delimiter = ", ") {
+    const joined = _.compact(children).reduce((prev, curr) => {
+      return [prev, delimiter, curr]
+    }, [])
```
damassi: Was missing `[]`.

damassi: Actually, this is a bit stranger now that I'm looking at it closer. For some reason infering a start value leads to correct test results, but compacting an end value _with_ a start value errors out. Looking...

damassi: Interesting: https://hackernoon.com/reduce-faster-with-no-starting-value-3d697c9464f9

damassi: And `arr.reduce(callback[, initialValue])` is the signature -- [no starting value required](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

damassi:
Going to close this since @kanaabe is aware and its on the list

PR #495, opened 2018-01-24
@kanaabe => Minor ui tweaks

xtina-starr:
This fixes the minor CSS items we saw related to the `line-height` in the titles as well the resizing that was happening to the images. In addition, it also supports a `redirectTo` prop to be passed Wizard component and down to the Budget component so that we can support redirecting the user to their entry point before account creation/onboarding after onboarding steps are completed.

### DIFF
--- a/src/Components/Onboarding/ItemLink.tsx
+++ b/src/Components/Onboarding/ItemLink.tsx
@@ -3,7 +3,7 @@ import styled from "styled-components"
 
 import Colors from "../../Assets/Colors"
 import * as fonts from "../../Assets/Fonts"
-import CircleIcon from "../CircleIcon";
+import CircleIcon from "../CircleIcon"
 import Icon from "../Icon"
 
 const Link = styled.a`
@@ -19,7 +19,7 @@ const Link = styled.a`
 `
 
 const Avatar = styled.img`
-  padding: 10px 15px 10px 10px;
+  margin: 10px 15px 10px 10px;
 `
 
 const FullWidthCol = styled.div`
@@ -77,7 +77,11 @@ export default class ItemLink extends React.Component<Props, State> {
         <Col>
           {
             <Avatar
-              src={this.props.image_url ? this.props.image_url : "https://www.artsy.net/images/icon-70.png"}
+              src={
+                this.props.image_url
+                  ? this.props.image_url
+                  : "https://www.artsy.net/images/icon-70.png"
+              }
               width={50}
               height={50}
             />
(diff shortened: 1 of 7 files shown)


src/Components/Onboarding/ItemLink.tsx
```diff
@@ -19,7 +19,7 @@ const Link = styled.a`
 `
 
 const Avatar = styled.img`
-  padding: 10px 15px 10px 10px;
+  margin: 10px 15px 10px 10px;
```
xtina-starr: We want to use `margin` instead of `padding` to prevent the images from being distorted in force.

src/Components/Onboarding/Steps/Layout.tsx
```diff
@@ -23,6 +23,7 @@ const Container = styled.div`
 
 const MainTitle = styled(StyledTitle) `
   text-align: center;
+  line-height: normal;
```
xtina-starr: This addresses the css reset that happens in force.

src/Components/Onboarding/ItemLink.tsx
```diff
@@ -77,21 +77,34 @@ export default class ItemLink extends React.Component<Props, State> {
...
+              color="black"
+              fontSize="39px"
+            />
+          ) : (
+            <CircleIconContainer>
+              <CircleIcon
+                name="close"
+                color="black"
+                fontSize="21px"
+                style={{ transform: "rotate(45deg)" }}
```
kanaabe: It looks like this was just a change from Prettier but I'll note that we could style this with `styled-components` since we're using them heavily in this project. It could look something like -- 
```
const StyledCircleIcon = styled(CircleIcon)`
  transform: rotate(45deg);
`
```
provided that `CircleIcon` passes in `className={this.props.className}` into the top level div. Not a blocker!

xtina-starr: Yeah I like that. I'm pretty sure @katarinabatina will have some QA items related to these icons so we could probably bundle that refactor in with those items.

src/Components/Onboarding/Wizard.tsx
```diff
@@ -28,7 +29,12 @@ class Wizard extends React.Component<Props, State> {
     }
 
     const CurrentStep = this.props.stepComponents[currentStep]
-    return <CurrentStep onNextButtonPressed={this.onNextButtonPressed.bind(this)} />
+    return (
+      <CurrentStep
+        onNextButtonPressed={this.onNextButtonPressed.bind(this)}
+        redirectTo={this.props.redirectTo}
+      />
```
kanaabe: Cool -- not sure if this will work, but is it possible to avoid passing in `redirectTo` to each step and instead redirect from the Wizard [here](https://github.com/artsy/reaction/pull/495/files#diff-f494b90c7872d093afb16d77f131ae8fR28)? It seems like `getCurrentStep` can tell if we're at the last step. I'm sure his works too but I'm curious and it might allow us to not pass down `redirectTo` to every step.

xtina-starr: Oh yeah I think we talked about that implementation when we were first looking at this but I forgot about it. I'll give it a try.

xtina-starr:
@kanaabe I added the suggested refactor. Is this good to merge?

kanaabe:
Yep I merged it yesterday 👍

xtina-starr:
haha I didn't refresh 🙈

PR #496, opened 2018-01-24
@damassi => Zero state artwork captions

kanaabe:
Fixes the zero-state on artwork captions, ie captions with no title, date, partner, etc. There's a lot of noise with Prettier changes/fixture updates but [here](https://github.com/artsy/reaction/compare/master...kanaabe:zero-state-artwork?expand=1#diff-109086d38202a7acf7b1dc711fd98a4eR25) is the main bit.

Artwork fixtures are also named now, so we can reference them by what the data looks like instead of arbitrarily with `Artworks[1]`.

### DIFF
--- a/src/Components/Publishing/Sections/ArtworkCaption.tsx
+++ b/src/Components/Publishing/Sections/ArtworkCaption.tsx
@@ -22,22 +22,18 @@ interface StyledArtworkCaptionProps {
 }
 
 export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
-  joinParts(children, delimiter = ', ') {
-    const joined = _.compact(children)
-      .reduce((prev, curr) => {
-        return [
-          prev,
-          delimiter,
-          curr
-        ]
-      })
+  joinParts(children, delimiter = ", ") {
+    const joined = _.compact(children).reduce((prev, curr) => {
+      console.log(prev)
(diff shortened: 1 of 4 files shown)


src/Components/Publishing/Sections/ArtworkCaption.tsx
```diff
@@ -254,10 +220,13 @@ const ArtistName = styled.span`
...
     }
-  `}
+  `};
 `
 
-const div: StyledFunction<StyledArtworkCaptionProps & React.HTMLProps<HTMLDivElement>> = styled.div
+const div: StyledFunction<
+  StyledArtworkCaptionProps & React.HTMLProps<HTMLDivElement>
+> =
+  styled.div
```
damassi: @kanaabe - with latest TypeScript version this form of typing passed props to styled-components can be way simplified: 

```typescript
interface Props {
  sectionLayout: string
}

const ArtworkCaption = styled.div`
  padding: ${(props: Props) => (props.sectionLayout === "fillwidth" ? "0 10px;" : "0;")}
`
```

kanaabe: Yea, I'm so excited for this! Gonna merge tho since this diff is from Prettier and not a code change. Thanks for bringing it up!

PR #499, opened 2018-01-25
[WIP] Onboarding snapshot tests

kanaabe:
TODO
- [ ] Mock metaphysics for more accurate snapshots

This is a big diff but it's mostly just generated code from the snapshots. I didn't add a snapshot test for Wizard because it doesn't contain any styles. Wizard will benefit from having lots of unit tests though.

### DIFF
--- a/src/Components/Onboarding/Steps/Artists/__tests__/__snapshots__/index.test.tsx.snap
+++ b/src/Components/Onboarding/Steps/Artists/__tests__/__snapshots__/index.test.tsx.snap
@@ -1,0 +1,275 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Artists snapshot renders the artist step 1`] = `
+.c6 {
+  font-family: artsy-icons;
+  color: #666666;
+  font-size: 24px;
+  margin: 0 5px;
+  display: inline-block;
+  -webkit-letter-spacing: 0px;
+  -moz-letter-spacing: 0px;
+  -ms-letter-spacing: 0px;
+  letter-spacing: 0px;
+}
+
+.c7 {
+  border: 0;
+  font-family: 'Adobe Garamond W08', 'adobe-garamond-pro', 'AGaramondPro-Regular', 'Times New Roman', 'Times', 'serif';
+  -webkit-font-smoothing: antialiased;
+  font-size: 17px;
+  outline: none;
+  -webkit-flex: 1;
+  -ms-flex: 1;
+  flex: 1;
+}
+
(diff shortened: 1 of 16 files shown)


alloy:
> Mock metaphysics for more accurate snapshots

@kanaabe Ah yes, I’ve been thinking long about this, it would be beneficial to actually use Relay to ensure that all queries are correct. I settled on the following:

* Unit tests benefit from data being stubbed local to the test, so a local stubbed copy of the MP schema that Relay queries against is probably the way to go. I was planning on wrapping https://github.com/relay-tools/relay-local-schema, but I just found that somebody else already did that: https://github.com/1stdibs/relay-mock-network-layer
* For storybooks we need more complete data and stubbing that becomes tedious as you’re adding/removing fields, so for this purpose I started writing some tooling that would use a ‘VCR’ library to record _actual_ network requests made by Relay to MP for later replay.

What do you think?

kanaabe:
@alloy interesting, coming from significantly less Relay experience, I was thinking of using https://github.com/jefflau/jest-fetch-mock to mock the metaphysics level fetch. But this seems like we would have slightly more control. Would snapshot tests use the network mock too?

alloy:
During a retro about snapshots/unit tests/storybooks we came to the conclusion that we’d want:

1. Storybooks for components that show their full behaviour
2. Use [storyshots](https://github.com/storybooks/storybook/tree/master/addons/storyshots) to create snapshots of those ^ stories
3. Plain unit tests for logic inside components, no snapshots whatsoever

Recording requests in a browser (storybooks) is hard-ish, which is what 2 could also nicely solve. Namely, when running the storyshots we allow it to record the requests, and on a static deploy of storybooks it can use those recorded requests.

orta:
Or even better, something like: https://github.com/oblador/loki so we get visual regressions rather than text

kanaabe:
@alloy Ah, so this PR contradicts what was agreed on before. Sorry I didn't know about that and was looking at https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=AR&modal=detail&selectedIssue=AR-43.

I'll close this for now then and try out storyshots.

alloy:
@kanaabe Sorry about that. I don’t think that retro outcome ever got codified :( Thanks for taking that on!

alloy:
@orta I definitely think we should be looking at such tooling in the future, but it doesn’t fix the specific issue that I mentioned is fixed by having storyshots and being able to record requests there.

orta:
Agree ^

PR #500, opened 2018-01-30
[WIP] Storyshots

kanaabe:
cc @alloy 

TODOs for things we broken on purpose in pursuit of a working storyshot setup:
- [ ] Removed isomorphic-fetch from global __mock__
- [ ] Removed react-relay from global __mock__
- [ ] Change import path of Event module
- [ ] Hardcoding the metaphysics url - need to setup sharify for tests

TODOs for our storyshot setup:
- [ ] API:
  - [ ] Let story specify a sentinel selector to look for to exist.
  - [ ] Let story specify a sentinel selector to look for to NOT exist.
  - [ ] Maybe make it possible to have a default sentinel selector, e.g.  look for when all ‘spinner’ components are removed.
- [ ] Need to call tree.unmount() after run, but consider case where test times
  out as well, so this should really run in `afterEach`. However, the
  current option to specify before/after filters through storyshots is that
  they are global, so no good way to scope e.g. `tree` to specific tests.
- [ ] I saw some ‘snapshot filename’ support in storyshot, so we should respect
  that and call `toMatchSnapshot` with the explicit filename.


src/__tests__/Storyshots.test.tsx
```diff
@@ -0,0 +1,10 @@
+import initStoryshots from "@storybook/addon-storyshots"
+
+initStoryshots({
+  storyNameRegex: /Gene Follow/,
+  test: ({ story, context }) => {
+    // console.log(story)
+    // console.log(context)
+    // const storyElement = story.render(context)
```
alloy: @kanaabe Is passing in `context` as the argument to `story.render(…)` the documented/required thing to do? If so, can you link me to it?

kanaabe: Nope, `context` is not necessary! It was using in the docs here so I had it but it works without: https://github.com/storybooks/storybook/tree/master/addons/storyshots#example

alloy:
Leaving this example of what an enzyme API could like here:

```js
function showing(selector) {
  return tree => tree.find(selector).length > 0
}

function noLongerShowing(selector) {
  return tree => tree.find(selector).length === 0
}

mount(<MyApp />)
  .when(showing('SuggestedGeneContent'))
  .then(tree => expect(tree).toMatchSnapshot())

mount(<MyApp />)
  .when(noLongerShowing('Spinner'))
  .then(tree => expect(tree).toMatchSnapshot())
```

orta:
dang - 80k LOC

alloy:
Yup. Might make more sense to snapshot the ‘debug’ output: http://airbnb.io/enzyme/docs/api/ReactWrapper/debug.html. I’ll push a copy of that instead.

PR #501, opened 2018-01-31
[WIP] Storyshots

kanaabe:
On artsy/reaction!

cc @alloy 

TODOs for things we broken on purpose in pursuit of a working storyshot setup:
- [ ] Removed isomorphic-fetch from global __mock__
- [ ] Removed react-relay from global __mock__
- [ ] Change import path of Event module
- [ ] Hardcoding the metaphysics url - need to setup sharify for tests

TODOs for our storyshot setup:
- [ ] API:
  - [ ] Let story specify a sentinel selector to look for to exist.
  - [ ] Let story specify a sentinel selector to look for to NOT exist.
  - [ ] Maybe make it possible to have a default sentinel selector, e.g.  look for when all ‘spinner’ components are removed.
- [ ] Need to call tree.unmount() after run, but consider case where test times
  out as well, so this should really run in `afterEach`. However, the
  current option to specify before/after filters through storyshots is that
  they are global, so no good way to scope e.g. `tree` to specific tests.
- [ ] I saw some ‘snapshot filename’ support in storyshot, so we should respect
  that and call `toMatchSnapshot` with the explicit filename.


alloy:
Leaving this example of what an enzyme API could like here:

```js
function showing(selector) {
  return tree => tree.find(selector).length > 0
}

function noLongerShowing(selector) {
  return tree => tree.find(selector).length === 0
}

mount(<MyApp />)
  .when(showing('SuggestedGeneContent'))
  .then(tree => expect(tree).toMatchSnapshot())

mount(<MyApp />)
  .when(noLongerShowing('Spinner'))
  .then(tree => expect(tree).toMatchSnapshot())
```

alloy:
@orta Using the [debug](http://airbnb.io/enzyme/docs/api/ReactWrapper/debug.html) output leads to 3800 LOC, rather than 80k: https://github.com/artsy/reaction/blob/724be7ccfa64f3cb5ec554e31ea595dce2af0d31/src/__tests__/__snapshots__/Storyshots.test.tsx.snap

damassi:
@kanaabe - is the idea that StoryShots is a way to automatically snapshot storybook output rather than writing snapshots at the component level?

kanaabe:
@damassi Yepp, we can eventually delete the manually written snapshot test code + generated snapshots as long as we have good stories. Storyshots would work pretty simply with the fixturized Publishing components but with Relay-based ones there's a little more work. Check out [this comment from @alloy](https://github.com/artsy/reaction/pull/499#issuecomment-360741761) which outlines his approach.

damassi:
Very cool! Wasn't aware of this util.

kanaabe:
Closing this since it hasn't been prioritized in a while.

PR #502, opened 2018-02-01
@damassi => Fixes <FeatureHeader> bug + Upgrade SC to 3.1.5, Jest to 22.1.4

kanaabe:
Force is on styled-components 2.4.0 which introduced a [change](https://www.styled-components.com/releases#v2.4.0) to the hash generating algorithm. A side effect of this change is that components that use the same base component with `extend` in the same file will have the same class name. More detail on that in [this issue](https://github.com/styled-components/styled-components/issues/1402) and [this one](https://github.com/styled-components/babel-plugin-styled-components/issues/79).

This PR upgrades SC to at least match Force + uses the `styled(Component)` wrapper instead of the problematic `extend` in the `FeatureHeader`. The rest is a lot of noise with dead code/prettier/snapshots. 

Happy "split" layout in the FeatureHeader component w/styled-components@2.4.0:
image

---------

After faffing for a long time with mismatching snapshot tests on local vs CI, I decided to upgrade Jest to 22.1.4 and it seemed to have fixed most problems. This PR now upgrades the following:
```
"babel-plugin-styled-components": "^1.4.0"
"jest": "^22.1.4",
"jest-styled-components": "^4.9.0",
"jest-styled-components": "^4.11.0-1",
"styled-components": "^3.1.5"
```
Tests and storybooks seem to be happy now. 😴


src/Components/Publishing/Header/FeatureHeader.tsx
```diff
@@ -178,20 +215,22 @@ const Div = styled.div`
...
-  ${Fonts.unica("s16", "medium")}
   margin-bottom: 10px;
+  ${Fonts.unica("s16", "medium")};
   ${pMedia.sm`
     ${Fonts.unica("s14", "medium")}
-  `}
+  `};
 `
-const HeaderTextContainer = Div.extend`
+const HeaderTextContainer = styled(Div)`
```
kanaabe: This line fixes the layout. `Div.extend` -> `styled(Div)`

alloy:
Re-assigning to @damassi because I’m out sick.

kanaabe:
Something is definitely up with this version of styled-components + consistent snapshots. Still looking at this!

kanaabe:
This should be good to go now @damassi! Note there are several upgrades in this PR.

package.json
```diff
@@ -91,7 +91,7 @@
     "react-dom": "^16.2.0",
     "react-test-renderer": "^16.0.0",
     "relay-compiler": "https://github.com/alloy/relay/releases/download/v1.3.0-artsy/relay-compiler-1.3.0-artsy.1.tgz",
-    "styled-components": "^2.2.3",
+    "styled-components": "^3.1.5",
```
damassi: This looks great and excited about latest styles-components features :) One issue: Could you copy this line to 'peerDependencies' and then open up a corresponding 3.1.5 pr update in force?

src/Components/Publishing/Display/DisplayPanel.tsx
```diff
@@ -399,71 +417,88 @@ const Image = styled.div`
...
+
   ${VideoCover} {
-    background: url(${p => p.coverUrl || ""}) no-repeat center center;
+    background: url(${(p: DivUrlProps) => p.coverUrl || ""}) no-repeat center
+      center;
     background-size: cover;
-    ${p =>
+  }
+
+  ${(p: DivUrlProps) =>
```
damassi: Yayy

src/Components/Publishing/Header/FeatureHeader.tsx
```diff
@@ -350,388 +387,8 @@ const FeatureHeaderContainer = DivWithProps`
...
         padding: 20px;
       }
-    `}
+    `};
   }
 `
 
 export const FeatureHeader = FeatureHeaderComponent
-
-// FIXME: Reeimplment below refactor once the heat is off....
```
damassi: I think the below can probably go since whatever should likely conform to a grid layout

src/Components/Publishing/Header/__test__/__snapshots__/FeatureHeader.test.tsx.snap
```diff
@@ -732,36 +714,36 @@ exports[`feature renders feature full header without a deck properly 1`] = `
   height: auto;
 }
 
-.c0[data-type="text"] .c8 {
+.c0[data-type="text"] .c9 {
```
damassi: I just realized that snapshots would be far more useful if, like babel-plugin-styled-components, it printed out the component name. I wonder if that could be flipped on in tests?

kanaabe: Looks like that was considered at one point: https://github.com/styled-components/jest-styled-components/issues/95

src/Components/Publishing/Header/FeatureHeader.tsx
kanaabe: @damassi what do you mean by this? Go as in be used?

damassi: Oh haha, no go meaning delete!

PR #503, opened 2018-02-01
[Toolchain] Add --allow-restart to concurrently

damassi:
When an error currently occurs (such as when moving a watched file) concurrently kills all running processes forcing dev to return to command-line and reboot. On closer look, however, there is the option `--allow-restart`, which will restart the dead process. 

conccurrently


damassi:
Actually, scratch that -- this configuration makes it impossible to send `ctrl+c` to kill the server; things will just boot back up.

PR #506, opened 2018-02-02
[Onboarding] Fix placeholder for category search

yuki24:
This fixes the placeholder for the category search input thanks to @bhoggard for pointing this out!


yuki24:
merging this PR sine it should be safe.

PR #511, opened 2018-02-06
@yuki24 => [Onboarding] Be able to force a step to start on

kanaabe:
Forgot that we need this piece of logic to get the Force side working! This allows you to pass a prop called `forceStep` which sets `state.currentStep` to be the one we ask for based on the slug. It can handle slugs that doesn't exist as a step (defaults to the first step) and updates the url.

Adds a story to represent forceStep: 
image

### DIFF
--- a/src/Components/Onboarding/Steps/Artists/index.tsx
+++ b/src/Components/Onboarding/Steps/Artists/index.tsx
@@ -28,7 +28,7 @@ interface State {
 }
 
 export default class Artists extends React.Component<StepProps, State> {
-  static slug = "artists"
+  static slug: "artists" = "artists"
 
   state = {
     inputText: "",
--- a/src/Components/Onboarding/Types.ts
+++ b/src/Components/Onboarding/Types.ts
@@ -12,3 +12,5 @@ export interface StepComponent extends React.ComponentClass<StepProps> {
 export interface FollowProps {
   updateFollowCount: (count: number) => void
 }
+
+export type StepSlugs = "artists" | "budget" | "categories" | "interests" | null
--- a/src/Components/Onboarding/Types.ts
+++ b/src/Components/Onboarding/Types.ts
@@ -1,3 +1,8 @@
+import Artists from "./Steps/Artists"
+import { BudgetComponent as Budget } from "./Steps/Budget"
+import { CollectorIntentComponent as CollectorIntent } from "./Steps/CollectorIntent"
+import Genes from "./Steps/Genes"
+
 /**
  * The props interface that the step needs to implement for the wizard.
  */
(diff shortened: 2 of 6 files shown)


src/Components/Onboarding/Wizard.tsx
```diff
@@ -4,6 +4,7 @@ import { StepComponent } from "./Types"
 interface Props {
   stepComponents: StepComponent[]
   redirectTo?: string
+  forceStep?: string
```
yuki24: This may be a bit exhausting to maintain, but I'd like a more strict type check here so human errors like a typo could be caught earlier:

```ts
import Budget from './Steps/Budget'
...

interface Props {
  ...
  forceStep?: Budget.slug | ... // 
}
```

kanaabe: 👍 Oh, I like it!

yuki24:
looks great, merging 👍

kanaabe:
@yuki24 Ok check out those updated types.

src/Components/Onboarding/Steps/Artists/index.tsx
```diff
@@ -28,7 +28,7 @@ interface State {
 }
 
 export default class Artists extends React.Component<StepProps, State> {
-  static slug = "artists"
+  static slug: "artists" = "artists"
```
yuki24: Is specifying `"artists"` as a type necessary to make TS happy?

src/Components/Onboarding/Types.ts
```diff
@@ -12,3 +17,10 @@ export interface StepComponent extends React.ComponentClass<StepProps> {
 export interface FollowProps {
   updateFollowCount: (count: number) => void
 }
+
+export type StepSlugs =
+  | typeof Budget.slug
+  | typeof Artists.slug
+  | typeof Genes.slug
+  | typeof CollectorIntent.slug
+  | null
```
yuki24: very nice 👍

src/Components/Onboarding/Wizard.tsx
```diff
@@ -14,11 +15,36 @@ interface State {
 class Wizard extends React.Component<Props, State> {
   static defaultProps = {
     redirectTo: "/",
+    forceStep: "",
```
yuki24: Should the default value be `null` or an empty string?

src/Components/Onboarding/Steps/Artists/index.tsx
kanaabe: Yeah, this is necessary to tell the type that we want a string value of "x". Otherwise, `typeof Artists.slug` will just infer that it is a String and not the actual value.

src/Components/Onboarding/Wizard.tsx
kanaabe: Ah good catch, this should be null.

yuki24:
looks great! I'll go ahead and merge.

PR #512, opened 2018-02-06
@xtina-starr => [Onboarding][AR-74] Autofocus on the artist step 

kanaabe:
Closes https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=AR&modal=detail&selectedIssue=AR-74

It looks like the Gene step already has autofocus on, so it's only the artist step.


xtina-starr:
easy peasy!

PR #513, opened 2018-02-07
@yuki24 => Fixes static variables on HOC

kanaabe:
We need to pass the slug onto the HOC that wraps BudgetComponent. The preferred way is to do `Budget.slug = BudgetComponent.slug` but that returns a TS error on `ContextConsumer` for not having defined slug. However, we probably don't want to litter the `ContextConsumer` types to do this so this works around the error by using string literals to set the variable.

I don't know if this is controversial to people, but it works 😄 

Ref: https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over

### DIFF
--- a/src/Components/Onboarding/Steps/Budget.tsx
+++ b/src/Components/Onboarding/Steps/Budget.tsx
@@ -99,4 +99,8 @@ export class BudgetComponent extends React.Component<
   }
 }
 
-export default ContextConsumer(BudgetComponent)
+const Budget = ContextConsumer(BudgetComponent)
+// tslint:disable:no-string-literal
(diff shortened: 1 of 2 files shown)


yuki24:
Ah that is tricky, but I like this workaounrd. Moving forward we may want to introduce the [`react-router`](https://github.com/ReactTraining/react-router), but I think what we have in place is a simple solution that could be implemented fairly quickly, so 👍

alloy:
Switching the consumer to use a ‘render prop’ rather than being a HOC and wrapping the component would probably fix this issue https://reactjs.org/docs/render-props.html.

PR #514, opened 2018-02-09
@yuki24 => [Onboarding] Wizard progress indicator

kanaabe:
https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=AR&modal=detail&selectedIssue=AR-91

Wasn't too sure about the stripes in the background of the indicator bc the comps were a little fuzzy but this is what I came up with:
image

Also the animation here uses ease-in but we can configure that to whatever.

onboarding-progress

### DIFF
--- a/src/Components/Onboarding/Wizard.tsx
+++ b/src/Components/Onboarding/Wizard.tsx
@@ -66,7 +66,12 @@ class Wizard extends React.Component<Props, State> {
     const { stepComponents, redirectTo } = this.props
 
     if (currentStep >= stepComponents.length - 1) {
-      window.location.href = redirectTo
+      this.setState({
+        percentComplete: 1,
+      })
+      setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/Components/Onboarding/Wizard.tsx
```diff
@@ -63,18 +66,31 @@ class Wizard extends React.Component<Props, State> {
     const { stepComponents, redirectTo } = this.props
 
     if (currentStep >= stepComponents.length - 1) {
-      window.location.href = redirectTo
+      this.setState({
+        percentComplete: 1,
+      })
+      setTimeout(() => {
+        window.location.href = redirectTo
+      }, 500)
```
kanaabe: This lets you see 100% completion before redirecting the page. I found it didn't feel as satisfying without this 😛

yuki24:
This LTGM!

PR #516, opened 2018-02-09
@yuki24 => [Onboarding] Analytics

kanaabe:
Addresses https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=AR&modal=detail&selectedIssue=AR-96

### DIFF
--- a/src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
+++ b/src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
@@ -39,7 +39,7 @@ class ArtistSearchResultsContent extends React.Component<RelayProps, null> {
   }
 
   onArtistFollowed(
-    artistId: string,
+    artist: any,
     store: RecordSourceSelectorProxy,
     data: SelectorData
   ): void {
@@ -57,7 +57,8 @@ class ArtistSearchResultsContent extends React.Component<RelayProps, null> {
       { term: this.props.term }
     )
     const updatedPopularArtists = popularArtists.map(
-      artist => (artist.getDataID() === artistId ? artistToSuggest : artist)
+      artistItem =>
+        artistItem.getDataID() === artist.__id ? artistToSuggest : artistItem
     )
 
     popularArtistsRootField.setLinkedRecords(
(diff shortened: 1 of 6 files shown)


alloy:
It might be a bit much for this PR, but Reaction could do with some upgrading to Emission’s state of analytics https://github.com/artsy/emission/blob/master/src/lib/utils/track.ts /cc @mennenia

src/Components/Onboarding/Steps/Artists/ArtistSearchResults.tsx
```diff
@@ -18,40 +26,72 @@ interface RelayProps extends React.HTMLProps<HTMLAnchorElement>, Props {
...
-    this.excludedArtistIds = new Set(this.props.viewer.match_artist.map(item => item._id))
+    this.excludedArtistIds = new Set(
+      this.props.viewer.match_artist.map(item => item._id)
+    )
   }
 
-  onArtistFollowed(artistId: string, store: RecordSourceSelectorProxy, data: SelectorData): void {
-    const suggestedArtistEdge = data.followArtist.artist.related.suggested.edges[0]
+  onArtistFollowed(
+    artist: any,
```
kanaabe: One change to note is that I'm passing in artist now instead of just `artist.__id` since we need `_id`, `id`, and `__id` in this function. This happens for all the subsequent files as well. Let me know if you see any problems with this!

src/Components/Onboarding/Wizard.tsx
```diff
@@ -61,17 +65,24 @@ class Wizard extends React.Component<Props, State> {
...
+    const { redirectTo } = this.props
+    this.setState({ percentComplete: 1 })
+    setTimeout(() => {
+      window.location.href = redirectTo
+    }, 500)
+
+    this.props.tracking.trackEvent({
+      action: "Completed Onboarding",
+    })
+  }
```
kanaabe: This is the only track call that will be lost when we remove the Wizard but I can follow up when we transfer to the react-router implementation.

yuki24:
LGTM! merging...

PR #517, opened 2018-02-12
@yuki24 => [Onboarding] Style QA PT.2

kanaabe:
Addresses the "In Progress" cards in: https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=AR&modal=detail&selectedIssue=AR-69

onboarding-qa

### DIFF
--- a/src/Components/Onboarding/Steps/Layout.tsx
+++ b/src/Components/Onboarding/Steps/Layout.tsx
@@ -1,7 +1,7 @@
 import React from "react"
 import styled from "styled-components"
-
 import Colors from "../../../Assets/Colors"
+import { primary } from "../../../Assets/Fonts"
 import InvertedButton from "../../Buttons/Inverted"
 import { media } from "../../Helpers"
 import StyledTitle from "../../Title"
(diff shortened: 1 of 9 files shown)


src/Components/Onboarding/Steps/Layout.tsx
```diff
@@ -25,49 +25,40 @@ const Container = styled.div`
...
   position: fixed;
   bottom: 0px;
   left: 0px;
-`
-
-/* Mobile safari doesn't support for the fixed position property:
- *   https://www.eventbrite.com/engineering/mobile-safari-why/
- **/
-const StickyButtonContainer = styled.div`
-  position: sticky;
```
yuki24: Have you tested this on different sizes of iOS devices? They are tricky and don't always respect `position: fixed`.

src/Components/Title.tsx
```diff
@@ -22,22 +23,16 @@ const Title: React.SFC<TitleProps> = props => {
...
 }
 
 const StyledTitle = styled(Title)`
   font-size: ${props => titleSizes[props.titleSize]};
   color: ${props => props.color};
   margin: 20px 0;
-  ${fonts.secondary.style}
-
-  ${media.sm`
+  ${fonts.secondary.style} ${media.sm`
```
yuki24: I wonder if this is prettier that auto-formats code, but at the same time, I'm not sure what the ideal formatting here is.

cc @orta

orta: This is prettier doing it, I'd call it as not perfect, but reasonable

kanaabe: Ah in the past I've used `;` to separate lines in these instances. Just updated the PR to reflect that!

src/Components/Onboarding/Steps/Layout.tsx
kanaabe: Not sure what's going on with storybooks on Safari rn (lmknow if you're seeing the same and I'll log an issue to keep track of it). But anyway I reverted back to the other implementation with just style updates so we should be covered. 👍

yuki24:
This is 💯

PR #521, opened 2018-02-14
@damassi => Add semantic-release

eessex:
- Adds semantic-release package
- Adds custom commit message `releaseRules` to package.json in the Ember standard
- Runs semantic-release after tests pass on master branch
- Updates readme to include explanation and examples for commit formats.

GH_TOKEN has also been added to reaction on Circle CI (npm token was already present)


circle.yml
```diff
@@ -27,6 +27,9 @@ test:
     - yarn test -- --runInBand
 
 deployment:
+  release:
+    branch: master
+      - yarn run semantic-release
```
damassi: I think there is some legacy stuff around [here](https://github.com/eessex/reaction/blob/0444c1109a60d8afb225c5034b840be83d191e7d/circle.yml#L38) that relates to [legacy deployment](https://github.com/artsy/reaction#deployments) process. Once we've verified that this works we should make sure to remove that and update README.

eessex: For sure!

eessex:
After merging conflicts, 17K lines were added to this file -- any idea where these came from?

orta:
It looks like they came from the https://github.com/artsy/reaction/pull/522 change - I think you `git add .`'d all the old files, run `find . -name __generated__ | xargs rm -rf` and then amend your commit 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@artsy/reaction-force",
-  "version": "0.23.1",
+  "version": "0.0.0-development",
```
damassi: Will this account for existing packages that have already been published to NPM? Does it determine from there that the existing release is `0.23.1` and go from there?

package.json
```diff
@@ -178,5 +181,16 @@
...
+  "release": {
+    "analyzeCommits": {
+      "preset": "ember",
+      "releaseRules": [
+        {"tag": "FIX", "release": "patch"},
+        {"tag": "DOC", "release": "patch"},
+        {"tag": "FEATURE", "release": "minor"},
+        {"tag": "BREAKING", "release": "major"}
+      ]
+    }
```
damassi: 👍

package.json
eessex: Yes it should fetch the version number from the existing package on npm.

README.md
```diff
@@ -50,11 +50,29 @@
...
-    $ git push --follow-tags
+Release versions (major, minor, patch) are triggered [by commit messages](https://github.com/semantic-release/semantic-release#commit-message-format), when they adhere to [Ember conventions](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-ember/readme.md):
 
-The package.json file will be updated and a commit staged. When a tag is pushed to `master`, Circle CI will publish to NPM. See [circle.yml](https://circleci.com/docs/1.0/npm-continuous-deployment/) and the [npm docs](https://docs.npmjs.com/getting-started/publishing-npm-packages).
+```
+[TAG context] commit message
+```
+
+[Valid tags](https://github.com/artsy/reaction/blob/master/package.json#L175) for release include DOC, FIX (patch), FEATURE (minor), and BREAKING (major). Commits that do not adhere to this convention will not trigger an NPM release.
+
```
damassi: What do you think about `FIX` vs `PATCH` as being the primary trigger? I always associate `FIX` with "BUGFIX" rathar than a minor change.

package.json
eessex: When merging, the build on Circle will log the build evaluation and which commits trigger changes, as well as the new version number.

README.md
eessex: Open to either, we can also add both or additional tags.

damassi: Yeah, lets go ahead and associate `PATCH` and `FIX` with a minor release

package.json
damassi: Sweet, just wanted to double check 👍

README.md
```diff
@@ -50,11 +50,29 @@
...
+##### Example Patch Release
+```
+[FIX onboarding]: increase plus-button size
+```
```
damassi: And update with additional example so that its clear:
```
[FIX onboarding] Modal does not open
[PATCH] Bump version

damassi: Also, can you remove the `:` from the docs so that its clear that this syntax is not required?

damassi:
This looks good to me -- @orta / @alloy, thoughts? Otherwise 👍

package.json
eessex: Also wanted to note it looks like this version number is being overwritten when I push a commit, but don't see where this is being set.

orta:
Have a think if there's a peril rule that could remind you about adding  `[FIX]` or `[PATCH]` to commits

damassi:
@orta - what I'm thinking is that those should only be added when a release is desired, though @eessex and I went back and forth about whether _every_ commit should be a release. I'm in the camp that for a project like Reaction it should be targeted and selective. For example, there could be many commits in a PR that has nothing to do with a release, or there could be many commits and then the last one is tagged signalling a release, or the whole thing could be squashed into one commit with a tag. But I think it should be conventional rather than required. Thoughts? Maybe a danger _warning_? But even then, i think having our README up to date in terms of how to release is sufficient -- a `git commit --allow-empty -m '[PATCH]` would be enough.

alloy:
I like it 👍 

I like that you still need to trigger a release, the future will tell us wether or not people in practice run into issues with that.

damassi:
Alright, gonna merge. Thanks for diving into this new area @eessex, will def help keep things efficient 👍

orta:
Yeah, I like the idea 👍

orta:
i think it's simple enough to not need a danger/peril rule

PR #525, opened 2018-02-15
Adds inline type definitions to all files exported by Reaction

yuki24:
Signed-off-by: Yuki Nishijima <yuki@artsymail.com>

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,8 @@
     "plugins": [{ "name": "typescript-styled-plugin" }],
     "rootDir": "src",
     "sourceMap": true,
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "plugins": [{ "name": "typescript-styled-plugin" }],
     "rootDir": "src",
     "sourceMap": true,
-    "target": "es2015"
+    "target": "es2015",
+    "declaration": true
```
damassi: Can you alphabetize this?

yuki24: cc @orta

PR #527, opened 2018-02-15
Move the <RoutesWithProgressBar> component from Force to Reaction

yuki24:
## WIP: Do not merge this yet

_When I tried to test this staging was missing popular artists data, which caused the `<PopularArtists>` component to fail to load._

This brings the `<RoutesWithProgressBar>` component (currently in Force) onto Reaction. We wanted to use the `@track` decorator in Force, but it turns out that not everything is set up properly and it seems a bit too early to adopt it. For now, it may make more sense to maintain TS classes and components in Reaction.

### DIFF
--- a/src/Components/Onboarding/Wizard.tsx
+++ b/src/Components/Onboarding/Wizard.tsx
@@ -1,108 +1,74 @@
 import React from "react"
+import { Redirect, Route } from 'react-router'
+
 import Events from "../../Utils/Events"
 import { track } from "../../Utils/track"
 import { ProgressIndicator } from "./ProgressIndicator"
-import { StepComponent, StepSlugs } from "./Types"
 
-interface Props {
-  stepComponents: StepComponent[]
-  redirectTo?: string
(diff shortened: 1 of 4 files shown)


src/Components/Onboarding/Wizard.tsx
```diff
@@ -1,108 +1,74 @@
...
-    setTimeout(() => {
-      window.location.href = redirectTo
-    }, 500)
+  onFinish = redirectTo => {
+    this.setState({ finished: true })
+    setTimeout(() => (window.location.href = redirectTo || "/"), 500)
 
     this.props.tracking.trackEvent({
       action: "Completed Onboarding",
     })
```
yuki24: I left the analytics code here so we should be good to go.

kanaabe: Ohhhhh I see now. Using @track directly in Force was the problem. Not that the track calls in each step didn't work, just this one in the main wrapper. Yea that totes makes sense.

PR #530, opened 2018-02-16
@eessex => Adds tracking to SeriesAbout description partner link

kanaabe:
Looks like we missed this analytics call when making Series.


src/Components/Publishing/Series/SeriesAbout.tsx
```diff
@@ -10,67 +11,90 @@ interface Props {
...
 
+  componentDidMount() {
+    const textLink = document.querySelector(".SeriesAbout__description a")
+    if (textLink) {
+      textLink.addEventListener("click", this.onClickFooterLink)
+    }
+  }
+
+  onClickFooterLink = event => {
+    this.props.tracking.trackEvent({
```
eessex: 👍

eessex:
Forgot to mention that we should use the new commit syntax for automatic NPM releases!

eessex:
about to do one more QA pass for the circle config, so this will be included in the next release -- no change needed on your side.

kanaabe:
OLALA!! 🥂

PR #531, opened 2018-02-16
[FIX semantic-release] consolidate deploy commands

eessex:
Circle was not running the storybooks command because the circle.yml `deployment` command only accepts one argument per branch -- this consolidates the release and demo blocks.


damassi:
LGTM 👍 will merge on 🍏

PR #533, opened 2018-02-20
@damassi => Slack integration settings

eessex:
Add whitelist for chatroom notifications to `circle.yml` -- only allows messages for the master branch.


damassi:
👍

PR #534, opened 2018-02-20
[Onboarding] Fix the Wizard component not pulling in realy environment

yuki24:
_This is unrelated to the build failures we observed last week and just a fix for the bug introduced by https://github.com/artsy/reaction/pull/527_

The `<CollectorIntentComponent>` and `<BudgetComponent>` components were imported and rendered in the `<Wizard>` component, but they weren't wrapped by the `ContextConsumer` function and don't know about `currentUser` or `relayEnvironment`. AS a result of this, clicking the NEXT button, which makes a call to MP, started failing due to `401 Unauthorized`.

This PR updates the `<Wizard>` component to import the right step components so it'll be able to make a request with an appropriate token.

### DIFF
--- a/src/Components/Onboarding/Wizard.tsx
+++ b/src/Components/Onboarding/Wizard.tsx
@@ -6,15 +6,15 @@ import { track } from "../../Utils/track"
 import { ProgressIndicator } from "./ProgressIndicator"
 
 import Artists from './Steps/Artists'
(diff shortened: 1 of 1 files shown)


kanaabe:
👍

kanaabe:
Oh I forgot to mention we should probably use tagged commits to publish a new version 🌳

PR #537, opened 2018-02-20
Adds a user agent to the metaphysics fetch

orta:
This at least handles the reaction side of https://github.com/artsy/force/issues/2191

Happy to change what the text actually is, but at least it's there.

### DIFF
--- a/src/Utils/__test__/metaphysics.test.ts
+++ b/src/Utils/__test__/metaphysics.test.ts
@@ -1,0 +1,22 @@
+jest.mock("isomorphic-fetch")
+import metaphysics from "../metaphysics"
+
+declare const global: any
+global.fetch = jest.fn(() =>
+  Promise.resolve({
+    status: 200,
+    json: () => Promise.resolve({ data: {} }),
+  })
+)
+
+it("Adds a user agent for reaction", () => {
+  expect.assertions(1)
+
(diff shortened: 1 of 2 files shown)


alloy:
Doesn’t matter much right now, but at some point Reaction will probably be used in other apps, so maybe the user-agent should also hint at what the host app is? I'm fine with merging now as-is, though.

kanaabe:
It already is in Positron 😄

joeyAghion:
Yeah same thought... it would be nice if this was a default/fall-back and including systems could override it with their own agent identifiers. But this is much better than nothing!

alloy:
@kanaabe My bad, thanks for rectifying! 🙏

alloy:
One very explicit option, albeit somewhat repetitive, is to add a prop to the Artsy component for when no `environment` prop is specified https://github.com/artsy/reaction/blob/11bbc2bbea08a99a072e3a4cfa63e47c9ed18dc7/src/Components/Artsy.tsx#L76

```tsx
<ContextProvider userAgent="Force" {...}>
  <SweetTree />
</ContextProvider>
```

joeyAghion:
Gonna merge this since it seems like an improvement on its own. Feel free to follow up with the configurability or not.

orta:
Agree, going to take a quick look at configurability now 👍

PR #540, opened 2018-02-21
@damassi => Rename NPM package 'Reaction'

eessex:
Renames package to 'reaction' in package.json -- the new version is now published here and up to date with the existing version: https://www.npmjs.com/package/@artsy/reaction

Will update consumers next 🎉 

TODO: add 'deprecated' message to reaction-force

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@artsy/reaction",
-  "version": "0.23.4",
+  "version": "0.0.0-development",
   "description": "Force’s React Components",
   "main": "dist/index.js",
   "repository": "https://github.com/artsy/reaction.git",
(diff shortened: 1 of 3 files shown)


damassi:
LGTM 👍 but seems like a snapshot failure?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@artsy/reaction-force",
+  "name": "@artsy/reaction",
   "version": "0.0.0-development",
```
l2succes: @eessex I'm a little confused as to where consumers well get the version from if `package.json` isn't updated with the correct version.

orta: @l2succes the actual version is generated at deploy time using semantic versioning - #521

l2succes: Got it, thanks @orta

orta:
Could snapshot names be based off the package.json name?

eessex:
@orta this does seem to be the case

PR #544, opened 2018-02-21
[FIX toolchain] User gulp-replace to fix relay paths

damassi:
This is a quick-fix for `src/__generated__` paths in output in order to unblock. (@alloy will follow up with a formal release in Relay)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,6 +4,7 @@ const clean = require("gulp-clean")
 const gulp = require("gulp")
 const mergeStream = require("merge-stream")
 const path = require("path")
+const replace = require("gulp-replace")
 const sourcemaps = require("gulp-sourcemaps")
 const tsc = require("gulp-typescript")
 const watch = require("gulp-watch")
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -26,16 +27,17 @@ gulp.task("compile", () => {
 
   const tsStream = tsResult.js
     .pipe(babel())
+    .pipe(replace("../src/__generated__", "./__generated__"))
```
damassi: This is the key change. Due to how https://github.com/artsy/reaction/blob/master/.babelrc#L4 encodes the full path, this will make that relative to output.

gulpfile.js
```diff
@@ -26,16 +27,17 @@ gulp.task("compile", () => {
 
   const tsStream = tsResult.js
     .pipe(babel())
+    .pipe(replace("../src/__generated__", "./__generated__")) // FIXME: Remove once Relay updates have been made
```
damassi: This is the key change. Due to how https://github.com/artsy/reaction/blob/master/.babelrc#L4 encodes the full path, this will make that relative to output.

gulpfile.js
alloy: I was thinking that we’d replace with just `__generated__`, but I guess that would break for files that reside in the top `src` directory?

alloy: So this will now also generate slightly awkward but correct paths like `../.././__generated__`, right?

damassi: Yup, then it would think its an npm module.

alloy: 👍

yuki24:
This looks great. Thank you! @damassi and @alloy

PR #547, opened 2018-02-22
[Toolchain] Replace TSC with Babel 7 

damassi:
This PR replace our TSC compilation with Babel 7.

Relates to https://github.com/artsy/force/pull/2199. 

Since working out of node_modules directly from consumer apps is filled with edge-cases best just to make compilation / dev process fast and stick with compiled code; new flow: 

```bash
cd artsy/reaction
yarn link
yarn watch // concurrently boot babel -w and babel-relay -w

cd force 
yarn link @artsy/reaction
yarn start 
```

Due to express-reloadable and webpack changes in reaction's `dist` folder are automatically picked up and hot-reloaded in. With `yarn watch`, and after initial babel compile, changes are pretty much instant across apps. 

#### TODO: 
- [x] Fix Jest


damassi:
Need to do some cleanup in https://github.com/artsy/force/pull/2199, but we're pretty much done! 😅

.babelrc
```diff
@@ -9,9 +20,12 @@
         "displayName": true,
         "minify": true,
         "transpileTemplateLiterals": true
-        // "preprocess": true // still experimental
       }
-    ]
+    ],
+    ["transform-rename-import", {
+      "original": "..\/src\/__generated__",
+      "replacement": "$1.\/__generated__"
```
damassi: This will eventually be replaced once relay plugin changes are merged upstream

package.json
```diff
@@ -23,36 +17,61 @@
...
+    "@babel/core": "7.0.0-beta.40",
+    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.40",
+    "@babel/plugin-proposal-decorators": "^7.0.0-beta.40",
+    "@babel/preset-env": "^7.0.0-beta.40",
+    "@babel/preset-es2015": "7.0.0-beta.40",
+    "@babel/preset-react": "7.0.0-beta.40",
+    "@babel/preset-stage-3": "7.0.0-beta.40",
+    "@babel/preset-typescript": "7.0.0-beta.40",
+    "@babel/template": "7.0.0-beta.40",
+    "@babel/types": "7.0.0-beta.40",
```
damassi: @alloy Are we using `template` or `types` anywhere? This was in previous `package.json` so i didn't want to remove.

alloy: Not anywhere in Force that I know of, only in deps.

src/Utils/track.ts
```diff
@@ -95,7 +95,9 @@ export function trackWithProps<P, T extends TrackingInfo.Global = TrackingInfo.G
  *        }
  *      }
  */
-export function trackWithoutProps<T extends TrackingInfo.Global = TrackingInfo.Global>(): Track<T, any> {
+export function trackWithoutProps<
+  T extends Global = Global
+>(): Track<T, any> {
   return _track
 }
```
orta: /cc @mennenia - we may eventually need to make similar changes (removing the `namespace`) in Emission - [context here in slack](https://artsy.slack.com/archives/C2TQ4PT8R/p1519255117000112)

alloy:
@damassi Can we have a single `start` script that does all the right things? (I was also wondering if it shouldn’t still start the storybooks server in reaction?)

package.json
```diff
@@ -19,7 +19,7 @@
   "scripts": {
     "clean": "rm -rf dist",
     "compile":
-      "babel src --out-dir dist --copy-files -s inline --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/__tests__,src/**/__stories__",
+      "babel src --out-dir dist --copy-files -s --source-map inline --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/__tests__,src/**/__stories__",
```
alloy: Isn’t the `-s` option for source-maps already?

damassi: that `inline` is a typo, which was moved over when I added in `--source-map` which is the command to output files. We were previously embedding sourcemaps into the files via `inline`, increasing weight; need to remove.

alloy: 👍

src/Components/Publishing/Series/SeriesAbout.tsx
```diff
@@ -43,6 +44,7 @@ export class SeriesAbout extends Component<Props, null> {
         ? sponsor.partner_dark_logo
         : sponsor.partner_light_logo)
 
+    // return <div>hi</div>
```
alloy: Stray?

tsconfig.json
```diff
@@ -9,10 +9,8 @@
     "module": "es2015",
     "moduleResolution": "node",
     "noUnusedLocals": true,
-    "outDir": "dist",
     "plugins": [{ "name": "typescript-styled-plugin" }],
     "rootDir": "src",
-    "sourceMap": true,
```
alloy: Is this not needed to generate source-maps?

damassi: Actually, let me see. For some reason I had presumed that Babel would handle this via the preset but let me double check.

damassi: @alloy - looks good: 

screen shot 2018-02-22 at 2 32 04 pm

alloy: 👍

alloy:
Great work! 👏 

Can you add yourself to the list of point people on this project? Once that’s in go ahead and merge on 💚

PR #552, opened 2018-02-23
[FIX tracking-context] Add context to contextTypes

damassi:
It looks like the compiler is a bit more strict. If defining a context it (with `@tracking()`) we need to be sure to add it in to the list. The fullScreenModal context was breaking things.

### DIFF
--- a/src/Components/Publishing/Layouts/ArticleLayout.tsx
+++ b/src/Components/Publishing/Layouts/ArticleLayout.tsx
@@ -1,11 +1,11 @@
-import { get } from 'lodash'
+import { get } from "lodash"
 import { cloneDeep, includes, map, omit } from "lodash"
 import PropTypes from "prop-types"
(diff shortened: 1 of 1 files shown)


src/Components/Publishing/Layouts/ArticleLayout.tsx
```diff
@@ -52,18 +52,22 @@ interface ArticleContainerProps {
...
+    }
+  },
+  {
+    dispatch: data => Events.postEvent(data),
   }
 )
 export class ArticleLayout extends React.Component<ArticleProps, ArticleState> {
   static childContextTypes = {
     onViewFullscreen: PropTypes.func,
+    tracking: PropTypes.object,
```
damassi: Since tracking injects a context, and we were requesting a context for `onFullScreen`, without defining it in the list it errors out.

PR #562, opened 2018-02-26
[DOC readme] Update with yarn watch instructions

damassi:
Adds docs around 

```
yarn link
cd ../force && yarn link @artsy/reaction && yarn start
```

(I'm still thinking about how I'd like to wrap all of these commands up)


orta:
👍

PR #563, opened 2018-02-27
@sarahscott => News text / Fixtures updates

eessex:
- Adds styles for News articles to `StyledText` component
- Adds 'news' layout to article typings
- Adds helper function `TextFromArticle` to get all html from text sections
- Reworks text fixtures to conform to section modeling, adds fixture articles with only text sections
- Uses cloneDeep in fixtures when duplicating articles

*** blockquote width is determined in sectionContainer, so will not show up in this screenshot 

screen shot 2018-02-27 at 11 58 06 am

### DIFF
--- a/src/Components/Publishing/Fixtures/Components.ts
+++ b/src/Components/Publishing/Fixtures/Components.ts
@@ -1,6 +1,5 @@
-import { compact, map } from "lodash"
+import { flatten } from "lodash"
 import { Props as ImageSetPreviewProps } from "../Sections/ImageSetPreview"
-import { NewsArticle } from "./Articles"
 
 export const ArtworkMissingInfo = {
   type: "artwork",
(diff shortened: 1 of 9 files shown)


src/Components/Publishing/Fixtures/Components.ts
```diff
@@ -555,41 +556,75 @@ export const Videos = [
...
-    "<p>Rines is a part of a recent migration that has galleries gravitating to the Lower East Side and Chinatown. She opened 56 Henry in December 2015, just four months after being forced to shutter her previous microgallery 55 Gansevoort—an elevator shaft a block east of the Whitney—after Restoration Hardware bought its building.</p>",
-    "<h2><a href='https://www.artsy.net/artist/ej-hill' class='is-follow-link' name='EJ_Hill'>EJ Hill</a><a data-id='ej-hill' class='entity-follow artist-follow'></a></h2>",
-    "<p><a href='https://www.artsy.net/artist/ej-hill' class='is-follow-link' name='EJ_Hill'>EJ Hill</a><a data-id='ej-hill' class='entity-follow artist-follow'></a></p>",
-  ],
-  // standard & feature close
-  [
-    "<p>Rines is a part of a recent migration that has galleries gravitating to the Lower East Side and Chinatown. She opened 56 Henry in December 2015, just four months after being forced to shutter her previous microgallery 55 Gansevoort—an elevator shaft a block east of the Whitney—after Restoration Hardware bought its building.<span class='content-end'></span></p>",
-  ],
+export const ArticleText: Object[] = [
```
kanaabe: Should this be ClassicText? Makes the subsequent Text stuff more clear I think?

src/Components/Publishing/Sections/StyledText.tsx
```diff
@@ -186,5 +219,5 @@ export const StyledText = div`
 `
 
 StyledText.defaultProps = {
-  color: 'black'
+  color: "black",
 }
```
kanaabe: Curious, did you find this hard to trace through and know where to add styles? I wonder if this will become increasingly difficult to manage as we add more layouts and cover more edge cases. Seems like separating them eventually will be nice.

kanaabe:
LGTM! Looping in @sarahscott to the review mix ☘️

src/Components/Publishing/Sections/StyledText.tsx
eessex: Most of the changes here were prettier -- was actually easier than expected to change (was only 2 lines to support news).  There are other components like 'Header' that might want to be broken out more as we add more layouts.

src/Components/Publishing/Sections/__tests__/Text.test.tsx
```diff
@@ -1,20 +1,45 @@
...
+  TextStandardArticle,
+} from "../../Fixtures/Articles"
+import { TextFromArticle } from "../../Fixtures/Helpers"
 import { Text } from "../Text"
 
 it("renders classic text properly", () => {
-  const text = renderer.create(<Text html={SectionText.classic} layout="classic" />).toJSON()
+  const text = renderer
+    .create(
+      <Text html={TextFromArticle(TextClassicArticle)} layout="classic" />
```
damassi: 👍

src/Components/Publishing/Sections/StyledText.tsx
kanaabe: True, i guess the styles are similar enough still that it doesn’t make sense yet to split out. Header for sure 👍🏼

eessex: The one that requires the most independent styles is 'classic' layout, so as formats grow we might consider using `StyledText` as a default, and extending it `ClassicStyledText` etc. rather than using conditionals in this file.

sarahscott:
Looks great to me! Merging ✨

PR #565, opened 2018-02-28
[WIP] News Headline component

sarahscott:
- [x] Basic news headline component
- [ ] Tests

screen shot 2018-02-28 at 2 11 55 pm

And some questions about web styling mysteries from this humble mobile dev :
- When do I need to use `pMedia`?
- Is "Headers" the right bucket for this component? Is there a difference between Headers and Headlines?

I expect that a lot of what I'm doing needs to be changed or tweaked so please don't be afraid to be harsh 😄. The specifics of web styling are all new for me.

Related to https://artsyproduct.atlassian.net/browse/AR-265

### DIFF
--- a/src/Components/Publishing/Header/Header.tsx
+++ b/src/Components/Publishing/Header/Header.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { ClassicHeader } from "./ClassicHeader"
 import { FeatureHeader } from "./FeatureHeader"
+import { NewsHeadline } from "./NewsHeadline"
 import { StandardHeader } from "./StandardHeader"
 
 interface HeaderProps {
@@ -29,9 +30,11 @@ const getVertical = (article, children) => {
 }
 
 const getLeadParagraph = (article, children) => {
-  const leadParagraph = article.lead_paragraph
-    ? <div dangerouslySetInnerHTML={{ __html: article.lead_paragraph }} />
-    : false
+  const leadParagraph = article.lead_paragraph ? (
+    <div dangerouslySetInnerHTML={{ __html: article.lead_paragraph }} />
+  ) : (
+    false
+  )
   return children ? children[1] : leadParagraph
 }
 
@@ -45,9 +48,10 @@ const getDeck = (article, children) => {
 export const Header: React.SFC<HeaderProps> = props => {
   const { article, children, date, height, isMobile } = props
   const title = getTitle(article, children)
+  const layout = article.layout
 
   // Classic Article
-  if (article.layout === "classic") {
+  if (layout === "classic") {
     const leadParagraph = getLeadParagraph(article, children)
     return (
       <ClassicHeader
@@ -58,13 +62,15 @@ export const Header: React.SFC<HeaderProps> = props => {
         leadParagraph={leadParagraph}
       />
     )
-    // Feature
-  } else {
-    const deck = getDeck(article, children)
-    const vertical = getVertical(article, children)
-    const image = children && children[3]
+  }
 
-    if (article.layout === "feature") {
+  const deck = getDeck(article, children)
+  const vertical = getVertical(article, children)
+  const image = children && children[3]
+
+  // tslint:disable-next-line:switch-default
+  switch (layout) {
+    case "feature": {
       return (
         <FeatureHeader
           article={article}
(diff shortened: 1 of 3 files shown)


src/Components/Publishing/Header/Header.tsx
```diff
@@ -29,9 +30,11 @@ const getVertical = (article, children) => {
 }
 
 const getLeadParagraph = (article, children) => {
-  const leadParagraph = article.lead_paragraph
-    ? <div dangerouslySetInnerHTML={{ __html: article.lead_paragraph }} />
-    : false
+  const leadParagraph = article.lead_paragraph ? (
```
sarahscott: prettier

src/Components/Publishing/Header/Header.tsx
```diff
@@ -58,13 +62,15 @@ export const Header: React.SFC<HeaderProps> = props => {
...
-    const vertical = getVertical(article, children)
-    const image = children && children[3]
+  }
 
-    if (article.layout === "feature") {
+  const deck = getDeck(article, children)
+  const vertical = getVertical(article, children)
+  const image = children && children[3]
+
+  // tslint:disable-next-line:switch-default
```
sarahscott: Before, I guess the `default` case was standard - does that sound right or does a `default`-less switch case make sense now that there are more `layout` possibilities?

src/Components/Publishing/Header/NewsHeadline.tsx
```diff
@@ -0,0 +1,38 @@
...
+  margin: 0 40px;
+`
+
+const NewsHeadlineContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  max-width: 780px;
+  width: 100%;
+  margin: 40px auto;
+  box-sizing: border-box;
```
sarahscott: Tbh not sure what `box-sizing` does; copied from `StandardHeader`.

src/Components/Publishing/Header/Header.tsx
kanaabe: Either works for me, I think it makes it more clear now to use `switch` so 👍.

src/Components/Publishing/News/NewsHeadline.tsx
```diff
@@ -0,0 +1,49 @@
...
+  max-width: 780px;
+  width: 100%;
+  margin: 40px auto;
+  box-sizing: border-box;
+  margin-bottom: 30px;
+`
+
+const Title = styled.div`
+  ${Fonts.garamond("s34")};
+  font-weight: 600;
```
kanaabe: Good point, lets leave this for now. I _think_ we'll want to use a different font here (something like Garamond Bold) instead of setting the font-weight but it looks like we don't support it in our `Fonts` file atm.

src/Components/Publishing/News/NewsHeadline.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <NewsHeadlineContainer>
+        <Title>{editTitle ? editTitle : title}</Title>
+      </NewsHeadlineContainer>
+    </NewsHeadlineParent>
+  )
+}
+
+const NewsHeadlineParent = styled.div`
+  margin: 0 40px;
+  ${pMedia.sm`
```
kanaabe: Nice nice nice nice

src/Components/Publishing/News/__tests__/NewsHeadline.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const component = renderer
+    .create(<NewsHeadline article={NewsArticle} />)
+    .toJSON()
+  expect(component).toMatchSnapshot()
+})
+
+it("renders a headline with children properly", () => {
+  const component = renderer
+    .create(
+      <NewsHeadline article={NewsHeadline} editTitle={EditableChild("title")} />
```
kanaabe: Woohoo! Pretty easy right? This will make a lot of sense when we move to Positron

src/Components/Publishing/Header/NewsHeadline.tsx
kanaabe: Ooh yeah box-sizing is interesting -- https://css-tricks.com/box-sizing/ and https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

You _might_ not need it here but maybe remove it and see if things change. It will be pretty obvious if things break because this controls sizing of content based on width/height/margin/padding

kanaabe:
Congrats on your foray into Reaction/Web! This looks awesome. 💯

PR #567, opened 2018-03-05
@sarahscott => News article parent component

kanaabe:
This adds a NewsLayout and NewsSections components to the mix. The NewsLayout gets used by the main Article component to render a full article. 

newsarticle

### DIFF
--- a/src/Components/Publishing/News/NewsSections.tsx
+++ b/src/Components/Publishing/News/NewsSections.tsx
@@ -2,11 +2,11 @@ import React, { Component } from "react"
 import { Col, Row } from "react-styled-flexboxgrid"
 import styled from "styled-components"
 import { pMedia } from "../../Helpers"
+import { Embed } from "../Sections/Embed"
+import { ImageCollection } from "../Sections/ImageCollection"
+import { Text } from "../Sections/Text"
+import { Video } from "../Sections/Video"
 import { Layout } from "../Typings"
-import { Embed } from "./Embed"
-import { ImageCollection } from "./ImageCollection"
-import { Text } from "./Text"
-import { Video } from "./Video"
 
 interface Props {
   article: {
--- a/src/Components/Publishing/News/NewsSections.tsx
+++ b/src/Components/Publishing/News/NewsSections.tsx
@@ -12,10 +12,8 @@ interface Props {
   article: {
     layout: Layout
     authors?: any
-    postscript?: string
     sections?: any
   }
-  isMobile?: boolean
 }
 
 interface ContainerProp {
@@ -23,10 +21,6 @@ interface ContainerProp {
 }
 
 export class NewsSections extends Component<Props, null> {
-  static defaultProps = {
-    isMobile: false,
-  }
-
   getSection(section, index) {
     const { article } = this.props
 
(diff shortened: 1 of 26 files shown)


src/Components/Publishing/News/NewsSections.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    return renderedSections
+  }
+
+  renderByline() {
+    const { article: { authors } } = this.props
+
+    if (authors) {
+      return
+      // Return a byline
```
kanaabe: This is for you!

src/Components/Publishing/Sections/ImageCollection.tsx
```diff
@@ -15,26 +15,29 @@ interface ImageCollectionProps {
   targetHeight?: number
   gutter?: number
   sectionLayout?: SectionLayout
+  articleLayout?: Layout
```
kanaabe: This component represents a section called `image_collection`. It's a group of `Image` and `Artwork` components in a row.

For Newsfeed, we don't want the ability to expand images into a fullscreen viewer. We have the ability to turn that feature off on the `Image` and `Artwork` components individually but we need to pass down this `articleLayout` prop to make that work within an `ImageCollection`.

src/Components/Publishing/Sections/ImageCollection.tsx
```diff
@@ -53,6 +56,7 @@ class ImageCollectionComponent extends React.PureComponent<ImageCollectionProps,
           <Image
             image={image}
             sectionLayout={sectionLayout}
+            layout={articleLayout}
```
kanaabe: This is where we pass it through. `Image` now knows that this has a `layout="news"` so it won't render that `IconExpand`.

src/Components/Publishing/Sections/Sections.tsx
```diff
@@ -109,54 +107,48 @@ export class Sections extends Component<Props, State> {
...
   }
 
   getSection(section, index) {
+    const { article } = this.props
+
     const sections = {
       image_collection: (
         <ImageCollection
           sectionLayout={section.layout}
+          articleLayout={article.layout}
```
kanaabe: This is actually all Prettier changes except this line -- we're now passing down the `articleLayout` again for the other Article types that use this more generic `Sections` component.

src/Components/Publishing/__stories__/Article.story.tsx
```diff
@@ -206,4 +211,7 @@ story.add("Feature", () => {
         relatedArticles={[StandardArticle, VideoArticle]}
       />
     )
-  })
\ No newline at end of file
+  })
+  .add("News Article", () => {
+    return <Article article={NewsArticle} />
+  })
```
kanaabe: 🗞

src/Components/Publishing/News/NewsSections.tsx
```diff
@@ -0,0 +1,102 @@
...
+    authors?: any
+    sections?: any
+  }
+}
+
+interface ContainerProp {
+  type: string
+}
+
+export class NewsSections extends Component<Props, null> {
```
damassi: No need to pass in `null` if not using a `State` interface.

sarahscott:
Looks awesome! Merging!

PR #568, opened 2018-03-06
@eessex => Use article id as injection marker for mobile display

kanaabe:
`uniqueId()` is problematic for SSR where sometimes the `id` generated on the server doesn't match the client and therefore can't mount the ad unit. Instead, lets use the article id which still works with infinite scroll to account for uniqueness.

### DIFF
--- a/src/Components/Publishing/Sections/Sections.tsx
+++ b/src/Components/Publishing/Sections/Sections.tsx
@@ -1,6 +1,6 @@
-import { clone, compact, once, uniqueId } from 'lodash'
+import { clone, compact, once } from "lodash"
 import React, { Component } from "react"
-import ReactDOM from 'react-dom'
+import ReactDOM from "react-dom"
 import styled, { StyledFunction } from "styled-components"
 import { pMedia } from "../../Helpers"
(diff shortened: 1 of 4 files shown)


src/Components/Publishing/Sections/Sections.tsx
```diff
@@ -71,32 +67,36 @@ export class Sections extends Component<Props, State> {
...
-    const tag = '</p>'
+    const articleId = this.props.article.id
+    const tag = "</p>"
     const updatedBody = compact(body.split(tag))
       .map(p => p + tag)
       .reduce((arr, block, paragraphIndex) => {
         if (paragraphIndex === MOBILE_DISPLAY_INJECT_INDEX) {
-          this.displayInjectId = uniqueId(MOBILE_DISPLAY_INJECT_ID_PREFIX)
+          this.displayInjectId = `${MOBILE_DISPLAY_INJECT_ID_PREFIX}${articleId}`
           return arr.concat([block, `<div id="${this.displayInjectId}"></div>`])
```
eessex: 👍

PR #569, opened 2018-03-06
@kanaabe => News Byline & updated typings

sarahscott:
This pr is super long because of prettier!! Apologies for that 🙈 

- [x] Byline dekstop
- [x] Byline mobile (except for share button; will wait on the asset)
- [x] Some exciting new typings
- [x] Rebase
- [x] Integrate into NewsArticle after @kanaabe's PR is merged
- [x] Tests

Desktop:
screen shot 2018-03-07 at 3 41 28 pm

Mobile:
screen shot 2018-03-07 at 3 41 42 pm

### DIFF
--- a/src/Components/Publishing/Byline/AuthorDate.tsx
+++ b/src/Components/Publishing/Byline/AuthorDate.tsx
@@ -19,14 +19,13 @@ interface DateProps {
 }
 
 export const Author: React.SFC<AuthorProps> = props => {
-  const includeBullet = props.articleLayout !== "news"
-
+  const { color, layout } = props
   return (
     <StyledAuthor
       className="author"
-      condensed={props.layout === "condensed"}
-      color={props.color}
-      withBullet={includeBullet}
+      condensed={layout === "condensed"}
+      color={color}
+      withBullet
     >
       By {getAuthorByline(props.authors)}
     </StyledAuthor>
--- a/src/Components/Publishing/Byline/AuthorDate.tsx
+++ b/src/Components/Publishing/Byline/AuthorDate.tsx
@@ -3,50 +3,78 @@ import styled, { StyledFunction } from "styled-components"
 import { pMedia } from "../../Helpers"
 import { getAuthorByline, getDate } from "../Constants"
 import { Fonts } from "../Fonts"
+import { ArticleLayout, BylineLayout, DateFormat } from "../Typings"
 
-interface AuthorDateProps {
+interface AuthorProps {
   authors?: any
+  layout?: BylineLayout
+  articleLayout?: ArticleLayout
   color?: string
-  date?: string
-  layout?: string
 }
 
-export const Author: React.SFC<AuthorDateProps> = props => {
+interface DateProps {
+  date?: string
+  layout?: BylineLayout
+  format?: DateFormat
+}
+
+export const Author: React.SFC<AuthorProps> = props => {
+  const includeBullet = props.articleLayout !== "news"
+
   return (
-    <StyledAuthor className="author" layout={props.layout} color={props.color}>
+    <StyledAuthor
+      className="author"
+      condensed={props.layout === "condensed"}
+      color={props.color}
+      withBullet={includeBullet}
+    >
       By {getAuthorByline(props.authors)}
     </StyledAuthor>
   )
 }
 
-export const Date: React.SFC<AuthorDateProps> = props => {
+export const Date: React.SFC<DateProps> = props => {
+  const condensed = props.layout === "condensed"
+  const format = props.format || (condensed ? "condensed" : "default")
+
   return (
-    <Text className="date" layout={props.layout}>
-      {getDate(props.date, props.layout)}
+    <Text className="date" condensed={condensed}>
+      {getDate(props.date, format)}
     </Text>
   )
 }
 
 Author.defaultProps = {
-  color: 'black'
+  color: "black",
 }
 
-const div: StyledFunction<AuthorDateProps & React.HTMLProps<HTMLInputElement>> = styled.div
+interface TextProps {
+  condensed?: boolean
+  withBullet?: boolean
+}
+
+const div: StyledFunction<TextProps & React.HTMLProps<HTMLInputElement>> =
+  styled.div
 
 const Text = div`
-  ${props => (props.layout === "condensed" ? Fonts.unica("s14", "medium") : Fonts.unica("s16", "medium"))}
+  ${props =>
+    props.condensed
+      ? Fonts.unica("s14", "medium")
+      : Fonts.unica("s16", "medium")}
   margin: 10px 30px 0 0;
   &.date {
     white-space: nowrap;
   }
   ${props => pMedia.sm`
-    ${props.layout === "condensed" ? Fonts.unica("s12", "medium") : Fonts.unica("s14", "medium")}
+    ${props.condensed
+      ? Fonts.unica("s12", "medium")
+      : Fonts.unica("s14", "medium")}
     ${Fonts.unica("s14", "medium")}
     margin: 10px 20px 0 0;
   `}
 `
-const adjustForCondensed = layout => {
-  return layout === "condensed"
+const adjustForCondensed = condensed => {
+  return condensed
     ? `
   &:before {
     min-width: 8px;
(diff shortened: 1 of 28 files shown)


src/Components/Publishing/Byline/NewsByline.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react"
+import styled from "styled-components"
+import { getArticleFullHref, getAuthorByline, getDate } from "../Constants"
+import { Fonts } from "../Fonts"
+import { Share } from "./Share"
+
+interface NewsBylineProps {
+  article: any
```
kanaabe: `ArticleData`? 😄

src/Components/Publishing/Byline/NewsByline.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+const DateAndSource: React.SFC<NewsBylineProps> = props => {
+  const { article } = props
+  const { date, news_source } = article
+
+  return (
+    <DateSourceContainer>
+      {getDate(date, "verbose")}
+      {`, via  ${news_source.title}`}
```
kanaabe: One thing to note is that not all articles will have a news_source so we want this to be a conditional

src/Components/Publishing/Byline/NewsByline.tsx
```diff
@@ -0,0 +1,57 @@
...
+  flex-direction: column;
+`
+
+const Poster = styled.div`
+  ${Fonts.unica("s14", "medium")};
+`
+
+const DateSourceContainer = styled.div`
+  display: flex;
+  ${Fonts.unica("s14")} color: #999999;
```
kanaabe: You can use Colors here: https://github.com/artsy/reaction/blob/master/src/Assets/Colors.tsx

src/Components/Publishing/Header/BasicHeader.tsx
```diff
@@ -208,41 +213,32 @@ const Container = div`
...
     margin-bottom: 15px;
-  `}
+  `};
 `
 
 const LeadParagraph = styled.div`
-  ${defaults}
-
-  ${breakpoint.xs`
+  ${defaults} ${breakpoint.xs`
```
kanaabe: Ah yeah Prettier does this but this can be separated by line by adding a semicolon. Aka this wouldn't condense those lines:
```
${defaults};
${breakpoint.xs``}
```

kanaabe: Don't feel like you have to go through and do those but just to note!

kanaabe:
Love it, this is a wonderful contribution to more types in Publishing! So far so good.

src/Components/Publishing/News/NewsSections.tsx
```diff
@@ -62,17 +59,20 @@ export class NewsSections extends Component<Props> {
     const { article: { authors } } = this.props
 
     if (authors) {
-      return
-      // Return a byline
+      return (
+        <BylineContainer>
+          <NewsByline article={this.props.article} />
+        </BylineContainer>
```
kanaabe: Thanks!

src/Components/Publishing/RelatedArticles/RelatedArticleFigure.tsx
```diff
@@ -39,6 +43,11 @@ export class RelatedArticleFigure extends React.Component<RelatedArticleFigurePr
     const { article } = this.props
     const href = getArticleHref(article.slug)
     const imageSrc = crop(article.thumbnail_image, { width: 510, height: 340 })
+    const bylineArticle = {
+      ...article,
+      layout: "series" as ArticleLayout,
```
kanaabe: Why do we manually set layout here?

sarahscott: Because the `RelatedArticleFigure` doesn't provide one but is passed in as an `article` anyway. This could also be remedied by making `layout` optional on `article` or adding `layout` to the `RelatedArticleFigure`.

kanaabe: Ah I see, yeah I'd prefer making layout optional then, because the query for RelatedArticleFigure on Force looks like [this](https://github.com/artsy/force/blob/master/src/desktop/apps/article/queries/relatedArticles.js) and we'd have to make sure those queries are updated if we wanted `layout` to work for RelatedArticleFigure. Probably not the most ideal solution, but once we get to a place where Reaction Publishing components don't need to pass down everything ie using a store/context then things will feel better.

kanaabe:
Looks great! I think after that one comment we can merge this. One thing I totally forgot about is our semantic versioning releases. When you make the next commit to update `layout`, please make your commit message like [this](https://github.com/artsy/reaction#example-minor-feature-release) to auto-publish a new package version to npm.

src/Components/Publishing/Byline/AuthorDate.tsx
```diff
@@ -3,50 +3,77 @@ import styled, { StyledFunction } from "styled-components"
...
       By {getAuthorByline(props.authors)}
     </StyledAuthor>
   )
 }
 
-export const Date: React.SFC<AuthorDateProps> = props => {
+export const Date: React.SFC<DateProps> = props => {
+  const condensed = props.layout === "condensed"
+  const format = props.format || (condensed ? "condensed" : "default")
+
```
eessex: Just out of curiosity, what formats are there besides condensed and default?

src/Components/Publishing/Byline/Share.tsx
```diff
@@ -120,6 +117,5 @@ const IconWrapper = styled.a`
   }
 `
 const ShareLabel = styled.span`
-  ${Fonts.unica("s16", "medium")}
-  margin: 10px 10px 10px 0px;
+  ${Fonts.unica("s16", "medium")} margin: 10px 10px 10px 0px;
 `
```
eessex: This is maybe not fixable because of prettier, but kind of weird to have these on one line.

src/Components/Publishing/Constants.ts
```diff
@@ -65,15 +68,36 @@ export const getAuthorByline = authors => {
   }
 }
 
-export const getDate = (date, layout = '') => {
-  switch (layout) {
-    case "monthYear": return moment(date).tz("America/New_York").format("MMMM YYYY")
-    case "condensed": return moment(date).tz("America/New_York").format("MMM D, YYYY")
-    default: return moment(date).tz("America/New_York").format("MMM D, YYYY h:mm a")
+export const getDate = (date, format: DateFormat = "default") => {
+  switch (format) {
```
eessex: I think this merits some unit tests

src/Components/Publishing/Byline/Share.tsx
kanaabe: It's fixable! Add a semicolon to the end of the interpolation: `${Fonts.unica("s16", "medium")};` I mentioned it in an earlier comment and said it's not necessary to address all of them in this PR.

src/Components/Publishing/Fixtures/Articles.ts
```diff
@@ -805,6 +805,7 @@ export const NewsArticle: ArticleData = {
     id: "50f4367051e7956b6e00045d",
     name: "Artsy Editorial",
   },
+  date: "2018-07-19T17:19:55.909Z",
   authors: [
```
eessex: Our api uses `published_at`, and `scheduled_publish_at` as date fields -- not sure if this is replicating return values from graphql, but might make sense to use a different field title for fixtures so it is more compatible with Positron.

src/Components/Publishing/RelatedArticles/RelatedArticleFigure.tsx
sarahscott: Sounds gut!

src/Components/Publishing/Byline/Share.tsx
sarahscott: I'll fix this one though 😄 and any other ones I catch.

src/Components/Publishing/Byline/__tests__/AuthorDate.test.tsx
```diff
@@ -1,31 +1,83 @@
...
+    let today = moment().tz("America/New_York")
+    today = today.set({
+      hour: 14,
+      minute: 22,
+    })
+
+    const date = getDate(today, "verbose")
+    expect(date).toEqual(expectedFormattedDates.verboseToday)
+  })
+})
```
kanaabe: Very nice set of tests, thanks!

PR #571, opened 2018-03-08
@sarahscott => Adds News Truncation

kanaabe:
Uses the familiar `isTruncated` prop to determine if an article is truncated. Once you click it stays expanded

- [x] Review Owen's CSS for exact hover state
- [x] Tests

On hover outline and click to expand:
news-truncation

### DIFF
--- a/src/Components/Publishing/Layouts/NewsLayout.tsx
+++ b/src/Components/Publishing/Layouts/NewsLayout.tsx
@@ -17,8 +17,12 @@ interface NewsContainerProps {
 }
 
 export class NewsLayout extends Component<Props, State> {
-  state = {
-    isTruncated: this.props.isTruncated || true,
+  constructor(props: Props) {
+    super(props)
+
+    this.state = {
+      isTruncated: this.props.isTruncated || true,
+    }
   }
 
   render() {
--- a/src/Components/Publishing/Layouts/NewsLayout.tsx
+++ b/src/Components/Publishing/Layouts/NewsLayout.tsx
@@ -22,7 +22,7 @@ export class NewsLayout extends Component<Props, State> {
     super(props)
 
     this.state = {
-      isTruncated: this.props.isTruncated || true,
+      isTruncated: this.props.isTruncated,
     }
   }
 
--- a/src/Components/Publishing/Layouts/NewsLayout.tsx
+++ b/src/Components/Publishing/Layouts/NewsLayout.tsx
@@ -5,25 +5,36 @@ import { NewsSections } from "../News/NewsSections"
 
 interface Props {
   article: any
-  expanded?: boolean
+  isTruncated?: boolean
 }
 
 interface State {
-  expanded: boolean
+  isTruncated: boolean
+}
+
+interface NewsContainerProps {
+  isTruncated: boolean
 }
 
 export class NewsLayout extends Component<Props, State> {
   state = {
-    expanded: this.props.expanded || true,
+    isTruncated: this.props.isTruncated || true,
   }
 
   render() {
     const { article } = this.props
+    const { isTruncated } = this.state
 
     return (
-      <NewsContainer>
+      <NewsContainer
+        onClick={() =>
+          this.setState({
+            isTruncated: false,
+          })}
+        isTruncated={isTruncated}
+      >
         <NewsHeadline article={article} />
-        <NewsSections article={article} />
+        <NewsSections article={article} isTruncated={isTruncated} />
       </NewsContainer>
     )
   }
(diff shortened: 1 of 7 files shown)


src/Components/Publishing/Layouts/NewsLayout.tsx
```diff
@@ -5,31 +5,51 @@ import { NewsSections } from "../News/NewsSections"
...
+}
+
+interface NewsContainerProps {
+  isTruncated: boolean
 }
 
 export class NewsLayout extends Component<Props, State> {
   state = {
-    expanded: this.props.expanded || true,
+    isTruncated: this.props.isTruncated || true,
```
alloy: I don’t know why this isn’t giving you a type error, but `this.props` is not what you expect it to be when assigning initial state statically, you have to do it from a `constructor` override instead.

Additionally, the assigned value is _always_ going to be `true`, because it’s essentially `(false || undefined) || true`. If the intention was to default to `true` than you’ll need to do `this.props.isTruncated == undefined ? true : this.props.isTruncated`

alloy: Oh, actually [I’m wrong](https://www.typescriptlang.org/play/#src=class%20X%20%7B%0D%0A%09props%3A%20%7B%20foo%3A%20string%20%7D%0D%0A%0D%0A%09constructor(props)%20%7B%0D%0A%09%09this.props%20%3D%20props%09%0D%0A%09%7D%0D%0A%7D%0D%0A%0D%0Aclass%20Y%20extends%20X%20%7B%0D%0A%09state%20%3D%20%7B%0D%0A%09%09foo%3A%20this.props.foo%0D%0A%09%7D%0D%0A%7D%0D%0A%0D%0Aconst%20y%20%3D%20new%20Y(%7B%20foo%3A%20%22ok%22%20%7D)%0D%0Aconsole.log(y.state.foo))!

Looking at the generated code you can see that the `this.props` static state assignment is moved into the constructor, so as long as the superclass’ constructor is being used (which is the case because you don’t override it) then `this.props` should be available.

Having said that, I find it very unintuitive because it’s inside an object literal and thus in the lexical scope I’d expect `this` to refer to the class (if such a thing were even possible with ES classes). I wonder if this is according to the spec or just a side-effect of the compiler implementing static instance variable assignment by moving the code inside the constructor.

alloy: Went a little deeper on this and asked for clarification on the proposal for class properties https://github.com/tc39/proposal-class-fields/issues/46#issuecomment-371440198.

damassi: Using `this` in this way (even if it is transpiled out into the proper form) very much strikes me as a code smell! It's also unintuitive because I don't think this kind of behavior is seen often in other languages due to lack of rewrite step. Why not set state in constructor and keep the control flow universal?

kanaabe: Ah yeah, this is not ideal I agree. Not sure why I did that - thanks for the catch!

src/Components/Publishing/Layouts/__tests__/NewsLayout.test.tsx
```diff
@@ -1,10 +1,27 @@
...
+  })
+
+  it("truncates the article", () => {
+    const component = mount(<NewsLayout article={NewsArticle} isTruncated />)
+    expect(component.find(NewsSectionContainer).length).toEqual(2)
+  })
+
+  it("expands the article on click", () => {
+    const component = mount(<NewsLayout article={NewsArticle} isTruncated />)
+    component.simulate("click")
```
sarahscott: sweet

src/Components/Publishing/News/NewsHeadline.tsx
```diff
@@ -34,9 +34,8 @@ const NewsHeadlineContainer = styled.div`
   text-align: left;
   max-width: 780px;
   width: 100%;
-  margin: 40px auto;
+  margin: 10px auto 30px auto;
```
sarahscott: Could this be shortened to `10px auto 30px` since the side margins are the same?

kanaabe:
Oh ha my bad @alloy  🤦‍♀️  thanks.

src/Components/Publishing/News/NewsHeadline.tsx
sarahscott: Sweet! I wasn't sure if I was right about that 😂

sarahscott:
👏🏼👏🏼👏🏼 Merging!

PR #572, opened 2018-03-09
@sarahscott => News Nav

kanaabe:
Fairly simple nav for News. Haven't added it to the Article because I think this should exist one level up, when we have an InfiniteScroll. Addresses https://artsyproduct.atlassian.net/browse/GROWTH-278

Desktop:
image

Mobile:
image


sarahscott:
👏

PR #574, opened 2018-03-14
@eessex => stricter section typings

sarahscott:
This doesn't affect functionality; it just makes it easier to keep section types in order 😊

### DIFF
--- a/src/Components/Publishing/Typings.ts
+++ b/src/Components/Publishing/Typings.ts
@@ -26,21 +26,32 @@ export type ArticleData = {
   postscript?: string
   date?: string
   published_at?: string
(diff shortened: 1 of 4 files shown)


src/Components/Publishing/Typings.ts
```diff
@@ -26,21 +26,32 @@ export type ArticleData = {
...
 }
+
+export type SectionType =
+  | "image_collection"
+  | "image_set"
+  | "embed"
+  | "social_embed"
+  | "text"
+  | "video"
+  | "default"
```
damassi: 👍

eessex:
Love this 👏

PR #575, opened 2018-03-14
@sarahscott => Handle non-valid links in SocialEmbed

kanaabe:
Handles the error we were seeing earlier -- we should not fetch anything if the url is null.


sarahscott:
🔥

PR #576, opened 2018-03-14
@kanaabe => Replace contributing_authors with authors

eessex:
Deprecates contributing_authors in Reaction.
Dependent on updated graphql fetch in Force.


src/Components/Publishing/Header/ClassicHeader.tsx
```diff
@@ -19,7 +19,7 @@ export const ClassicHeader: React.SFC<ClassicHeaderProps> = props => {
       <Title>{title}</Title>
       {leadParagraph}
       <AuthorDateClassic
-        authors={article.contributing_authors}
+        authors={article.authors}
```
kanaabe: We'll still need to support contributing_authors for Classic articles since the concept of `authors` only exists for Editorial.

eessex: 👍  will update this now

PR #577, opened 2018-03-15
@eessex => Editable source

kanaabe:
Makes the source of a news article editable. I also separated DateSource from NewsByline since there were two components in that file. Makes it a little easier to test and read.

### DIFF
--- a/src/Components/Publishing/Byline/DateSource.tsx
+++ b/src/Components/Publishing/Byline/DateSource.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { Fragment } from "react"
 import styled from "styled-components"
 import colors from "../../../Assets/Colors"
 import { pMedia } from "../../Helpers"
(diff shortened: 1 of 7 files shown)


src/Components/Publishing/Byline/DateSource.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { article, editSource } = props
+  const { news_source, published_at } = article
+
+  const getNewsSource = source => {
+    if (!source || !source.url) return null
+    return (
+      <Fragment>
+        {", via"}&nbsp;
+        {editSource ? editSource : <a href={source.url}>{source.title}</a>}
+      </Fragment>
```
eessex: Usually `editSouce` is a function, so would probably need to be called here

kanaabe: Hmm it's a component so I think it can be returned as is. To verify I looked at some other places we're doing the same thing and we don't call it. Am I missing something?

eessex: That might be an older version that uses children, here is a recent example:
https://github.com/artsy/reaction/blob/master/src/Components/Publishing/Sections/Image.tsx#L46

And this is the positron side:
https://github.com/artsy/positron/blob/master/client/apps/edit/components/content/sections/images/components/edit_image.jsx#L62

kanaabe: Ah the reason that one is a func is because we want to do a conditional in Positron -- that makes sense. But this recent example renders the component: https://github.com/artsy/reaction/blob/master/src/Components/Publishing/Series/SeriesTitle.tsx#L24

Is func a preference for you? It seems like there are varied usages and that we haven't established a standard.

eessex: Oh yes i see -- in that case this looks good!

PR #579, opened 2018-03-19
@kanaabe => Rethink content-end

eessex:
Content-end spans have always been problematic in Positron, so I've reworked Reaction's text component to insert them into html on the client side.  Dots will no longer appear in Positron, as this only applies to `dangerouslySetHtml` and not to editable children.

- Sections component identifies last text section if feature/standard layout
- Text section removes existing `content-end` spans (can be removed if we backfill them out of articles)
- Text section inserts `content-end` span for the last paragraph of html if `isContentEnd` prop is true

Will require some follow-up removing related logic from Positron, but does not block this PR.


src/Components/Publishing/Sections/Text.tsx
```diff
@@ -33,3 +43,22 @@ export const Text: React.SFC<TextProps> = props => {
 Text.defaultProps = {
   color: "black",
 }
+
+const htmlMaybeWithContentEnd = (html, isContentEnd) => {
+  const doc = document.createElement("div")
+  doc.innerHTML = html
+  // remove existing content-end spans, remove after backfill
```
kanaabe: Ah we should address this backfill. Want to put some time in the calendar so I can walk you through it?

eessex: Yes sounds great, would be glad for a walkthrough of backfilling sections!

eessex:
For the release, because this changes article fixtures, I thought it needed a version change -- do you think patch is more appropriate?

kanaabe:
Oh no, I think using `FEATURE` is appropriate. Just that the last two commits wouldn't be included in that release since they come after that commit. Or am I wrong?

eessex:
Anything that is merged to master should be included in the release, you can trigger from any pr -- and also add a second release commit within the same pr

kanaabe:
Can you elaborate on that behavior? Like in this scenario:
```
1 -> some edit
2 -> [FEATURE thing] add feature
3 -> oops forgot this thing [wip]
4 -> [PATCH thing] adds forgotten thing
5 -> ah forgot something else
```
Would commit #5 will be included in the patch or feature or none?

eessex:
I believe each PR has one version number attached to it on merge to master, and it is cumulative based on all triggering commits (assuming a triggering commit is included).

If starting from 0, the version number would be 0.1.1, and should include all commits, even if they are after.

kanaabe:
Hm interesting! I was thinking about it chronologically, not from a whole PR. Thank you for clarifying and sorry for the tangent but I feel more in the know now 👍

eessex:
There is a tab on the circle build for `yarn semantic-release` -- you can follow the process by line there. It looks at the commit 'merge to master' as the final info to be included in each release, and generates version numbers based on what comes before it.

PR #580, opened 2018-03-20
@eessex => nowrap for single artist names; wrap for multiple

sarahscott:
Before:

screen shot 2018-03-20 at 1 09 53 pm

After:
screen shot 2018-03-21 at 12 28 20 pm

Fixes https://artsyproduct.atlassian.net/browse/GROWTH-254

### DIFF
--- a/src/Components/Publishing/Sections/ArtworkCaption.tsx
+++ b/src/Components/Publishing/Sections/ArtworkCaption.tsx
@@ -21,6 +21,10 @@ interface StyledArtworkCaptionProps extends React.HTMLProps<HTMLDivElement> {
   sectionLayout?: SectionLayout
 }
 
+interface ArtistNamesProps extends React.HTMLProps<HTMLDivElement> {
+  multipleNames?: boolean
+}
+
 export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
   joinParts(children, delimiter = ", ") {
     const compacted = _.compact(children)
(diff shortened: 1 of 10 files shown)


src/Components/Publishing/Sections/ArtworkCaption.tsx
```diff
@@ -39,7 +43,7 @@ export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
     const { artwork: { artist, artists } } = this.props
 
     // Multiple artists
-    if (artists && artists.length > 0) {
```
sarahscott: The `artists` array contains an `artist` even if there is just one.

sarahscott:
Actually it might make sense to do 'nowrap' on the individual artist names instead of the whole string.. will try that now.

src/Components/Publishing/Sections/ArtworkCaption.tsx
eessex: This conditional could be a const, something list `hasMultipleArtists` -- noticed you use it again on 207

src/Components/Publishing/Sections/ArtworkCaption.tsx
```diff
@@ -196,11 +204,11 @@ export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
           sectionLayout={sectionLayout}
           className="display-artwork__caption"
         >
-          <ArtistName>{this.renderArtists()}</ArtistName>
-
+          <ArtistNames multipleNames={artists && artists.length > 1}>
+            {this.renderArtists()}
```
eessex: Related to above, names is not super-specific, I think `hasMultipleArtists` is more explicit about the expectations

sarahscott:
I figured out a way to make the individual names `nowrap` instead of all of the names together and I think it looks better during resizing (bc the whole name will go to the next line instead of splitting up last/first names).

PR #583, opened 2018-03-21
@kanaabe => Content end refactor without 'document'

eessex:
Reworks[ reverted previous PR](https://github.com/artsy/reaction/pull/582), and updates `content-end` span insertion to happen on `componentDidMount` to avoid conflicts with SSR.


kanaabe:
Oo I'm curious about your choice to intro a new lib as opposed to the other options that were mentioned in #front-end. It doesn't seem critical that this gets rendered server-side to me. The option of moving it to `componentDidMount` is used in some places already like in [this method](https://github.com/artsy/reaction/blob/d3fc20b6b8cb7b8c6777b6f8d94c783b131f887d/src/Components/Publishing/Sections/Sections.tsx#L56) where we access `document`

eessex:
The component did mount option seems like a good one as well -- tried looking for document to exist, as a link @l2succes sent suggested but wasn't able to make it to work for me. JSDOM would have also required another library.  Have found `underscore.string` useful in positron and is also lightweight, using `document` wasn't fully necessary to achieve the end goal.  Open to go back to the other route if you think it would be better.

kanaabe:
Yeah I'm leaning towards using the pattern we already have that doesn't require a new dependency.

eessex:
Cool, this has been updated to use document again.

kanaabe:
Nice! This LGTM but you prob want to ammend a commit to include the release message! Feel free to merge after that.

eessex:
OOh yes good point!

PR #586, opened 2018-03-22
@damassi => Skip images-loaded fade if editing

eessex:
Adds an optional boolean prop `editing` to images and artworks, which disables the fade in on loading. 

Once we switched to Redux in Positron, changes to the article can trigger re-rendering in the tree, which causes erratic fade-ins that can confuse users.

### DIFF
--- a/src/Components/Publishing/Sections/ImageWrapper.tsx
+++ b/src/Components/Publishing/Sections/ImageWrapper.tsx
@@ -12,6 +12,7 @@ interface Props extends React.HTMLProps<HTMLImageElement> {
   width?: string | number
   height?: string | number
   alt?: string
+  editing?: boolean
   index?: number
 }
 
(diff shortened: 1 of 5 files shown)


src/Components/Publishing/Sections/ImageWrapper.tsx
```diff
@@ -55,10 +56,16 @@ export class ImageWrapper extends React.PureComponent<Props, any> {
   }
 
   render() {
-    const { layout, linked, index, ...blockImageProps }: any = this.props
+    const {
+      layout,
+      linked,
+      editing,
+      index,
+      ...blockImageProps,
```
eessex: @damassi I'm having an issue where tslint keeps adding a trailing comma after `...blockImageProps`, but commas are not allowed for the spread. 

I tried adding `tslint:disable` to the top of the file, but it still gets added.  Any ideas of what rules I could look at to prevent this?

damassi: Hmm, no clue... I would dig around online!

eessex: Yeah is definitely coming from prettier, stops if I disable for this file.  Supposedly adding `--parser typescript` should stop adding it, but haven't managed to make it work.

PR #588, opened 2018-03-23
[FIX onboarding] Fixed a bug where the NEXT button doesn't show up on Edge

yuki24:
I believe [the original issue](https://artsyproduct.atlassian.net/browse/GROWTH-361) I was looking into was fixed by https://github.com/artsy/force/pull/2291, but even with that fix there are many other issues in MS Edge or IE 11, and this PR fixes one of them.

The issue this PR addresses is that on the onboarding screen the NEXT button does not show up. This PR removes the `position: sticky;` property that seems to fix the issue. I was worried about mobile safari not respecting `position: fixed`. I tested it on iOS 9 and 11, and apparently that is no longer an issue. See the screenshots below:

## Before

edge-next-button

## After

screen shot 2018-03-23 at 2 56 18 pm screen shot 2018-03-23 at 2 56 15 pm

screen shot 2018-03-23 at 2 56 51 pm

### DIFF
--- a/src/Components/Onboarding/Steps/Layout.tsx
+++ b/src/Components/Onboarding/Steps/Layout.tsx
@@ -51,20 +51,11 @@ const ItemContainer = styled.div`
   padding-bottom: 50px;
 `
 
-/* MS IE11 and Edge don't support for the sticky position property */
-const FixedButttonContainer = styled.div`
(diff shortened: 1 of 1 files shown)


xtina-starr:
LGTM!

PR #590, opened 2018-03-26
@l2succes => Authorization directory skeleton + authform

erikdstock:
This PR creates an `Authorization` components directory and story for the various auth forms and states. It's mainly intended to give a place for us to work off. I'm not sure if some of my approaches (eg commonElements.ts) are standard.

I was thinking that our desktop components would be `Form`s and our mobile components would be `Wizards`, ie `LoginWizard.tsx`- based on my understanding that we use wizards elsewhere and they could fit anywhere.

current structure:
```
src/Components/Authorization/
├── AuthForm.tsx
├── ForgotPasswordForm.tsx
├── LoginForm.tsx
├── RegisterForm.tsx
├── Types.ts
└── commonElements.ts
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,6 +122,7 @@
     "webpack-merge": "^4.1.0"
   },
   "dependencies": {
+    "formik": "^0.11.11",
     "history": "^4.6.1",
     "isomorphic-fetch": "^2.2.1",
     "jsonp": "^0.2.1",
@@ -145,7 +146,8 @@
     "react-waypoint": "^7.3.3",
     "relay-runtime": "https://github.com/alloy/relay/releases/download/v1.5.0-artsy.3/relay-runtime-1.5.0-artsy.3.tgz",
     "sharify": "^0.1.6",
-    "superagent": "^3.6.3"
+    "superagent": "^3.6.3",
+    "yup": "^0.24.1"
   },
   "prettier": {
     "semi": false,
--- a/src/Components/Authorization/AuthForm.tsx
+++ b/src/Components/Authorization/AuthForm.tsx
@@ -1,0 +1,299 @@
+import React from "react"
+import styled from "styled-components"
+// import colors from "../../Assets/Colors"
+
+import FacebookButton from "../Buttons/Facebook"
+import InvertedButton from "../Buttons/Inverted"
+import TwitterButton from "../Buttons/Twitter"
+import Input from "../Input"
+
+type Mode = "log_in" | "register" | "forgot_password"
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  mode: Mode
+  signupIntent?: string
+  destination?: string
+  formFields?: FormFields
+}
+
+interface State extends React.HTMLProps<HTMLDivElement> {
+  mode?: Mode
+  formFields: FormFields
+}
+
+class AuthForm extends React.Component<Props, State> {
+  static defaultProps: Partial<Props> = {
+    signupIntent: "sign_up",
+    destination: "/personalize", // ?
+    formFields: {
+      email: { value: "", error: "" },
+      name: { value: "", error: "" },
+      password: { value: "", error: "" },
+    },
+  }
+
+  constructor(props: Props) {
+    super(props)
+    this.state = {
+      mode: this.props.mode,
+      formFields: this.props.formFields,
+    }
+    this.handleUpdateInput = this.handleUpdateInput.bind(this)
+    this.handleChangeMode = this.handleChangeMode.bind(this)
+    this.handleSubmit = this.handleSubmit.bind(this)
+  }
+
+  validateField = (fieldName: string, value: string) => {
+    return {
+      name: val => (val.length > 0 ? "" : "Please enter your name"),
+      email: val => (val.indexOf("@") > -1 ? "" : "Please enter your email"),
+      password: val =>
+        val.length > 6 ? "" : "Password must be 6 characters long",
+    }[fieldName](value)
+  }
+
+  handleUpdateInput(fieldName: string) {
+    return event => {
+      console.log("here")
+      const { value } = event.target
+      const error = this.validateField(fieldName, value)
+      this.setState((prevState, props) => {
+        const { formFields, ...rest } = prevState
+        const newFields = { ...formFields, [fieldName]: { value, error } }
+        const newState = { ...rest, formFields: newFields }
+        console.log(newState)
+        return newState
+      })
+    }
+  }
+
+  handleChangeMode(newMode: Mode) {
+    return event => {
+      event.preventDefault()
+      this.setState({ mode: newMode })
+    }
+  }
+
+  handleSubmit(e: React.FormEvent<HTMLFormElement>) {}
+
+  render() {
+    const { formFields } = this.state
+    let Form: FormComponentType
+    switch (this.state.mode) {
+      case "log_in":
+        Form = LoginForm
+        break
+      case "register":
+        Form = RegisterForm
+        break
+      case "forgot_password":
+        Form = ForgotPasswordForm
+        break
+      default:
+        throw new Error(`${this.state.mode} mode needs a component`)
+    }
+    return (
+      <div>
+        <Form
+          {...formFields}
+          handleUpdateInput={this.handleUpdateInput}
+          handleChangeMode={this.handleChangeMode}
+          handleSubmit={this.handleSubmit}
+        />
+      </div>
+    )
+  }
+}
+
+export default AuthForm
+
+interface FormField {
+  value: string
+  error: string
+}
+
+interface FormFields {
+  name?: FormField
+  email?: FormField
+  password?: FormField
+}
+
+interface FormProps extends FormFields {
+  handleSubmit: (e: React.FormEvent<HTMLFormElement>) => void
+  handleUpdateInput: (
+    key: string
+  ) => (event: React.FocusEvent<HTMLInputElement>) => void
+  handleChangeMode: (mode: Mode) => (event) => void
+}
+
+type FormComponentType = React.SFC<FormProps> //only refers to a type, but is being used as a value here
+
+/** class components can't extend type like below but I don't think I need a class so...
+type FormComponentType = React.Component<FormProps, any> //only refers to a type, but is being used as a value here
+class LoginForm extends React.Component<FormProps, any> {
+  // public emailInput: React.ReactHTMLElement
+  static defaultProps: Partial<FormProps> = {}
+  constructor(props) {
+    super(props)
+  }
+
+  render() {
+    const {
+      email,
+      password,
+      handleUpdateInput,
+      handleChangeMode,
+      handleSubmit,
+     } = this.props
+*/
+const LoginForm: FormComponentType = props => {
+  const {
+    email,
+    password,
+    handleChangeMode,
+    handleSubmit,
+    handleUpdateInput,
+  } = props
+  return (
+    <FormContainer onSubmit={handleSubmit}>
+      <StyledFacebookButton>Log in with Facebook</StyledFacebookButton>
+      <StyledTwitterButton />
+      <StyledInput
+        block
+        value={email.value}
+        placeholder="Email"
+        onChange={handleUpdateInput("email")}
+        error={email.error.length > 0}
+        errorMessage={email.error}
+      />
+      <StyledInput
+        block
+        value={password.value}
+        type="password"
+        placeholder="Password"
+        onChange={handleUpdateInput("password")}
+        error={password.error.length > 0}
+        errorMessage={password.error}
+      />
+      <p>
+        Uh oh I{" "}
+        <a onClick={handleChangeMode("forgot_password")} href="#">
+          Forgot My Password
+        </a>
+      </p>
+      <InvertedButton>Log In</InvertedButton>
+      <p>
+        Don't have an account?{" "}
+        <a onClick={handleChangeMode("register")} href="#">
+          Sign Up
+        </a>
+      </p>
+    </FormContainer>
+  )
+}
+
+const RegisterForm: FormComponentType = props => {
+  const { email, name, password } = props
+  return (
+    <FormContainer onSubmit={props.handleSubmit}>
+      <StyledInput
+        block
+        value={name.value}
+        placeholder="Name"
+        error={name.error.length > 0} // will cause field to turn red
+        onChange={props.handleUpdateInput("name")}
+        errorMessage={name.error} // errorMessage will cause input to fail on form submission
+      />
+      <StyledInput
+        block
+        value={email.value}
+        placeholder="Email"
+        error={email.error.length > 0}
+        onChange={props.handleUpdateInput("email")}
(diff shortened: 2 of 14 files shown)


src/Components/__stories__/Input.story.tsx
```diff
@@ -5,41 +5,58 @@ import colors from "../../Assets/Colors"
...
       <Input type="password" placeholder="Password" block />
       <Button block>Submit</Button>
     </div>
-  )
+  ))
+// .add("With Custom Validation", () => {
+//   // const validateOnBlur: (event: React.FocusEvent<HTMLInputElement>) => void = (
+//   //   e => e.target.value === 'cow' ?
+//   // )
+//   return (
```
erikdstock: will revert this file as well

src/Components/Authorization/AuthForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import ForgotPasswordForm from "./ForgotPasswordForm"
+import LoginForm from "./LoginForm"
+import RegisterForm from "./RegisterForm"
+import { FormComponentType, FormFields, Mode } from "./Types"
+
+type Validator = (val: string) => string
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  mode: Mode
```
yuki24: I'd prefer to use a more generalized router rather than implementing our own router. Have you looked into react-router's [`<HashRouter>`](https://reacttraining.com/react-router/web/api/HashRouter) or [`<MemoryRouter>`](https://reacttraining.com/react-router/web/api/MemoryRouter)?

src/Components/Authorization/AuthForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+    this.handleChangeMode = this.handleChangeMode.bind(this)
+    this.handleSubmit = this.handleSubmit.bind(this)
+  }
+
+  validators = {
+    name: val => (val.length > 0 ? "" : "Please enter your name"),
+    email: val => (val.indexOf("@") > -1 ? "" : "Please enter your email"),
+    password: val =>
+      val.length > 6 ? "" : "Password must be 6 characters long",
+  }
```
yuki24: It is common that a single input field has multiple validation rules. We should probably make this a collection of validation objects so it can show different error messages depending on the user input.

src/Components/Authorization/commonElements.ts
```diff
@@ -0,0 +1,31 @@
...
+export const StyledFacebookButton = styled(FacebookButton)`
+  width: ${buttonWidth};
+  background: #4e65b1;
+`
+export const StyledTwitterButton = styled(TwitterButton)`
+  width: ${buttonWidth};
+`
+export const StyledButton = styled(InvertedButton)`
+  width: ${buttonWidth};
+`
```
yuki24: You can probably use [the `<Button block>` component](https://github.com/artsy/reaction/blob/fa102eb655e3fd1165941bd1d6c579bc18945fd7/src/Components/__stories__/Buttons.story.tsx#L55) to make the button fill out the whole width.

src/Components/__stories__/AuthForm.story.tsx
```diff
@@ -0,0 +1,29 @@
...
+//             ...Artwork_artwork
+//           }
+//         }
+//       `}
+//       variables={{ artworkID: props.artworkID }}
+//       render={readyState =>
+//         readyState.props && <Artwork {...readyState.props as any} />}
+//     />
+//   )
+// }
```
yuki24: I wouldn't add the actual submission logic to reaction. There are so many things that can't be taken care of on reaction - how to launch a modal from old-school plain JS or coffee script, possible style conflict, where to redirect after submission, etc. I think reaction should stay as a style guide like the Bootstrap that allows us to quickly build up a UI with low-level components _on Force_.

WDYT? @l2succes @xtina-starr @eessex

src/Components/Authorization/AuthForm.tsx
damassi: There are a few examples of React Router in Force, too, notably in the /Consignments app

src/Components/Authorization/AuthForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  constructor(props: Props) {
+    super(props)
+    this.state = {
+      mode: this.props.mode,
+      formFields: this.props.formFields,
+    }
+    this.handleUpdateInput = this.handleUpdateInput.bind(this)
+    this.handleChangeMode = this.handleChangeMode.bind(this)
+    this.handleSubmit = this.handleSubmit.bind(this)
```
damassi: Can use => instead of binding in this way

```
handleUpdateInput = (event) => { ... } 
```

Fat arrow binds to the outer scope (`this`)

src/Components/Authorization/AuthForm.tsx
damassi: @yuki24 - Are we using any React form libs anywhere in Reaction? 

@anandaroop - What are ya'll using in Volt? I remember seeing it a while back but might be worthwhile introducing here rather than rolling own validation.

yuki24: @damassi No, everything is a plain React input/form. Do you think it is something we should look into now?

src/Components/Authorization/AuthForm.tsx
erikdstock: wow, it never occurred to me that this is something that would be _routed_ in that way.

l2succes: Good point, react router would make sense in terms of switching between various forms

src/Components/Authorization/AuthForm.tsx
damassi: Def -- IMO there's no reason to roll your own form stuff these days with such nice / simple react libraries around that reduce so much complexity. I _think_ this is what Roop et al used: https://github.com/jaredpalmer/formik

anandaroop: I can recommend [Formik](https://github.com/jaredpalmer/formik) as a solution to form handling & validation in React.

It's got a bit of learning curve, but much less than anything involving Redux-ifying your forms.

src/Components/Authorization/AuthForm.tsx
```diff
@@ -0,0 +1,109 @@
+import React from "react"
+
+import ForgotPasswordForm from "./ForgotPasswordForm"
+import LoginForm from "./LoginForm"
+import RegisterForm from "./RegisterForm"
+import { FormComponentType, FormFields, Mode } from "./Types"
+
+type Validator = (val: string) => string
```
l2succes: 👍Still thinking through the actual API but depending on the experience it may make sense to support multiple return value types as such 
`(val: string) => string | Promise<string>`

that way we can also support both sync & async validation. A use case is validating emails or looking up if the account already exists

src/Components/Authorization/AuthForm.tsx
erikdstock: 🙄 thanks!

src/Components/Authorization/AuthForm.tsx
erikdstock: Oh woops- I was trying to find the point in react history where people started using that style, and in the process was writing it both ways. I guess I kept it.

xtina-starr: Shout out to public class fields syntax.

src/Components/Authorization/AuthForm.tsx
```diff
@@ -0,0 +1,109 @@
+import React from "react"
+
+import ForgotPasswordForm from "./ForgotPasswordForm"
+import LoginForm from "./LoginForm"
+import RegisterForm from "./RegisterForm"
+import { FormComponentType, FormFields, Mode } from "./Types"
+
+type InputValidator = (val: string) => string
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
```
xtina-starr: Is this going to be the form itself? Wondering if the props here should extends `<HTMLFormElement>` vs `<HTMLDivElement>`?

src/Components/Authorization/AuthForm.tsx
xtina-starr: I get a little confused about this too @erikdstock the [react docs](https://reactjs.org/docs/handling-events.html) for handling events explains pretty well how and why you can use this syntax.

src/Components/Authorization/AuthForm.tsx
xtina-starr: Just noticed you are pulling in some of the form types. Maybe you can extend the props there.

src/Components/__stories__/AuthForm.story.tsx
l2succes: I would agree but by using Relay containers and colocating data requirements with components, we've already made reaction more than just a UI library, it handles the data/model layer as well. 

So IMO, encapsulating submission logic in reaction is fine. However, I do get the argument to handle data submission and redirecting in Force. Basically I think either is fine, as long as we don't have multiple implementations of site-wide login and sign up forms lying around.

src/Components/Authorization/AuthForm.tsx
erikdstock: Ah yes, good point.

erikdstock:
Thanks for the input all. I've started on a stab at making a Formik form and it seems pretty nice so far, but would mean some major top-level changes to the AuthForm component here. I'll update tomorrow morning.

erikdstock:
Thanks for the input all. I've started on a stab at making a Formik form and it seems pretty nice so far, but would mean some major top-level changes to the AuthForm component here. I'll update tomorrow morning. Oh... I forgot to submit this.

erikdstock:
I've updated the form to use formik- setup is very easy, though I hit some typescript snags that don't stop the forms from working. I think the most efficient way to approach these would be for me to not do these alone.

I think the multi-step wizard forms will require a bit of cleverness- but this wasn't bad at all.

src/Components/Authorization/AuthForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface State extends React.HTMLProps<HTMLFormElement> {
+  mode?: Mode
+}
+
+class AuthForm extends React.Component<Props, State> {
+  static defaultProps: Partial<Props> = {
+    errors: {},
+    values: {},
+    handleSubmit: () => alert("submitting!"),
```
erikdstock: for now, this is just to give some feedback when testing the form in storybook. I guess I could have put it in storybook.

src/Components/Authorization/LoginForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const LoginForm: FormComponentType = props => {
+  const { email: emailValidator } = inputValidators
+  return (
+    <Formik
+      initialValues={props.values}
+      onSubmit={props.handleSubmit}
+      validationSchema={Yup.object().shape({
+        email: emailValidator,
+        password: Yup.string().required("Please enter your password."),
```
erikdstock: I did not reuse the password validator since this ticket raises the minimum password length from 6 to 8, and I figured it is best not to impose restrictions on what might be a password.

damassi:
If you're hitting typescript snags then is it possible to remove the typings for now? Those can be added in another PR perhaps?

Regarding React Router, it's best to think of it as URL driven development, where the state is encoded in the URL bar. Since this is going to be a new global login form with multiple states we want to be able to deep-link in where necessary, or at least enable that kind o functionality in the background through RR.

src/Components/Authorization/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,58 @@
+import { Formik } from "formik"
+import React from "react"
+import Yup from "yup"
```
damassi: What is `yup`?

src/Components/Authorization/RegisterForm.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <p>
+              Already have an account?
+              <ChangeMode handleClick={props.handleChangeMode("log_in")}>
+                Log In
+              </ChangeMode>
+            </p>
+          </FormContainer>
+        )
+      }}
+    </Formik>
```
damassi: Form stuff looks great 👍

src/Components/Authorization/LoginForm.tsx
damassi: @mzikherman - do we have any password requirements on the gravity side?

anandaroop:
Wow, you made quick work of that. Nice job integrating Yup and Formik here 👏

l2succes:
Nice work @erikdstock! Let's go ahead and merge as it's a good base to work from. We can iterate on introducing react router in a separate PR. Happy to pair on that.

src/Components/Authorization/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <ChangeMode handleClick={props.handleChangeMode("log_in")}>
+              Log In
+            </ChangeMode>
+          </Form>
+        )
+      }}
+    </Formik>
+  )
+}
+export default LoginForm
```
damassi: We should avoid using `export default` because then we loose some static analysis. Also, best to name the export after the filename: `export const ForgotPasswordForm`

src/Components/Authorization/LoginForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+  BlockButton as Button,
+  ChangeMode,
+  FormContainer as Form,
+  inputValidators,
+  StyledFacebookButton,
+  StyledInput as Input,
+} from "./commonElements"
+import { FormComponentType } from "./Types"
+
+const LoginForm: FormComponentType = props => {
```
damassi: `export const LoginForm`

src/Components/Authorization/commonElements.tsx
```diff
@@ -0,0 +1,67 @@
...
+  display: flex;
+  flex-direction: column;
+  width: 400px;
+  height: 500px;
+  padding: 15px 20px;
+`
+
+const buttonWidth = "100%"
+
+export const StyledFacebookButton = styled(FacebookButton)`
```
damassi: In [the docs](https://www.styled-components.com/docs/basics#extending-styles) they recommend using `.extend` (`FacebookButton.extend`) over the `styled()` HOC.

src/Components/Authorization/AuthForm.tsx
damassi: (Here where the plugin is located: https://github.com/artsy/reaction/blob/master/.babelrc#L13)

package.json
```diff
@@ -145,7 +146,8 @@
     "react-waypoint": "^7.3.3",
     "relay-runtime": "https://github.com/alloy/relay/releases/download/v1.5.0-artsy.3/relay-runtime-1.5.0-artsy.3.tgz",
     "sharify": "^0.1.6",
-    "superagent": "^3.6.3"
+    "superagent": "^3.6.3",
+    "yup": "^0.24.1"
```
yuki24: Yup 👍

PR #592, opened 2018-03-27
Adds label to Input component

l2succes:
This PR adds support for labels on our Input component by setting a `label` prop.

screen shot 2018-03-26 at 7 03 36 pm

- [x] Add fade in/out animation to label

### DIFF
--- a/src/Assets/Animations.tsx
+++ b/src/Assets/Animations.tsx
@@ -1,25 +1,21 @@
-import { css, keyframes } from "styled-components"
+import { keyframes } from "styled-components"
 
-export const fadeOut = css`
-  animation: ${keyframes`
-    from {
-      opacity: 1;
-    }
+export const fadeIn = keyframes`
+  from {
+    opacity: 0;
+  }
 
-    to {
-      opacity: 0;
-    }
-  `} 0.5s both;
+  to {
+    opacity: 1;
+  }
 `
 
-export const fadeIn = css`
-  animation: ${keyframes`
-    from {
-      opacity: 0;
-    }
+export const fadeOut = keyframes`
+  from {
+    opacity: 1;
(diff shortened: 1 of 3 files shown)


src/Components/Input.tsx
```diff
@@ -75,51 +91,32 @@ class Input extends React.Component<InputProps, InputState> {
     }
   }
 
-  render() {
-    if (this.props.leftView && this.props.rightView) {
-      const { className, ref, ...newProps } = this.props
+  onChange(e: React.FocusEvent<HTMLInputElement>) {
```
xtina-starr: perhaps `onChange = () => {...}` here to bind the callback? idk if it's possible to overuse this but this is one pattern I'd like to see us consistently use in reaction.

l2succes: Yeah, unfortunately it's incompatible with the type (`React.FocusEvent<HTMLInputElement>`) but I think that's ok

src/Assets/Animations.tsx
```diff
@@ -1,25 +1,21 @@
...
+    opacity: 1;
+  }
 
-    to {
-      opacity: 1;
-    }
-  `} 0.5s both;
+  to {
+    opacity: 0;
+  }
```
yuki24: This animation stuff is a bit out-dated. Feel free to drop it entirely and add whatever makes more sense!

l2succes: Yeah I thought about it but I ended  updated it since it worked pretty well for the fadeIn/Out animation I needed to implement instead of using `react-addons-css-transition-group`

erikdstock:
When this is ready to go, I'll use this in my existing popover/tooltip errors! Very cool.

PR #593, opened 2018-03-27
@kanaabe => Fix in editSource of NewsByline

sarahscott:
This makes sure `editSource` can still render if `news_source` is null.


kanaabe:
👍 Reminder to change commit message with `[FIX news] Fix in editSource` so it can be versioned!

kanaabe:
Coool, thank you! Feel free to merge on 🍏

sarahscott:
😄 ok now i am getting the hang of the semantic versioning

PR #595, opened 2018-03-27
@sarahscott => [News] Updates on NewsLayout + Expand Button

kanaabe:
Adds in the Expand button and brings the layout up to date with latest designs.

updated-expand


src/Components/Publishing/Layouts/NewsLayout.tsx
```diff
@@ -25,45 +30,103 @@ export class NewsLayout extends Component<Props, State> {
...
+        onMouseEnter={() => {
+          if (!isMobile) {
+            this.setState({ isHovered: true })
+          }
+        }}
+        onMouseLeave={() => {
+          if (!isMobile) {
+            this.setState({ isHovered: false })
+          }
+        }}
```
kanaabe: Graduated from using simple CSS `hover` properties to this since we'll need a little more custom behavior to handle mobile that uses viewport to set the hover state: 
image

src/Components/Publishing/Layouts/NewsLayout.tsx
```diff
@@ -25,45 +30,103 @@ export class NewsLayout extends Component<Props, State> {
...
+    const { article, isMobile } = this.props
+    const { isTruncated, isHovered } = this.state
 
     return (
       <NewsContainer
-        onClick={() =>
-          this.setState({
-            isTruncated: false,
-          })}
         isTruncated={isTruncated}
```
eessex: Is it intended that the entire box does not expand on click in desktop?

eessex: IMO would be more user friendly to keep in fullscreen, and use an isMobile flag like below.

src/Components/Publishing/Layouts/NewsLayout.tsx
```diff
@@ -9,14 +11,17 @@ interface Props {
...
+  isHovered?: boolean
 }
 
 interface State {
   isTruncated: boolean
+  isHovered: boolean
 }
 
 interface NewsContainerProps {
   isTruncated: boolean
```
sarahscott: In Emission we call this state `collapsed`/`expanded` and reserve `truncated` for text, but that's ofc up to you!

src/Components/Publishing/Layouts/NewsLayout.tsx
sarahscott: 🔥

sarahscott:
Looks good to merge!

PR #599, opened 2018-03-28
@sarahscott => Add marginTop, onExpand to NewsLayout

eessex:
- Adds `marginTop` optional prop to NewsLayout, used on first article in a list to avoid menu overlap
- Adds `onExpand` optional prop to NewsLayout, called on click, used to change `window.location` in Force


kanaabe:
LGTM!

PR #602, opened 2018-03-29
@yuki24 => Error Popovers

erikdstock:
cc @l2succes @xtina-starr @yuki24 
This component requires props:
- `show`
- `message` (string or function returning jsx)
- `anchorRef` - the relevant component (eg a bad `Input`)

With optional props:
- `dismissOnClick` - defaults to true
- `minWidth` - string, allows longer text to wrap lines and fill a larger box.

Things I considered but did not implement:
- `maxWidth` - to make messages fit more intelligently than `minWidth` can (eg, wrapping text automatically at a certain point)
- `dismissOnTimeout` - `int` for an automatically dismissing message

Caveats:
- `span` tags inside the rendered message get a pointer-style cursor. `onClick` handlers on internal message elements can stop propogation if necessary to prevent the tip from being hidden.
- _Animations don't seem to work_ - I think I missed something there.
- If the `show` prop is passed immediately on first render as `true` things seem to get confused around the tip's placement, which depends on a ref to anchor to. In retrospect I could probably have just factored that into whether to ultimately render the tip or not.


erikdstock:
Thanks! There is a little more styling to do, and for naming there was already a much different `ToolTip` so I wasn't sure what to do so picked different and top-level for now. I'd like to change it though. For the original `ToolTip` see the Market Insights story:
image

erikdstock:
Here is where I'm at. Very close, but the main black rectangle is always *one update behind its actual width*:
tooltip
(I also renamed it `Popover`, prepared to change that though).

src/Components/Popover.tsx
```diff
@@ -0,0 +1,154 @@
...
+    hide: true,
+    maxWidth: "200px",
+    dismissOnClick: true,
+    // dismissTimeout: null
+  }
+
+  constructor(props) {
+    super(props)
+    const message = this.getMessage(props)
+    const hide = props.hide === true || !message
```
damassi: For consistency sake, maybe 
```
const anchorPosition = this.getAnchorPosition(props.anchorRef),
```
which is then passed in similarly like the above two vars?

src/Components/Popover.tsx
```diff
@@ -0,0 +1,154 @@
...
+      const el = ReactDOM.findDOMNode(this.messageRef)
+      return el.getBoundingClientRect().width || fallback
+    }
+    return fallback
+  }
+
+  getMessage = props =>
+    (typeof props.children === "function"
+      ? props.children()
+      : props.children) || ""
```
damassi: This is a bit odd; what is going on here and how is it used in context? I would also recommend always using `{ }` for functions, and include `return` since implicit return gets future-buggy and somewhat clever when its spread across more than one line.

src/Components/__stories__/TooltipError.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const Container = styled.div`
+  padding: 50px;
+`
+
+class ErrorShower extends React.Component<any, any> {
+  constructor(props) {
+    super(props)
+    this.state = { count: 0, errorMessage: "" }
+  }
```
damassi: No need for a constructor if not setting state from props; can simplify to 

```
state = {
  count: 0,
  errorMessage: ""
}
```

src/Components/__stories__/TooltipError.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  showTip = () => {
+    const count = this.state.count
+    this.setState({
+      errorMessage: this.props.message + ` (${count})`,
+      count: count + 1,
+    })
```
damassi: Does this mean the error message and the count state are now out of sync? Should it be:

```javascript
const count = this.state.count + 1

this.setState({
  errorMessage: this.props.message + ` (${count})`,
  count
})
```

src/Components/__stories__/TooltipError.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+`
+class InputWithError extends React.Component<any, any> {
+  private inputRef
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      errorMessage: "",
+    }
+  }
```
damassi: No need for a constructor, can use classProperties to set state directly

src/Components/__stories__/TooltipError.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <p>
+          Just some unassuming content below the button. Just some unassuming
+          content below the button.
+        </p>
+      </Container>
+    )
+  }
+}
+
+const NoMarginInput = styled(Input)`
```
damassi: `Input.extend`

src/Components/Popover.tsx
```diff
@@ -0,0 +1,154 @@
...
+    const hide = !newMessage || newMessage === this.state.message
+    this.setState({
+      hide,
+      message: newMessage,
+      anchorPosition: this.getAnchorPosition(nextProps.anchorRef),
+      messageWidth: this.getMessageWidth(),
+    })
+  }
+
+  getAnchorPosition: (ref: any) => PositionProps = ref => {
```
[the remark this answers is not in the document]
erikdstock: I think this is a misunderstanding- `(ref: any) => PositionProps` is the type of the method itself?

src/Components/Popover.tsx
erikdstock: :+1: maybe not necessary anyways since i'm not currently passing any functions to my children. I think prettier made it look like this though.

src/Components/__stories__/TooltipError.story.tsx
erikdstock: I didn't think about this that much- it was my first story and I needed something to make sure the text was changing. I'll think about this.

src/Components/Popover.tsx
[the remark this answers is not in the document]
damassi: Oh weird! I see -- i wasn't aware that syntax was possible; what happens when you wrap that argument in parens? Does prettier strip it out?  

```
getAnchorPosition: ((ref: any) => PositionProps) = ref => { ... }
```

src/Components/Popover.tsx
```diff
@@ -0,0 +1,161 @@
...
+    window.removeEventListener("scroll", this.handleScroll)
+  }
+
+  handleScroll = () => {
+    const anchorPosition = this.getAnchorPosition(this.props.anchorRef)
+    this.setState({ anchorPosition })
+  }
+
+  getMessage = message =>
+    (typeof message === "function" ? message() : message) || ""
```
damassi: Can use lodash util (which is already in reaction) for greater readability

```javascript 
import { isFunction } from 'lodash'`

...

getMessage = (message = '') => isFunction(message) ? message() : message
```

src/Components/Popover.tsx
```diff
@@ -0,0 +1,161 @@
...
+    anchorPosition && anchorPosition.width / 2 - size.width / 2}px;
+  padding: 13px 15px;
+  background: #000;
+  color: #fff;
+  font-family: "Courier New", Courier, monospace;
+  cursor: default;
+  & span {
+    cursor: pointer;
+  }
+`
```
damassi: #minor - elsewhere in Reaction StyledComponents typically go below Class / Component definitions, at the bottom of the file

src/Components/Popover.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  show?: boolean
+  minWidth?: string
+  anchorRef?: HTMLElement
+  dismissOnClick?: boolean
+  message?: (() => any) | string
+}
+
+export class Popover extends React.Component<Props, any> {
```
damassi: If using `state`, should also define a `State` interface and pass into `any`

src/Components/__stories__/Popover.story.tsx
```diff
@@ -0,0 +1,113 @@
...
+import Button from "../Buttons/Default"
+import Input from "../Input"
+import { Popover } from "../Popover"
+
+const Container = styled.div`
+  padding: 50px;
+  height: 1400px;
+`
+
+class ErrorShower extends React.Component<any, any> {
```
damassi: `<any, any>`

damassi: Ahh this is just a storybook, ignore me 👍

src/Components/__stories__/Popover.story.tsx
```diff
@@ -0,0 +1,113 @@
...
+import React from "react"
+
+import styled from "styled-components"
+import Button from "../Buttons/Default"
+import Input from "../Input"
+import { Popover } from "../Popover"
+
+const Container = styled.div`
+  padding: 50px;
+  height: 1400px;
```
yuki24: Why does this container have to have a very long height?

src/Components/__stories__/Popover.story.tsx
```diff
@@ -0,0 +1,113 @@
...
+`
+
+class ErrorShower extends React.Component<any, any> {
+  public buttonRef
```
yuki24: Could you add a type here?

src/Components/__stories__/Popover.story.tsx
```diff
@@ -0,0 +1,113 @@
...
+  state = { count: 0, errorMessage: "", showError: false }
+  showTip = () => {
+    const count = this.state.count + 1
+    this.setState({
+      showError: true,
+      count,
+    })
+  }
```
yuki24: Add a line break between function definitions for readability.

src/Components/Popover.tsx
```diff
@@ -0,0 +1,164 @@
...
+  minWidth?: string
+  anchorRef?: HTMLElement
+  dismissOnClick?: boolean
+  message?: (() => any) | string
+}
+
+interface State {
+  anchorPosition?: any
+  show: boolean
+  message: any
```
yuki24: Replace `any`s with a more appropriate type.

src/Components/Popover.tsx
```diff
@@ -0,0 +1,164 @@
...
+import styled, { StyledFunction } from "styled-components"
+import { fadeIn, fadeOut } from "../Assets/Animations"
+import { secondary as secondaryFont } from "../Assets/Fonts"
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  show?: boolean
+  minWidth?: string
+  anchorRef?: HTMLElement
+  dismissOnClick?: boolean
+  message?: (() => any) | string
```
yuki24: Replace the `any` with `string` (i'm guessing it should return a string)

src/Components/Popover.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const anchorPosition = this.getAnchorPosition(nextProps.anchorRef)
+    this.setState({
+      show,
+      message: newMessage,
+      anchorPosition,
+    })
+  }
+
+  componentDidMount() {
+    window.addEventListener("scroll", this.handleScroll)
```
yuki24: This means the `getAnchorPosition` gets called and the `setState` possibly re-renders this component on every scroll event, which may not be as performant as it should.

src/Components/Popover.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const Arrow = Positionable`
+  width: 0;
+  height: 0;
+  border-bottom: 10px solid #000;
+  border-left: 10px solid transparent;
+  border-right: 10px solid transparent;
+  position: absolute;
+  left: ${({ anchorPosition }) =>
+    anchorPosition && anchorPosition.width / 2 - 10}px;
```
yuki24: Would this possibly fail if `anchorPosition` is missing? It looks like it is an optional prop.

src/Components/__stories__/Popover.story.tsx
erikdstock: I was testing scrolling.

erikdstock:
Thanks for the comments @yuki24 - just to be clear I'm sure I left some errors/bad form in my typings. There are cases where the anchor depends on a ref from some other object and so I left them as options, which may not be correct. My intent was that the component basically should not render if it does not have an anchor. Also as a reminder I'm out until next wednesday. Thanks again!

yuki24:
closing in favor of #615

PR #604, opened 2018-03-30
@xtina-starr => Restore expand on article click for desktop

eessex:
Puts back article expand click action, only for desktop.


xtina-starr:
LGTM!

PR #608, opened 2018-04-02
Add a <Checkbox> component

yuki24:
This updates the existing `<Checkbox>` component:

screen shot 2018-04-02 at 18 43 58

And here is how to use it:

```tsx
<Checkbox onChange={event => console.log("I'll remember you!", event)}>
  Remember me
</Checkbox>

<Checkbox checked>
  Remember me
</Checkbox>

<Checkbox disabled>
  Remember me
</Checkbox>

<Checkbox checked disabled>
  Remember me
</Checkbox>
```

This is an existing component, so the interface for `props` are unchagend, but now it can take a `children` prop to show the label for a checkbox input.

This also updates the styling, so for example, the  `<FilterBar>` in the `<GeneArtworks>` component would look slightly different:

### Before:

screen shot 2018-04-02 at 18 44 35

### After:

screen shot 2018-04-02 at 18 43 44

### DIFF
--- a/src/Components/ArtworkFilter/ForSaleCheckbox.tsx
+++ b/src/Components/ArtworkFilter/ForSaleCheckbox.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { Component, HTMLProps } from "react"
 
 import Checkbox from "../Checkbox"
 
@@ -6,54 +6,25 @@ import styled from "styled-components"
 import colors from "../../Assets/Colors"
 import { primary } from "../../Assets/Fonts"
 
-interface Props extends React.HTMLProps<ForSaleCheckbox> {
-  onChange?: any
-}
-
-interface State {
-  isChecked: boolean
-}
-
-export class ForSaleCheckbox extends React.Component<Props, State> {
-  state = {
-    isChecked: false,
-  }
-
-  onClick() {
-    this.setState({
-      isChecked: !this.state.isChecked,
-    })
-    this.props.onChange()
-  }
-
+export class ForSaleCheckbox extends Component<HTMLProps<Checkbox>, null> {
   render() {
-    const { isChecked } = this.state
-    const { checked } = this.props
     return (
-      <div className={this.props.className} onClick={() => this.onClick()}>
-        <StyledCheckbox checked={isChecked || checked} />
-        <label>Only for Sale</label>
-      </div>
+      <CheckboxContainer>
+        <Checkbox {...this.props}>Only for Sale</Checkbox>
+      </CheckboxContainer>
     )
   }
 }
 
-const StyledCheckbox = styled(Checkbox)`
-  margin-right: 15px
-`
-
-const StyledForSaleCheckbox = styled(ForSaleCheckbox)`
-  display: inline-block;
+const CheckboxContainer = styled.div`
   border: 1px solid ${colors.grayRegular};
-  font-size: 13px;
-  line-height: 160%;
   padding: 15px 18px;
   font-size: 13px;
   vertical-align: middle;
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-  ${primary.style}
+  ${primary.style};
 `
 
-export default StyledForSaleCheckbox
+export default ForSaleCheckbox
--- a/src/Components/Checkbox.tsx
+++ b/src/Components/Checkbox.tsx
@@ -1,71 +1,114 @@
-import React from "react"
-
+import React, { Component, HTMLProps } from "react"
 import styled from "styled-components"
 import colors from "../Assets/Colors"
 
-interface Props extends React.HTMLProps<Checkbox> {
+interface CheckboxState {
   checked: boolean
 }
 
-export class Checkbox extends React.Component<Props, null> {
-  static defaultProps = {
-    checked: true,
+export class Checkbox extends Component<HTMLProps<Checkbox>, CheckboxState> {
+  constructor(props) {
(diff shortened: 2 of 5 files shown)


src/Components/Checkbox.tsx
```diff
@@ -1,71 +1,112 @@
...
 
-interface Props extends React.HTMLProps<Checkbox> {
+export interface CheckboxState {
   checked: boolean
 }
 
-export class Checkbox extends React.Component<Props, null> {
-  static defaultProps = {
-    checked: true,
+export class Checkbox extends Component<HTMLProps<Checkbox>, CheckboxState> {
```
damassi: 👍

src/Components/ArtworkFilter/ForSaleCheckbox.tsx
```diff
@@ -1,59 +1,30 @@
...
+  render() {
+    return (
+      <CheckboxContainer>
+        <Checkbox {...this.props}>Only for Sale</Checkbox>
+      </CheckboxContainer>
+    )
+  }
+}
+
+export default ForSaleCheckbox
```
yuki24: I kept this `default` export for compatibility. Open to suggestions as to how it could go through the deprecation cycle.

src/Components/Checkbox.tsx
```diff
@@ -1,71 +1,112 @@
...
+const Label = styled.label`
+  position: relative;
+  line-height: 135%;
+  cursor: pointer;
+  display: flex;
+  align-items: center;
 `
 
-export default StyledCheckbox
+export default Checkbox
```
yuki24: Same here. A few components still depend on the `default` export.

src/Components/Checkbox.tsx
```diff
@@ -1,71 +1,112 @@
...
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      checked: props.checked || false,
+    }
+  }
+
+  onChange = event => {
+    this.props.onChange && this.props.onChange(event)
```
damassi: While this is totally valid and looks clean I think the literal version would be more maintainable long-term:

```javascript
if (this.props.onChange) {
  this.props.onChange(event)
}
```

src/Components/__stories__/Input.story.tsx
```diff
@@ -49,6 +50,30 @@ storiesOf("Components/Input", module)
       <TextArea placeholder="Your Message" disabled />
     </div>
   ))
+  .add("Check Boxes", () => (
+    <div>
+      <div style={{ padding: 10 }}>
+        <Checkbox onChange={event => console.log("I'll remember you!", event)}>
```
yuki24: I left this `console.log` intentionally but the CI doesn't like it...

yuki24:
@damassi my bad, I put the before/after screenshots in the opposite order. I've just flipped them. Is this good design-wise?

src/Components/__stories__/Input.story.tsx
damassi: `// tslint:disable-next-line`?

damassi:
Oh yeah! that looks great. I was thinking that the old version was better, but that was actually the new version 😄

src/Components/Checkbox.tsx
```diff
@@ -1,71 +1,114 @@
...
 
-export class Checkbox extends React.Component<Props, null> {
-  static defaultProps = {
-    checked: true,
+export class Checkbox extends Component<HTMLProps<Checkbox>, CheckboxState> {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      checked: props.checked || false,
```
kanaabe: I don't think we need `|| false` here?

src/Components/Checkbox.tsx
```diff
@@ -1,71 +1,114 @@
...
-    border: 2px solid black;
-    border-top: none;
-    border-right: none;
-    opacity: 0;
-    transform: rotate(-45deg) translateZ(0);
-    transition: opacity 0.25s;
+    left: 0;
+    width: 1rem;
+    height: 1rem;
+    background: #fff;
```
kanaabe: All these hex definitions could be updated to use `colors._` like you do in the line below

yuki24: :+1: I think I should replace all `color` occurrences with a `colors._`

src/Components/__stories__/Input.story.tsx
yuki24: I ended up getting rid of it since tslint doesn't recognize `tslint:disable` comments in JSX.

src/Components/Checkbox.tsx
yuki24: This is intentional so the `checked` will either be `true` or `false`, not `undefined` or `null`.

yuki24:
This PR is ready for a merge.

src/Components/ArtworkFilter/ForSaleCheckbox.tsx
```diff
@@ -1,59 +1,30 @@
...
-  }
-}
-
-const StyledCheckbox = styled(Checkbox)`
-  margin-right: 15px
-`
-
-const StyledForSaleCheckbox = styled(ForSaleCheckbox)`
-  display: inline-block;
+const CheckboxContainer = styled.div`
```
damassi: @yuki24 - #superminor - most components have the style definition below the main component for API readability.

yuki24: done.

PR #610, opened 2018-04-03
[FIX onboarding] Do not suggest after artists after following an artist

yuki24:
finishes https://artsyproduct.atlassian.net/browse/GROW-369

From the ticket:

> Right now when a user follows an artist we recommend a suggested artist to follow. We want to remove "after" artists from the list of possible suggestions. You can see a video of "after andy warhol" being suggested after I followed Andy Warhol
>
> We have many of these "after artists" and we want to eliminate them from recs during onboarding.

Initially, I thought about excluding all after artists whose name **includes** `After` (about +1400), but this caused MP to make a GET request that is too long (most HTTP servers only accept up to 8KB of GET requests and the list size exceeded 43KB). Then I pulled in a list of all artists with a name that **starts with** `After` (197 artists in Looker). This is still too long, so I sorted it by `follow_count` to limit the list. Out of the 197 artists there are only 59 artists that have at least one follower, so I compiled a list of the first 100 artists sorted by `foillow_count` with a name that starts with `After`. I pulled in 100 just to be safe, but it's probably ok to limit even further.

I'm assuming that those without followers don't show up as a suggestion because they are probably not considered to be popular or related artists to the artist that the user just followed. Search results still include After artists since it'd require a bit of data structure change.


xtina-starr:
@yuki24 this looks good to me. It might be worth having @katarinabatina or someone glance over the list you compiled/ confirm the assumptions you've made. After that and resolving the merge conflict (which I think it because of suppressing the unique key error in https://github.com/artsy/reaction/pull/609) feel free to merge.

katarinabatina:
Can't we use the popular artist endpoint to help filter vs follow count?

katarinabatina:
@yuki24 you said you sorted the list by follow_count to limit the list and I was wondering why we couldn't sort/filter the list by popular_artist. You are correct, we want to recommend related artists.

yuki24:
@katarinabatina I'll look to see if it's possible to sort by `popular_artist`.

yuki24:
I have looked into the `/popular` endpoint and it seems that the trending score is calculated based on the follow count for the last two weeks:

 * https://github.com/artsy/gravity/blob/d2be21611d779f501a7eeebd8da267d9d7fc455b/app/api/v1/artists_endpoint.rb#L203
 * https://github.com/artsy/delta/blob/master/lib/delta.rb#L73

The list I generated is sorted by all time `follow_count` for each after eartist, so there may be a rank change for the after artists that have at least one follower, but the 59 afer artists (which have a follower) would probably stay in the top 100 popular artists. The rest of the after artists are basically artists that have never been followed by anyone (which we don't really care).

Given that, I believe sorting by popularity would give us the same set of the after artists as the list in this PR. We could double-check with @cavvia if we want to be extra cafeful, but I feel pretty good about this.

xtina-starr:
@katarinabatina @yuki24 where are we with this? I thought the code looked good it was just a matter if the implementation was okay with product right?

cavvia:
`After...` artists for which we also have canonical artists, such as the `After Andy Warhol` example in the description, no longer exist. So that use case no longer exists. We merged them into their canonical parents [here](https://github.com/artsy/gravity/pull/11637). 

However, as Yuki says, a long list of `After...` artists for which we _don't_ have canonical artists could still potentially show up. I don't think filtering by popular artists will help you if you want to filter those out.

yuki24:
I just confirmed artists with a name that starts with `After` doesn't show up as a suggestion on https://www.artsy.net/personalize/artists. I think this PR is no longer relevant and could be closed.

damassi:
@yuki24 / @katarinabatina - can we close this?

PR #615, opened 2018-04-05
Adds a new <Popover> component

yuki24:
This is a continuation of @erikdstock's PR https://github.com/artsy/reaction/pull/602.

* [x] Remove the need of specifying `container={this}`
* [x] Minor styling changes
* [x] Polish the interface for `<OverlayTrigger>` (add more doc, remove `any`s, etc)
* [x] ~Rename it to `<Tooltip>`?~ Deferring this as there is a `<Tooltip>` component (which looks more like a `<Popover>` component) and they need to be renamed in one go
* [x] Add an example with an animation transition

## Popover

screen shot 2018-04-05 at 16 45 52

### Code: 

```tsx
<Popover placement="right">Post to Facebook</Popover>
<Popover placement="top">Post to Facebook</Popover>
<Popover placement="left">Post to Facebook</Popover>
<Popover placement="bottom">Post to Facebook</Popover>
```

## Display on click

display-on-click

### Code: 

```tsx
class DisplayPopoverOnClickExample extends React.Component {
  state = {
    show: false,
  }

  render() {
    const popover = <Popover>Post to Facebook</Popover>

    return (
      <div style={{ padding: "70px 140px" }}>
        <OverlayTrigger show={this.state.show} placement="top" overlay={popover}>
          <Button style={{ margin: 0 }} onClick={() => this.setState({ show: !this.state.show })}>
            Button
          </Button>
        </OverlayTrigger>
      </div>
    )
  }
}
```

## Display on hover

display-on-hover

### Code: 

```tsx
class DisplayPopoverOnHoverExample extends React.Component {
  state = {
    show: false,
  }

  render() {
    const popover = <Popover>Save</Popover>

    return (
      <div style={{ padding: "30px 30px" }}>
        <OverlayTrigger show={this.state.show} placement="bottom" overlay={popover}>
          <CircleIcon
            name="heart-small"
            color="black"
            fontSize="30px"
            ratio={0.5}
            onMouseEnter={() => this.setState({ show: true })}
            onMouseLeave={() => this.setState({ show: false })} />
        </OverlayTrigger>
      </div>
    )
  }
}
```

## Display on blur

display-on-blur

### Code: 

```tsx
class DisplayPopoverOnBlurExample extends React.Component {
  state = {
    value: "",
    show: false,
  }

  render() {
    const popover = <Popover>Is your email {this.state.value}?</Popover>

    return (
      <div style={{ padding: "50px" }}>
        <OverlayTrigger show={this.state.show} placement="bottom" overlay={popover}>
          <Input
            placeholder="Email"
            onChange={e => this.setState({ show: false, value: e.currentTarget.value })}
            onBlur={() => this.setState({ show: true })} />
        </OverlayTrigger>
      </div>
    )
  }
}
```


src/Components/Popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const PopoverContainer = styled.div`
+  text-align: left;
+  text-align: start;
+  position: relative;
+  padding: 5px;
+  display: inline-block;
+`
+
+const PopoverArrow: any = styled.div`
```
yuki24: using `any` just to make the CI happy. Open to suggestions as to what type this should be.

src/Components/Popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+    left: <PopoverLeftArrow />,
+  }
+
+  render() {
+    const arrowElement = this.Arrows[this.props.placement]
+    const { arrowOffsetLeft, arrowOffsetTop } = this.props
+
+    return (
+      <PopoverContainer {...this.props as any}>
+        {cloneElement(arrowElement, { arrowOffsetLeft, arrowOffsetTop })}
```
yuki24: I wonder this is an overuse of `cloneElement`

src/Components/Popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import * as fonts from "../Assets/Fonts"
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  placement?: "top" | "right" | "bottom" | "left"
+  arrowOffsetTop?: string
+  arrowOffsetLeft?: string
+}
+
+export class Popover extends React.Component<Props, null> {
```
damassi: Don't need `null`

yuki24:
@erikdstock @damassi @xtina-starr Alright this PR is ready! let me know if there's anything that needs to be fixed before merging.

src/Components/Popover.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        <PopoverInner>{this.props.children}</PopoverInner>
+      </PopoverContainer>
+    )
+  }
+}
+
+const PopoverContainer = styled.div`
+  text-align: left;
+  text-align: start;
```
kanaabe: I think you meant to only have one `text-align`?

src/Components/__stories__/PopoverExamples/DisplayPopoverOnBlurExample.tsx
```diff
@@ -0,0 +1,30 @@
...
+    value: "",
+    show: false,
+  }
+
+  render() {
+    const { show } = this.state
+    const popover = <Popover>Is your email {this.state.value}?</Popover>
+
+    return (
+      <div style={{ padding: "50px" }}>
```
kanaabe: I know it's a small/minor style, but I would probably use styled-components here anyway to stay consistent.

src/Components/Popover.tsx
yuki24: Yeah, this looks weird. Maybe we don't need both of them so it'll just default to whatever is set to the entire `<body>`.

src/Components/__stories__/PopoverExamples/DisplayPopoverOnBlurExample.tsx
yuki24: Paddings and margins are always tricky to consolidate and make consistent (not in terms of coding style but the actual length of padding or margin), so there would be a lot of one-off components and the entire file could be longer than it should if they all should be componentized. I'll componentize them anyway since It does look like most of the `style`s in the storybook can be consolidated.

Ideally, there should be a consolidated margin/padding definitions (e.g. [basscss-margin](http://basscss.com/#basscss-margin)). This is a wider conversation, but I wanted to leave a note that most of these inline `style`s for padding/margin are actually intentional.

damassi: This is something we should def discuss as we begin implementing artsy-wide style guidelines

src/Components/Popover.tsx
```diff
@@ -0,0 +1,104 @@
...
+  padding: 5px;
+  display: inline-block;
+`
+
+const PopoverArrowFactory: StyledFunction<
+  PopoverArrowPositionProps & React.HTMLProps<HTMLDivElement>
+> =
+  styled.div
+
+const PopoverArrow = PopoverArrowFactory`
```
l2succes: FYI, another way to add props to a styled component without declaring an interface.
Found it [here](https://github.com/styled-components/styled-components/issues/630#issuecomment-375013384)

You can do this 
```typescript
styled.div.attrs<PopoverArrowPositionProps>({
  arrowOffsetTop: string
})`
  position: absolute;
  ...
`
```

l2succes:
@yuki24 LGTM! Clean API too, let's merge after addressing @kanaabe's comments

src/Components/Popover.tsx
yuki24: Interesting, I'll give it a try.

yuki24:
Alright, I think I fixed them all. I'll merge once the CI is ✅

PR #617, opened 2018-04-06
@yuki24 => [Component] Add <StepMarker /> 

damassi:
Adds a new `<StepMarker />` component in support of a new order form as seen here: 

screen shot 2018-04-05 at 7 08 13 pm

Additionally provides a render-props API like:

```javascript
const steps = [
  {
    label: "Shipping",
    isActive: false,
    isComplete: false,
  },
  {
    label: "Payment",
    isActive: false,
    isComplete: false,
  },
  {
    label: "Review",
    isActive: false,
    isComplete: false,
  },
]

return (
  <Container>
    <StepMarker steps={steps}>
      {({ nextStep, previousStep, gotoStep, stepState }) => {
          ...
      }}
    </StepMarker>
  <Container>
)
```

#### TODO
- [x] Tests

**Spec:**
- https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5abd3f9478dd7103788ecc99

### DIFF
--- a/src/Components/StepMarker/index.tsx
+++ b/src/Components/StepMarker/index.tsx
@@ -1,0 +1,193 @@
+import Icon from "../Icon"
+import React, { Component } from "react"
+import Text from "../Text"
+import colors from "../../Assets/Colors"
+import styled from "styled-components"
+import { Fonts } from "../Publishing/Fonts"
+import { Props, State, StepProps } from "./types"
+import { isUndefined } from "lodash"
+
+export class StepMarker extends Component<Props, State> {
+  constructor(props) {
+    super(props)
+    this.validate(props)
+    this.state = this.getStepState(props)
+  }
+
+  validate(props: Props) {
+    const isInvalid = props.steps.filter(step => step.isActive).length > 1
+
+    if (isInvalid) {
+      console.error(
+        "(Components/StepMarker) Error: Step configuration has more than",
+        "one active item."
+      )
+    }
+  }
+
+  getStepState(propsOrState: State) {
+    let { currentStep, steps } = propsOrState
+
+    // If currentStep isn't passed in attempt to infer it from configuration
+    if (isUndefined(currentStep)) {
+      currentStep = steps.reduce((acc, step, index) => {
+        return step.isActive || step.isComplete ? index : acc
+      }, 0)
+    }
+
+    steps = steps.reduce((acc, step, index) => {
+      let stepToUpdate = {
+        ...step,
+        isActive: false,
+        isComplete: false,
+      }
+
+      if (index < currentStep) {
+        stepToUpdate.isComplete = true
+      }
+      if (index === currentStep) {
+        stepToUpdate.isActive = true
+      }
+
+      return acc.concat([stepToUpdate])
+    }, [])
+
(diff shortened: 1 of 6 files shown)


src/Components/StepMarker/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+        top: 1px;
+        position: absolute;
+        content: " ";
+      }
+
+      &:last-child:after {
+        border-top: none;
+      }
+    `
+  }};
```
yuki24: This is interesting.

src/Components/StepMarker/types.ts
```diff
@@ -0,0 +1,24 @@
+import { ReactNode } from "react"
+
+export interface StepProps {
+  label?: string
+  isActive?: boolean
+  isComplete?: boolean
```
yuki24: Are these all optional or if they are `undefined` it'll just look like a  "⚪️▫️⚪️▫️⚪️"?

damassi: Heh, no they're all required, need to update this as it was a dev thing

src/Components/StepMarker/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+    }
+  }
+
+  updateStep(currentStep) {
+    this.setState(
+      {
+        currentStep,
+      },
+      () => {
+        this.setState(this.getStepState(this.state))
```
damassi: TODO: Remove this extra render cycle

damassi:
Aightt, added some tests so this is now ready to go

src/Components/StepMarker/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      &:after {
+        border-top: 2px solid ${connectingBorderColor};
+        width: 100%;
+        left: 50%;
+        top: 1px;
+        position: absolute;
+        content: " ";
+      }
```
yuki24: Is the `<Step>` component completely static or it should have animations on a page transition? If so, `&:after` could make it tricky to add animations.

src/Components/StepMarker/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  }};
+`
+
+const StyledText = Text.extend`
+  ${Fonts.unica("s12", "regular")};
+  margin-top: 0;
+  margin-bottom: 0;
+`
+
+const StyledIcon = Icon.extend`
```
yuki24: This is minor, but you could use [a `CircleIcon`](https://artsy.github.io/reaction/?selectedKind=Components%2FIcons&selectedStory=All%20Icons&full=0&down=1&left=1&panelRight=0) to eliminate a few CSS properties:

```tsx
<CircleIcon name='check' color='black' />
```

damassi: @yuki24 - due to positioning and so on this still requires a similar amount of CSS; the regular icon makes it a bit simpler

src/Components/StepMarker/index.tsx
damassi: Not going to worry about animations for now

PR #613, opened 2018-04-04
@sarahscott => NewsPanel component

eessex:
Adds 'NewsPanel' component for entry point to `/news` on the main editorial page.

screen shot 2018-04-04 at 3 03 31 pm

screen shot 2018-04-04 at 3 03 37 pm

PR #618, opened 2018-04-06
@kanaabe => Remove snapshot that renders todays date

eessex:
A [PR from yesterday](https://github.com/artsy/reaction/pull/613) broke the master, because a snapshot includes rendering today's date. 

This removes the snapshot from the test file, it still has unit tests for non-date items.


kanaabe:
Ah i see, could we fix the snapshot to use a fixed date instead of deleting it?

eessex:
The problematic component always prints today's date as text

kanaabe:
Ok, i guess that's fine then! If we're covering with unit then I'm 👍

PR #621, opened 2018-04-11
@kanaabe => Signup form

l2succes:
https://artsyproduct.atlassian.net/browse/GROW-394


l2succes:
Just realized there were 3 PR's assigned to @xtina-starr, so re-assigned that one to you @kanaabe :)

src/Components/Authorization/DesktopModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+  padding: 10px;
+  margin: 20px 0 0;
+`
+
+const Content = styled.div`
+  box-sizing: border-box;
+  margin: 0px 60px;
+`
+
+const Logo = styled(Icon).attrs({
```
kanaabe: Interesting, why do we do this? Wouldn't it be the same as:

```
const Logo = styled(Icon)`
  color: "black",
  font-size: "34px"
`
```

src/Components/Authorization/RegisterForm.tsx
```diff
@@ -75,19 +96,20 @@ export const RegisterForm: FormComponentType = props => {
               onBlur={handleBlur}
               errorMessage={errors.acceptedTermsOfService}
             >
-              I Agree to the TOS And PP
+              <Text color={colors.grayDark}>
+                I Agree to the <TextLink>Terms Of Service</TextLink> And{" "}
+                <TextLink>Privacy Policy</TextLink>
+              </Text>
             </TOSCheckbox>
             {/* touched.password && errors.password && <div>{errors.password}</div> */}
```
kanaabe: Is this meant to stay as a comment?

src/Components/Authorization/Validators.ts
```diff
@@ -0,0 +1,14 @@
...
+  password: Yup.string()
+    .required("Password required")
+    .min(8, "Your password must be at least 8 characters"),
+  acceptedTermsOfService: Yup.boolean().required(
+    "You must agree to our terms to continue."
+  ),
+  email: Yup.string()
+    .email("Please enter a valid email.")
+    .required("Please enter a valid email."),
+})
```
kanaabe: Sweet!

src/Components/Buttons/Default.tsx
```diff
@@ -76,8 +79,7 @@ export const StyledButton = styled(Button)`
     }};
   }
 
-  ${fonts.primary.style}
-  ${block()}
+  ${fonts.primary.style} ${block()};
```
kanaabe: Minor, but adding a semicolon after fonts to prevent Prettier changes might be nicer to read

src/Components/Buttons/Facebook.tsx
```diff
@@ -4,19 +4,23 @@ import Icon from "../Icon"
...
+  return (
+    <Button {...props} icon={icon}>
+      {props.children || "Log in with Facebook"}
+    </Button>
+  )
 }
 
 export default styled(FacebookButton)`
-  background: #39439C;
+  background: #39439c;
```
kanaabe: Whoah what are these mysterious colors lol. Are they repeated anywhere? We might want to add to that Colors file for default social colors if they are used elsewhere.

src/Components/Authorization/DesktopModal.tsx
l2succes: Actually no it isn't the same, in this case it would have the same effect but this is a shorthand for passing props to the styled component, it also supports type definitions. For instance,

```typescript
import {IconProps, Icon} from './Icon'

interface LogoProps extends IconProps {
  secondaryColor: string
}

const Logo = styled(Icon).attrs<LogoProps>({
  secondaryColor: 'green'
})``;
```
this is equivalent to `<Logo secondaryColor="green" />` in this case

l2succes:
Thanks for the review @kanaabe! I agree about the hardcoded colors and.

I'm gonna go ahead and merge this with the commented out error handling as I'm currently consolidating fonts/colors across reaction and I'd like to keep this work separate.

kanaabe:
👍

PR #623, opened 2018-04-11
Wizard for mobile auth form

xtina-starr:
This implements a wizard needed for the mobile auth form. This wizard iterates through the different step inputs, but does not yet have validation hooked up.


src/Components/Authorization/Mobile/Wizard.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const stepIndex = this.state.currentStep + 1
+    this.setState({ currentStep: stepIndex })
+  }
+
+  validate(): boolean {
+    const step = this.stepComponents[this.state.currentStep] as any
+
+    const hasErrors = Children.toArray(step.props.children)
+      .map((input: any) => {
+        return !!this.props.errors[input.props.name]
```
damassi: This is just a matter of opinion, but I always find 
```javascript
return Boolean(this.props.errors[input.props.name])
```
to be more readable at a glance

src/Components/__stories__/ProgressIndicator.story.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Button from "../Buttons/Default"
+import { ProgressIndicator } from "../ProgressIndicator"
+
+interface State {
+  count: number
+  finished: boolean
+}
+
+class Steps extends React.Component<any, State> {
+  public numberOfSteps: number = 4
```
damassi: Might not have gotten to this point yet, but is there any way we could pass this in as props or infer it from the `<Wizard />` container?

xtina-starr: Yeah this wasn't really finished yet. I wonder if the progress bar even needs it's on story and could just be demonstrated in the wizard story...?

damassi: @xtina-starr - curious if you have a comp of the progress bar I could peek at? I'm curious if there's any overlap here: https://artsy.github.io/reaction/?selectedKind=Components%2FStepMarker&selectedStory=With%20renderProps%20API&full=0&down=1&left=1&panelRight=0

l2succes: @damassi The progress bar is just that black bar at the top in the follow design (zpl.io/aw05Dod). So it's more simple than the one you built. 
What I had in mind is to have the Wizard component support rendering an arbitrary component to display the current step

```
<Wizard progressComponent={props => <ProgressBar {...props} />} />
```

```
props = {
  currentStep: 1,
  numberOfSteps: 4
}
```

damassi: Ooh, that design in zeplin is simple and elegant, also the approach to be able to pass in a component 👍 I wonder if at some point we could unify all of this behind the `<Wizard>`, because it looks like it could be a good stateful host for a lot of various step-like UIs.

l2succes:
Looks good! We'll still need to add support for custom progress bars and potentially support for nested input components but we can do so in a subsequent PR.

PR #625, opened 2018-04-12
@kanaabe => Refactor ArticleLayout

eessex:
While looking into adding series/sponsor components to Features, I realized that a lot of non-overlapping logic was housed in the `ArticleLayout` component, making it pretty difficult to follow. This PR separates some of these concerns into separate components, so that we can more easily add to feature and standard articles independently.  No significant changes were made in logic, mostly just moving things around.

- Moves the body of features into the already existing `FeatureLayout`
- Moves the body of standard articles into the already existing `StandardLayout`
- Moves sidebar content into the `Sidebar` component
- Adds more unit tests for each layout and sidebar

The `ArticleLayout` file might better be described as `ArticleWithFullscreenViewer` -- however renaming it meant the changes were not side-by-side, making it harder to review.  We might consider renaming this later.

### DIFF
--- a/src/Components/Publishing/Layouts/StandardLayout.tsx
+++ b/src/Components/Publishing/Layouts/StandardLayout.tsx
@@ -6,12 +6,10 @@ import { Responsive } from "../../../Utils/Responsive"
 import { pMedia } from "../../Helpers"
 import { DisplayCanvas } from "../Display/Canvas"
 import { DisplayPanel } from "../Display/DisplayPanel"
-import { EmailPanel } from "../Email/EmailPanel"
 import { Header } from "../Header/Header"
 import { ReadMore } from "../ReadMore/ReadMoreButton"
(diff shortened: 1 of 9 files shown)


src/Components/Publishing/Layouts/StandardLayout.tsx
```diff
@@ -1,16 +1,196 @@
...
+
+export interface ArticleProps {
+  article: ArticleData
+  seriesArticle?: ArticleData
+  closeViewer?: () => void
+  display?: {
+    name: string
+    panel: object
+    canvas: any
+  }
```
kanaabe: Would be nice to get this typed eventually

kanaabe:
👍 LGTM! Nice one.

kanaabe:
Ah oops I realized you missed versioning on this.

PR #627, opened 2018-04-12
@xtina-starr => Series nav on feature articles

kanaabe:
I noticed that the Articles story was getting hard to navigate, so this also breaks out this file into multiple stories for easier navigation + development. 

Addresses https://artsyproduct.atlassian.net/browse/GROW-488

Adds this transparent nav to Feature articles with Fullscreen if they belong to a Series. 
image


eessex:
Good call breaking out the stories!

src/Components/Publishing/Nav/Nav.tsx
```diff
@@ -79,23 +70,29 @@ export class NavComponent extends React.Component<Props, State> {
...
+  styled.div
 
 const NavContainer = Div`
-  background-color: ${props => props.transparent ? "transparent" : "black"};
+  background-color: ${props => (props.transparent ? "transparent" : "black")};
   border-bottom: 1px solid white;
-  ${props => props.transparent && !props.isFixed && `
+  ${props =>
+    props.transparent &&
+    !props.isFixed &&
```
xtina-starr: I'm assuming that this is suppose to be if there is a transparent property and `isFixed` is false apply those style rules?

Minor but I found this a bit tricky to read. I don't know if this is syntactically correct but maybe wrap the first two statements in parenthesis?

kanaabe: Yep, this is syntactically correct. It's a little trick -- && will stop and return false at a false value, otherwise it will return the last value. So this returns the string literal if it passes the first two &&'s

xtina-starr: Right and what I was proposing was something like...

``` js
${props =>
  (props.transparent && !props.isFixed) &&
  `
  ...
`}
```

to group the two conditionals for readability.

kanaabe: Oh gotcha, ya I'd be glad to do that!

kanaabe: Ah well I tried to do this but Prettier just removed them anyway 😅

xtina-starr: Haha no worries. It was just a minor suggestion. In general this PR looks good to me. I think we can merge.

kanaabe:
Cool, going to merge then!

PR #628, opened 2018-04-16
@eessex => Use seriesArticle from article instead of a prop

kanaabe:
Use the `seriesArticle` that's already available from `article` instead of as a prop.


src/Components/Publishing/Layouts/FeatureLayout.tsx
```diff
@@ -24,8 +23,8 @@ export class FeatureLayout extends React.Component<ArticleProps> {
       isMobile,
       isSuper,
       relatedArticlesForCanvas,
-      seriesArticle,
     } = this.props
+    const { seriesArticle } = article
```
eessex: Could also do this inline as `article: { seriesArticle }`

eessex: 👍

PR #629, opened 2018-04-16
@kanaabe => Type system update

l2succes:
The PR is an initial pass at consolidating our uses of fonts and styles across reaction in light of the new Design system that's been introduced recently.

It replaces the `primary` and `secondary` mixins for ones named after the fonts `avantgarde`, `garamond` and `unica`. 

This also adds [`babel-plugin-module-resolver`](https://github.com/tleunen/babel-plugin-module-resolver) which allows us to use path aliases instead of 
`import Fonts from '../../../Assets/Fonts'` we can now write `import Fonts from 'Assets/Fonts'`
this was verified to work in force as well.

Todo:

- [ ] Add support for responsive styles

https://www.notion.so/artsy/Design-92030f16ed7c4c72bb3eb832b4243d04


src/Assets/Fonts.tsx
```diff
@@ -1,36 +1,178 @@
...
+}
+
+export const unica = (size: keyof typeof UnicaSizes, family = "regular") => {
+  const evaluatedSize = UnicaSizes[size]
+  return css`
+    font-family: ${fontFamily.unica[family]}, Arial, serif;
+    -webkit-font-smoothing: antialiased;
+    font-size: ${evaluatedSize.size};
+    line-height: ${evaluatedSize.height};
+  `
```
damassi: While I think this might be superceded by some of our discussion today this is where I was thinking about introducing some layer or responsiveness into the system

src/Components/Artist/MarketInsights/MarketInsights.tsx
```diff
@@ -3,15 +3,15 @@ import * as React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
 
-import colors from "../../../Assets/Colors"
-import { Fonts } from "../../Publishing/Fonts"
+import colors from "Assets/Colors"
+import { unica } from "Assets/Fonts"
```
damassi: We should make 100% sure that absolute paths are cross-app compatible; maybe there's something easy we can do if not?

Can you try to run this while linked with force?

l2succes: I did get some errors in force but I'm not 100% sure it's related to the use of absolute paths. Will confirm, if anything i'll revert back to relative paths

src/Components/Gene/Artists.tsx
```diff
@@ -40,8 +40,7 @@ const LoadMoreContainer = styled.div`
 `
 
 const LoadMoreButton = styled.a`
-  font-family: ${fonts.primary.fontFamily};
-  font-size: 14px;
+  font-family: ${avantgarde("s13")};
```
kanaabe: Do we need `font-family` here?

src/Components/Publishing/Display/Canvas/CanvasSlideshow.tsx
```diff
@@ -280,6 +280,6 @@ const Image = responsiveImage`
   `}
 `
 const Caption = styled.div`
-  ${Fonts.garamond("s11")} color: ${Colors.grayMedium};
+  ${garamond("s11")} color: ${Colors.grayMedium};
```
kanaabe: Could use a `;`

src/Components/Publishing/Email/InstantArticleEmailSignup.tsx
```diff
@@ -106,19 +110,17 @@ const InstantArticleEmailContainer = styled.div`
...
 `
 const Title = styled.div`
-  ${Fonts.garamond("s23")}
-  margin-bottom: 30px;
+  ${Fonts.garamond("s23")} margin-bottom: 30px;
 `
 const Subtitle = styled.div`
-  ${Fonts.avantgarde('s11')}
-  margin-bottom: 10px;
+  ${Fonts.avantgarde("s11")} margin-bottom: 10px;
```
kanaabe: `;` and one above

src/Components/Publishing/Fonts.tsx
```diff
@@ -1,184 +0,0 @@
...
-    line-height: ${textSizesForGaramond[size].height};
-  `
-  return style.concat(sizeStyles)
-}
-
-export const Fonts = {
-  garamond,
-  avantgarde,
-  unica,
-}
```
kanaabe: 😴

src/Components/Publishing/Header/BasicHeader.tsx
```diff
@@ -134,7 +134,7 @@ const Description = styled.div`
 `
 
 const defaults = css`
-  ${Fonts.unica("s16", "medium")} line-height: 1.1em;
+  ${unica("s16", "medium")} line-height: 1.1em;
```
kanaabe: 👀

src/Components/Publishing/Header/StandardHeader.tsx
```diff
@@ -40,19 +40,17 @@ const StandardHeaderContainer = styled.div`
   box-sizing: border-box;
   ${pMedia.sm`
     margin: 30px auto;
-  `}
+  `};
 `
 const Title = styled.div`
-  ${Fonts.garamond("s50")}
-  margin-bottom: 50px;
+  ${garamond("s50")} margin-bottom: 50px;
```
kanaabe: 👁

kanaabe:
Nice so far! There are a few places where it gets hard to read with those Prettier-ed interpolated styles but I spared you of commenting on all of them lol. I'm surprised that we didn't have to sacrifice any font sizes in the new system - happy about that! Given that visually nothing should change, it would be great to get @eessex and I to take a look on staging to make sure we didn't miss anything before we ship to Force.

l2succes:
Thanks @kanaabe. Yeah I realize prettier makes it difficult to review the actual changes.

Unfortunately I actually haven't updated switched/added all the new values yet so some of them are subject to change. I already spoke to Owen about what to replace them with. Hoping to get it all done today. Will keep you posted.

src/Assets/Fonts.tsx
l2succes: Yeah, that's what I was thinking as well. Just to keep the scope of this PR manageable I'll do it in another PR.

src/Components/Gene/Artists.tsx
l2succes: No we don't, good catch :)

l2succes:
@kanaabe Ok I think this finally good to go, I tested things in force to make sure everything still works but I'll take another pass once it's merged. As I mentioned, this is preliminary work that sets us up to support the new Design & Typography system.

In another PR, I will introduce the new font names yet as we want to also add support for responsiveness using (https://github.com/jxnblk/styled-system)

src/Components/Artist/MarketInsights/MarketInsights.tsx
damassi: @l2succes - how did this end up working in force?

kanaabe:
🎉 LGTM!!!!!!!!!!!! 🎉  (we discussed doing a major version change for this since it will be breaking)

src/Components/Artist/MarketInsights/MarketInsights.tsx
l2succes: @damassi Tested it locally on two machines and it works great. There was an issue with jest not resolving the path aliases but I fixed it

damassi: 👍

PR #633, opened 2018-04-17
@xtina-starr => Conditionally show partner info

eessex:
Adds a few conditionals to the PartnerBlock component to prevent showing title text when no image is present.


xtina-starr:
👍

PR #634, opened 2018-04-17
@eessex => Updates analytics to include new naming

kanaabe:
Addresses https://artsyproduct.atlassian.net/browse/GROW-460


eessex:
👍

PR #611, opened 2018-04-03
[Order2/WIP] Add new Order Form

damassi:
### WIP

This PR builds out a new order form for use in Force under the `/order2` route. Built using [React Router 4](https://reacttraining.com/react-router/) and [Formik](https://github.com/jaredpalmer/formik) (similar to Redux Form). 

What needs to be done:
- Form validation and submission. Formik has been dropped in but the particulars (validation before NEXT btn is enabled on a per-form basis) need to be wired up
- New top-level `onSubmit` API to handle submission / purchase on Force side, which connects or remaps existing API
- Style pixel pushing
- Tests 

**Related PRs:**
- https://github.com/artsy/force/pull/2326 -- (`/order2` route) 
- https://github.com/artsy/reaction/pull/617

**Storybooks:**
- http://localhost:9001/?selectedKind=Components%2FForms%2FOrder%20Form&selectedStory=Order%20Form&full=0&down=1&left=1&panelRight=0

**Specs:**
- https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=35&projectKey=BUY&selectedIssue=BUY-130&sprint=50
- https://app.zeplin.io/project/5ab94282c124db8d58365b47/dashboard

When developing in force, make sure to open a new tab and run `yarn link && yarn watch` so that changes are synced over on the force side during development.

PR #636, opened 2018-04-17
@sweir27 => New order form

erikdstock:
Tracked in [PURCHASE-13](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-13)

This PR updates @damassi 's previous order form PR to use the new Wizard component developed by @xtina-starr and @l2succes. The form is visible in development mode on force at `/order2` if you `yarn link` it and `yarn watch` this source.

On the matter of file structure, this is organized under a new `Components/Forms` folder.  It also creates a form wizard component under `src/Components/Forms` following the pattern laid out in the formik repo examples: https://github.com/jaredpalmer/formik/blob/master/examples/MultistepWizard.js


src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -10,7 +10,7 @@ import {
   inputValidators,
   StyledFacebookButton,
   StyledInput as Input,
-  TOSCheckbox,
+  TOSCheckbox
```
damassi: @erikdstock - was there no solution to the trailing comma issue? We don't want to remove this setting but rather fix prettier / eslint / tslint so that conflicts don't arise.

erikdstock: I'm unclear- do we WANT trailing commas? the specific issue came from object destructuring with `...rest` props at the end which was handled inconsistently from the other pieces.

damassi: We do, they're super useful when refactoring and this is the setting we have across all other projects.

erikdstock: oh, my mistake then. this seems to be the issue: https://github.com/prettier/prettier/issues/3528

erikdstock: Yeah, this was fixed in [1.9](https://prettier.io/blog/2017/12/05/1.9.0.html#don-t-print-a-trailing-comma-after-object-rest-spread-3313-https-githubcom-prettier-prettier-pull-3313-by-duailibe-https-githubcom-duailibe)

I'm able to avoid it for now with a `// disable` comment.

src/Components/Forms/FormikInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { FormikProps, Field } from "formik"
+import styled from "styled-components"
+import { default as Input, InputProps } from "../Input"
+
+// export interface FormikInputProps<T> extends InputProps {}
+
+/**
+ * An input with the standard FormimkProps added for rendering niceness
+ */
+export const FormikInput: (props: any) => any = props => {
```
l2succes: we should name this component `Field` if it's an abstraction for our `Input` component + `Formik`, and use a concrete interface instead of `any` (maybe just reuse `InputProps`)

sweir27:
It feels like this PR is growing a bit-- is there any way to separate the `Wizard` updates from the `OrderForm` specifics? Might be a little easier to review in isolation if that's a possibility.

erikdstock:
I'll see if I can pull the wizard into its own PR this morning.

src/Components/Wizard/types.ts
```diff
@@ -5,11 +5,23 @@ export type FormErrors = { [key: string]: any } | null
...
 
+// Extending WizardContext here causes
+// `Type '(e: FormEvent<HTMLFormElement>) => any' is not assignable to type '(event: MouseEvent<Button>) => void'
+// in form.handleSubmit for button onClick prop
+// OR '(e: FormEvent<any>, values: { [key: string]: any; }) => void' is not assignable to type '(event: MouseEvent<Button>) => void'
+// for wizard.previous  for button OnClick prop
+export interface WizardStepChildProps {
+  form: any
+  wizard: any
+}
```
erikdstock: The props passed via context to a Wizard's Step's Child somehow cause type errors as descibed above, basically a conflict between MouseEvents and FormEvents. This isn't a problem in the wizard example form, where we don't use context from Step, so I'm solving it by setting these props to `any`. This avoids needing to say `< ... onClick={ form.handleSubmit as any }>` in the shipping, review and payment forms.

src/Components/Forms/support/index.tsx
```diff
@@ -0,0 +1,3 @@
+import { Field } from "./Field"
+
+export { Field }
```
sweir27: I keep wishing there was a way to import/export in the same line but I'm not sure there's a better way than this.

src/Components/Forms/OrderForm/Forms/ShippingForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+        <Col xs>
+          <Title titleSize="small">Shipping address</Title>
+        </Col>
+      </Row>
+
+      <AddressFormInputs />
+
+      <Row>
+        <Col xs>
+          <InvertedButton block onClick={form.handleSubmit}>
```
sweir27: Still wondering what the best practice is for adding logic in the `submit` function on each step, since it seems that `handleSubmit` is defined on the parent. Would you just add a function that does an action and then calls `handleSubmit` explicitly?

I'm also happy to look into this when I pick up the next task, just wondering if there's already a preferred pattern.

src/Components/Forms/support/index.tsx
erikdstock: you can actually combine those 2 lines (`~import~ export { Blah } from "../Blahs"`)

src/Components/Forms/OrderForm/Forms/ShippingForm.tsx
erikdstock: Yeah, its hard to explain but:
`<Wizard>`'s children get `wizard` and `form` props which contain the same props as the Wizard `Step`'s child. In the latter case those props come down from context, which seems to cause some typing confusion. Ultimately I see it as- if you want a simple control in your layout (the `Wizard` child) then you can use it there. If you would rather have it in the `Step` (eg, "continue to step 2") you can use the same props there (but under the hood they come from context via the wizard)

src/Components/Forms/OrderForm/Forms/ReviewForm.tsx
```diff
@@ -0,0 +1,260 @@
...
+              </Text>
+            </Col>
+          </Row>
+        </Fragment>
+      </Grid>
+    )
+  }
+}
+
+// // Submission Complete
```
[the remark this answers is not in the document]
erikdstock: Left over from chris' work. I figured I would leave it until I've checked what comes at that step. I haven't checked yet.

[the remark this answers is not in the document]
sweir27: If his work is still captured in a closed PR (i.e. https://github.com/artsy/reaction/pull/611) then I'd rather we remove the commented-out code so we can avoid messy history.

src/Components/Forms/OrderForm/Forms/ReviewForm.tsx
```diff
@@ -0,0 +1,260 @@
...
+import { Grid, Row, Col } from "react-styled-flexboxgrid"
+import { WizardStepChildProps } from "../../../Wizard/types"
+
+export class ReviewForm extends Component<WizardStepChildProps> {
+  render() {
+    return (
+      <Grid fluid>
+        <Fragment>
+          <ShippingAddress editable />
+          <PaymentMethod editable />
```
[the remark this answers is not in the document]
erikdstock: credit @damassi

src/Components/Forms/OrderForm/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { SFC } from "react"
+import { App } from "./App"
+// import { BrowserRouter } from "react-router-dom"
+// import { StaticRouter } from "react-router"
+
+export const OrderForm: SFC<{ onSubmit: any }> = ({ onSubmit }) => {
+  // const isClient = typeof window !== "undefined"
```
sweir27: ditto here

src/Components/Publishing/Byline/__tests__/__snapshots__/NewsByline.test.tsx.snap
```diff
@@ -158,7 +158,7 @@ exports[`News Byline renders on mobile 1`] = `
   >
     <a
       className="c6"
-      href="https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fwww.artsy.net%2Farticle%2Fundefined"
+      href="https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fwww.artsy.net%2Farticle%2Fnews-article"
```
sweir27: Is it intentional that all of these snapshots got updated?

erikdstock: No, i haven't even run tests- all i've done is rebase off master. i'll double check this.

PR #639, opened 2018-04-18
@kanaabe => Add ErrorBoundary component

eessex:
The caption component seems to bug out differently based on the caption it gets -- I'm going to make a separate ticket for looking at this more closely. 

For expediencies sake on the Emerging Artists feature/slideshow bug, this PR adds an ErrorBoundary component, and wraps the Truncator component in it to prevent irregular captions from breaking their parent components. 

This is the same bug that is causing articles not to render [referenced in this ticket](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-529)

### DIFF
--- a/src/Components/ErrorBoundary.tsx
+++ b/src/Components/ErrorBoundary.tsx
@@ -1,0 +1,15 @@
+import React from "react"
+
+interface Props {
+  children?: any
+}
+
+export class ErrorBoundary extends React.Component<Props, null> {
+  componentDidCatch(error, errorInfo) {
+    console.error(error)
+  }
+
+  render() {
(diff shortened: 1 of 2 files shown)


src/Components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react"
+
+interface Props {
+  children?: any
+}
+
+export class ErrorBoundary extends React.Component<Props, null> {
```
damassi: Can omit `null` if not passing in `State` interface

PR #645, opened 2018-04-20
@eessex => Adds back read more tracking

kanaabe:
Oops! I removed this before thinking we don't need it since we don't want to make a track call, but instead we should keep this and omit it here so we still get a pageview: https://github.com/artsy/force/blob/e69298df7f9f9a321ee75142e7fc7bce3d62b733/src/desktop/assets/analytics.coffee#L17


eessex:
Not sure why on hold, but feel free to merge when ready.

PR #646, opened 2018-04-20
@kanaabe => Show series title in feature header 

eessex:
- Shows series title/link in fullscreen feature header if article is in series
- Adds `VerticalOrSeriesTitle` component, to show series title if eligible over vertical
- Removes `Title` component from `RelatedArticleCards` and replaces with `VerticalOrSeriesTitle`

screen shot 2018-04-20 at 12 14 36 pm
screen shot 2018-04-20 at 1 15 24 pm


src/Components/Publishing/RelatedArticles/ArticleCards/Block.tsx
```diff
@@ -45,8 +56,15 @@ ArticleCardsBlock.defaultProps = {
 const ArticleCardsContainer = styled.div`
   color: ${props => props.color};
 
-  ${Link} {
-    color: ${props => props.color};
+  ${Vertical} {
+    ${Fonts.unica("s32")} width: 100%;
```
kanaabe: Could use a semi here

src/Components/Publishing/Sections/VerticalOrSeriesTitle.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Fonts } from "../Fonts"
+
+interface Props {
+  article?: ArticleData
+  color?: string
+  prefix?: string
+  vertical?: any
+}
+
+export const VerticalOrSeriesTitle: React.SFC<Props> = props => {
```
kanaabe: Is there a design name for this? Like a Tagline or something? I definitely don't mind the verboseness of the component but just a thought if this expands to include more possible text types.

eessex: Thought about this, but decided against because it is used in both the article header and the series footer atm. I'm open to changing, but feel we may want to wait until the opportunity for more uses arises.

PR #650, opened 2018-04-20
Form Wizard

erikdstock:
This PR creates a Wizard component that can use the Formik library to create a multi-step form. It also passes general information about progress through the wizard and some controls to its children. It's still a little bit cumbersome but the basic form definition is workable- very open to suggestions for improvement:

form

```
export const FormWizard = () => {
  const submitForm = (values, actions) => {
    window.alert(`Name: ${values.name} | Age: ${values.age}`)
    actions.setSubmitting(false)
  }

  const pages: WizardSchema = [
    {
      label: "Name",
      component: props => (
        <FormPage>  // A styled.div
          <Field autoFocus name="name" type="text" placeholder="Your Name" />
          <Button onClick={props.form.handleSubmit}>Continue to Age</Button>
        </FormPage>
      ),
      stepName: "name",
      validate: validateYupSchemaSync({
        name: yup.string().required("Please enter your name"),
      }),
    },
    {
      label: "Age",
      component: ({ form, wizard }) => (
        <FormPage>
          <Field autoFocus name="age" type="text" placeholder="Your Age" />
          <Button onClick={form.handleSubmit}>Finish</Button>
          <Button onClick={wizard.previous}>Back</Button>
        </FormPage>
      ),
      stepName: "age",
      validate: validateYupSchemaSync({
        age: yup
          .number()
          .min(18, "You must be at least 18 to proceed")
          .required("Age is required"),
      }),
    },
    {
      label: "Review",
      component: ({ wizard, form }) => (
        <FormPage>
          {JSON.stringify(form.values)}
          <Button onClick={form.handleSubmit}>Submit</Button>
          <Button onClick={wizard.previous}>Back</Button>
        </FormPage>
      ),
      stepName: "review",
    },
  ]
  return (
    <Wizard onComplete={submitForm} pages={pages}>
      {(wizardBag: RenderProps) => (
        <Container>
          <StepMarker
            steps={wizardBag.pages}
            currentStepIndex={wizardBag.pageIndex}
          />
          <WizardForm {...wizardBag} container={Form} />  // Container is a styled.form
          <pre>{JSON.stringify(wizardBag.values, null, 2)}</pre>
        </Container>
      )}
    </Wizard>
  )
}
```


src/Components/Forms/support/Field.tsx
```diff
@@ -0,0 +1,46 @@
...
+      placeholder={placeholder}
+      // prettier-ignore
+      // because restProps should not have a trailing comma (fixed in prettier 1.9)
+      render={({
+        field, // { name, value, onChange, onBlur }
+        form: { touched, errors, values }, // also values, setXXXX, handleXXXX, dirty, isValid, status, etc.
+        ...restProps
+      }) => (
+        <div>
+          <Input
```
erikdstock: The Formik `<Field>` wraps an `input` or other component and passes the formik form + field props into it.

src/Components/Wizard/StepMarker/index.tsx
```diff
@@ -1,112 +1,60 @@
...
+import { Fonts } from "../../Publishing/Fonts"
+import { Props, State, StepState } from "./types"
+
+export * from "./types"
 
 export class StepMarker extends Component<Props, State> {
+  static defaultProps = {
+    disableInternalState: false,
+  }
+
```
erikdstock: I took @damassi 's Stepmarker, moved it and simplified it to a point that it probably could be an SFC now, but this was the last piece of what I did before submitting the PR and thought I shouldn't undo too much of what was already there. I did move it inside of the `Wizard`.

src/Components/Wizard/StepMarker/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  currentStepIndex: number
+  steps: StepState[]
+}
+
+export type StepState = Step & { isActive: boolean; isComplete: boolean }
+
+export interface Step {
+  label: string
+  onClick?: any
+}
```
erikdstock: I tried but didn't ultimately succeed yet at making a StepMarker, Generic Step and Form Step that would share/extend from each other. `WizardSchema`s are actually a `Step[]` so it would be nice to pass them around similarly to interested components.

src/Components/Wizard/Wizard.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Wrapper
+        pages={this.props.pages}
+        pageIndex={pageIndex}
+        next={this.next}
+        previous={this.previous}
+        onComplete={this.props.onComplete}
+        values={this.state.values}
+        // goToPage=
+        // reset=
+        // exit=
```
erikdstock: The Wizard passes its controls via render props - `next` vs `onComplete` vs `handleSubmit` (which chooses which one for you), so I pass both down... In fact it looks like I may not be using handleSubmit at all.  There are some unimplemented methods that seem like they would be useful but I'm not sure how best to implement.

src/Components/Wizard/types.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+// This interface is shared with StepMarker and could be two separate things
+export interface Step {
+  label: string
+  // TODO: Fix typing here and in WizardForm
+  component: any | React.ComponentType<RenderProps>
+  stepName: string
+  validate?: any
+}
+export type WizardSchema<S = Step> = S[]
```
erikdstock: As I mentioned elsewhere the typings between WizardSchema, Step, FormStep and the Steps used for StepMarker need to be better.

src/Components/Wizard/types.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export interface RenderProps {
+  validate?: () => any | FormikHandler
+  next: (v: any) => void
+  previous: (v: any) => void
+  onComplete: (v: any) => void
+  pageIndex: number
+  pages?: WizardSchema
+  values?: any
+}
```
erikdstock: The standard bag of props handed down to a wizard's child. Note that I'm not using the handleSubmit method I created in Wizard, leaving it up to the children to decide when to `next`/`onComplete`

*I really think it would be semantically nice to pass the pages in as children to the wizard itself* but I wasn't able to do that at first pass. Open to thoughts on it.

src/Components/Forms/support/Field.tsx
```diff
@@ -0,0 +1,46 @@
...
+ */
+export const Field: React.SFC<FormikInputProps> = props => {
+  const { name, type, placeholder, block, autoFocus } = props
+  return (
+    <FormikField
+      name={name}
+      type={type}
+      placeholder={placeholder}
+      // prettier-ignore
+      // because restProps should not have a trailing comma (fixed in prettier 1.9)
```
damassi: @erikdstock - can you upgrade to latest prettier to fix this?

src/Components/__stories__/WizardExamples/FormWizard.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <Wizard onComplete={submitForm} pages={pages}>
+      {(wizardBag: RenderProps) => (
+        <Container>
+          <StepMarker
+            steps={wizardBag.pages}
+            currentStepIndex={wizardBag.pageIndex}
+          />
+          <WizardForm {...wizardBag} container={Form} />
+          <pre>{JSON.stringify(wizardBag.values, null, 2)}</pre>
```
damassi: 👍

damassi:
This is really superb work @erikdstock 💯 I'll pull it down on Monday to give it a spin, but so far so good

src/Components/Wizard/Wizard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  pages: WizardSchema
+  children: React.ComponentType<RenderProps>
+}
+
+interface State {
+  pageIndex: number
+  values?: any
+}
+
+export class Wizard extends React.Component<Props, State> {
```
anandaroop: I like the API for this component 👍

src/Components/Forms/support/WizardForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  render() {
+    // Don't pass props.next to the child- use formik's onSubmit
+    const { initialValues, container: Form, next, ...wizardProps } = this.props
+    const ActiveComponent = this.activePage.component
+
+    return (
+      <Formik
+        initialValues={initialValues}
+        enableReinitialize={false}
```
anandaroop: #minor I think this defaults to `false`, so maybe not necessary to include here, unless being purposely explicit.

erikdstock:
One lingering question I have is around the `Field` I created in `src/Components/Forms`- It's really an amalgam of our `Input` and formik's `Field` (which hooks into Formik automa~g~tically via context)- but we will need similar components for bits like checkboxes and other inputs. Aside from their explicit integration with formik, I'm wondering what the meaning is to put them here when we already have Input and Checkbox files.

{ @damassi & @roop- thanks. I've resolved your comments }

l2succes:
@erikdstock Yes, I've thought about that as well, I think we should add a `component` prop to `Field` so the same way formik's `Field` component behaves so we can configure it to wrap other input types.
For instance

```typescript
   <Field component="select" name="color">
      <option value="red">Red</option>
      <option value="green">Green</option>
      <option value="blue">Blue</option>
    </Field>
```

We're introducing a lot of new abstractions in this PR already though so IMO It'd make sense to tackle that in another one.

src/Components/StepMarker/index.tsx
```diff
@@ -4,109 +4,57 @@ import Text from "../Text"
 import colors from "../../Assets/Colors"
 import styled from "styled-components"
 import { Fonts } from "../Publishing/Fonts"
-import { Props, State, StepProps } from "./types"
-import { isUndefined } from "lodash"
+import { Props, State, StepState } from "./types"
```
l2succes: Since these are external interfaces I would suggest using more specific names. (i.e. `StepProps`, `StepState`)

src/Components/Wizard/types.tsx
```diff
@@ -0,0 +1,23 @@
...
+// This interface is shared with StepMarker and could be two separate things
+export interface Step {
+  label: string
+  // TODO: Fix typing here and in WizardForm
+  component: any | React.ComponentType<RenderProps>
+  stepName: string
+  validate?: any
+  validationSchema?: any
+}
+export type WizardSchema<S = Step> = S[]
```
l2succes: I'd name this variable `steps: WizardSteps`. Just to be aligned with the `Step` component

src/Components/Wizard/types.tsx
```diff
@@ -0,0 +1,23 @@
+import { FormikHandler } from "../Forms/support"
+
+export interface RenderProps {
+  validate?: () => any | FormikHandler
+  validationSchema?: () => any | FormikHandler
+  next: (v: any) => void
+  previous: (v: any) => void
+  onComplete: (v: any) => void
+  pageIndex: number
+  pages?: WizardSchema
```
l2succes: I'd name this variable `steps: WizardSteps`. Just to be aligned with the `Step` component

src/Components/Forms/support/index.tsx
```diff
@@ -0,0 +1,5 @@
+import { FormikActions } from "formik"
+import { WizardForm } from "./WizardForm"
+export { WizardForm }
+
+export type FormikHandler = <V>(values: V, bag?: FormikActions<V>) => any
```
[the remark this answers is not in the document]
erikdstock: You know I was trying to have an abstract type for defining all formik actions and I'm not sure this really does it- the official formik types themselves aren't nearly this simple. I'm also not clear on when types need to be exported- Will remove for now.

src/Components/Forms/support/WizardForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <Formik
+        initialValues={initialValues}
+        validate={this.activePage.validate}
+        validationSchema={this.activePage.validationSchema}
+        onSubmit={this.handleSubmit}
+        render={formikRenderProps => {
+          const { handleSubmit } = formikRenderProps
+          return (
+            <Form onSubmit={handleSubmit}>
+              <ActiveComponent form={formikRenderProps} wizard={wizardProps} />
```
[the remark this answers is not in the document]
erikdstock: I've wondered about this but formik examples always pass the handler to both the form and other buttons.

PR #651, opened 2018-04-24
Revert "@kanaabe => Type system update"

l2succes:
Reverts artsy/reaction#629

Need to revert and bump version since that PR has some breaking changes

PR #653, opened 2018-04-24
Major version bump

l2succes:
Reverts artsy/reaction#651


l2succes:
@eessex Awesome. I'll merge that one first then the README fix PR

PR #654, opened 2018-04-24
[Dev] updates to compatible version of babel-loader

l2succes:
`babel-loader@8.0.0-beta.0` was breaking the module-resolver plugin which, in turn, was causing compilation errors in webpack


l2succes:
Nevermind! false alert. Tested on a clean pull and no errors.

PR #655, opened 2018-04-26
@xtina-starr => ToolTip parent component, GeneToolTip

eessex:
- Adds ToolTip that accepts a model and entity data, to display styled box w/ arrow and select sub-component
- Adds GeneToolTip component to show gene data
- Adds placeholder FollowButton with styles, assuming this will be replaced with Relay logic
- Adds NewFeature component to show feedback info

One thing I noticed is that the image is blurry -- changing the quality doesn't fix, maybe there is a better query for this?

Also note that the text is in italics because it is markdown from rich text.

screen shot 2018-04-26 at 12 44 54 pm

### DIFF
--- a/src/Components/Publishing/ToolTip/Components/ArrowDown.tsx
+++ b/src/Components/Publishing/ToolTip/Components/ArrowDown.tsx
@@ -1,0 +1,39 @@
+import styled from "styled-components"
+import React from "react"
+
+export const ArrowDown: React.SFC = () => {
+  return (
+    <ArrowContainer>
+      <Arrow />
+      <ArrowShadow />
+    </ArrowContainer>
+  )
+}
+
+export const ArrowContainer = styled.div`
+  position: relative;
+`
+
+const Arrow = styled.div`
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 0;
+  height: 0;
+  border-left: 20px solid transparent;
+  border-right: 20px solid transparent;
+  border-top: 20px solid white;
+  z-index: 1;
+`
+
+const ArrowShadow = styled.div`
+  position: absolute;
+  top: 5px;
+  left: 5px;
+  width: 0;
+  height: 0;
+  border-left: 15px solid transparent;
+  border-right: 15px solid transparent;
+  border-top: 15px solid white;
+  filter: drop-shadow(0 0 4px rgba(0, 0, 0, 0.15));
+`
--- a/src/Components/Publishing/ToolTip/Components/FollowButton.tsx
+++ b/src/Components/Publishing/ToolTip/Components/FollowButton.tsx
@@ -1,0 +1,18 @@
+import styled from "styled-components"
+import React from "react"
+import { unica } from "Assets/Fonts"
+import Colors from "Assets/Colors"
+
+export const FollowButton: React.SFC = () => {
+  return <FollowButtonContainer>Follow</FollowButtonContainer>
+}
+
+const FollowButtonContainer = styled.div`
+  border: 1px solid ${Colors.grayRegular};
+  ${unica("s12", "medium")};
+  width: 80px;
+  height: 24px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`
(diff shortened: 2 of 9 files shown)


src/Components/Publishing/ToolTip/Components/ArrowDown.tsx
```diff
@@ -0,0 +1,39 @@
...
+const ArrowShadow = styled.div`
+  position: absolute;
+  top: 5px;
+  left: 5px;
+  width: 0;
+  height: 0;
+  border-left: 15px solid transparent;
+  border-right: 15px solid transparent;
+  border-top: 15px solid white;
+  filter: drop-shadow(0 0 4px rgba(0, 0, 0, 0.15));
```
kanaabe: 👍

src/Components/Publishing/ToolTip/Components/FollowButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const FollowButtonContainer = styled.div`
+  border: 1px solid ${Colors.grayRegular};
+  ${unica("s12", "medium")};
+  width: 80px;
+  height: 24px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`
```
kanaabe: I think ideally we will want to refactor [this](https://github.com/artsy/reaction/blob/master/src/Components/Follow.tsx) instead? We can push that task to the Follow ticket, but just as an FYI.

src/Components/Publishing/ToolTip/Gene.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Truncator } from "../Sections/Truncator"
+import { FollowButton } from "./Components/FollowButton"
+import { NewFeature, NewFeatureContainer } from "./Components/NewFeature"
+
+export interface GeneProps {
+  description?: string
+  href?: string
+  image?: {
+    cropped: {
+      url: string
```
kanaabe: Can you link to the query for this? Maybe there are other options available in Gravity that just hasn't been exposed to MP?

src/Components/Publishing/ToolTip/Components/FollowButton.tsx
eessex: Yeah this is is a placeholder component as noted in the ToolTip file -- i expect we will just wrap the existing one in these styles.

src/Components/Publishing/ToolTip/Gene.tsx
kanaabe: (for blurriness)

eessex: I'm not sure what you mean?  There is no query being made atm, assuming that will be covered by the relay work.

eessex:
```{
  gene(id: "graphic-design") {
  	name
    image {
      cropped(width: 240, height: 160) {
        url
      }
    }
    description
    href
  }
}
```

eessex: Although I'm inserting an exact size, crop seems to be upscaling from a smaller image.

kanaabe: Yes, I know there is no query being made, just wanted to see the query you got the mock data from so we can see if there is a non-blurry version we can use instead. The query you provided is what I was looking for -- feel free to task the cropping issue for MP in Jira so we don't forget if you don't want to look into that now.

src/Components/Publishing/ToolTip/Components/NewFeature.tsx
```diff
@@ -0,0 +1,28 @@
+import styled from "styled-components"
+import React from "react"
+import { unica } from "Assets/Fonts"
+import Colors from "Assets/Colors"
+
+const FEEDBACK_EMAIL = "mailto:productfeedback@artsy.net?subject=Feedback"
```
damassi: 👍

xtina-starr:
This looks great @eessex!

PR #656, opened 2018-04-26
[WIP] Wizard update for mobile sign up form

l2succes:


xtina-starr:
Oops, I didn't mean to rebuild this! Looks like there is a legit failure though. Wondering should the `validationSchema` prop be optional in the wizard or at least passed everywhere?

src/Components/Wizard/Wizard.tsx
```diff
@@ -17,6 +17,7 @@ interface WizardProps {
       ) => React.ReactElement<any>)
     | StepElement
     | StepElement[]
+  validationSchema: object
```
erikdstock: Why is this needed? I've validationSchema is a property on `Step`.

erikdstock: Ah, ok. Thanks.

PR #658, opened 2018-04-27
@l2succes => Caption bug

eessex:
Adds an error boundary up one level -- captions will still fail, but the truncated section that triggers and error will display as white rather than crashing the page.


src/Components/Publishing/Sections/ArtworkCaption.tsx
```diff
@@ -42,14 +43,18 @@ export class ArtworkCaption extends React.Component<ArtworkCaptionProps, null> {
     }
     const delimSpan = <span>{delimiter}</span>
 
-    return names.slice(1).reduce((prev, curr, i) => {
-      return prev.concat([delimSpan, curr])
-    },
-    [names[0]])
+    return names.slice(1).reduce(
```
l2succes: Umm... prettier still creeping in our PRs 🤔

PR #661, opened 2018-04-30
Mobile ui components

xtina-starr:
Included in this PR:
- Extend Metaphysics module to accept XAPP-TOKEN
- Wires up validations and includes custom email validation for mobile
- Adds progress bar and other styling

ToDO to complete mobile UI Auth components:

- [x] Address inputs changing sizes on different steps
- [x] Data from the previous step persists on the next step when it shouldn't.

mobile signup component

### DIFF
--- a/src/Components/Authorization/Mobile/RegisterForm.tsx
+++ b/src/Components/Authorization/Mobile/RegisterForm.tsx
@@ -78,6 +78,7 @@ export const MobileRegisterForm: FormComponentType = props => {
             block
             error={errors.password}
             name="password"
+            label="Password"
             placeholder="Password"
             type="password"
             value={values.password}
@@ -98,6 +99,7 @@ export const MobileRegisterForm: FormComponentType = props => {
             block
             error={errors.name}
             name="name"
+            label="Name"
             placeholder="Name"
             type="text"
             value={values.name}
@@ -116,13 +118,7 @@ export const MobileRegisterForm: FormComponentType = props => {
         const { currentStep } = wizard
         return (
           <Container>
-            <ProgressIndicator
-              percentComplete={
-                wizard.isLastStep
-                  ? 1
-                  : (wizard.currentStepIndex + 1) / wizard.steps.length
-              }
-            />
+            <ProgressIndicator percentComplete={wizard.progressPercentage} />
             <BackButton onClick={wizard.previous as any}>
               <Icon name="chevron-left" color="#cccccc" fontSize="20px" />
             </BackButton>
@@ -150,6 +146,7 @@ const Container = styled.div`
   flex: 1;
   align-items: stretch;
   position: relative;
+  min-width: 260px;
 `
 
 const Header = styled.div`
--- a/src/Components/Authorization/Mobile/RegisterForm.tsx
+++ b/src/Components/Authorization/Mobile/RegisterForm.tsx
@@ -2,7 +2,7 @@ import React from "react"
 import { Step, Wizard } from "../../Wizard"
 import { ProgressIndicator } from "../../ProgressIndicator"
 
-import { GrayFacebookButton } from "../commonElements"
+import { GrayFacebookButton, TOSCheckbox } from "../commonElements"
 import styled from "styled-components"
 import Input from "../../Input"
 import Button from "../../Buttons/Inverted"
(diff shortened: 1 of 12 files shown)


src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -40,58 +68,73 @@ export const MobileRegisterForm: FormComponentType = props => {
...
+              <Icon name="chevron-left" color="#cccccc" fontSize="20px" />
+            </BackButton>
             <Header>
-              <BackButton onClick={wizard.previous as any}>Back</BackButton>
               <Logo name="logotype" />
             </Header>
             {currentStep}
-            <Button onClick={wizard.next as any} block>
+            <Button
+              onClick={form.handleSubmit as any}
```
erikdstock: I had this issue as well :/

src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -40,58 +68,73 @@ export const MobileRegisterForm: FormComponentType = props => {
...
         return (
           <Container>
+            <ProgressIndicator
+              percentComplete={
+                wizard.isLastStep
+                  ? 1
+                  : (wizard.currentStepIndex + 1) / wizard.steps.length
+              }
+            />
+            <BackButton onClick={wizard.previous as any}>
```
l2succes: @erikdstock We need to figure out the right TS type so we don't need to add `as any` on those. Will look into it

src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -40,58 +68,73 @@ export const MobileRegisterForm: FormComponentType = props => {
...
+    <Wizard steps={steps}>
       {context => {
-        const { wizard } = context
+        const { form, wizard } = context
         const { currentStep } = wizard
         return (
           <Container>
+            <ProgressIndicator
+              percentComplete={
+                wizard.isLastStep
```
l2succes: I think we could abstract that logic into the wizard and expose it as `wizard.progressPercentage: number`.
which is a value from 0 to 1

src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -12,10 +13,36 @@ import Text from "../../Text"
...
+      return false
+    }
+    return true
+  })
+}
 
 export const MobileRegisterForm: FormComponentType = props => {
   const steps = [
-    <Step>
+    <Step validationSchema={Validators.email} onSubmit={checkEmail}>
```
kanaabe: 👍

src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -40,58 +68,69 @@ export const MobileRegisterForm: FormComponentType = props => {
...
+    <Wizard steps={steps}>
       {context => {
-        const { wizard } = context
+        const { form, wizard } = context
         const { currentStep } = wizard
         return (
           <Container>
+            <ProgressIndicator percentComplete={wizard.progressPercentage} />
+            <BackButton onClick={wizard.previous as any}>
+              <Icon name="chevron-left" color="#cccccc" fontSize="20px" />
```
kanaabe: Could use a pre-defined color here

src/Components/Input.tsx
```diff
@@ -79,9 +92,19 @@ class Input extends React.Component<InputProps, InputState> {
   }
 
   onChange = e => {
+    const inputFieldTouched = {}
+    inputFieldTouched[`${this.props.name}`] = true
+
+    if (this.props.setTouched) {
+      this.props.setTouched(inputFieldTouched)
```
l2succes: I would just do with ES6
```
this.props.setTouched({
  [this.props.name]: true
})
```

l2succes:
Assigned from that one comment, Looks Good!

src/Components/Authorization/Mobile/RegisterForm.tsx
```diff
@@ -12,25 +13,53 @@ import Text from "../../Text"
...
       {({
         wizard,
-        form: { errors, touched, values, handleChange, handleBlur },
+        form: { errors, touched, values, handleChange, handleBlur, setTouched },
       }) => (
         <div>
           <Input
             block
-            error={touched.email && errors.email}
+            error={errors.email}
```
erikdstock: Do we want the error to show immediately rather than after focus is lost? I noticed in the screen cap that as soon as you begin typing you get validation errors. You might also be able to use the new Artsy Input wrapped in a Formik Field at https://github.com/artsy/reaction/blob/master/src/Components/Forms/support/Field.tsx

kanaabe:
Based on screenshots, it looks like this is missing the TOSCheckbox

PR #662, opened 2018-04-30
@l2succes => Parse text for artsy links

eessex:
Util methods for tooltips, parses article text for artsy links and returns arrays of artsy ids by models.

### DIFF
--- a/src/Components/Publishing/Constants.ts
+++ b/src/Components/Publishing/Constants.ts
@@ -1,4 +1,7 @@
+import { compact, last } from "lodash"
+import cheerio from "cheerio"
 import moment from "moment-timezone"
+import url from "url"
 import { DateFormat } from "../Publishing/Typings"
 
 /**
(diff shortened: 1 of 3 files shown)


src/Components/Publishing/Constants.ts
```diff
@@ -133,3 +136,39 @@ export const formatTime = time => {
   const secondsStr = seconds < 10 ? "0" + seconds : seconds
   return minutesStr + ":" + secondsStr
 }
+
+export const getArtsySlugsFromArticle = article => {
```
l2succes: let's add some types in the method signatures to make public API use clearer

```typescript
interface ArtsySlugs {
  articleSlugs?: string[]
  geneSlugs?: string[]
  ...
}

export const getArtsySlugsFromArticle = (article: ArticleData): ArtsySlugs => {
  ...
}
```

src/Components/Publishing/Constants.ts
```diff
@@ -133,3 +136,39 @@ export const formatTime = time => {
...
+  let artists = getArtsySlugsFromHTML(articleBody, "artist")
+  let genes = getArtsySlugsFromHTML(articleBody, "gene")
+
+  return {
+    artists,
+    genes,
+  }
+}
+
+export const getArtsySlugsFromHTML = (html, model) => {
```
l2succes: see above ^

PR #665, opened 2018-05-01
@l2succes => Artist tooltip

eessex:
Artist genes will be covered in a separate PR once Metaphysics work is in.  For now, if `showMarketData` is true and no data is available, it reverts to artist `blurb`.

screen shot 2018-05-01 at 10 55 33 am
screen shot 2018-05-01 at 10 56 02 am


l2succes:
Looks great! 👏

PR #666, opened 2018-05-01
adding more options to Tooltip component

oxaudo:
I'm planning to reuse this component for the other `?` roll overs in force as requested here: https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=39&projectKey=EV&modal=detail&selectedIssue=EV-96 and made possible by @damassi here: https://github.com/artsy/stitch/pull/8

I need to be able to specify if the roll over should be shown left or right of the question mark. So adding those as props here. The current version will work because those props are optional and defaulted to current values.

Considering that is my first PR - all critique is welcome:)

### DIFF
--- a/src/Components/Tooltip.tsx
+++ b/src/Components/Tooltip.tsx
@@ -5,7 +5,7 @@ import colors from "../Assets/Colors"
 interface Props extends React.HTMLProps<HTMLDivElement> {
   message: string
   hoverAlign?: string
-  hoverWidth?: Number
+  hoverWidth?: number
 }
 
 const Div: StyledFunction<Props & React.HTMLProps<HTMLDivElement>> = styled.div
--- a/src/Components/Tooltip.tsx
+++ b/src/Components/Tooltip.tsx
@@ -4,6 +4,8 @@ import colors from "../Assets/Colors"
 
 interface Props extends React.HTMLProps<HTMLDivElement> {
   message: string
+  hoverAlign?: string
+  hoverWidth?: Number
 }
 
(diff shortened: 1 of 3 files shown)


oxaudo:
And also added this story to show the difference in settings. Not sure that is what we usually do here:)
screen shot 2018-05-01 at 3 05 19 pm

l2succes:
@oxaudo So actually we were planning on deprecating the `Tooltip` component and replacing it with `Popover` (https://github.com/artsy/reaction/blob/master/src/Components/Popover/index.tsx) which already supports placement in top, down, left and right position. The API is pretty flexible in terms of styling. See if it would work for your usecase, I can help with setting it up.

https://artsy.github.io/reaction/?selectedKind=Components%2FPopover&selectedStory=CSS&full=0&down=1&left=1&panelRight=0

l2succes:
@oxaudo If you already have `Tooltip` implemented in force, we can also merge this in. I just pointed it out cause it's not obvious we have another component designed for this exact use-case. Either way, let me know :)

oxaudo:
@l2succes - the idea here is that currently in force we are using this `?` tooltip in several places. And it's style id configured separately there for each usage. I wanted to reuse the style of this Tooltip component that is defined here in reaction and replace the usage here: https://github.com/artsy/force/blob/master/src/desktop/apps/artwork/components/commercial/templates/price.jade#L12 with this reaction one since it's exactly the same how design team want this `?` to look now.

oxaudo:
Ok @l2succes - merge?

l2succes:
Yup! 👏🎉 on your first reaction PR @oxaudo  🤗

PR #669, opened 2018-05-03
giving tooltip an option to be aligned vertically and horizontally

oxaudo:
@briansw also approved some font/spacing change here. This should be the roll over for the `?` that we use in force across the board.

### DIFF
--- a/src/Components/Tooltip.tsx
+++ b/src/Components/Tooltip.tsx
@@ -1,16 +1,19 @@
 import React from "react"
 import styled, { StyledFunction } from "styled-components"
 import colors from "../Assets/Colors"
+import { unica } from "Assets/Fonts"
 
 interface Props extends React.HTMLProps<HTMLDivElement> {
-  message: string
-  hoverAlign?: string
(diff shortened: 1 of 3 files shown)


src/Components/Tooltip.tsx
```diff
@@ -1,16 +1,19 @@
...
 import colors from "../Assets/Colors"
+import { unica } from "Assets/Fonts"
 
 interface Props extends React.HTMLProps<HTMLDivElement> {
-  message: string
-  hoverAlign?: string
+  hAlign?: string
   hoverWidth?: number
+  message: string
+  vAlign?: string
```
alloy: Is this ‘horizontal’ and ‘vertical’? If so, I’d prefer to call them that and have the options be entirely self-explanatory.

oxaudo: yea. I thought it's a pretty common abbreviation. But you are right - explicit is better. Updated

alloy:
Ace 👍

PR #670, opened 2018-05-03
@kanaabe => Relay Follow button

eessex:
Moved our tooltip `FollowButton` into the general /Components directory, and added a wrapper component `FollowArtistButton` that can attach Relay props and handle the follow mutation.

Will add another variation in a separate PR for `FollowGeneButton` once https://github.com/artsy/metaphysics/pull/1035 is merged.

### DIFF
--- a/src/Components/FollowButton/FollowArtistButton.tsx
+++ b/src/Components/FollowButton/FollowArtistButton.tsx
@@ -7,20 +7,13 @@ import {
 } from "react-relay"
 import { FollowButton } from "./Button"
 import * as Artsy from "../Artsy"
+import { FollowArtistButton_artist } from "../../__generated__/FollowArtistButton_artist.graphql"
 
-export interface Props
-  extends RelayProps,
-    React.HTMLProps<FollowArtistButton>,
+interface Props
+  extends React.HTMLProps<FollowArtistButton>,
     Artsy.ContextProps {
   relay?: RelayProp
-}
-
-interface RelayProps {
-  artist?: {
-    __id: string
-    id: string
-    is_followed: boolean | null
-  }
+  artist?: FollowArtistButton_artist
 }
 
 interface State {
@@ -80,7 +73,7 @@ export class FollowArtistButton extends React.Component<Props, State> {
 }
 
 export default createFragmentContainer(
-  FollowArtistButton,
+  Artsy.ContextConsumer(FollowArtistButton),
   graphql`
     fragment FollowArtistButton_artist on Artist {
       __id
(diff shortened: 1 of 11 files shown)


src/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+  RelayProp,
+} from "react-relay"
+import { FollowButton } from "./Button"
+import * as Artsy from "../Artsy"
+
+export interface Props
+  extends RelayProps,
+    React.HTMLProps<FollowArtistButton>,
+    Artsy.ContextProps {
+  relay?: RelayProp
```
alloy: I see you export this interface, but didn’t see where it was being used. I’d suggest not exporting an interface that has the `relay` prop, because that’s normally really a detail internal to the component.

src/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+interface Props
+  extends RelayProps,
+    React.HTMLProps<FollowArtistButton>,
+    Artsy.ContextProps {
+  relay?: RelayProp
+}
+
+interface RelayProps {
+  artist?: FollowArtistButton_artist
+}
```
alloy: Btw, this use of a separate `RelayProps` interface is a legacy pattern we used for a tool we previously had that would in-place update an interface by that name with the fields and types that were being queried. It’s not necessary to keep using that pattern, so your interface could probably just look like:

```ts
interface Props extends React.HTMLProps<FollowArtistButton>, Artsy.ContextProps {
  relay: RelayProp
  artist: FollowArtistButton_artist
}
```

(Note that I also removed the optionality from the `relay` and `artist` props, because I assume these will always be present, but that may well be an incorrect assumption.)

src/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <FollowButton
+        isFollowed={artist && artist.is_followed}
+        handleFollow={this.handleFollow}
+      />
+    )
+  }
+}
+
+export default createFragmentContainer(
+  FollowArtistButton,
```
alloy: Should this component not first be wrapped by the Artsy context component?

src/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <FollowButton
+        isFollowed={artist && artist.is_followed}
+        handleFollow={this.handleFollow}
+      />
+    )
+  }
+}
+
+export default createFragmentContainer(
+  Artsy.ContextConsumer(FollowArtistButton),
```
l2succes: 👍

PR #675, opened 2018-05-08
@l2succes => ToolTip stories

eessex:
- Adds story for `FollowArtistButton`
- Reworks `ArtistToolTip` stories
- Adds `showToolTipMarketData` prop to Article and pass it down to `LinkWithToolTip`
- Removes unused `node` prop from `LinkWithTooltip` component


l2succes:
👍 Well done

PR #676, opened 2018-05-08
@l2succes => Use relay for Gene Tooltip

eessex:
Updates GeneToolTip to use relay


l2succes:
👍Looks good, let's merge after resolving conflicts

PR #677, opened 2018-05-08
@alloy => FollowGeneButton

eessex:
Adds a button with mutations to follow a gene.

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -76,7 +76,7 @@ type Artist implements Node {
   _id: String!
   cached: Int
   alternate_names: [String]
-  articles(sort: ArticleSorts, limit: Int): [Article]
+  articles(sort: ArticleSorts, limit: Int, in_editorial_feed: Boolean): [Article]
   artists(
     # The number of Artists to return
     size: Int
@@ -280,7 +280,7 @@ type ArtistItem implements Node {
   _id: String!
   cached: Int
   alternate_names: [String]
-  articles(sort: ArticleSorts, limit: Int): [Article]
+  articles(sort: ArticleSorts, limit: Int, in_editorial_feed: Boolean): [Article]
   artists(
     # The number of Artists to return
     size: Int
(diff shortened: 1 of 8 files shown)


alloy:
Looks good! 👍 

Future thought: we could probably make the follow button generic by adding a ‘followable’ interface to these types, because the button react components and relay code seem otherwise identical.

alloy:
@eessex Let me know if you want to take a stab at that.

eessex:
@alloy yes definitely-- was trying to find a way to reuse parts of the `handleFollow` but got tripped up by naming conventions for the queries.  Are there examples in the code now of this kind of interface?

alloy:
Ah sorry, my comment was ambiguous. I was referring to a GraphQL interface. I.e. making it possible to use the same query for any ‘followable’ entity. I’ll write up a ticket with more details after dinner!

alloy:
Consider the following schema:

```graphql
interface Favourable {
  __id: ID!
  id: ID!
  isFavourited: Boolean!
}

type Artist implements Favourable {
  # ...
}

type Artwork implements Favourable {
  # ...
}

type Gene implements Favourable {
  # ...
}

type Profile implements Favourable {
  # ...
}
```

By using an interface and stating that all these entities can be favourited, we have now make the act of favouriting an entity generic and so a favourite button would not need to know about the real type and can query like so:

```graphql
fragment FavouriteButton_favorable on Favourable {
  __id
  id
  isFavourited
}
```

Parent components would compose the fragment onto concrete types, e.g.:

```graphql
fragment ArtistFollowButton_artist on Artist {
  name
  ... FavouriteButton_favorable
}
```

…and a component that uses it like so:

```jsx
<FavouriteButton title={this.props.artist.name} favourable={this.props.artist} />
```

I have a working example that you can play with here https://launchpad.graphql.com/7llzqqx88j

Let me know if this is clear enough or whether we should jump on a call.

alloy:
(Btw, the word ‘favourite’ is an example, I think these are all ‘collections’ in the core API, so maybe that’s the term we should use.)

PR #678, opened 2018-05-09
@eessex => [Tooltips] adds slight delay on hiding tooltips

l2succes:
This isn't perfect yet, we're gonna need to keep tweaking the logic on when to show/hide tooltips but it's better than where we are now.


eessex:
Possible to show a gif of this interaction?  Looks ok to be besides the ❌

eessex:
This should also include a release tag

src/Components/OverlayTrigger/index.tsx
```diff
@@ -16,7 +17,7 @@ export class OverlayTrigger extends React.Component<OverlayTriggerProps> {
           {children}
         </Container>
 
-        <Overlay target={this.target} {...props}>
+        <Overlay {...props} target={() => findDOMNode(this.target)}>
           {cloneElement(overlay, {
```
eessex: Is this related to a refs complaint?

src/Components/Publishing/ToolTip/LinkWithTooltip.tsx
```diff
@@ -50,32 +52,34 @@ export class LinkWithTooltip extends Component<Props, State> {
...
       <div />
     )
 
     const { show } = this.state
     return (
-      <OverlayTrigger show={show} placement="top" overlay={toolTip}>
+      <OverlayTrigger
+        show={show}
+        placement="top"
+        // container={this}
```
eessex: stray comment or intentional?

src/Components/Publishing/ToolTip/ToolTip.tsx
```diff
@@ -36,16 +36,21 @@ export class ToolTip extends React.Component<Props> {
     if (!this.props.entity) return null
 
     return (
-      <ToolTipContainer>
-        {this.getToolTip()}
-        <ArrowDown />
+      <ToolTipContainer style={{ ...this.props.style }}>
```
eessex: Where is props.style coming from?

src/Components/Publishing/ToolTip/LinkWithTooltip.tsx
l2succes: it's a stray comment from experimenting

l2succes:
@eessex I'm still working out some kinks to make this better, will ping you when it's ready

l2succes:
@eessex Ok I think it's good enough for now

src/Components/OverlayTrigger/index.tsx
l2succes: Yes, I believe that's where that error was coming from.

PR #679, opened 2018-05-09
@xtina-starr => Use nested directories in publishing stories

eessex:
A quick re-org of our stories to use the recently added nested subdirectories.  Makes navigation easier, and breaks out layout variations more legibly. 

screen shot 2018-05-09 at 1 12 16 pm


src/Components/Publishing/__stories__/Video.story.tsx
```diff
@@ -38,10 +39,12 @@ storiesOf("Publishing/Video", module)
...
     )
   })
-  .add("Video About", () => {
+
+storiesOf("Publishing/Video/About", module)
+  .add("About", () => {
     return <VideoAbout article={VideoArticle} />
   })
-  .add("Video About with edit props", () => {
+  .add("With edit props", () => {
```
damassi: 💯 on the cleanup

xtina-starr:
Nice!

PR #680, opened 2018-05-10
@l2succes => import URL

eessex:
The url parsing was causing 500 errors because URL is part of the window object.


l2succes:
@eessex Makes sense but just trying to understand. Didn't we agree to not enable SSR for tooltips as the data isn't needed for the initial page render?

PR #681, opened 2018-05-10
use onClick property instead of handleClick for changing the type of modal rendered

xtina-starr:
This simplifies the `<ChangeMode>` component to just use the `onClick` property.

### DIFF
--- a/src/Components/Authorization/DesktopModal.tsx
+++ b/src/Components/Authorization/DesktopModal.tsx
@@ -2,12 +2,13 @@ import React, { Component } from "react"
 import styled from "styled-components"
 import Icon from "../Icon"
 import Modal, { ModalProps } from "../Modal/Modal"
+import { ModalType } from "./Types"
 import Colors from "Assets/Colors"
 import { DesktopHeader } from "./DesktopHeader"
 
 export interface DesktopModalProps extends ModalProps {
   subtitle?: string
-  onTypeChange?: (type: any) => void
+  onTypeChange?: (type: ModalType) => void
 }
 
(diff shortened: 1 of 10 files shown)


src/Components/Authorization/DesktopModal.tsx
```diff
@@ -7,6 +7,7 @@ import { DesktopHeader } from "./DesktopHeader"
 
 export interface DesktopModalProps extends ModalProps {
   subtitle?: string
+  onTypeChange?: (type: any) => void
```
xtina-starr: typing this as `any` for now until we decide whether we want to use `register` or `signup`.

l2succes: @xtina-starr let's go with `signup`. I think it's clearer then `register` for the Sign Up Form

src/Components/Authorization/LoginForm.tsx
```diff
@@ -95,7 +95,7 @@ export const LoginForm: FormComponentType = props => {
                 <Text color={Colors.grayDark}>Remember me</Text>
               </Checkbox>
               <ForgotPasswordLink
-                handleClick={props.handleChangeMode("reset_password")}
+                onClick={() => props.handleChangeMode("reset_password")}
```
l2succes: we should also rename `handleChangeMode` to something clearer like `presentModal`

src/Components/Authorization/DesktopModal.tsx
xtina-starr: Was leaning towards signup myself.

PR #683, opened 2018-05-11
@mzikherman => [Artwork/Save] Support saves in non-relay artwork grids 

damassi:
Adds the ability to save artworks within contexts that don't directly support Relay. Depends on https://github.com/artsy/force/pull/2512.

How it works:
- Old force section wants to use new artwork bricks / grid
- Arbitrary data is fed into the grid from the outside as props
- Since there's no outer relay context managing data fetching and population, saves don't work since the typical Relay HOC environment is missing. Successful mutations don't re-inject props
- On the Force side of things we use `<ContextConsumer>` and pass in a user
- This boots up a new Relay environment at runtime that we can piggy back on for a `commitMutation` function call (saves), ignoring that the rest of the component display tree isn't using it directly
- During this update cycle we use `setState` to change the display state of the favorite, depending on the outcome of the mutation

saves

### DIFF
--- a/src/Components/Artwork/Save.tsx
+++ b/src/Components/Artwork/Save.tsx
@@ -1,4 +1,6 @@
+import { isNull } from "lodash"
 import React from "react"
+import * as RelayRuntimeTypes from "relay-runtime"
 import {
   commitMutation,
   createFragmentContainer,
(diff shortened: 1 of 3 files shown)


src/Components/Artwork/Save.tsx
```diff
@@ -1,92 +1,154 @@
...
+  relayEnvironment?: RelayRuntimeTypes.Environment
   useRelay?: boolean
 }
 
-export class SaveButtonContainer extends React.Component<Props, null> {
-  static defaultProps = {
-    useRelay: true,
-  }
+interface State {
+  is_saved: boolean
```
alloy: 🐫 ?

src/Components/Artwork/Save.tsx
```diff
@@ -1,92 +1,154 @@
...
+    static defaultProps = {
+      useRelay: true,
+    }
+
+    state = {
+      is_saved: null,
+    }
+
+    get isSaved() {
+      const isSaved = isNull(this.state.is_saved)
```
alloy: Just out of JS curiosity, does this do anything else besides `this.state.is_saved == null`

src/Components/Artwork/Save.tsx
```diff
@@ -1,92 +1,154 @@
...
+            },
           },
-        },
-        // TODO: Relay Modern: This is not working yet
-        optimisticResponse: {
-          saveArtwork: {
-            artwork: {
-              __id: artwork.__id,
-              is_saved: !this.props.artwork.is_saved,
+          // TODO: Relay Modern: This is not working yet
```
alloy: This actually does work, no? So this comment seems out of date?

alloy:
Nice one 👍

src/Components/Artwork/Save.tsx
damassi: I was matching this to the relay response since it will be refactored out, but will add a TODO so that's clearer

src/Components/Artwork/Save.tsx
damassi: [Nope](https://github.com/lodash/lodash/blob/master/isNull.js#L17) -- just reads more at a glance, but not necessary

src/Components/Artwork/Save.tsx
damassi: Will remove

src/Components/Artwork/Save.tsx
alloy: Thanks 👍

PR #684, opened 2018-05-14
@l2succes => Tooltip QA

eessex:
- Changes subject line for feedback email
- Sends artsy context to follow buttons and ToolTipData
- Shows tooltips in Feature articles

### DIFF
--- a/src/Components/Publishing/ToolTip/TooltipsDataLoader.tsx
+++ b/src/Components/Publishing/ToolTip/TooltipsDataLoader.tsx
@@ -1,19 +1,16 @@
 import React, { Component } from "react"
 import PropTypes from "prop-types"
 import { QueryRenderer, graphql } from "react-relay"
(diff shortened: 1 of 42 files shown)


src/Components/Publishing/ToolTip/TooltipsDataLoader.tsx
```diff
@@ -74,18 +81,24 @@ export class TooltipsDataLoader extends Component<Props> {
...
+      currentUser,
     }
   }
 
   render() {
     return this.props.children
   }
 }
+
+export const TooltipsData = Artsy.ContextConsumer(TooltipsDataLoader)
```
l2succes: I wonder if just exporting `ContextConsumer(TooltipsDataLoader` as `TooltipsDataLoader` in terms of keeping the name of the module the same as the main export as you would in redux but that's just a naming preference.

PR #686, opened 2018-05-15
Create MarketDataSummary component

oxaudo:
This adds MarketDataSummary that is essentially a copy of MarketInsights component but with a bit less data and slightly different logic to display Gallery Representation. (only top category of representation is displayed here)

@mzikherman - Not sure if there is a good way to consolidate somehow the shared code between MarketInsights and MarketDataSummary component. For now it was basically copy/pasted with bunch of code deleted. Let me know what you think.

This will be used in new artist header that will be introduced here:
https://github.com/artsy/force/pull/2520


oxaudo:
And that is how this component will look like: 
marketdata

oxaudo:
Yea... I was thinking about that. It also only shows the highest level of gallery representation for now(also totally possible to configure through code). I think it's still being debated if the other one have to do the same.

I'm up for trying to reuse the existing component, but in discussion with @alloy we kinda agreed that might be a better way before things shake out - do an updated copy and then consolidate if those stay similar enough.

But I'm game to refactor now. What do you think @alloy?

alloy:
> But I'm game to refactor now. What do you think @alloy?

Please do! 👍 Now that you have done it the copy/paste route and know exactly what’s needed to make the code shared, it should be straightforward.

As to @mzikherman’s point on how to do the query, here’s an example of a query with a conditional and how the parent component configures the child to render differently: https://stackoverflow.com/a/44792842/95397

src/Components/Artist/MarketDataSummary/MarketDataSummary.tsx
```diff
@@ -0,0 +1,204 @@
...
+              display(format: "0a")
+            }
+          }
+        }
+      }
+    }
+  `
+)
+
+interface RelayProps {
```
alloy: We no longer need to manually define types now that we have TypeScript support in our Relay fork. The Relay compiler emits typings as you make changes, [here’s](https://github.com/oxaudo/reaction/blob/06752b9212ae386c3f81208a58b824a870fc4406/src/__generated__/MarketDataSummary_artist.graphql.ts#L4-L48) the typing for your component. Use it like so:

```ts
import { MarketDataSummary_artist } from "../../../__generated__/MarketDataSummary_artist.graphql"

interface Props extends React.HTMLProps<MarketDataSummary> {
  artist: MarketDataSummary_artist
}
```

(Note that I also removed the `export` as you don't actually seem to need it in any other files and renamed it to the idiomatic `Props` name.)

oxaudo:
Cool. Will work on that then. Just need to confirm with @katarinabatina and @briansw if the logic of displaying only the top tier gallery representation work for both or should be only on the summary, not on expanded version.

oxaudo:
But also maybe it might make sense to merge as is here @alloy - so that we have a working version to test on Force side. And I can work on refactoring in parallel. Especially because this refactor will affect current Market Insights on the artwork page so might make sense to test separately.

I might even go with creating a third version of this component - which is combination of 2 and test it on staging with Force so that we are not interfering with current working version.

oxaudo:
@mzikherman, @alloy - I'm merging this as is since the other component is about to change. Will work on merging those 2 next sprint.

alloy:
👍

PR #687, opened 2018-05-16
@l2succes => Tooltip hover/position changes

eessex:
- Only open one tooltip at a time
- Removes Overlay component
- Adds logic to place components in links, aware of page boundaries
- Adds `target="_blank"` to tooltip links

tooltips


src/Components/Publishing/ToolTip/LinkWithTooltip.tsx
```diff
@@ -48,51 +63,101 @@ export class LinkWithTooltip extends Component<Props, State> {
...
 
-export const Link = styled.a`
+interface AProps {
+  onMouseEnter: any
+  onMouseLeave: any
+}
+
+const A: StyledFunction<AProps> = styled.div
+
+export const Link = A`
```
l2succes: I think is the appropriate syntax for this
 ```typescript
export const Link = styled.div.attrs<{onMouseEnter: any; onMouseLeave: any}>({})`
...
`
```

File: src/Components/Publishing/ToolTip/ToolTip.tsx [hunk 2]
```diff
@@ -46,8 +55,19 @@ export class ToolTip extends React.Component<Props> {
...
-export const ToolTipContainer = styled.div`
+interface DivProps {
+  onMouseEnter: any
+  onMouseLeave: any
+  positionLeft: number
+}
+
+const Div: StyledFunction<DivProps> = styled.div
+
+export const ToolTipContainer = Div`
```
l2succes: see above ^

src/Components/Publishing/ToolTip/GeneToolTip.tsx
```diff
@@ -36,6 +36,9 @@ export const GeneToolTip: React.SFC<GeneProps> = (props, context) => {
 
 const Wrapper = styled.div`
   width: 240px;
+  a:hover {
```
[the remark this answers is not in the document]
eessex: Tried putting it in the `styled.a` just below but it doesn't seem to take, I think this may be because of some styles in Force that are overriding it.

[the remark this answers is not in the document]
l2succes: Got it ok makes sense, I was just saying because we lose a bit of context when having this selector nested inside of `Wrapper`

src/Components/Publishing/Sections/__tests__/Text.test.tsx
```diff
@@ -8,72 +9,102 @@ import {
   TextFeatureArticle,
   TextStandardArticle,
 } from "../../Fixtures/Articles"
+import { ContextProvider } from "../../../Artsy"
```
damassi: Since @l2succes got absolute paths working we should start the process of incrementally migrating imports

eessex: Good to know, didn't realize we had this -- @l2succes have you posted this to any dev channels?

l2succes: agreed. @eessex Yes I did, I made an announce in the #dev channel

PR #689, opened 2018-05-16
@l2succes => Tooltip QA

eessex:
- Adds fade transitions
- Adds positioning above or below text
- Updates storybooks with context
- Uses absolute paths for gene links
- Tooltip links are clickable and have hover styles
- Linked text is not stripped from descriptions

tooltips-fade


src/Components/Publishing/ToolTip/LinkWithTooltip.tsx
```diff
@@ -44,10 +47,11 @@ export class LinkWithTooltip extends Component<Props, State> {
...
 
   componentDidMount() {
-    if (this.link) {
-      const position = findDOMNode(this.link).getBoundingClientRect()
-      this.setState({ position })
-    }
+    const SetupToolTipPosition = () => defer(this.setupToolTipPosition)
+    this.setupToolTipPosition()
+
+    window.addEventListener("scroll", SetupToolTipPosition)
```
l2succes: we should add `removeEventListener` calls in `componentWillUnmount` for those

src/Components/Publishing/ToolTip/ToolTip.tsx
```diff
@@ -2,10 +2,11 @@ import styled from "styled-components"
 import React from "react"
 import { ArtistTooltipContainer } from "./ArtistToolTip"
 import { GeneToolTipContainer } from "./GeneToolTip"
-import { ArrowDown, ArrowContainer } from "./Components/ArrowDown"
+import { Arrow, ArrowContainer } from "./Components/Arrow"
 
 interface Props extends React.HTMLProps<HTMLDivElement> {
   entity: object
+  isBelowContent?: boolean
```
l2succes: just a suggestion but I would use `position: "up" | "down"` as the API for placing the tooltip instead of a boolean here as I think it's clearer.

src/Components/Publishing/__stories__/ArticleStandard.story.tsx
```diff
@@ -1,7 +1,7 @@
 import { storiesOf } from "@storybook/react"
 import React from "react"
 import styled from "styled-components"
-import { Article } from "../Article"
+import { Article } from "Components/Publishing/Article"
```
l2succes: Thanks for adding support for this

src/Components/Animation/FadeTransition.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react"
+import { Transition } from "react-transition-group"
+
+const duration = 300
```
l2succes: Maybe make the duration overrideable via a prop

src/Components/Publishing/ToolTip/LinkWithTooltip.tsx
```diff
@@ -44,10 +47,11 @@ export class LinkWithTooltip extends Component<Props, State> {
   }
 
   componentDidMount() {
-    if (this.link) {
-      const position = findDOMNode(this.link).getBoundingClientRect()
-      this.setState({ position })
-    }
+    const SetupToolTipPosition = () => defer(this.setupToolTipPosition)
```
damassi: Aliases to other methods are typically prefixed with an underscore so that most of the body still matches 1:1 -- `_setupToolTipPosition`

src/Components/Publishing/__stories__/ArticleStandard.story.tsx
damassi: Ah, just remembered about the `roots` field in jest config: https://github.com/artsy/prediction/blob/master/package.json#L38

eessex: can try that too

l2succes:
👍

PR #690, opened 2018-05-17
@essex => [Display Ad] adds support for pixel tracking in ads

l2succes:
https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-596

One thing to note, for sizmek we assume the `pixel_tracking_code` is a URL here, so we'll have to make sure to let the team know to only include the url and not the script tags.

### DIFF
--- a/src/Components/Publishing/Display/DisplayPanel.tsx
+++ b/src/Components/Publishing/Display/DisplayPanel.tsx
@@ -88,7 +88,7 @@ export class DisplayPanel extends Component<Props, State> {
 
   @trackImpression(() => "panel")
   trackImpression() {
-    // Add 3rd party impression support
+    // noop
   }
 
   // TODO: This could be shared with <CanvasVideo />
--- a/src/Components/Publishing/Display/DisplayPanel.tsx
+++ b/src/Components/Publishing/Display/DisplayPanel.tsx
@@ -18,6 +18,7 @@ interface Props extends React.HTMLProps<HTMLDivElement> {
   isMobile?: boolean
   unit: any
   tracking?: any
+  renderPixelTracker?: (unit: any) => React.ReactElement<any>
 }
 
 interface State {
@@ -87,7 +88,7 @@ export class DisplayPanel extends Component<Props, State> {
 
   @trackImpression(() => "panel")
   trackImpression() {
-    // noop
+    // Add 3rd party impression support
   }
 
   // TODO: This could be shared with <CanvasVideo />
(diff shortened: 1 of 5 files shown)


src/Components/Publishing/Layouts/StandardLayout.tsx
```diff
@@ -177,3 +197,7 @@ const FooterContainer = styled.div`
     margin: 0 20px;
   `};
 `
+
+const TrackerImage = styled.img`
+  display: none;
```
eessex: if you use `display: none;` does it still track?

src/Components/Publishing/Layouts/StandardLayout.tsx
```diff
@@ -40,6 +40,23 @@ export class StandardLayout extends React.Component<
...
+    }
+    const isSizmek = url.indexOf("serving-sys") !== -1
+    if (isSizmek) {
+      url = url.replace(
+        "[timestamp]",
+        String(new Date().getMilliseconds() / 1000)
+      )
+    }
+
+    // handle doubleclick
```
eessex: assuming this is a TODO, if so would be good to label as such

src/Components/Publishing/Layouts/StandardLayout.tsx
```diff
@@ -40,6 +40,23 @@ export class StandardLayout extends React.Component<
     this.setState({ isTruncated: false })
   }
 
+  renderPixelTracker(unit) {
+    let url = unit.pixel_tracking_code
```
eessex: could this be `renderPixelTracker({ pixel_tracking_code })`

src/Components/Publishing/Display/DisplayPanel.tsx
```diff
@@ -87,7 +88,7 @@ export class DisplayPanel extends Component<Props, State> {
 
   @trackImpression(() => "panel")
   trackImpression() {
-    // noop
+    // Add 3rd party impression support
```
eessex: do we need to use the `track` library for this?  Looks like it isn't implemented either way, but assumed we would need to inject a script tag as well.

l2succes: Na, I thought we did but we actually don't need this her

src/Components/Publishing/Layouts/StandardLayout.tsx
l2succes: Yeah it still makes the request as long as width and height aren't set to 0

l2succes: Actually apparently even that is fine https://stackoverflow.com/a/18448053

PR #691, opened 2018-05-18
[Authentication] Rename components to disambiguate from authorization.

alloy:
Also made sure all stories are grouped together.

* Authentication is about identifying a person (sign in/up).
* Authorization is about whether a person has access to a resource.
* Auth is an ambiguous short-hand, the proper ones are authn and authz, but even better is no short-hand versions at all.

----

@kanaabe @xtina-starr @l2succes @erikdstock As fas as I could tell these are not yet being used outside of Reaction, is that correct? Because if they this change would be breaking and would need some more care.


alloy:
Also, should I start using this component yet for [my authentication needs](https://artsyproduct.atlassian.net/browse/EV-118)?

l2succes:
@alloy Yes! Totally agree the naming was off, so thanks for cleaning it up. 

This component is being used in this force WIP PR https://github.com/artsy/force/pull/2441 where we're going through all existing authentication models and making sure everything still works with the react ones but that's the only place so far.

alloy:
@l2succes Got it, dope 👍 Ok, I’ll just keep using the existing Force authentication form in my work and then it should automatically get migrated with your Force PR.

PR #692, opened 2018-05-21
[FEATURE] Add MarketDataSummaryre component

oxaudo:
This is to publish MarketDataSummary component that used here: https://github.com/artsy/force/pull/2520


damassi:
@oxaudo - I _think_ there needs to be a second argument in the brackets -- `[FEATURE something]`. Can you let me know if it successfully publishes?

eessex:
Yes, the second arg of context is required for a version bump!

oxaudo:
Oh.. thanks here! It indeed ignored this build. Redoing it.

PR #697, opened 2018-05-22
@l2succes => Tooltip Link QA

eessex:
Addresses remaining items from https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-627

- links can click through
- no weird z-index during fade animations
- fixup underline styles


src/Components/Publishing/Sections/StyledText.tsx
```diff
@@ -38,17 +40,30 @@ export const StyledText = div`
...
     background-repeat: repeat-x;
     background-position: bottom;
     &:hover {
-      color: ${props => (props.color === "black" ? "#999" : props.color)};
+      color: ${props =>
+        props.color === "black" ? Colors.grayDark : props.color};
       opacity:  ${props => (props.color === "black" ? "1" : ".65")};
     }
   }
+  ${PrimaryLink} {
```
l2succes: Is that how you're rendering a dotted line? can we keep this styling contained in the `PrimaryLink` styled component instead?

I understand why the `StyledText` component is written that way but it re-introduces some problems you get with CSS where it's hard to track down where some rules are coming from, that styled-components was designed to fix.

eessex: Yeah I can move it in there, the trade off is we need to use `!important` to override style declarations from the `PrimaryLink` component

eessex: Also thinking it is time to refactor the `StyledText` component to separate files by layout.

PR #698, opened 2018-05-23
@l2succes => Fix Gene follow

eessex:
Mutation for following a gene does not use the 'unfollow' input arg (found on artist follows).  Removing the arg fixes saving issues with following genes.


src/Components/FollowButton/FollowGeneButton.tsx
```diff
@@ -33,7 +33,6 @@ export class FollowGeneButton extends React.Component<Props> {
         variables: {
           input: {
             gene_id: gene.id,
-            unfollow: gene.is_followed,
```
l2succes: Does that mean we cannot toggle between followed/unfollowed on genes?

eessex: Gene works slightly differently from artist, the api accepts only the ID and it takes either the 'follow' or 'unfollow' action based on the existing value.

eessex: https://github.com/artsy/metaphysics/blob/b52664c65de27e587cefda5e27a74f06b679f906/src/schema/me/follow_gene.js

PR #699, opened 2018-05-23
@oxaudo => Use MarketDataSummary component in Article ToolTips

eessex:
A few small updates to the `MarketDataSummary` component so we can use it in the article tooltips as well, 
- adds 'showGenes' prop to `MarketDataSummary`
- updates article fixtures re: market data
- adds some additional unit tests for `MarketDataSummary`
- removes `ArtMarketData` component

With market data:
screen shot 2018-05-23 at 3 10 28 pm

With genes:
screen shot 2018-05-23 at 3 13 04 pm


src/Components/Publishing/ToolTip/ArtistToolTip.tsx
```diff
@@ -120,6 +122,10 @@ export const ArtistContainer = styled.div`
       color: black;
     }
```
oxaudo: I'm wondering if it makes sense to make this `new feature, tell us what you think` part of market data component. What you now have in `<NewFeature />` component - does that refer to market data or the whole roll over in general?

src/Components/Publishing/ToolTip/ArtistToolTip.tsx
```diff
@@ -120,6 +122,10 @@ export const ArtistContainer = styled.div`
...
   }
+  ${MarketDataSummaryContainer} {
+    ${unica("s12")};
```
oxaudo: It's a bit annoying that we have to make the size overwrite here. Wondering if there is a way to pass a prop - something like `text size - xsm, sm or medium` and based on that define style in the MarketData component?

src/Components/Artist/MarketDataSummary/MarketDataSummary.tsx
```diff
@@ -114,7 +115,18 @@ export class MarketDataSummary extends React.Component<Props, null> {
     return <div>Collected by major museums</div>
   }
 
+  renderArtistGenes = () => {
```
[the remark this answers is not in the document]
eessex: For the genes, would it make more sense to you if it was passed as a prop, for alternate data to show if empty? Can move the gene fetch back into the tooltip component that way.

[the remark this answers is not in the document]
oxaudo: In my mind - yes... Something like `noDataDisplay` or some other name along those lines. Don't feel too strongly about it though...

src/Components/Artist/MarketDataSummary/MarketDataSummary.tsx
```diff
@@ -123,6 +126,10 @@ export class MarketDataSummary extends React.Component<Props, null> {
           {this.renderPermanentCollection()}
         </MarketDataSummaryContainer>
       )
+    } else if (onEmptyText) {
+      return (
+        <MarketDataSummaryContainer>{onEmptyText}</MarketDataSummaryContainer>
+      )
```
oxaudo: 👍

src/Components/Artist/MarketDataSummary/__tests__/__snapshots__/MarketDataSummary.test.js.snap
```diff
@@ -12,16 +12,18 @@ exports[`MarketDataSummary renders correctly 1`] = `
   className="c0"
 >
   <div>
-    $63m
+    $63,312,500
```
oxaudo: Why is this changed?

eessex: Not sure, was there some other change outside of my pr that could have affected this?  I didn't make any changes to the market data rendering.

eessex: Actually this is because I am using a different fixture here: https://github.com/artsy/reaction/pull/699/files/13fb2ce08a6ac59da4db28d7bf33202eed683931#diff-9529e5842b7bcdd0f9b413d7120b6baeR13

eessex: Are you returning the data rounded from the api?  this is what returned from metaphysics for me: https://github.com/artsy/reaction/pull/699/files/13fb2ce08a6ac59da4db28d7bf33202eed683931#diff-e9af1869999b3d835b6944bbbfe05a8cR928

oxaudo: I think MP takes a parameter of `format` when `display` is requested there:

https://github.com/eessex/reaction/blob/13fb2ce08a6ac59da4db28d7bf33202eed683931/src/Components/Artist/MarketDataSummary/MarketDataSummary.tsx#L174

Not sure what changed in test data that we lost it here.

eessex: ok in that case i am not using that parameter in the fixture, and will update to reflect that.

PR #700, opened 2018-05-23
@kanaabe => News source does not require link to display

eessex:
Adds support for displaying a news source's title when a link is not present.
Will follow up with some minor changes to support in Positron once this is merged.


kanaabe:
LGTM! Btw we could probably condense these merge commits with a rebase but that can be next time 👍

PR #701, opened 2018-05-23
Gene test fix

eessex:
Not sure why these tests passed in https://github.com/artsy/reaction/pull/698, but they are failing in master (as they should). A quick fix for a broken build.


eessex:
Self merging since trivial

PR #704, opened 2018-05-29
@alloy => SSR Relay 

damassi:
#### WIP (but just about done) 

This adds a light routing scaffold around Relay via [`found-relay`](https://github.com/4Catalyzer/found-relay) for isomorphic Relay rendering. Additionally adds a UX helper `<PreloadLink />` which behaves just like react router's `<Link to='/some-path>` but instead of transitioning immediately on click will fetch data in the background and then transition on complete. 

It has two additional props: 

- `immediate`: if true, immediately fetch route data
- `onLoadingToggle`: a callback that executes when loading starts and stops that one can tie into an external loading setup

 (I can imagine adding a few more like `onHover` in the future.)

#### Questions / Comments: 
- Right now I placed these files under `src/Router` but it seems like it could be scoped / named a bit better? Maybe just put everything inside `src/Relay`? 
- This is v1 -- I imagine stuff will evolve once we implement first new app section

reaction-example

#### Example App: 

```javascript
// routes.js 
// ---------

export const routes = [
  {
    Component: Layout,
    path: '/isomorphic-relay-example',
    children: [
      {
        path: '/',
        Component: HomeRoute,
      },
      {
        path: '/about',
        Component: () => <div>About page!</div>,
      },
      {
        path: '/artist/:id',
        Component: ArtistRoute,
        query: graphql`
          query routes_ArtistRouteQuery($id: String!) {
            artist(id: $id) {
              ...ArtistRoute_artist
            }
          }
        `,
      }
    ]
  }
]

// server.js 
// --------


import { buildServerApp } from 'reaction/Router'
import { routes } from './routes'

app.get('/isomorphic-relay-example*', (req, res, next) => {
  buildServerApp(routes, req.url)
    .then(({ ServerApp }) => {
      res.send(renderToString(<ServerApp />))
    })
}

// client.js 
// --------

import { buildClientApp } from 'reaction/Router'
import { routes } from './routes'

buildClientApp(routes)
  .then(({ ClientApp }) => {
    ReactDOM.hydrate(<ClientApp />, document.getElementById('react-root'))
  })
```

workin-in-force


src/Components/Artsy.tsx
```diff
@@ -75,13 +96,16 @@ export class ContextProvider extends React.Component<ContextProps, null>
 
     this.relayEnvironment =
       props.relayEnvironment || createEnvironment(this.currentUser)
+
+    this.reactionRouter = props.reactionRouter
```
damassi: Need to come up with a better name. I was going to use `router` but when we use `withRouter` HOC from found it occupies that key.

One way around this would be to refactor `ContextConsumer` so that it supports renderProps as opposed to HOC injection.

package.json
```diff
@@ -19,6 +19,7 @@
   "peerDependencies": {
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
+    "react-router": "^4.2.0",
```
damassi: There are two areas in force that use React router, but the API between found and RR is basically the same. Will need to follow up and swap out.

src/Router/AppShell.tsx
```diff
@@ -0,0 +1,40 @@
...
+            ${loadableState ? loadableState.getScriptTag() : ""}
+
+            <script>
+              var __RELAY_BOOTSTRAP__ = ${serialize(
+                JSON.stringify(relayData || {}),
+                {
+                  isJSON: true,
+                }
+              )};
+            </script>
```
damassi: I came up with this approach when working on stitch's [`componentRenderer`](https://github.com/artsy/stitch/blob/master/src/componentRenderer.js#L35) and realized that we can just inline SSR bootstrap data directly in containing div rather than having to worry about the modifying the overall page shell. It would throw warnings due to clientside rehydration mismatch, but I learned tonight that React has a light escape hatch via `suppressHydrationWarning`; I just need to figure out how to type the div correctly so that TS doesn't get angry.

src/Router/PreloadLink.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    render() {
+      const _props = omit(
+        [
+          "currentUser",
+          "immediate",
+          "relayEnvironment",
+          "reactionRouter",
+          "routeConfig",
```
damassi: React will complain if props are passed down through to `<a>` tags, so ensure things are stripped.

src/Router/PreloadLink.tsx
```diff
@@ -0,0 +1,146 @@
...
+        ],
+        this.props
+      )
+
+      return (
+        <Found.Link onClick={this.handleClick} {..._props}>
+          {this.props.children}
+          {this.state.isLoading ? " [loading...]" : null}
```
damassi: This will eventually call back to `props.onLoading`, and apparently `react-relay-modern-network-layer` has `onProgress` functionality. This will improve UX.

src/Router/buildServerApp.tsx
```diff
@@ -0,0 +1,91 @@
...
+      const loadableState = await getLoadableState(<AppContainer />)
+
+      resolve({
+        ServerApp: props => (
+          <AppContainer
+            relayData={relayData}
+            loadableState={loadableState}
+            {...props}
+          />
+        ),
```
damassi: This intermediate component isn't the nicest compared `resolve({ ClientApp })` but it's needed to inject hydration data during the second pass.

package.json
```diff
@@ -37,8 +38,7 @@
     "start": "yarn storybook",
     "storybook": "node verify-node-version.js && concurrently --raw --kill-others 'yarn relay --watch' 'start-storybook -p 9001'",
     "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
-    "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && cp .env.example .env && yarn install && yarn dump-schema -- ../../data",
-    "sync-schema:localhost": "cd ../metaphysics && yarn dump-schema -- ../reaction/data",
+    "sync-schema": "graphql-fetch-schema https://metaphysics-staging.artsy.net -o data",
```
damassi: `graphql-cli get-schema` seems to be a bit broken right now (willing to debug if its a blocker or can factor it back in when we need something more extensive since this does exactly what it says.)

src/Router/AppShell.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { SFC } from "react"
+import serialize from "serialize-javascript"
+import { ContextProvider } from "../Components/Artsy"
+
+interface Props {
+  loadableState?: any
+  relayData?: any
+  provide?: any
+}
```
damassi: #todo

src/Router/relayEnvironment.ts
```diff
@@ -0,0 +1,53 @@
+// TODO: move this with other relay stuff
```
damassi: #todo

damassi:
@alloy - this is now ready for a pass

src/Components/Artsy.tsx
alloy: I guess I don’t really understand the problem. Do context keys within one tree have to be unique?

.storybook/webpack.config.js
```diff
@@ -78,7 +78,7 @@ module.exports = (baseConfig, env) => {
       stats: "errors-only",
     },
     resolve: {
-      extensions: [".js", ".jsx", ".ts", ".tsx"],
+      extensions: [".mjs", ".js", ".jsx", ".ts", ".tsx"],
```
alloy: Do we use `mjs`?

src/Components/Artsy.tsx
alloy: …and if that’s the problem, why do we need both?

src/Components/Artsy.tsx
```diff
@@ -120,8 +145,14 @@ export function ContextConsumer<P>(
...
-      const { currentUser, relayEnvironment } = this.context
-      const props = Object.assign({ currentUser, relayEnvironment }, this.props)
+      const { currentUser, relayEnvironment, reactionRouter } = this.context
+
+      const props = Object.assign(
+        { currentUser, reactionRouter, relayEnvironment },
+        this.props
+      )
+
+      // console.log(this.context)
```
alloy: Stray loggy

src/Router/relayEnvironment.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import {
+  RelayNetworkLayer,
+  urlMiddleware,
+  cacheMiddleware,
+  // loggerMiddleware,
+} from "react-relay-network-modern"
+
+interface RelayEnvironment extends Environment {
+  relaySSRMiddleware: RelayServerSSR | RelayServerSSR
```
alloy: This should probably be `RelayClientSSR | RelayServerSSR`?

.storybook/webpack.config.js
damassi: No, but `react-relay-network-modern` packages / publishes `.mjs` files and webpack will complain. I've increasingly started seeing other libraries do the same (I believe `graphql-js`, for example).

alloy: I see 👍

src/Router/AppShell.tsx
zephraph: Clever. I had some initial reservations about this, but looking [in the docs](https://reactjs.org/docs/react-dom.html#hydrate), it seems like it's fine. It only ignores the warning at that level so any hydration warnings for children _should_ be preserved. 

I think longer term hoisting it out of the AppShell would be ideal though.

damassi: for sure, I think once we replace the outer page shell with react and establish a proper line of communication thats def the way to go

src/Components/Artsy.tsx
damassi: @alloy - inside of `PreloadLink` we need access to `found`, which inserts itself as `router` inside of the context. But we also need data from far exterior such as `routes` and the `resolver` (for relay). If we set that outer context var to `router` then it will clobber props coming in due to duplicate keys. As a sort of wip i'm storing those outer values under `reactionRouter`.

src/Components/__stories__/RelayRouter.story.tsx
```diff
@@ -0,0 +1,61 @@
...
+            }
+          }
+        `,
+        prepareVariables: params => ({
+          artistID: "andy-warhol",
+        }),
+
+        // TODO: Figure out how to get around Storybooks + iFrame. For now only
+        // a single-route example is possible since cant push into top-level history.
+        // children: []
```
damassi: @alloy - This is a limitation I just discovered in Storybooks -- we can't push to history due to it being a sub-window iframe. Regardless, I added an example app with one route to show how a routing app can be built out in Reaction (meaning, render the page with the top level query and then go from there down the tree).

damassi:
@alloy - This PR is ready to go

src/Components/__stories__/RelayRouter.story.tsx
damassi: Actually -- kinda scratch that. Farce supports in memory routing updates so i've updated `buildClientApp` to support different history types and updated the example

src/Components/Artsy.tsx
alloy: Does found make itself publicly available through context or is it meant only for internal use of the router? If the latter, then found should imo rename their variable to not be so generic. Something like `_foundRouter`.

src/Relay/createEnvironment.ts
```diff
@@ -1,15 +1,91 @@
...
+  const isServer = typeof window === "undefined"
+  const relaySSRMiddleware = isServer
+    ? new RelayServerSSR()
+    : new RelayClientSSR(cache)
+
+  relaySSRMiddleware.debug = false
+
+  const headers = {
+    "Content-Type": "application/json",
+    "User-Agent": "Reaction",
```
alloy: Would be nice (in a next PR) to include the Reaction version number here.

src/Relay/createEnvironment.ts
```diff
@@ -1,15 +1,91 @@
...
+            "X-ACCESS-TOKEN": user && user.accessToken,
+          }
+        : headers,
+    }),
+    relaySSRMiddleware.getMiddleware({
+      lookup: true,
+    }),
+    cacheMiddleware({
+      size: 100, // max 100 requests
+      ttl: 900000, // 15 minutes
```
alloy: The TTL could probably be increased greatly, although I guess it doesn’t persist and so a user’s session doesn’t last much longer? FYI While we may still need to fine-tune it, for now we have the response caching TTL in Emission set to a day.

src/Router/AppShell.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const AppShell: SFC<AppShellProps> = props => {
+  const { loadableState, data, children, provide } = props
+
+  return (
+    <ContextProvider {...provide}>
+      <div>
+        <div
+          // FIXME: type
+          {...{ suppressHydrationWarning: true }}
```
alloy: Is there a specific reason to spread here?

src/Router/buildClientApp.tsx
```diff
@@ -0,0 +1,77 @@
...
+          case "browser":
+            return new BrowserProtocol()
+          case "hash":
+            return new HashProtocol()
+          case "memory":
+            return new MemoryProtocol("/")
+          default:
+            return new BrowserProtocol()
+        }
+      }
```
alloy: Just out of curiosity, why did you make this a function?

src/Router/types.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface Router {
+  routes: RouteConfig
+  resolver: any // FIXME
+}
+
+export interface AppShellProps {
+  loadableState?: {
+    getScriptTag: () => string
+  }
+  data?: Object[]
```
alloy: This should probably be `object[]`, very rarely should you use the uppercase version of these types.

typings/sharify.d.ts
```diff
@@ -16,6 +16,10 @@ declare module "sharify" {
      */
     export interface GlobalData {
       readonly METAPHYSICS_ENDPOINT: string
+      readonly CURRENT_USER?: {
+        id: string
+        accessToken: string
+      }
```
alloy: This doesn’t seem right, because this will make it possible again to do the wrong thing in SSR code. Is this meant for client-side code?

src/Router/AppShell.tsx
damassi: This was a quick way to get around a missing type, need to update

src/Router/buildClientApp.tsx
damassi: Just for list readability, like 

```javascript
const getHistoryProtocol = () => {
  switch (historyProtocol) {
    case "browser":
      return new BrowserProtocol()
    case "hash":
      return new HashProtocol()
    case "memory":
      return new MemoryProtocol("/")
    default:
      return new BrowserProtocol()
  }
}
```

reads better to my eye than 

```javascript 
if (historyProtocol === 'browser') {
  return new BrowserProtocol()
} else if (historyProtocol === 'hash') {
  return new HashProtocol()
} else if (historyProtocol === 'memory') {
  return new BrowserProtocol()
}
```

typings/sharify.d.ts
damassi: I'll remove this for now and type `(sd as any)` in the client code, and add a fixme for the next PR.

PR #705, opened 2018-05-29
@xtina-starr => Tooltip auth analytics

eessex:
Passes tracking args to `onOpenAuth`, used as options to `mediator` in Force.


xtina-starr:
This looks good to me!

PR #709, opened 2018-05-30
@l2succes => Refactors impression/click tracking for 3rd Party to be more generic

kanaabe:
This fixes the double impression view by passing in `renderTime` based on the unix timestamp at the moment it is rendered. Since this comes from a top-level component, we shouldn't have to worry about re-renders causing multiple impressions. 

Previously I had a version that used the article's published_at since it is a unique value but I realized that in order to bust the cache across multiple sessions it needs to be unique for every render. Sorta obvious now but that tripped me up at first :) 

On the Force side, we need to bump the version and add the `renderTime` prop for /news.


kanaabe:
@l2succes are we good to go on this one?

kanaabe:
It doesn't explicitly prevent re-renders, but it fixes the problem. We are now using a "static" cache buster (`renderTime` which is calculated on `componentDidMount` from the parent component). This value doesn't change on re-renders to the Display units which prevents the impression from firing multiple times.

l2succes:
Right makes sense, ok merging

PR #710, opened 2018-05-31
@kanaabe => Auth forms submission and redirect

l2succes:
The idea with moving some of the auth components from force to reaction is we have a system of auth components ready to be dropped in not only in force but in other apps as well.

This also moves the submission logic here.

TODO

- [x] Add tests


src/Components/Authentication/Types.ts
```diff
@@ -1,6 +1,10 @@
 import { FormikProps } from "formik"
 
-export type ModalType = "login" | "signup" | "reset_password"
+export enum ModalType {
```
xtina-starr: Nice!

xtina-starr:
@l2succes I'm definitely in favor of adding tests to this PR, but how do you feel about handling the errors for the backend in a separate PR? I'm suggesting that because this is a pretty significant change in that it moves logic that was previously in force into reaction. Once this is in it might make it easier for the other items to be worked on in parallel.

l2succes:
@xtina-starr Yes I agree, i just put it there to not forget but yes let's put that in a separate PR

src/Components/Authentication/helpers.ts
```diff
@@ -0,0 +1,36 @@
...
+      "Content-Type": "application/json",
+      "X-Requested-With": "XMLHttpRequest",
+    },
+    method: "POST",
+    credentials: "same-origin",
+    body: JSON.stringify(values),
+  })
+  const data = await res.json()
+  return data
+}
```
kanaabe: We'll want to make sure all of the same steps for signup/register in https://github.com/artsy/force/blob/master/src/desktop/models/logged_out_user.coffee are covered here too -- maybe this seems a little backwards, but have we considered passing up the Formik data up to Force and simply reuse the Backbone logic we already have for login/signup?

src/Components/Authentication/Desktop/Components/DesktopHeader.tsx
```diff
@@ -1,10 +1,9 @@
 import React, { SFC } from "react"
 import styled from "styled-components"
-import Text from "../Text"
-import Icon from "../Icon"
-import { DesktopModalProps } from "./DesktopModal"
+import Text from "Components/Text"
+import Icon from "Components/Icon"
```
damassi: nice

src/Components/Authentication/Desktop/ModalManager.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { Component } from "react"
+import { DesktopModal } from "./Components/DesktopModal"
+import { FormSwitcher } from "./FormSwitcher"
+import { handleSubmit } from "Components/Authentication/helpers"
+import { ModalType } from "Components/Authentication/Types"
+
+interface Props {
+  submitUrls: { [P in ModalType]: string }
```
damassi: @l2succes curious, what does this signature do?

l2succes: @damassi  `ModalType` is an `enum` type that looks like the following. 
```typescript
enum ModalType {
  login = "login"
  signup = "signup"
}
```
`{ [P in ModalType]: string }` essentially tells typescript that the values in the enum (denoted as `P`) are the only accepted keys in the `submitUrls` object.

PR #711, opened 2018-05-31
@kanaabe => [FIX tooltips] track clicks on tooltip hover background

eessex:
Addresses missing click event for tooltips.  (Hover background covers in-text link while tooltip is open.)


kanaabe:
LGTM!

PR #713, opened 2018-05-31
@orta => Remove circle slack notification setup

eessex:
To remove slack notifications on builds to the #front-end channel, also have disabled this in our circle/slack settings.


orta:
ace, thanks!

PR #715, opened 2018-05-31
@damassi => Upgrades typescript and enables type emitting

l2succes:
screen shot 2018-05-31 at 5 31 33 pm
This allows us to emit type definitions alongside compiled js files in our reaction builds so consumers that also use typescript (e.g. 'force') have access to them.

TODO
- [x] swap `build` from babel 7 back to `tsc`


tsconfig.json
```diff
@@ -22,21 +16,14 @@
     "rootDir": "src",
     "target": "es2015",
     "paths": {
-      "Assets/*": [
-        "./Assets/*"
-      ]
+      "Assets/*": ["./Assets/*"]
```
damassi: One thing we should try to do if possible is use a root directory similar to [here](https://github.com/artsy/reaction/blob/master/.babelrc#L28) so that we don't have to define individual keys.

tsconfig.json
```diff
@@ -22,21 +16,14 @@
...
     }
   },
   "include": [
+    "node_modules/styled-components/typings/styled-components.d.ts",
```
damassi: oh cool, i remember that for a while they didn't have a ts definition

damassi: Would it be possible to use this property rather than include? https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

damassi:
@l2succes  - @zephraph pointed out that the reason decorators might be breaking with babel is because the `legacy` option might not be set. Can you try [setting this](https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-decorators#legacy) in our babel config and see if switching from [`typescript-babel-jest`](https://github.com/artsy/reaction/blob/3e9f7a4b96b611403e18c672eb916cff5473e471/package.json#L169) to `babel-jest` fixes everything and all tests run?

l2succes:
@damassi I can look into it but switching to [`ts-jest`](https://github.com/kulshekhar/ts-jest) worked great.

I go through all of the type errors and verified `.d.ts` files are generating properly.

tsconfig.json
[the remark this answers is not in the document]
l2succes: Right, I removed that line as it was just an experiment to fix TS errors

damassi:
@l2succes - all good, was just curious if it compiled without errors, but nbd 👍

src/Components/index.tsx
```diff
@@ -1,18 +1,29 @@
...
+// @ts-ignore
+import Nav, { NavBarProps } from "./Nav"
+// @ts-ignore
+import Text, { TextProps } from "./Text"
 import TextArea from "./TextArea"
-import TextLink from "./TextLink"
-import Title from "./Title"
+// @ts-ignore
+import TextLink, { LinkProps } from "./TextLink"
+// @ts-ignore
```
damassi: @l2succes - I wonder if there is some way we could ignore imported interfaces without having to flag each unused import -- do you know if there is a TSLint / ts config option that would allow us to keep this clean?

Also, i'm pretty sure this is a shot in the dark, but any way to auto-export interfaces? 

@orta / @alloy - have y'all ever come across a good way to do this?

l2succes:
@artsy/engineering 

We just upgraded typescript to `2.9` which allows us to enable emitting types for reaction. What that means is you'll see some new errors if you don't make certain things public. Some changes,
for any new components you work on, you'll need to export the `Props` interface associated with the component. For example: 

```typescript
export interface IconProps {
  name: string;
}

const Icon: React.SFC<IconProps> = props => {
  ...
}
```

src/Components/index.tsx
l2succes: @damassi unfortunately, I don't think there's a way to ignore a whole block 😔 (Related: https://github.com/Microsoft/TypeScript/issues/19573)

They just fixed the main issue that enables us to emit types in the first place with typescript@2.9.1 since this issue is related I think they should give us a way to auto-import interfaces when generating types

damassi: Ahh cool cool, so I imagine we'll just have to track this as the feature becomes more mature 👍

damassi:
@l2succes - Can you update README with some info on type emission and notify #dev since errors due to non-exported types could be confusing?

src/Components/index.tsx
damassi: @alloy - in order for the types to be automatically emitted, it needs an in-scope reference to the interface  -- in the example above, `LinkProps` needs to be made available BUT because we're not using it directly in the code (only during the compilation process) it needs to be ignored due to the `noUnusedLocals` setting. 

This aspect of type emission kinda bugs me because it clutters up code with some developer vs compiler indirection, but I suspect now that this feature has landed there will be some overall tooling improvements coming soon that will help clean the process up.

PR #722, opened 2018-06-05
add authentication modal options interface

xtina-starr:
Related to https://artsyproduct.atlassian.net/browse/GROW-628

This adds a modal authentication interface and passes the entire options object via the openModal function.

### DIFF
--- a/src/Components/Authentication/Types.ts
+++ b/src/Components/Authentication/Types.ts
@@ -24,11 +24,27 @@ export interface FormProps {
   handleTypeChange?: (modalType: ModalType) => void
 }
 
+/**
+ * ModalOptions Description
+ *
+ * copy: the free string copy of the modal that was triggered.
+ *
+ * mode: the type of modal to display.
+ *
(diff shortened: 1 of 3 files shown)


src/Components/Authentication/Types.ts
```diff
@@ -24,4 +24,11 @@ export interface FormProps {
   handleTypeChange?: (modalType: ModalType) => void
 }
 
+export interface ModalOptions {
+  copy?: string
+  signupIntent?: string
```
l2succes: @xtina-starr I've seen `signupReferer` roaming around too. Like here https://github.com/artsy/force/blob/master/src/desktop/components/auth_modal/view.coffee#L58

Should we keep it? or is it safe to remove

xtina-starr: Good catch. I'll add it.

kanaabe:
Nice, this just needs a rebase with a commit message updated to use semantic versioning

PR #723, opened 2018-06-05
@kanaabe => update tsconfig.json

l2succes:
This fixes some TS errors we get in force from switching to `tsc`. 

Specifically, imports were not being converted to CommonJS/Node `require` calls.
Also

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
     "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && cp .env.example .env && yarn install && yarn dump-schema -- ../../data",
     "sync-schema:localhost": "cd ../metaphysics && yarn dump-schema -- ../reaction/data",
-    "test": "node verify-node-version.js && jest --no-cache",
+    "test": "node verify-node-version.js && jest",
     "test:watch": "jest --watch --runInBand",
     "type-check": "tsc --noEmit --pretty",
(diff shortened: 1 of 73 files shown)


kanaabe:
Looks like several failed tests on this, otherwise this LGTM

src/Components/Authentication/Desktop/ModalManager.tsx
```diff
@@ -9,28 +9,36 @@ import {
...
 }
 
-interface State {
-  currentType: ModalType
+export interface ModalManagerState {
+  currentType?: ModalType
 }
 
-export class ModalManager extends Component<Props, State> {
+export interface ModalOptions {
```
l2succes: @xtina-starr We'll need to use the interface you wrote here. Just putting a placeholder for now

kanaabe:
Ah, a couple new merge conflicts ^

package.json
```diff
@@ -41,10 +41,10 @@
     "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
     "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && cp .env.example .env && yarn install && yarn dump-schema -- ../../data",
     "sync-schema:localhost": "cd ../metaphysics && yarn dump-schema -- ../reaction/data",
-    "test": "node verify-node-version.js && jest",
+    "test": "node verify-node-version.js && jest --no-cache",
```
damassi: Should we keep `--no-cache` here by default or just leave it to devs to clear it on their own? Since we have so many tests these days the cache could help increase speed.

l2succes: I'll remove it, it was supposed to fix the decorators error but we can call it adhoc

PR #726, opened 2018-06-06
@zephraph => Move static pages to /Styleguide and add basic Docs scaffold

damassi:
This adds a basic Styleguide setup via `@storybooks/addons-info` and moves our files over from https://github.com/artsy/responsive-static-a-a-pages. 

screen shot 2018-06-06 at 10 37 10 pm

screen shot 2018-06-06 at 10 42 41 pm

Note that stuff in `public/assets/*` will eventually go away. This is just a quick fix for serving up our static asset image files.


src/Styleguide/theme.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+    body {
+      margin: 0;
+      padding: 0;
+    }
+
+    @font-face {
+      font-family: Unica77LL-Regular;
+      src: url(/assets/fonts/Unica77LL-Regular.otf);
+    }
```
damassi: These fonts will eventually move / correspond to our webfonts CDN.

package.json
```diff
@@ -150,6 +151,8 @@
     "react-markdown": "^2.5.0",
     "react-oembed-container": "^0.3.0",
     "react-overlays": "^0.8.3",
+    "react-primatives": "^0.0.5",
```
zephraph: Probably should remove this...

zephraph:
For now, will you remove the dependencies too?

package.json
damassi: roger that

tslint.json
```diff
@@ -5,6 +5,7 @@
     "interface-name": [true, "never-prefix"],
     "jsx-boolean-value": [true, "never"],
     "jsx-no-lambda": false,
+    "jsx-no-bind": false,
```
damassi: Not sure why rebasing from master started err'ing out on this; there are areas in our code that do use `bind` and would like to avoid refactoring those right now. (This should probably be a warning anyway since its something of a premature optimization, but would still fail in CI if so.)

src/Styleguide/Elements/Typography.tsx
```diff
@@ -0,0 +1,73 @@
...
+// Example: <Sans size="2" weight="medium" italic>
+
+export const Sans: any = props => (
+  <StyledText family="unica" typeSize={`sans${props.size || 4}`} {...props} />
+)
+
+export const Serif: any = props => (
+  <StyledText
+    family="garamond"
+    typeSize={`serif${props.size || 4}`}
```
alloy: @zephraph Why are we calling this prop `size` instead of `modifier` as the design team does?

alloy: Also, looking at https://www.notion.so/artsy/Typography-d1f9f6731f3d47c78003d6d016c30221, it seems to me that `3` is the default (body), am I reading that incorrectly?

alloy:
It looks like the react-primitives dep is still in, but as the branch for this PR lives on a fork I can’t easily amend it, so I’m going to merge this as-is so I can get started and we can follow up when y’all wake up.

package.json
alloy: Woah, what even is this package that’s spelled ‘primitives’ instead of ‘primitives’? https://github.com/DavidWells/react-dom-primitives

PR #727, opened 2018-06-07
[Styleguide] prep cleanup

alloy:
* Removes unused dependency
* Splits single styleguide story file into story files colocated with their components, otherwise we end up needing to rewrite all imports later on when we move them


src/Styleguide/Elements/__stories__/Typography.story.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <Sans size="4">Donald Judd</Sans>
+            <Sans size="4t">Donald Judd</Sans>
+            <Sans size="5">Donald Judd</Sans>
+            <Sans size="5t">Donald Judd</Sans>
+            <Sans size="6">Donald Judd</Sans>
+            <Sans size="8">Donald Judd</Sans>
+            <Sans size="10">Donald Judd</Sans>
+            <Sans size="12">Donald Judd</Sans>
+            <Sans size="14">Donald Judd</Sans>
+            <Sans size="16">Donald Judd</Sans>
```
zephraph: @alloy, you'd asked about size in the other PR. I sat down w/ Brian and asked him about what semantics would make the most sense for this component. That's what recommended. Still, it is a valid point that the Typography doc uses `modifier` instead of `size`.

zephraph: I'll ask in the design systems chat.

PR #728, opened 2018-06-07
[WIP] Add new Artwork Banner

zephraph:


src/Styleguide/Artwork/Banner.tsx
```diff
@@ -0,0 +1,40 @@
...
+  width: ${(props: RoundedImageProps) => props.size};
+  height: ${props => props.size};
+  border-radius: ${props => props.size};
+`
+
+export class Banner extends React.Component {
+  render() {
+    return (
+      <Row>
+        <Col sm={8}>
```
damassi: @zephraph - curious about your take on this, but if possible we should keep all Row / Column dependencies out of components so that they're pure, and can be used in a variety of environments. Seems like these elements should be hoisted up to the page level.

src/Styleguide/Artwork/Banner.tsx
```diff
@@ -0,0 +1,40 @@
...
+  display: flex;
+  flex-direction: row;
+`
+
+interface RoundedImageProps {
+  size: string
+}
+
+const RoundedImage = styled.img.attrs<RoundedImageProps>({})`
+  width: ${(props: RoundedImageProps) => props.size};
```
damassi: This isn't needed now that `styled.img.attrs<RoundedImageProps>` is there, right?

src/Styleguide/Artwork/Banner.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export class Banner extends React.Component {
+  render() {
+    return (
+      <Row>
+        <Col sm={8}>
+          <Flex>
+            <RoundedImage
+              size="110px"
+              src="https://picsum.photos/110/110/?random"
```
oxaudo: Wondering if/where/how we would deal with possibility of image is not present here?

Not sure it's a very likely scenario in this case, but it will be very much likely if we display gallery logo/artist image in  this avatar style.

src/Styleguide/Artwork/Banner.tsx
```diff
@@ -0,0 +1,28 @@
...
+  flex-direction: row;
+`
+
+export class Banner extends React.Component {
+  render() {
+    return (
+      <Row>
+        <Col sm={8}>
+          <Flex>
+            <Avatar size="110px" src="https://picsum.photos/110/110/?random" />
```
oxaudo: I know we are building a static page for now but wondering if/where/how we would deal with possibility of image is not present here?

Not sure it's a very likely scenario in this case, but it will be very much likely if we display gallery logo/artist image in this avatar style.

src/Styleguide/Artwork/Banner.tsx
zephraph: Yeah, I think that makes a lot of sense.

damassi:
Going to merge per @oxaudo's request, we can iterate

PR #729, opened 2018-06-07
Add stylelintrc and commands to run linter

zephraph:
Ports over Emission's stylelint setup


alloy:
This looks good 👍 

We might also need to add these suggested vscode extensions https://github.com/artsy/emission/blob/e59c5c65aced1bfbf67446a1395a978231ab62f6/.vscode/extensions.json#L11-L12. Can you see which work for us here too and PR that as well?

PR #730, opened 2018-06-07
@kanaable: Mobile wizard

xtina-starr:
This resolves the merge conflicts for the out dated mobile register form and brings it to a working state. We will still want to fix the acceptedTermsOfService checkbox, but I think that can be addressed in a separate PR.


kanaabe:
`Yup`! 😸

Also, reminder to make a commit for semantic release with `FEATURE`. Can't wait to get this merged!

PR #731, opened 2018-06-07
@xtina-starr => Text input QA

kanaabe:
Addresses work in: https://artsyproduct.atlassian.net/browse/GROW-533

Password messaging
image

Tabbing (i promise i'm tabbing here 😄). Modal heights will be addressed next in https://artsyproduct.atlassian.net/browse/GROW-693 
input-tabbing

### DIFF
--- a/src/Components/Input.tsx
+++ b/src/Components/Input.tsx
@@ -147,8 +147,8 @@ export class Input extends React.Component<InputProps, InputState> {
         <Container>
           <InputContainer
             hasLabel={!!label}
-            focused={this.state.focused}
             hasError={!!error}
+            className={this.state.focused ? "focused" : ""}
           >
             <Label out={!showLabel}>{label}</Label>
             {!!leftView && leftView}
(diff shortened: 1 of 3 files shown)


src/Components/Input.tsx
```diff
@@ -146,6 +148,13 @@ export class Input extends React.Component<InputProps, InputState> {
               ? this.getRightViewForPassword()
               : !!rightView && rightView}
           </InputContainer>
+          {!error && showPasswordMessage ? (
```
l2succes: @kanaabe Why not just use the standard `error` field/system? Did it not work?

kanaabe: Ah, because this is separate from an error - we want to show that password message always, _unless_ there is an error. Is that what you're wondering or something else?

xtina-starr:
@kanaabe this looks good to me! Looks like there is a conflict...probably due to the mobile work that got merged in yesterday.

PR #734, opened 2018-06-07
Implement banner for new artwork page

zephraph:
Second pass at refining the banner for the new artwork page.

The Flex component could likely be shared... also, we'll want to start breaking out the types for styled-system being as it doesn't provide its own.

### DIFF
--- a/src/Styleguide/Artwork/Banner.tsx
+++ b/src/Styleguide/Artwork/Banner.tsx
@@ -1,54 +1,14 @@
 import React from "react"
-import styled from "styled-components"
-import {
-  flexDirection,
-  alignItems,
-  alignContent,
-  justifyContent,
-  space,
-} from "styled-system"
 import { Avatar } from "../Elements/Avatar"
 import { Sans, Serif } from "../Elements/Typography"
-
-interface FlexProps {
-  flexDirection?: "row" | "column"
-  alignItems?: "flex-start" | "flex-end" | "center" | "baseline" | "stretch"
-  alignContent?:
-    | "flex-start"
-    | "flex-end"
-    | "center"
-    | "space-between"
-    | "space-around"
-    | "stretch"
-  justifyContent?:
-    | "flex-start"
-    | "flex-end"
-    | "center"
-    | "space-between"
-    | "space-around"
-    | "space-evenly"
-  m?: string
-  mr?: string
-  ml?: string
-  mt?: string
-  mb?: string
-}
-
-const Flex = styled.div.attrs<FlexProps>({})`
-  display: flex;
-  ${flexDirection};
-  ${alignItems};
-  ${alignContent};
-  ${justifyContent};
-  ${space};
-`
+import { Flex } from "../Elements/Flex"
 
 export class Banner extends React.Component {
   render() {
     return (
       <Flex flexDirection="row">
         <Avatar size="110px" src="https://picsum.photos/110/110/?random" />
-        <Flex flexDirection="column" justifyContent="center" ml="20px">
+        <Flex flexDirection="column" justifyContent="center" ml={4}>
           <Sans weight="medium" size="2">
             In show
           </Sans>
--- a/src/Styleguide/Artwork/Banner.tsx
+++ b/src/Styleguide/Artwork/Banner.tsx
@@ -1,22 +1,30 @@
 import React from "react"
 import styled from "styled-components"
+import { flexDirection } from "styled-system"
 import { Avatar } from "../Elements/Avatar"
+import { Sans } from "../Elements/Typography"
 
-const Flex = styled.div`
+interface FlexProps {
+  flexDirection: "row" | "column"
+}
+
+const Flex = styled.div.attrs<FlexProps>({})`
   display: flex;
-  flex-direction: row;
+  ${flexDirection};
 `
 
 export class Banner extends React.Component {
   render() {
     return (
-      <Flex>
+      <Flex flexDirection="row">
         <Avatar size="110px" src="https://picsum.photos/110/110/?random" />
-        <div>
-          <div>In show</div>
+        <Flex flexDirection="column">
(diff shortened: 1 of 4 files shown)


src/Styleguide/Artwork/Banner.tsx
```diff
@@ -1,22 +1,62 @@
...
-        </div>
+        <Flex flexDirection="column" justifyContent="center" ml="20px">
+          <Sans weight="medium" size="2">
+            In show
+          </Sans>
+          <Serif size="4t">Francesca DiMattio: Boucherouite</Serif>
+          <Serif size="4t" color="black60">
+            Salon 94
+          </Serif>
+        </Flex>
```
damassi: Ah, now I understand why this `<Flex />` is broken out -- so that composing elements can remain neutral and live within the system. @yuki24 did something similar over in reaction. 

Would you mind pushing this into a separate component before merge?

zephraph: Sure thing. Elements?

src/Styleguide/Artwork/Banner.tsx
```diff
@@ -1,22 +1,62 @@
...
     return (
-      <Flex>
+      <Flex flexDirection="row">
         <Avatar size="110px" src="https://picsum.photos/110/110/?random" />
-        <div>
-          <div>In show</div>
-          <div>Francesca DiMattio: Boucherouite</div>
-          <div>Salon 94</div>
-        </div>
+        <Flex flexDirection="column" justifyContent="center" ml="20px">
```
damassi: Ahh -- `ml='20px'` -- can we stick to [palette's `space` values](https://github.com/artsy/palette/blob/master/src/Theme.tsx#L89) at all times and religiously avoid one-off's? I spoke to brian and he said that in instances where the pixel values slightly diverge between what is in the zeplin redlines and what is in the design that we should round. (Also, at this point, any time an artibitrary `px` value is seen it should be considered a code smell, and in styled-component classes we should always be using the `getTheme()` styled-system helper and using constants.)

src/Styleguide/Artwork/Banner.tsx
damassi: 👍

src/Styleguide/StyledSystemTypes.ts
```diff
@@ -0,0 +1,26 @@
+export interface MarginProperties {
```
zephraph: I don't think this is the best place for this file... It should probably go in typings? I also don't know exactly what to name it.

src/Styleguide/Elements/Flex.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from "styled-components"
+import { FlexboxProperties, MarginProperties } from "../StyledSystemTypes"
```
damassi: Since its somewhat conventional to see `SomethingProps` around our TS codebases, can we shortent `Properties` a bit?

src/Styleguide/Artwork/Banner.tsx
```diff
@@ -1,22 +1,22 @@
...
-`
+import { Sans, Serif } from "../Elements/Typography"
+import { Flex } from "../Elements/Flex"
 
 export class Banner extends React.Component {
   render() {
     return (
-      <Flex>
+      <Flex flexDirection="row">
         <Avatar size="110px" src="https://picsum.photos/110/110/?random" />
```
zephraph: The `Avatar` size is being set via pixels... should we not?

src/Styleguide/Elements/Flex.tsx
zephraph: Yeah, I can do that. I was trying to distinguish between react "props" and CSS properties.

src/Styleguide/Artwork/Banner.tsx
alloy: Seems fine for now 👍

damassi: This seems like a good candidate [for here](https://github.com/artsy/palette/blob/master/src/Theme.tsx#L94), but i would maybe run by brian?

alloy: I’m thinking that we need to take some more stuff into account in the future (such as fetching the image of the right size, also taking retina into account, etc) but the layout value, to me, seems very local to the component.

src/Styleguide/StyledSystemTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+    | "space-around"
+    | "stretch"
+  justifyContent?:
+    | "flex-start"
+    | "flex-end"
+    | "center"
+    | "space-between"
+    | "space-around"
+    | "space-evenly"
+}
```
damassi: @zephraph - wanted to point this out as @yuki24 has already typed much of this in his spike work: https://github.com/artsy/emission/blob/master/src/lib/Components/Bidding/Elements/types.d.ts

yuki24: I'm sure some incorrect definitions slipped in due to a human error... I wonder if there's any better way to auto-generate all these type definitions.

cc @alloy

zephraph: Oh nice! Good catch. Can we share that? 😄

alloy: @yuki24 Are these typings for the styled-system API? If so, no typings are maintained manually, but you should consider publishing them to http://github.com/DefinitelyTyped/DefinitelyTyped instead so others can help maintaining them.

alloy: Right. Publish it on DT so we are at least maintainers and then invite that person to be a maintainer too?

damassi: Once I get a little farther on the comps stuff I'll pull down and figure out how to add a lib to it; i haven't done that before.

PR #735, opened 2018-06-07
[Dev] fix ts error

l2succes:


src/Styleguide/Elements/Avatar.tsx
```diff
@@ -1,6 +1,8 @@
+// @ts-ignore
+import React from "react"
 import styled from "styled-components"
 
-interface AvatarProps {
+export interface AvatarProps {
```
l2succes: @zephraph FYI this causes an error because the interface needs to be exported as well 😩

```typescript
src/Styleguide/Elements/Avatar.tsx:7:14 - error TS4023: Exported variable 'Avatar' has or is using name 'React.ClassAttributes' from external module "/home/ubuntu/reaction/node_modules/@types/react/index" but cannot be named.

7 export const Avatar = styled.img.attrs<AvatarProps>({})`
               ~~~~~~
```
https://circleci.com/gh/artsy/reaction/3578

zephraph: Sorry about that!

PR #739, opened 2018-06-08
Starting to add Artists to the sidebar

oxaudo:
This is my attempt to start with the Sidebar code. I'm not sure I'm doing the right thing here. But my thinking was to create `<Artists/>` component that will take an array of artists and display them properly.

Realizing the we already have a `<TextLink/>` component I attempted to reuse that but seems like it is coming with some styles attached to it.

Also we have a follow button component in reaction which we probably should reuse as well, but seems like I'm not using the right one here. Or not using it correctly:(

Anyway - committing this work in progress to get some feed back on the approach here.

Just to show I've added it to the main artwok page story to make it look like that for now:
screen shot 2018-06-07 at 11 16 30 pm


src/Styleguide/Artwork/Sidebar/Artists.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class Artists extends React.Component<ArtistsProps> {
+  renderArtistName(artist) {
+    return artist.href ? (
+      <TextLink href={artist.href}>
+        <Serif size="5t">{artist.name}</Serif>
+      </TextLink>
+    ) : (
+      <Serif>{artist.name}</Serif>
```
zephraph: You probably want to preserve the size here

alloy:
@damassi @zephraph Question, were you hoping to have the component do even less for now (in terms of handling variable artists input data)?

zephraph:
I think this covers what we need. It really only handles the display logic for having an artist or multiple artists...

@alloy was there something specific that made you ask?

oxaudo:
@zephraph, @damassi - new updated version here:

screen shot 2018-06-11 at 4 03 26 pm

I think this might be mergeable:)

zephraph:
Awesome! I'll check it out.

src/Styleguide/Elements/FollowIcon.tsx
```diff
@@ -0,0 +1,69 @@
...
+    return (
+      <FollowIconContainer data-followed={isFollowed}>
+        <Icon
+          name={iconName}
+          style={{
+            verticalAlign: "left",
+            color: "inherit",
+            margin: "-2px 0 0 -5px",
+          }}
+        />
```
damassi: @oxaudo - if you rebase from master I've mixed in new `styled-system` fns so you can just do 

```
<Icon my={3} /> (margin Y) 
```
and control margins / padding and so on that way. 

See https://github.com/artsy/reaction/blob/master/src/Components/Icon.tsx#L36-L40

zephraph: Waaa. I didn't know that was a thing. Mx/Px work too?

src/Styleguide/Elements/FollowIcon.tsx
```diff
@@ -0,0 +1,69 @@
+import React from "react"
+import styled from "styled-components"
+import { themeGet } from "styled-system"
+import Icon from "../../Components/Icon"
+import { Flex } from "./Flex"
+import { Serif } from "@artsy/palette"
+
+interface FollowIconProps {
```
oxaudo: Figured we can move this `+ Follow` into an element here. The idea is that it's just a display and then some other component like `ArtistFollow` or whatever will deal with data will actually handle the change in status.

src/Styleguide/Elements/__stories__/Typography.story.tsx
```diff
@@ -122,15 +122,28 @@ stories.add("Typography", () => {
...
         </Sans>
 
         <Serif mx={5}>
-          <a href="#" className="colorLink">
+          <a href="#" className="purpleLink">
+            This is a link
+          </a>
+        </Serif>
+      </Section>
+      <Section title="Links .blackLink">
```
oxaudo: In our case the Artist name is a Serif link but black, so added another option here and renamed the original colored one into more explicit purpleLink. Hope it's ok @damassi

src/Styleguide/Pages/Artwork/Sidebar/Artists.tsx
```diff
@@ -0,0 +1,59 @@
...
+    ) : (
+      <Serif size="5t" display="inline-block">
+        {artist.name}
+      </Serif>
+    )
+  }
+
+  render() {
+    const { artists } = this.props
+    if (artists.length === 1) {
```
oxaudo: I guess we would have to handle a CulturalMaker and no artists scenario here as well, but we don't have designs for it so postponed.

src/Styleguide/Pages/Artwork/index.js
```diff
@@ -5,6 +5,8 @@ import { AboutTheWork } from "./AboutTheWork"
 import { Bibliography } from "./Bibliography"
 import { ExhibitionHistory } from "./ExhibitionHistory"
 import { Banner } from "./Banner"
+import { Artists } from "./Sidebar/Artists"
+import { SingleFollowedArtist } from "./__stories__/Sidebar/Artists.story"
```
oxaudo: Not sure what is the best way to show those different option here. For now one option will do.

src/Styleguide/Elements/FollowIcon.tsx
damassi: Yup :)

src/Styleguide/Elements/__stories__/Typography.story.tsx
damassi: @oxaudo - It defaults to black, so no need to add an additional class. Re `purpleLink`Its ok with me but @briansw should chime in as I'm following the link spec precisely (meaning, following naming). If `colorLink` is meant to refer specifically to purple (and thus eliminate the possibility of someone arbitrarily introducing a new color class) then we should keep things aligned, but if `colorLink` means "pass in a color" then that's something else. 

@briansw - thoughts?

briansw: presently we'd never make a link any color but purple.

damassi: @zephraph - this also addresses our discussion about `<Link>` primitives. Less options is more (so to speak)

src/Styleguide/Pages/Artwork/index.js
damassi: We could add a stateful click switch as we're developing. What i've done in the past for areas that have multiple options is to drop in a container that when clicked cycles through the different options.

src/Styleguide/Elements/__stories__/Typography.story.tsx
oxaudo: The default is actually `color: ${themeGet("colors.black60")};` which makes it gray. That is why I added this black color.

damassi: @oxaudo - I see what you mean now: 

screen shot 2018-06-11 at 1 34 42 pm

damassi: So this is my mistake, rather than assigning a color it should be `color: inherit` (contextual). I was specifically assigning a color to the link.

oxaudo: I'll revert the other change then and let it inherit the color from parent. Thanks for checking here!

src/Styleguide/Elements/FollowIcon.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+      <FollowIconContainer data-followed={isFollowed}>
+        <Icon
+          name={iconName}
+          style={{
+            verticalAlign: "left",
+            color: "inherit",
+            margin: "-2px 0 0 -5px",
+          }}
```
zephraph: I think at some point it'd be good to consolidate this on styled-system's semantics instead of doing inline styles, but I think it's okay for the first pass.

alloy:
> @alloy was there something specific that made you ask?

@zephraph Just trying to get a sense for how static vs dynamic y’all want this first phase to be.

PR #742, opened 2018-06-08
Add styled-system.d.ts

damassi:
Adds a `styled-system.d.ts` file to the repo and removed our interim typings. I'm noticing that its a little out of date so i've updated it in a few places. Once stuff shakes out we should send it to DefinitelyTyped.


src/Styleguide/Elements/Flex.tsx
```diff
@@ -1,20 +1,28 @@
...
-  alignItems,
   alignContent,
+  alignItems,
+  AlignItemsProps,
+  flexDirection,
+  FlexDirectionProps,
   justifyContent,
+  JustifyContentProps,
   space,
+  SpaceProps,
```
damassi: @zephraph - now that we're pulling interfaces into our components / primitives lets make sure our styled-system stuff is all alpha so its clear at a glance what's in scope and what still needs types.

typings/styled-system.d.ts
```diff
@@ -0,0 +1,328 @@
+declare namespace StyledSystem {
```
alloy: FYI Use of namespaces is discouraged, rather use `declare module` here directly and remove it altogether once it moves to DT.

PR #743, opened 2018-06-08
@zephraph => Add <Button /> element 

damassi:
Adding a few elements to the lib. Primary API: 

```jsx
<Button>Click me</Button>
<Button variant='outline'>Click me</Button>
<Button size='small'>Click me</Button>
```
Other common props can be passed through via styled-system, e.g.,: 
```jsx
<Button width='50%' />
```

screen shot 2018-06-08 at 12 46 10 am

### DIFF
--- a/src/Styleguide/Elements/Button.tsx
+++ b/src/Styleguide/Elements/Button.tsx
@@ -1,50 +1,128 @@
-import React, { SFC } from "react"
+import React, { Component, ReactNode } from "react"
 import styled from "styled-components"
-import { Sans } from "@artsy/palette"
+import { Sans, TypographyProps } from "@artsy/palette"
 import { pick } from "lodash/fp"
 
 import {
   border,
+  BorderProps,
+  borderColor,
+  BorderColorProps,
   borderRadius,
-  color,
-  maxWidth,
-  space,
-  textAlign,
-  width,
   BorderRadiusProps,
-  PseudoStyleValue,
+  color,
+  ColorProps,
+  space,
   SpaceProps,
+  textAlign,
+  TextAlignProps,
+  width,
+  WidthProps,
 } from "styled-system"
 
-export type ButtonProps = SpaceProps & PseudoStyleValue & BorderRadiusProps
-
-export const Button: SFC<ButtonProps> = ({ children, ...rest }) => {
-  const textProps = pick(["color", "size", "weight"], rest)
-
-  return (
-    <Container {...rest}>
-      <Sans {...textProps}>{children}</Sans>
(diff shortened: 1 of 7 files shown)


src/Styleguide/Elements/Button.tsx
```diff
@@ -1,40 +1,126 @@
...
+  large = "large",
+  default = "default",
+  small = "small",
+  xsmall = "xsmall",
+}
+
+export interface ButtonProps extends ButtonBaseProps {
+  children: ReactNode
+  size?: any // FIXME, ButtonSize
+  variant?: any // FIXME, ButtonVariant?
```
damassi: @alloy - I'm having trouble assigning a possible enum to a prop. When I set 
```
variant?: ButtonVariant`
```
and then try to use the button like 
```
<Button variant='outline' />
```
it throws this error: 

screen shot 2018-06-08 at 12 50 01 am

Not quite sure I understand how enums / options work.

typings/styled-system.d.ts
```diff
@@ -321,8 +325,23 @@ declare namespace StyledSystem {
...
+  export interface BorderProps {
+    border?: BorderValue
+  }
+
+  // FIXME: Should be able to use ColorProps above
+  export interface ColorProps {
+    color?: ColorValue
+    bg?: ColorValue
+    backgroundColor?: ColorValue
+  }
```
damassi: @zephraph - I added this section at the bottom of the file for WIP things. Unless you have a better idea we should put all of our additions here so that we can keep track / iterate.

src/Styleguide/Elements/Button.tsx
alloy: The error may be that you are passing a string value where an enum key is expected. You should probably just use a string union instead of an enum. Is there any value in defining the enum as is done anyways?

damassi: My thinking around enums is that you can also use them as values, so in situations like this where there will eventually be a large number of variants it keeps the drift level to a minimum. I remember luc doing something similar recently and it felt clean.

This totally answers the question tho, I was thinking it was looking for the value when really it is looking for the key.

typings/styled-system.d.ts
zephraph: Ahhh, makes sense. Yeah, I like it. 👍

zephraph:
LGTM. @damassi are you ready to merge, or do you want to do the enum changes first?

damassi:
We can merge so that `<Button />` is unblocked -- i'll update after 👍

PR #744, opened 2018-06-08
Add tentive props to banner

zephraph:
Add props to banner. Names tentative, but we can refine later.


src/Styleguide/Artwork/__stories__/Banner.story.tsx
```diff
@@ -3,5 +3,14 @@ import { storiesOf } from "storybook/storiesOf"
 import { Banner } from "../Banner"
 
 storiesOf("Styleguide/Artwork", module).add("Banner", () => {
-  return <Banner />
+  return (
+    <Theme>
```
damassi: We don't need `<Theme />` any more if you `import { storiesOf } from 'storybook/storiesOf`

zephraph: Sounds good

PR #745, opened 2018-06-08
First pass at artwork gallery component

zephraph:
Takes a first pass at fleshing out the desktop version of the artwork gallery component. Would note that there is some similarities between this and the banner.

### DIFF
--- a/src/Styleguide/Artwork/Gallery.tsx
+++ b/src/Styleguide/Artwork/Gallery.tsx
@@ -1,0 +1,27 @@
+import React from "react"
+import { Flex } from "../Elements/Flex"
+import { Avatar, Button } from "../Elements"
+import { Serif, Sans } from "@artsy/palette"
+
+interface GalleryProps {
+  src: string
+  headline: string
+  subHeadline: string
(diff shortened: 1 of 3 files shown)


src/Styleguide/Artwork/Gallery.tsx
```diff
@@ -0,0 +1,27 @@
...
+  render() {
+    return (
+      <Flex>
+        <Avatar src={this.props.src} size="100px" />
+        <Flex flexDirection="column" justifyContent="center" ml={4}>
+          <Serif size="5t">{this.props.headline}</Serif>
+          <Sans size="2">{this.props.subHeadline}</Sans>
+          <Button variant="outline" size="small" mt={3} width="80px">
+            Follow
+          </Button>
```
damassi: 😄

zephraph: I wanted to pull the `width` size from space, but it wasn't really working out for me. Didn't spent _too_ much time trying to figure it out though. I'll likely revisit it shortly.

PR #747, opened 2018-06-08
@zephraph => Add GlobalStyles pattern + some text

damassi:
Adds a new GlobalStyles component to `/Elements` while we build this out. this will eventually get hoisted up. Also an `<InfoWindow />` component which should be used when using the `addons-info` component (keeps things properly aligned throughout styleguide generated content)

### DIFF
--- a/src/Styleguide/Elements/GlobalStyles.tsx
+++ b/src/Styleguide/Elements/GlobalStyles.tsx
@@ -1,0 +1,56 @@
+import styled from "styled-components"
+import { css } from "styled-components"
+import { themeGet } from "styled-system"
+import { Sans, Serif, Display } from "@artsy/palette"
+
+const Links = css`
+  ${Sans} {
+    a {
+      color: ${themeGet("colors.black60")};
+
+      &:hover {
+        color: ${themeGet("colors.black100")};
+      }
+
+      &:active {
+        color: ${themeGet("colors.black100")};
+      }
+
+      &:focus {
+        border: 1px solid ${themeGet("colors.purple100")};
+        outline: 0;
+        padding: 2px;
+      }
+    }
+  }
+
+  ${Serif} {
+    a {
+      color: ${themeGet("colors.black60")};
+
+      &:hover {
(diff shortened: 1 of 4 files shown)


src/Styleguide/Elements/GlobalStyles.tsx
```diff
@@ -0,0 +1,56 @@
+import styled from "styled-components"
+import { css } from "styled-components"
+import { themeGet } from "styled-system"
+import { Sans, Serif, Display } from "@artsy/palette"
+
+const Links = css`
```
damassi: Idea is to define top-level categories using `css` and then ...

src/Styleguide/Elements/GlobalStyles.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  ${Display} {
+    a {
+      color: ${themeGet("colors.black100")};
+    }
+  }
+`
+
+export const GlobalStyles = styled.div`
+  ${Links};
```
damassi: ...mix them in here. This object will then go back into the Palette when ready.

src/__stories__/storiesOf.js
```diff
@@ -2,9 +2,10 @@ import React from "react"
 import { storiesOf as _storiesOf } from "@storybook/react"
 import { Theme, injectGlobalCSS } from "@artsy/palette"
 
+injectGlobalCSS()
+
 export function storiesOf(desc, mod) {
   return _storiesOf(desc, mod).addDecorator(storyFn => {
-    injectGlobalCSS()
     return <Theme>{storyFn()}</Theme>
```
damassi: For some reason `<GlobalStyles>` doesn't like this `storyFn()`; behaves inconsistently during HMR. Will return.

zephraph:
LGTM

src/__stories__/storiesOf.js
damassi: I've tested it with various trees outside of this and it works all over, there's just something weird about this one spot.. 🤔

PR #750, opened 2018-06-08
@eessex => Mobile Login + Reset Password

kanaabe:
Flows for Login + Reset Password

- [x] Still needs a few style updates and tests

mobile-login

image

### DIFF
--- a/src/Components/Authentication/Mobile/LoginForm.tsx
+++ b/src/Components/Authentication/Mobile/LoginForm.tsx
@@ -1,4 +1,5 @@
 import React from "react"
+import styled from "styled-components"
 import { Step, Wizard } from "../../Wizard"
 import { ProgressIndicator } from "../../ProgressIndicator"
 import Input from "../../Input"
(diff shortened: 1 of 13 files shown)


src/Components/Authentication/Mobile/LoginForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+    }
+    actions.setFieldError("email", "Email does not exist.")
+    actions.setSubmitting(false)
+    return false
+  })
+}
+
+export const MobileLoginForm: FormComponentType = props => {
+  const steps = [
+    <Step validationSchema={LoginValidator.email} onSubmit={checkEmail}>
```
damassi: I just love this wizard / step component; very nicely designed.

kanaabe:
@eessex I made you the reviewer so you can get more acquainted with this but adding @xtina-starr here for more eyes too!

src/Components/Authentication/helpers.ts
```diff
@@ -1,3 +1,6 @@
+import * as sharify from "sharify"
```
eessex: Interesting, didn't realize we had sharify in reaction too -- does this pick up data from force's config?

src/Components/Authentication/helpers.ts
```diff
@@ -34,3 +37,37 @@ export async function sendAuthData(
...
+    {
+      appToken: sharify.data.XAPP_TOKEN,
+    }
+  ).then(({ data }: any) => {
+    if (data.user.userAlreadyExists) {
+      if (shouldExist) {
+        return true
+      } else {
+        actions.setFieldError("email", "Email already exists.")
+        actions.setSubmitting(false)
```
eessex: Curious to know where these actions come from -- are they stored in context?

src/Components/Authentication/helpers.ts
kanaabe: I assumed so but now that I look for it, I'm not sure if it'd actually work once in Force. @xtina-starr is working on integrating `MobileSignUp` into Force so we'll see if it's working as expected.

src/Components/Authentication/helpers.ts
kanaabe: These actions are passed down as args from Formik's `onSubmit` here: https://github.com/artsy/reaction/pull/750/files#diff-789fe1e619c39fe8bf9ce02131a25dddR26

PR #754, opened 2018-06-09
@zephraph => <ReadMore /> expandable component

damassi:
This adds a new `<ReadMore>` component to our set. Note that "read more" is an old, tricky, edge-case filled problem that hasn't yet (to my knowledge) been solved in an performant way that a) addresses browser dimensions; and b) includes SSR, since we need to measure the viewport. @zephraph - have you had to deal with this before? Any thoughts? For now however I think this will be sufficient, but all in all its pretty whatevs as you can see by dragging the viewport width around... 

I'm using the same component that we use elsewhere in publishing (https://www.artsy.net/news), but I had to make a PR to the underlying lib to support custom "read more" links: https://github.com/xiaody/react-lines-ellipsis/pull/19. We're currently pointing at this PR commit in `package.json`, so I'll make sure to track this. 

**NOTE:** In the [Artist zeplin spec](zpl.io/bPdMXLm) it requests that we cap at a particular character count, which is different than the [Artwork spec](zpl.io/25y0NGQ) in that it requests a full line count. Character count is definitely solvable as its context/display independent. I think that this feature can take place in a follow up PR though. 

screen shot 2018-06-09 at 12 32 39 am

So we're supposed to cap after a character count, as opposed to a line count. 

readmore

### DIFF
--- a/src/Styleguide/Components/ReadMore.tsx
+++ b/src/Styleguide/Components/ReadMore.tsx
@@ -1,0 +1,82 @@
+import React, { Component } from "react"
+import styled from "styled-components"
+import { Truncator } from "../../Components/Truncator"
+import { Theme, Sans } from "@artsy/palette"
+
+interface ReadMoreProps {
+  isExpanded?: boolean
+  maxLineCount?: number
+}
+
+interface ReadMoreState {
+  isExpanded: boolean
+}
+
+export class ReadMore extends Component<ReadMoreProps, ReadMoreState> {
+  state = {
+    isExpanded: false,
+  }
+
+  static defaultProps = {
+    isExpanded: false,
+    maxLineCount: 3,
+  }
+
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      isExpanded: props.isExpanded,
+    }
+  }
+
+  expandText = () => {
+    this.setState({
+      isExpanded: true,
+    })
+  }
+
+  render() {
+    const { isExpanded } = this.state
+    const { maxLineCount } = this.props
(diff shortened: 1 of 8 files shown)


damassi:
Gonna go ahead and merge this, can loop back if there's anything we need.

zephraph:
👍

damassi:
@zephraph curious if you've ever dealt with a widget like this that needed to be SSR rendered?

src/Styleguide/Components/ReadMore.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+const ReadMoreLink = ({ children }) => {
+  return (
+    // TODO: Investigate why <Truncator />, when calling `renderToStaticMarkup`,
+    // breaks the context chain requiring us to wrap ReadMore in a <Theme />
+
+    <Theme>
+      <ReadMoreLinkContiner>
```
alloy: Little spelling error

zephraph:
I've delt with truncation in various forms. Both character based and line based. I've never had to solve line based truncation w/ SSR though. I'd likely just fall back on character based truncation and let it expand out on CSR.

damassi:
I wonder whats better in terms of SEO, truncating at the server level and finding a way to elegantly collapse the content on mount, or truncate / then rehydrate and find a way to bury some hidden text

alloy:
@damassi The latter sounds like the best solution to address both SEO and UX (collapsing after render isn’t the prettiest).

PR #756, opened 2018-06-09
[PATCH checkDeclarations] Ensure declaration in precommit

damassi:
Runs `tsc emit-types` in the `precommit` hook so that we can see if there are any issues with `--declaration` files.


zephraph:
Awww yuus 👏 

This has bit me a few times already.

PR #758, opened 2018-06-09
Experimental responsive handler component

zephraph:
This is an experimental responsive api built off of react's context api and the web's [matchMedia](https://caniuse.com/#search=matchmedia) api. 

It provides a framework for describing application level breakpoints, but doesn't enforce specific values. It uses dom events around `matchMedia` so it's not polling to figure out when to update a component. 

**Responsive Context Api**

`<ResponsiveProvider/>` is called near the root of the app. It can take two props. 

1. `initialBreakpoint` -- What breakpoint (if any) to initialize the app to. i.e. `"xs"`, `"sm"`, `"md"`
2. `breakpoints` -- An object with a key of the breakpoint's name and a value of the breakpoint expression. i.e. `{ xs: "(max-width: 767px)" }`

`<ResponsiveProvider/>` has the following state

1. `breakpoints` -- An object with key of the breakpoint's name and a boolean value of whether or not that breakpoint is currently active.
2. `mediaMatches` -- An array of [MediaQueryList](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList) objects that are responsible for managing the event listeners that fires when its corresponding media query is triggered
2. `mqHandler` -- This method is triggered whenever one of the `MediaQueryList` objects emits an event. The function which iterates through all the [MediaQueryList](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList) objects in `mediaMatches` and checks to see if they're triggered. Each breakpoint's state is updated when this method is called. 

`<ResponsiveProvider/>` has the following functionality flow

1. On instantiation, the breakpoint state object is initialized with all values to `false` except the `initialBreakpoint` (if one is provided). 
2. When (or if) `componentDidMount` is called, the provider creates the `MediaQueryList` for each media query provided, sets up the event listeners to react to those media queries, and runs `mqHandler` to set the correct initial state based on the browsers current viewport. 
3. When the provider renders, it passes the breakpoint state object to the _actual_ provider, which then passes the values down to the context component (`<Responsive/>`). 
4. on `componentWillUnmount` the media query event listeners are cleaned up.

### DIFF
--- a/src/Styleguide/Elements/Responsive.tsx
+++ b/src/Styleguide/Elements/Responsive.tsx
@@ -4,20 +4,20 @@ import createContext, { ConsumerProps } from "create-react-context"
 // FIXME: Replace with React.createContext after React 16.3+ upgrade
 const ResponsiveContext = createContext({})
 
-interface Breakpoints {
+export interface Breakpoints {
   [breakpoint: string]: string
 }
 
-interface ResponsiveProviderProps {
+export interface ResponsiveProviderProps {
   initialBreakpoint?: string
   breakpoints: Breakpoints
 }
 
-interface BreakpointState {
+export interface BreakpointState {
   [breakpoint: string]: boolean
 }
 
-interface ResponsiveProviderState {
+export interface ResponsiveProviderState {
   breakpoints: BreakpointState
   mediaMatchers: MediaQueryList[]
   mqHandler: MediaQueryListListener | null
--- a/src/Styleguide/Elements/Responsive.tsx
+++ b/src/Styleguide/Elements/Responsive.tsx
@@ -18,7 +18,7 @@ interface BreakpointState {
 interface ResponsiveProviderState {
   breakpoints: BreakpointState
   mediaMatchers: MediaQueryList[]
-  mqHandler: null | MediaQueryListListener
+  mqHandler: MediaQueryListListener | null
 }
 
 export class ResponsiveProvider extends React.Component<
--- a/src/Styleguide/Elements/Responsive.tsx
+++ b/src/Styleguide/Elements/Responsive.tsx
@@ -1,6 +1,8 @@
 import React from "react"
+import createContext, { ConsumerProps } from "create-react-context"
 
-const ResponsiveContext = React.createContext({})
+// FIXME: Replace with React.createContext after React 16.3+ upgrade
+const ResponsiveContext = createContext({})
 
 interface Breakpoints {
   [breakpoint: string]: string
@@ -80,7 +82,5 @@ export class ResponsiveProvider extends React.Component<
   }
 }
 
-export const Responsive: React.ComponentType<
-  React.ConsumerProps<BreakpointState>
-> =
+export const Responsive: React.ComponentClass<ConsumerProps<BreakpointState>> =
   ResponsiveContext.Consumer
--- a/src/Styleguide/Elements/Responsive.tsx
+++ b/src/Styleguide/Elements/Responsive.tsx
@@ -7,6 +7,7 @@ interface Breakpoints {
 }
 
 interface ResponsiveProviderProps {
+  initialBreakpoint?: string
   breakpoints: Breakpoints
 }
 
@@ -30,7 +31,7 @@ export class ResponsiveProvider extends React.Component<
       breakpoints: {
         ...Object.keys(props.breakpoints)
           .map(breakpoint => ({
-            [breakpoint]: false,
+            [breakpoint]: breakpoint === props.initialBreakpoint ? true : false,
           }))
           .reduce((acc, curr) => ({ ...acc, ...curr }), {}),
       },
--- a/src/Styleguide/Elements/Responsive.tsx
+++ b/src/Styleguide/Elements/Responsive.tsx
@@ -1,0 +1,85 @@
+import React from "react"
+
+const ResponsiveContext = React.createContext({})
+
+interface Breakpoints {
+  [breakpoint: string]: string
+}
+
+interface ResponsiveProviderProps {
+  breakpoints: Breakpoints
+}
+
+interface BreakpointState {
+  [breakpoint: string]: boolean
+}
+
+interface ResponsiveProviderState {
+  breakpoints: BreakpointState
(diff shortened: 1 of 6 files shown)


src/Styleguide/Elements/Responsive.tsx
```diff
@@ -0,0 +1,85 @@
...
+      mediaMatchers: [],
+      mqHandler: null,
+    }
+  }
+
+  componentDidMount() {
+    this.setupObservers(this.props.breakpoints)
+  }
+
+  componentWillUnmount() {
```
damassi: Heads up that this lifecycle is [going away](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html)

damassi: Actually, scratch that...

src/Styleguide/Elements/Responsive.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const mqHandler = () => {
+      for (let i = 0; i < mediaMatchers.length; ++i) {
+        this.setState({
+          breakpoints: { [breakpointKeys[i]]: mediaMatchers[i].matches },
+        })
+      }
+    }
+    this.setState({ mediaMatchers, mqHandler })
+    mediaMatchers.forEach(mediaQuery => {
+      mediaQuery.addListener(mqHandler)
```
zephraph: Being that this listener will be called on every state change, it might be a good idea to debounce it. It's a bit of a nuanced case and might not make a ton of difference. Thoughts?

damassi:
@zephraph this thing is a real beauty 👍. We'll def be able to remove my bloated version from `/Utils` with this, and i love that we no longer need the dependency. Can you make sure that it accepts initial state for SSR?

src/Styleguide/Elements/Responsive.tsx
damassi: No need to debounce IMO as things are performant enough these days. If an issues arises we can tune tho...

zephraph:
It now accepts `initialBreakpoint`. Now it can be used like 
```jsx
<ResponsiveProvider initialBreakpoint="xs" breakpoints={...}/>
```

src/Styleguide/Elements/Responsive.tsx
```diff
@@ -0,0 +1,86 @@
...
+    })
+    mqHandler()
+  }
+
+  render() {
+    return (
+      <ResponsiveContext.Provider value={this.state.breakpoints}>
```
damassi: We'll need to upgrade force to `16.3` but that shouldn't be a problem; there are no breaking changes, just a deprecated context API that's still available for use. This kind of context tends to be locally scoped and deliberate, so no real conflict worry anyway.

damassi: Actually @zephraph, I take that back. Lets use [this](https://github.com/jamiebuilds/create-react-context) for now and then once Force is upgraded we can use native API.

zephraph:
**Note**: I accidentally committed the metaphysics submodule. **Please don't merge until that's reverted.**

src/Styleguide/Elements/Responsive.tsx
zephraph: Done. I'll create a PR both on reaction and force to upgrade `React`

damassi: word 👍

damassi:
Lets get this in! 👍 Feel free to merge this once its 🍏

damassi:
lol its that time of day again.. date failures, gonna merge

src/Styleguide/Elements/Responsive.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export interface BreakpointState {
+  [breakpoint: string]: boolean
+}
+
+export interface ResponsiveProviderState {
+  breakpoints: BreakpointState
+  mediaMatchers: MediaQueryList[]
+  mqHandler: MediaQueryListListener | null
```
alloy: I know that `mq` is an often used abbreviation for `media-query` but all these abbreviations being introduced in the context of breakpoints aren’t helping readability imo :/ Can we use actual words or is that an uphill battle with regards to 3rd party dependencies?

src/__stories__/storiesOf.js
```diff
@@ -1,11 +1,26 @@
...
+import { ResponsiveProvider } from "../Styleguide/Elements/Responsive"
 
 injectGlobalCSS()
 
+const breakpoints = {
+  xl: "(min-width: 1192px)",
+  lg: "(min-width: 1024px) and (max-width: 1191px)",
+  md: "(min-width: 900px) and (max-width: 1023px)",
+  sm: "(min-width: 768px) and (max-width: 889px)",
+  xs: "(max-width: 767px)",
```
alloy: 😞

alloy:
This looks really great! 👏 

I do think lib code like this should have some test coverage. Was there a plan for that yet?

src/Styleguide/Elements/Responsive.tsx
zephraph: Yeah, we can. This component has no 3rd party dependencies so we can update the language to whatever we want it to be.

zephraph:
Yeah, definitely. I'll work on it.

src/__stories__/storiesOf.js
alloy: It’s the same thing as above about abbreviations. I’ll leave it up to y’all to determine what is the preferred style.

src/Styleguide/Elements/Responsive.tsx
damassi: @alloy the old version piggy backed on https://github.com/damassi/react-responsive-decorator but this version skips all of that.

damassi: @zephraph which make me think, once we have some time can you make a note to replace the old `<Responsive>` component where its used, or add a note deprecating it at the call sites?  I believe there's only a few places in Publishing but should be a pretty seamless transition.

zephraph: I can certainly do that!

PR #759, opened 2018-06-10
Update React Types

zephraph:
Reaction is on React 16.4. This is just a bump to types. I'm going to open a similar PR on force, but that'll be actually updating react too.

### DIFF
--- a/src/Components/ArtworkGrid.tsx
+++ b/src/Components/ArtworkGrid.tsx
@@ -54,7 +54,7 @@ export class ArtworkGridContainer extends React.Component<
 
   maybeLoadMore() {
     const threshold = window.innerHeight + window.scrollY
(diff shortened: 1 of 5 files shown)


alloy:
Aww yeah, dates in tests 🤘

zephraph:
@alloy can you kick off the test again? It should pass now.

I need to get perms for circleci.

alloy:
Hmm, looks like I don’t have the permission either, which is weird. @damassi @l2success do you know more about this setup?

damassi:
@zephraph and @alloy - not sure exactly; normally you just have to log in and it will understand team permissions.

alloy:
Weird, I do have the permissions on desktop, just not on mobile :| Well, it’s green now, so 👍

PR #760, opened 2018-06-10
Banner, Gallery fully responsive

zephraph:
Both the artwork banner and the artwork gallery are now fully responsive. I went through and corrected minor things like font and avatar sizes between the different breakpoints. 

Also, because I got tired of dragging the browser, I created a Section component to wrap the component in. It's an accordion type thing that makes it easier for one story to show multiple variations on a component. 

image

### DIFF
--- a/src/Styleguide/Pages/Artwork/__stories__/Banner.story.tsx
+++ b/src/Styleguide/Pages/Artwork/__stories__/Banner.story.tsx
@@ -1,14 +1,35 @@
 import React from "react"
 import { storiesOf } from "storybook/storiesOf"
-import { Banner } from "../Banner"
+import { Banner, SmallBanner, LargeBanner } from "../Banner"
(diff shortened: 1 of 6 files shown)


src/Styleguide/Pages/Artwork/__stories__/Banner.story.tsx
```diff
@@ -1,14 +1,35 @@
...
 storiesOf("Styleguide/Artwork", module).add("Banner", () => {
   return (
-    <Banner
-      src="https://picsum.photos/110/110/?random"
-      badge="In show"
-      headline="Francesca DiMattio: Boucherouite"
-      subHeadline="Salon 94"
-    />
+    <React.Fragment>
+      <Section title="Responsive">
```
damassi: Nice 😎 I'll go ahead and update the other areas with this 👍

PR #761, opened 2018-06-10
Add <Pagination /> component; Add <Section> wrappers

damassi:
Styles a `<Pagination component />` (still needs interactivity though)

screen shot 2018-06-09 at 11 14 58 pm


damassi:
Just dropped in `<Responsive />` -- A++

src/Styleguide/Elements/Button.tsx
```diff
@@ -42,62 +42,66 @@ export interface ButtonProps extends ButtonBaseProps {
...
+        ...this.getVariant(),
+      }
+
+      return <ButtonBase {...buttonProps}>{this.props.children}</ButtonBase>
+    }
   }
-}
+)`
+  ${space};
+`
```
damassi: @zephraph - following that pattern for wrapping our type in order to control it elsewhere, this seems like a good pattern for somewhat transparently yielding control back to styled components and mixing in stuff too:

```javascript
export const SomeUIClass = styled(
  class extends Component {
    ...
  }
)`
  {color};
  {flex};
  {space};
`
```

PR #762, opened 2018-06-10
Make artist card responsive

zephraph:
Implements the small artist card view as well as wiring it up w/ the responsive component. 

Does a little clean-up/consistence around the avatar. There are only two sizes in the designs currently (100px, 70px) so I added a size prop to that. @damassi it might be worth having a discussion about hoisting that into palette.


zephraph:
I'm gonna go ahead and push this one through so I can use the avatar stuff on the auction card. Can come back and address any feedback if anything comes up.

PR #767, opened 2018-06-11
@l2succes => Auth Analytics

kanaabe:
cc @l2succes @xtina-starr. See [this doc for more details](https://docs.google.com/spreadsheets/d/1hwI_npaZK1sf9o4lsPDky50K5bD79tlK_Xg-O0uT5y4/edit?usp=sharing).

This doesn't include the submission success calls since that will all happen in Force.

### DIFF
--- a/src/Components/Authentication/Desktop/ModalManager.tsx
+++ b/src/Components/Authentication/Desktop/ModalManager.tsx
@@ -26,7 +26,7 @@ export interface ModalManagerProps {
 
 export interface ModalManagerState {
   currentType?: ModalType
-  copy?: string | null
+  options?: ModalOptions
 }
 
 export class ModalManager extends Component<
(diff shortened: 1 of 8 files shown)


src/Components/Authentication/Desktop/ModalManager.tsx
```diff
@@ -14,6 +16,9 @@ export interface ModalManagerProps {
   submitUrls: { [P in ModalType]: string }
   csrf: string
   redirectUrl?: string
+  tracking?: any
+  trackingData?: any
```
l2succes: @kanaabe could you define the type for `trackingData`? Would be helpful to look at a glance what's being logged

kanaabe:
Lets chat tomorrow about it but I've run into an interesting problem with the way this uses `react-tracking` as a decorator + how we're using refs to open the modal.

PR #768, opened 2018-06-11
@anandaroop => [Dev] fixes `yarn watch`

l2succes:
Fixes `yarn watch` in reaction


anandaroop:
Nice – I just cherrypicked & tried out in my local Rails setup and all looks good 👍 

Thanks for the quick fix!

PR #769, opened 2018-06-11
Slightly tweaking the styles for Follow component so that it's easier to use it for our new responsive static page

oxaudo:
The idea is that we should reuse this component on the new Artwork page as is. For now it's tricky to do the static set up so this change will make it easier. And Thanks @zephraph for helping shuffle things around here.

This should have no effect on existing component.

### DIFF
--- a/src/Components/Follow.tsx
+++ b/src/Components/Follow.tsx
@@ -26,6 +26,32 @@ interface Props
   relay?: any
 }
 
+export const StyledFollowButton = styled.div`
+  display: flex;
+  cursor: pointer;
+  color: black;
+  font-size: 16px;
+  align-items: center;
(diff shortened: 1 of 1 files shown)


zephraph:
From a functionality perspective, this should be exactly the same behavior as what was there before. Just semantically how the component is exported, the named component actually contains all the visuals now whereas before the only way to get all the visuals was by using the fragment container in the default export.

alloy:
Nice 👌

oxaudo:
Actually closing this one after discussion in slack. The consensus is that we should split visual from actually updating data in the component. So for this round I'm creating `FollowArtist` component inside styleguide that is going to be responsible only for rendering the proper visual.

PR #770, opened 2018-06-11
Remove create-react-context dependency

zephraph:
We temporarily added a polyfill of sorts for React's context api while we were on 16.2.0. Now that we're on 16.4.0 it's no longer needed.


zephraph:
Being as 16.4 is already merged, I'm gonna go ahead and merge this. We're just shipping a dependency that we don't need at this point.

PR #771, opened 2018-06-11
[PATCH tsconfig] Add declaration to config

damassi:
Adds declaration so we get VSCode warnings about missing interface imports: 

screen shot 2018-06-11 at 12 16 19 pm

### DIFF
--- a/src/Styleguide/Components/ReadMore.tsx
+++ b/src/Styleguide/Components/ReadMore.tsx
@@ -1,19 +1,19 @@
-import React, { Component } from "react"
+import React from "react"
 import styled from "styled-components"
 import { DisplayProps } from "styled-system"
 import { Truncator } from "../../Components/Truncator"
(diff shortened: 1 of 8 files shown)


src/Styleguide/Components/ReadMore.tsx
```diff
@@ -1,19 +1,19 @@
...
   maxLineCount?: number
 }
 
-interface ReadMoreState {
+export interface ReadMoreState {
   isExpanded: boolean
 }
 
-export class ReadMore extends Component<ReadMoreProps, ReadMoreState> {
+export class ReadMore extends React.Component<ReadMoreProps, ReadMoreState> {
```
damassi: Little things for consistency sake across @zephraph and my code

zephraph: I'm not married to that pattern. I'd be fine using it the other way around too.

PR #773, opened 2018-06-11
@l2succes => Match Desktop Modals to latest design

kanaabe:
This should cover all the latest changes for Desktop Modals in Zeplin. Here are some screenshots...

I think the only visual difference is the button heights (which would mean changing the line height of the `Default` button -- something I'd like to address more carefully later on).

One more thing that needs to be handled further is the `subtitle` -- it seems like we want to use the "Log in to your account" copy for Login screens always, but I need to check with @owendodd if that is indeed the case. If it is, we can include a default for Login screens to avoid having to overwrite in many places. 

Also went ahead and extracted TOS stuff to it's own component so it can be reused for mobile.. `src/Components/Authentication/TermsOfServiceCheckbox.tsx`

Login:
image

Signup:
image

Reset Password:
image

### DIFF
--- a/src/Components/Authentication/Desktop/Components/DesktopHeader.tsx
+++ b/src/Components/Authentication/Desktop/Components/DesktopHeader.tsx
@@ -1,7 +1,7 @@
 import React, { SFC } from "react"
 import styled from "styled-components"
 import Icon from "Components/Icon"
-import { garamond } from "../../../../Assets/Fonts"
+import { garamond } from "Assets/Fonts"
 
 export const DesktopHeader: SFC<{ subtitle?: string }> = props => {
   const subtitle = props.subtitle || "The art world online"
--- a/src/Components/Authentication/Desktop/Components/DesktopHeader.tsx
+++ b/src/Components/Authentication/Desktop/Components/DesktopHeader.tsx
@@ -1,10 +1,10 @@
 import React, { SFC } from "react"
 import styled from "styled-components"
-import Text from "Components/Text"
 import Icon from "Components/Icon"
+import { garamond } from "../../../../Assets/Fonts"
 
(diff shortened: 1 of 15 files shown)


src/Components/Authentication/Desktop/Components/DesktopHeader.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { SFC } from "react"
 import styled from "styled-components"
-import Text from "Components/Text"
 import Icon from "Components/Icon"
+import { garamond } from "../../../../Assets/Fonts"
```
l2succes: you can use `"Assets/Fonts"`

src/Components/Authentication/Desktop/ResetPasswordForm.tsx
```diff
@@ -28,7 +36,7 @@ export const ResetPasswordForm: FormComponentType = props => {
         isValid,
       }: FormikProps<InputValues>) => {
         return (
-          <Form onSubmit={handleSubmit}>
+          <Form onSubmit={handleSubmit} height={180}>
```
l2succes: Shouldn't `height` be `maxHeight`? 

I thought we wanted to compute the height of the forms based on the children and animate height changes when errors show up. At least that's how I remember seeing it on Owen's prototype.

kanaabe: Yeah, the tricky thing is we want to change the height of the form, but not of the modal itself. The form needs a fixed height that accommodates for the maximum number of errors. So the desired behavior looks like this:

max-errors

l2succes: Got it, what you did makes sense then. Thanks for the explanation!

PR #775, opened 2018-06-11
@zephraph => Add <CurrentEvent /> and <Image /> primitive 

damassi:
current-event

### DIFF
--- a/src/Styleguide/Elements/Image.tsx
+++ b/src/Styleguide/Elements/Image.tsx
@@ -1,0 +1,19 @@
+import styled from "styled-components"
+
+import {
+  space,
+  width,
+  SpaceProps,
+  WidthProps,
+  height,
+  HeightProps,
+} from "styled-system"
+
+export interface ImageProps extends SpaceProps, WidthProps, HeightProps {
+  src: string
+}
+export const Image = styled.img.attrs<ImageProps>({})`
+  ${space};
+  ${width};
+  ${height};
+`
--- a/src/Styleguide/Pages/Artist/Components/CurrentEvent.tsx
+++ b/src/Styleguide/Pages/Artist/Components/CurrentEvent.tsx
@@ -1,26 +1,8 @@
 import React from "react"
-import styled from "styled-components"
 import { Responsive } from "../../../Utils/Responsive"
 import { Flex } from "../../../Elements/Flex"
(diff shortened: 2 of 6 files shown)


src/Styleguide/Elements/Image.tsx
```diff
@@ -0,0 +1,19 @@
...
+  SpaceProps,
+  WidthProps,
+  height,
+  HeightProps,
+} from "styled-system"
+
+export interface ImageProps extends SpaceProps, WidthProps, HeightProps {
+  src: string
+}
+export const Image = styled.img.attrs<ImageProps>({})`
```
zephraph: 👏

src/Styleguide/Pages/Artist/Components/CurrentEvent.tsx
```diff
@@ -0,0 +1,44 @@
...
+  location: string
+  date: string
+}
+
+export class CurrentEvent extends React.Component<CurrentEventProps> {
+  render() {
+    return (
+      <Responsive>
+        {({ xs }) => {
+          if (xs) return null
```
zephraph: Why does it not render on `xs`?

damassi: Part of the spec is to hide it

src/Styleguide/Pages/Artist/__stories__/CurrentEvent.story.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { Section } from "../../../Utils/Section"
+import { CurrentEvent, LargeCurrentEvent } from "../Components/CurrentEvent"
+
+storiesOf("Styleguide/Artist", module).add("Current Event", () => {
+  return (
+    <React.Fragment>
+      <Section title="Responsive">
```
zephraph: I'd started naming these just `Responsive` to begin with, but I think we should fully name them. `Responsive Current Event` or w/e. That way if you're taking screen shots or gifs or something it captures the full context.

damassi: That makes sense, will update

zephraph:
LGTM Merge on green

PR #776, opened 2018-06-11
[PATCH Dev] fix build

l2succes:
Fixes some TS errors

https://circleci.com/gh/artsy/reaction/3725

### DIFF
--- a/src/Components/Authentication/commonElements.tsx
+++ b/src/Components/Authentication/commonElements.tsx
@@ -1,3 +1,5 @@
+// @ts-ignore
+import React from "react"
 import styled from "styled-components"
 import Colors from "Assets/Colors"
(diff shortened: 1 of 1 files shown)


src/Components/Authentication/commonElements.tsx
```diff
@@ -1,3 +1,5 @@
+// @ts-ignore
```
kanaabe: Ooh so it's legit to ignore or is this meant to be temporary?

l2succes: Yeah it's because of how Typescript type emission work, a fix is being considered (https://github.com/Microsoft/TypeScript/issues/9944) but in the meantime that's the recommended way to silence the compiler.

Here's a longer explanation https://github.com/artsy/reaction/pull/715/files#r193201970

PR #777, opened 2018-06-11
[Dev] fix `yarn type-check` #trivial

l2succes:
Type emission errors weren't getting exposed in `yarn type-check` so errors don't get outputted in PR builds.

merge on green


damassi:
Oh interesting, i guess it needs to be in a comment?

PR #779, opened 2018-06-12
@zephraph => Add <SelectMenu>

damassi:
select


zephraph:
Isn't this backwards? The bigger input is on desktop and the smaller one is on mobile?

zephraph:
Oh... nope. Okay, weird. The artwork page uses the larger one on desktop. 
image

damassi:
I’ll switch the names

zephraph:
It's not actually backwards. I was just looking at the wrong designs. On the artist page, the `Sort` UI is small on desktop. 

image

But I'm assuming on mobile it's intended to be full sized. 

image

However, there are instances of the larger select appearing on desktop. Should the dropdown menu just be a separate component?

src/Styleguide/Elements/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export interface SelectProps extends PositionProps, SpaceProps {}
+
+export class Select extends React.Component<SelectProps> {
+  render() {
+    return (
+      <Responsive>
+        {({ xs }) => {
+          if (xs) return <LargeSelect />
+          else return <SmallSelect />
```
zephraph: This behavior really only happens on the sort field, right? Might want to ask Brian tomorrow, but I don't think it's a general Select behavior.

damassi: Can you elaborate? In the spec it says to display the native select menu on click.

src/Styleguide/Elements/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+      </select>
+    </LargeSelectContainer>
+  )
+}
+
+export const SmallSelect = () => {
+  return (
+    <SmallSelectContainer>
+      <Sans size="2" display="inline" mr={2}>
+        Sort:
```
zephraph: Seems like this would be the label field for whatever the select box is. I can't find the label replicated for the larger dropdown in the designs, but something like it is on production.

image

https://www.artsy.net/collect?sort=-decayed_merch

damassi:
Going to merge this for a quick deploy; will follow up in subsequent PR if any other needed changes.

PR #780, opened 2018-06-12
Add <Slider /> component, <Arrow /> element

damassi:
slider


src/Components/Publishing/Header/__tests__/__snapshots__/FeatureHeader.test.tsx.snap
```diff
@@ -2837,6 +2837,7 @@ exports[`feature renders superArticle full header properly 1`] = `
   -ms-letter-spacing: 0px;
   letter-spacing: 0px;
   position: relative;
+  color: white;
```
damassi: From adding `${color}` to `<Icon />`. Will now check the theme first for color key (like `purple100'`) and if not found fall back to the value passed in. Backwards compatable with existing code -- e.g., 

```
<Icon color='black' />
```
no "black" in theme, but black is a valid CSS color value for `color` and passes right through.

PR #785, opened 2018-06-12
Initial pass at responsive layout of Other Auctions

zephraph:
First pass at the "Other Auctions" section

Also...

- Adds some types to our styled-system definition
- Fixes a bug in the responsive component


src/Styleguide/Elements/Flex.tsx
```diff
@@ -1,19 +1,24 @@
...
+  maxHeight,
   space,
   SpaceProps,
   HeightProps,
   WidthProps,
   height,
   width,
   style,
+  MaxHeightProps,
+  MaxWidthProps,
```
damassi: Can these be moved up near their associated functions?

src/Styleguide/Sections/__stories__/OtherAuctions.story.tsx
```diff
@@ -0,0 +1,42 @@
...
+  },
+]
+
+storiesOf("Styleguide/Sections", module).add("Other Auctions", () => {
+  return (
+    <Section title="Responsive Other Auctions">
+      <OtherAuctions>
+        {auctions.map(auction => <AuctionCard {...auction} />)}
+      </OtherAuctions>
+    </Section>
```
damassi: Will `LargeOtherAuctions` and `SmallOtherAuctions` eventually be shown here?

src/Styleguide/Utils/Responsive.tsx
```diff
@@ -2,6 +2,13 @@ import React from "react"
 
 const ResponsiveContext = React.createContext({})
 
+const shallowEqual = (a, b) => {
+  for (let key in a) {
+    if (a[key] !== b[key]) return false
+  }
+  return true
+}
```
damassi: https://lodash.com/docs/4.17.10#isEqual

src/Styleguide/Utils/Responsive.tsx
```diff
@@ -40,6 +47,14 @@ export class ResponsiveProvider extends React.Component<
     }
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextProps.children !== this.props.children) return true
+    if (shallowEqual(this.state.breakpoints, nextState.breakpoints)) {
+      return false
+    }
+    return true
+  }
```
damassi: This requires some testing (especially because i'm not sure about shallow comparisons of  `children` as a data structure), but maybe `React.PureComponent` would work well here.

src/Styleguide/Utils/Responsive.tsx
```diff
@@ -38,10 +45,15 @@ export class ResponsiveProvider extends React.Component<
...
   }
 
-  componentDidMount() {
-    this.setupObservers(this.props.breakpoints)
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextProps.children !== this.props.children) return true
+    if (shallowEqual(this.state.breakpoints, nextState.breakpoints)) {
+      return false
+    }
+    return true
```
damassi: This requires some testing (especially because i'm not sure about shallow comparisons of children as a data structure), but maybe React.PureComponent would work well here.

src/Styleguide/Sections/__stories__/OtherAuctions.story.tsx
zephraph: Could be... that component is already responsive. It's just relying on that.

src/Styleguide/Utils/Responsive.tsx
zephraph: Yeah, I'm just wanting to keep this component clean so it could be released separately.

damassi: Yeah this would make a great lib 👍

src/Styleguide/Utils/Responsive.tsx
zephraph: I agree that it'll require testing. This approach reduces the render duplicating over what `PureComponent` does though.

src/Styleguide/Elements/Flex.tsx
zephraph: Sure thing.

src/Styleguide/Utils/Responsive.tsx
```diff
@@ -34,20 +34,80 @@ export class ResponsiveProvider extends React.Component<
...
+  setupMatchers = (breakpoints, breakpointKeys) => {
+    return breakpointKeys.map(breakpoint =>
+      window.matchMedia(breakpoints[breakpoint])
+    )
+  }
+
+  /**
+   * Uses the mediaMatchers list to build a map of the states of each breakpoint
+   */
+  checkBreakpoints = (breakpoints, breakpointKeys, mediaMatchers) => {
```
alloy: In case you’re taking these in as parameters for ease of testing purposes, then I’d suggest making it a standalone function and type the params. Otherwise as a method I’d suggest just accessing `this.state.*` directly because it means you’ll automatically get the right types.

PR #786, opened 2018-06-12
[Artwork/ImageCarousel] First static pass.

alloy:
https://artsyproduct.atlassian.net/browse/EV-190

### DIFF
--- a/src/Components/Icon.tsx
+++ b/src/Components/Icon.tsx
@@ -26,11 +26,13 @@ export interface IconProps
   onClick?: () => void
 }
 
-const Icon: React.SFC<IconProps> = props => (
-  <div className={props.className} style={props.style} onClick={props.onClick}>
-    {icons[props.name]}
-  </div>
-)
+const Icon: React.SFC<IconProps> = ({
+  color,
+  font,
+  fontSize,
+  name,
+  ...props
+}) => <div {...props}>{icons[name]}</div>
 
 export default styled(Icon)`
   font-family: ${props => props.font || "artsy-icons"};
--- a/src/Styleguide/Artwork/ImageCarousel.tsx
+++ b/src/Styleguide/Artwork/ImageCarousel.tsx
@@ -25,15 +25,14 @@ const ButtonsContainerHeight = 60
 const ImageContainerViewportMargin = 100 + ButtonsContainerHeight
 
 interface ImageCarouselProps {
-  src: string
+  src: string[]
 }
 
 const Container = styled(Flex)`
   /* -debug-background-color: #0fdb82; */
 `
 
-const ZoomHitTarget = styled.a`
-  cursor: zoom-in;
+const ImageAreaContainer = styled(Flex)`
   width: 100%;
   min-height: 450px;
   height: calc(100vh - ${ImageContainerViewportMargin}px);
@@ -46,6 +45,7 @@ const ImageContainer = styled(Flex)`
 `
 
 const Image = styled.img`
+  cursor: zoom-in;
   max-height: 100%;
   max-width: 100%;
 `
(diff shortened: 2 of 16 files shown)


src/Styleguide/Artwork/ImageCarousel.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+const Image = styled.img`
+  cursor: zoom-in;
+  max-height: 100%;
+  max-width: 100%;
+`
+
+// TODO: Should Icon have this styling by default?
+const Button = styled.a`
+  ${Icon /* sc-selector */} {
```
damassi: @alloy - I saw your message the other day but still don't quite understand what this comment does? Is this a magic comment similar to how webpack does it?

src/Styleguide/Artwork/ImageCarousel.tsx
```diff
@@ -0,0 +1,209 @@
...
+                justifyContent="center"
+              >
+                <Button
+                  href="#"
+                  onClick={e => {
+                    e.preventDefault()
+                    this.changeCurrentImage(changeCurrentImageBy)
+                  }}
+                >
+                  <Icon name={iconName} color="black" />
```
damassi: We now have an arrow element as it's used all over: https://github.com/artsy/reaction/blob/master/src/Styleguide/Elements/Arrow.tsx

src/Components/Icon.tsx
```diff
@@ -26,11 +26,13 @@ export interface IconProps
...
 
-const Icon: React.SFC<IconProps> = props => (
-  <div className={props.className} style={props.style} onClick={props.onClick}>
-    {icons[props.name]}
-  </div>
-)
+const Icon: React.SFC<IconProps> = ({
+  color,
+  font,
+  fontSize,
```
zephraph: I think these props aren't needed here. So, it's a bit of an odd case. `color`, `font`, and `fontSize` are actually being handled by the `styled` wrapper that's put around this component (see [#L37](https://github.com/artsy/reaction/pull/786/files#diff-8299e56061299f400f96969da00c2b56L35)). It shouldn't actually receive these props.

src/Components/Icon.tsx
```diff
@@ -26,11 +26,13 @@ export interface IconProps
...
+  name,
+  ...props
+}) => <div {...props}>{icons[name]}</div>
 
 export default styled(Icon)`
```
zephraph: @damassi Do you know if we can type this? I guess the `.attrs<>({})` thing wouldn't work here. Maybe because we're calling this like a function it'd just be `styled<IconProps>(Icon)`?

src/Styleguide/Artwork/ImageCarousel.tsx
```diff
@@ -0,0 +1,209 @@
...
+const ImageAreaContainer = styled(Flex)`
+  width: 100%;
+  min-height: 450px;
+  height: calc(100vh - ${ImageContainerViewportMargin}px);
+  line-height: 0; /* Don’t introduce visual margins */
+`
+
+const ImageContainer = styled(Flex)`
+  height: 100%;
+  width: 100%;
```
zephraph: You don't actually need to create a wrapper component for this. The `Flex` component already has prop hooks for width and height via styled-systems.

```jsx
<Flex height="100%" width="100%"/>
```

src/Components/Icon.tsx
damassi: @alloy might have an idea but i was playing with `styled<IconProps>(Icon)` a few days ago and it didn't seem valid.

src/Styleguide/Artwork/ImageCarousel.tsx
alloy: Might be easier to compare to a magic comment that disables a linter rule.

src/Components/Icon.tsx
alloy: Another option might be to switch it around? i.e. create the styled container first and export the icon component which uses that styled container. That way it’s easy to type with `React.SFC<…>`.

typings/styled-system.d.ts
```diff
@@ -419,6 +419,27 @@ declare module "styled-system" {
...
+
+  export interface FlexBasisProps {
+    // TODO: The FlexBasisValue currently really only exists for documentation
+    //       purposes, because flex-basis also accepts `Nem` and `Npx` strings.
+    //       Not sure there’s a way to still have the union values show up as
+    //       auto-completion results.
+    flexBasis?: FlexBasisValue | string
+  }
+
+  export function flexBasis(...args: any[]): any
```
alloy: I didn’t end up actually using this. Let me know if you think I should take it out.

src/Components/Icon.tsx
alloy: This can be done separately from this PR imo.

alloy:
@zephraph Ok, this is ready for final review.

zephraph:
LGTM 👍

PR #789, opened 2018-06-13
Pull mediaQueries from theme file

zephraph:
There's a corresponding PR in palette that'll need to be merged for this to be complete. 

Moves the mediaQueries to the theme file so that they can be easily used in the styled-components css.


damassi:
👍

PR #792, opened 2018-06-13
@kanaabe => Mobile auth server errors

eessex:
Addresses [GROW-690](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-690)

This uses the same `context.form.status` as the desktop component, assuming the data will be structured in the same way.

Q - there are no tests at the moment for the existing form components, are we planning to add those separately as a work item?

### DIFF
--- a/src/Components/Authentication/Mobile/LoginForm.tsx
+++ b/src/Components/Authentication/Mobile/LoginForm.tsx
@@ -9,6 +9,7 @@ import Colors from "Assets/Colors"
 import { MobileLoginValidator } from "../Validators"
 import {
   BackButton,
+  Error,
   Footer,
   ForgotPassword,
   MobileContainer,
@@ -96,7 +97,10 @@ export const MobileLoginForm: FormComponentType = props => {
   return (
     <Wizard steps={steps}>
       {context => {
-        const { wizard, form } = context
+        const {
+          wizard,
+          form: { handleSubmit, status },
+        } = context
         const { currentStep } = wizard
 
         return (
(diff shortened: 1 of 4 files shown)


kanaabe:
This PR adds tests https://github.com/artsy/reaction/pull/750 -- I'll go ahead and merge that one in so you can cover the errors for Login and Reset as well.

eessex:
Cool, added the error to Login and Reset -- having some issues setting the Formik context manually for the test thought -- my usual methods for mocking context keep returning empty.

kanaabe:
Yeah, I've been having a lot of difficulty with this as well. I'm planning on looking at this again towards the end of sprint if you're interested. Maybe see if any of these help? https://github.com/jaredpalmer/formik/tree/master/test

kanaabe:
Are you still working on a test for this?

eessex:
Yeah still trying to figure this out -- realized we don't need to mock the context, but instead rewire the return value from `sendAuthData` when called inside `handleSubmit` to trigger the `formik.setStatus` method here- https://github.com/artsy/reaction/blob/master/src/Components/Authentication/helpers.ts#L10

I think is ok to merge, can keep exploring tests separately if this is pressing.

kanaabe:
That specific helper method isn't used at all -- we handle the submit in Force [here](https://github.com/artsy/force/blob/master/src/desktop/apps/auth2/helpers.ts) instead. How about passing in a custom `handleSubmit` that sets the status?

Either way, I agree we can get back to this -- we logged better testing as a todo item in one of the tickets. Feel free to merge after resolving the conflict!

eessex:
Is that file used anywhere?

kanaabe:
Yes, the helpers file is used.

eessex:
Can you show me an example of where, is it used ? Still trying to understand how the auth app works, not seeing what the Reaction helpers `handleSubmit` is for if it is not used by this app, I don't see it imported anywhere in Force or Reaction.

kanaabe:
The `helpers.ts` file is used, and `handleSubmit` is not. [Here](https://github.com/artsy/reaction/blob/master/src/Components/Authentication/Mobile/SignUpForm.tsx#L18) is an example of where we are using that file. We eventually want to make Authentication stand alone, which means handling submission in Reaction, but since there are many edge cases to handle, we decided to handle it in Force. The handleSubmit that is currently there works as a happy path so we are keeping it for future use.

PR #793, opened 2018-06-13
@zephraph => Update font weights, Mobile pagination, Render props in <Tabs /> component

damassi:
Update sans medium fonts where necessary and replace `<Icon>` with `<Arrow>`. Finishes up mobile pagination: 

screen shot 2018-06-13 at 12 00 06 pm

Also adds new API around `<Tabs />`:

```jsx
<Tabs labels=['hello', 'how', 'are', 'you']>
  {({ activeTab }) => {
    return <div>{activeTab.label}</div>
  }}
</Tabs>
```

tabs

### DIFF
--- a/src/Styleguide/Components/Tabs.tsx
+++ b/src/Styleguide/Components/Tabs.tsx
@@ -51,13 +51,17 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
 
 const Tab = ({ children, ...props }) => (
   <TabContainer {...props}>
-    <Sans size="3t">{children}</Sans>
+    <Sans size="3t" weight="medium" color="black30">
+      {children}
+    </Sans>
   </TabContainer>
 )
 
 const ActiveTab = ({ children }) => (
   <ActiveTabContainer>
-    <Sans size="3t">{children}</Sans>
+    <Sans size="3t" weight="medium">
+      {children}
+    </Sans>
   </ActiveTabContainer>
 )
 
--- a/src/Styleguide/Components/Tabs.tsx
+++ b/src/Styleguide/Components/Tabs.tsx
@@ -1,50 +1,91 @@
-import React from "react"
+import React, { ReactNode } from "react"
 import styled from "styled-components"
-import { borders, themeGet } from "styled-system"
+import { borders, themeGet, WidthProps } from "styled-system"
 import { Sans } from "@artsy/palette"
+import { Flex } from "../Elements/Flex"
 
-export interface TabsProps {
-  labels: string[]
+export interface ActiveTabProps {
+  activeTab: {
+    index: number
+    label: string
+  }
+}
+
+export interface TabsProps extends WidthProps {
+  children?: (activeTab: ActiveTabProps) => ReactNode
+  onChange?: (activeTab?: ActiveTabProps) => void
   activeTabIndex?: number
+  labels: string[]
 }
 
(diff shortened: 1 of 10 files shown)


src/Styleguide/Components/__stories__/Tabs.story.tsx
```diff
@@ -24,6 +26,27 @@ storiesOf("Styleguide/Components", module).add("Tabs", () => {
...
+      <Section title="With renderProps and onChange handler">
+        <Tabs
+          activeTabIndex={1}
+          labels={["Hello", "How", "Are", "You?"]}
+          onChange={activeTab => {
+            // tslint:disable-next-line
+            console.log(activeTab)
+          }}
+        >
+          {({ activeTab }) => {
```
zephraph: I can't help but wonder... should this be a [FaCC](https://medium.com/merrickchristensen/function-as-child-components-5f3920a9ace9)? I guess I **do** see the value in passing down the active tab. Still, this pattern creates a new function on every render. 

How often are views associated with the tabs going to need the name? Moreover, should the tabs component be responsible for rendering the views associated with it? I've got no actionable changes, just questions, heh.

src/Styleguide/Components/Tabs.tsx
```diff
@@ -1,70 +1,113 @@
...
+              <ActiveTab key={index}>{label}</ActiveTab>
+            ) : (
+              <Tab key={index} onClick={() => this.setActiveTab(index, label)}>
+                {label}
+              </Tab>
+            )
+          })}
+        </TabsContainer>
+
+        {children && (
```
zephraph: 👍 I like this

src/Styleguide/Utils/Section.tsx
```diff
@@ -46,8 +48,8 @@ export class Section extends React.Component<SectionProps> {
             flexDirection="column"
             alignItems="center"
             my={4}
-            mx="auto"
-            maxWidth="1192px"
+            mx={3}
+            maxWidth={maxWidth}
```
zephraph: :+1:

src/Styleguide/Components/__stories__/Tabs.story.tsx
zephraph: Further... what's the api for creating a tab? How do we intend this to be used? When I was first poking at it and added the `labels` prop, I wasn't 100% happy with that, but it got the job done as far as displaying the different tabs.

damassi: > Still, this pattern creates a new function on every render.

I think this is probably a premature optimization at this point, but we should keep it in mind. 

My thinking with the `<Tabs>` component was to provide a way to quickly render tab content based on values provided as well as provide a hook `onChange` for wiring up to things like a router. 

> what's the api for creating a tab's content

For right now being able to [match against the label](https://github.com/artsy/reaction/blob/master/src/Styleguide/Pages/Artist/index.js#L68) is sufficient for this sprint, but I could imagine us replacing a label with an array of object (for example). If you're ok with it we can shake this stuff out in the next sprint.

zephraph: Yep, I don't think it's something we have to immediately tackle. Just thinking out loud. Might be valuable to take a peek at how some other libraries/frameworks solve the problem from an api perspective.

PR #795, opened 2018-06-13
@eveessex => More auth errors + modal animations

l2succes:
This PR covers

- some more auth server errors for desktop modals
- adds animations for modals

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,6 +168,7 @@
     "react-router": "^4.2.0",
     "react-sizeme": "^2.4.4",
     "react-slick": "^0.14.11",
+    "react-spring": "^5.3.15",
     "react-styled-flexboxgrid": "^2.2.0",
     "react-tracking": "^4.2.1",
     "react-transition-group": "^2.3.0",
--- a/package.json
+++ b/package.json
@@ -123,7 +123,6 @@
     "relay-compiler-language-typescript": "^0.9.0",
     "semantic-release": "^12.4.1",
     "storybook-addon-scissors": "^4.0.1",
-    "styled-bootstrap-grid": "damassi/styled-bootstrap-grid#5a115a48a7f4d5608bc73097d52e14265edc6dd3",
     "styled-components": "^3.3.2",
     "stylelint": "^9.2.1",
     "stylelint-config-standard": "^18.2.0",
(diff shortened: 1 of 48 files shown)


src/Components/Authentication/Desktop/SignUpForm.tsx
```diff
@@ -39,9 +39,11 @@ export const SignUpForm: FormComponentType = props => {
         handleBlur,
         handleSubmit,
         isSubmitting,
-        isValid,
+        dirty,
         status,
```
eessex: what is `dirty`?

l2succes: `dirty` is set to `false` if none of the fields in the form have been touched (the value has been changed), `true` if one of those has changed.

package.json
```diff
@@ -167,6 +167,7 @@
     "react-router": "^4.2.0",
     "react-sizeme": "^2.4.4",
     "react-slick": "^0.14.11",
+    "react-spring": "^5.3.15",
```
damassi: 🎉

src/Components/Modal/Modal.tsx
```diff
@@ -16,16 +17,20 @@ const ModalContainer = styled.div`
...
   height: 100%;
-  top: 0px;
-  left: 0px;
+  top: 0;
+  left: 0;
   z-index: 9998;
-  background: rgba(0, 0, 0, 0.3);
+  background: rgba(200, 200, 200, 0.5);
+  -webkit-backdrop-filter: blur(20px);
+  backdrop-filter: blur(20px);
```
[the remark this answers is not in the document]
l2succes: Yeah, it's still very new. Doesn't have great browser support but it works great for the ones that have it implemented

src/Components/Modal/Modal.tsx
```diff
@@ -44,17 +49,30 @@ export class Modal extends React.Component<ModalProps, any> {
...
 
-    if (!this.props.show) {
-      return null
-    }
     return (
       <div>
-        <ModalContainer {...newProps}>{this.props.children}</ModalContainer>
-        <Overlay onClick={this.close} />
+        <Spring {...transitions as any}>
+          {(styles: any) => (
```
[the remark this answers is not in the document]
l2succes: yes, I've been investigating the library for a bit. cross-platform support was one main reason why I went that route especially since we want modals on modal as well.

src/Components/Modal/Modal.tsx
damassi: @l2succes - I just tried out the modal and noticed that these CSS blur props aren't applying to the page background -- anything I need to do?

PR #797, opened 2018-06-13
Add responsive image component

zephraph:
So we had a need to create an image container that could stay square and scale but retain still show the full image inside of it. This is a first implementation of that pattern. It's implemented in the auction card component and in the image story. 

Note that I still need to work with Brian to figure out the details of the mobile version of the auction card so that looks a little.... weird at the moment. Will also need to work through the details of where else this component will need to be used. 

Here's the component rendering in the _Other Auction_ section.

Notice that even though all the images have different sizes, they all fit in a square container.

### DIFF
--- a/src/Styleguide/Elements/__stories__/Image.story.tsx
+++ b/src/Styleguide/Elements/__stories__/Image.story.tsx
@@ -1,7 +1,8 @@
 import React from "react"
 import { Section } from "../../Utils/Section"
-import { Image } from "../Image"
+import { Image, ResponsiveImage } from "../Image"
 import { storiesOf } from "storybook/storiesOf"
+import { BorderBox } from "../Box"
 
 storiesOf("Styleguide/Elements", module).add("Image", () => {
   return (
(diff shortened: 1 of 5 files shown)


src/Styleguide/Elements/__stories__/Image.story.tsx
```diff
@@ -16,6 +17,11 @@ storiesOf("Styleguide/Elements", module).add("Image", () => {
           src="https://picsum.photos/300/300/?random"
         />
       </Section>
+      <Section title="Responsive Image">
+        <BorderBox maxWidth="400px" width="100%" height="auto" p={0}>
```
damassi: `p={0}` can probably go away, right?

damassi:
Danger, merge on green

src/Styleguide/Elements/__stories__/Image.story.tsx
zephraph: Nah, it's got a default padding.

PR #799, opened 2018-06-14
[Artwork] => Add <ImageCarousel> to artwork

damassi:
slideshow


src/Styleguide/Pages/Artwork/index.js
```diff
@@ -37,6 +46,7 @@ export class Artwork extends Component {
         <Row>
           <Col sm={8}>
             <ArtworkSlider>
+              <ImageCarousel src={slideshowImages} />
```
damassi: @alloy - seems like this prop could be a bit more plural for clarity, like `images=[...]`

PR #802, opened 2018-06-14
[Artwork/ImageCarousel] Use aspect ratio so layout doesn’t jump.

alloy:
See the inline comment https://github.com/artsy/reaction/compare/more-artwork-image-carousel-work?expand=1#diff-182d6f0b0c129e245b5c501ddad78c93R29

### DIFF
--- a/src/Styleguide/Pages/Artwork/ImageCarousel.tsx
+++ b/src/Styleguide/Pages/Artwork/ImageCarousel.tsx
@@ -15,7 +15,7 @@ const Spacer = styled.div`
   flex-grow: 1;
 `
 
-const BaseImageArea = styled.div`
+const BaseImageArea = styled.div.attrs<{ aspectRatio: number }>({})`
   line-height: 0; /* Don’t introduce visual margins */
   width: 100%;
   display: flex;
@@ -24,6 +24,14 @@ const BaseImageArea = styled.div`
 
 const SmallImageArea = styled(BaseImageArea)`
   max-height: calc(100vh - 120px);
+
+  /**
+   * This will make sure the image area already exists prior to image loading
+   * and the layout won’t jump later on, but it does rely on the current design
+   * where this image is supposed to be shown from the left side of the screen
+   * to the right side of the screen, i.e. the width of the viewport.
+   */
+  height: calc(100vw / ${({ aspectRatio }) => aspectRatio});
 `
 
 const LargeImageArea = styled(BaseImageArea)`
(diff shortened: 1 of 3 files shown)


src/Styleguide/Components/Artwork/ImageCarousel.tsx
```diff
@@ -24,6 +24,14 @@ const BaseImageArea = styled.div`
...
 const SmallImageArea = styled(BaseImageArea)`
   max-height: calc(100vh - 120px);
+
+  /**
+   * This will make sure the image area already exists prior to image loading
+   * and the layout won’t jump later on, but it does rely on the current design
+   * where this image is supposed to be shown from the left side of the screen
+   * to the right side of the screen, i.e. the width of the viewport.
+   */
+  height: calc(100vw / ${({ aspectRatio }) => aspectRatio});
```
oxaudo: This all looks good to me. Just trying to understand - here we are trying to make the image height proportional to view width? This will guarantee that really wide images do not take the whole width of the screen? That is an idea?

But should not the height of the image be the same for all the images in the carousel?  Or the idea is that it will jump in size when going image to image?

src/Styleguide/Components/Artwork/ImageCarousel.tsx
```diff
@@ -121,7 +129,7 @@ const ActionButtonsContainer = styled.div`
 `
 
 interface ImageCarouselProps {
-  src: string[]
+  images: Array<{ uri: string; aspectRatio: number }>
```
oxaudo: I like this rename. Makes it much easier to read.

oxaudo:
Also needs a master merge:)

src/Styleguide/Components/Artwork/ImageCarousel.tsx
alloy: Oh yeah absolutely, the previous one wasn’t really meant to be final in any way, neither is this one really :)

src/Styleguide/Components/Artwork/ImageCarousel.tsx
alloy: On mobile, the image carousel does not have a height relative to the vertical viewport, like it does on desktop. So when the page initially loads, the image area has a 0 height, then once the browser has downloaded the image the page jumps as the image area grows to fit the image. This bit of CSS immediately sets the height on the initial page load, because we know the aspect ratio and the width.

alloy:
Updated 👍 

Merge on green, jolly ol’ danger bot.

PR #803, opened 2018-06-14
Static implementation of Artist Info

zephraph:
Adds the Artist info... includes name, bio, market info

Updated both the artist and artwork pages.

### DIFF
--- a/src/Styleguide/Components/MarketInsights.tsx
+++ b/src/Styleguide/Components/MarketInsights.tsx
@@ -1,0 +1,52 @@
+import React from "react"
+import { Responsive } from "../Utils/Responsive"
+import { BorderBox } from "../Elements/Box"
+import { Sans } from "@artsy/palette"
+import { themeGet } from "styled-system"
+import styled from "styled-components"
+
+const TextWrap = styled.div`
+  display: block;
+
+  @media ${themeGet("mediaQueries.xs")} {
+    display: flex;
+    flex-direction: column;
+    margin-bottom: 10px;
+  }
+`
+
+export interface MarketInsight {
+  primaryLabel: string
+  secondaryLabel: string
+}
+
+export interface MarketInsightsProps {
+  insights: MarketInsight[]
+}
+
+export class MarketInsights extends React.Component<MarketInsightsProps> {
+  render() {
+    return (
+      <Responsive>
+        {({ xs }) => {
+          return (
+            <BorderBox flexDirection="column">
+              {this.props.insights.map(insight => {
+                return (
+                  <TextWrap>
+                    <Sans size="2" weight="medium" display="inline" mr={3}>
+                      {insight.primaryLabel}
+                    </Sans>
+                    <Sans size="2" display="inline" color="black60">
+                      {insight.secondaryLabel}
+                    </Sans>
+                  </TextWrap>
+                )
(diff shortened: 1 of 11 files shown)


src/Styleguide/Components/MarketInsights.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Responsive } from "../Utils/Responsive"
+import { BorderBox } from "../Elements/Box"
+import { Sans } from "@artsy/palette"
+import { themeGet } from "styled-system"
+import styled from "styled-components"
+
+const TextWrap = styled.div`
+  display: block;
+
+  @media ${themeGet("mediaQueries.xs")} {
```
orta: I worry when I see code like this that it won't work in react native (which doesn't support media queries)

src/Styleguide/Pages/Artwork/index.js
```diff
@@ -114,6 +118,16 @@ export class Artwork extends Component {
             </TabContent>
           </Col>
         </Row>
+        <Row>
+          <Col xl={8} lg={8} md={8} sm={12} xs={12}>
```
alloy: Are these colspans at various breakpoints?

zephraph: Yep, that exactly.

src/Styleguide/Components/MarketInsights.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <BorderBox flexDirection="column">
+        <Responsive>
+          {({ xs }) => {
+            const TextWrap = wrapper(xs)
+            return this.props.insights.map(insight => {
+              return (
+                <TextWrap>
+                  <Sans size="2" weight="medium" display="inline" mr={3}>
+                    {insight.primaryLabel}
+                    k{" "}
```
alloy: Is this intentional?

src/Styleguide/Components/ReadMore.tsx
```diff
@@ -65,7 +65,10 @@ const ReadMoreLink = ({ children }) => {
 
     <Theme>
       <ReadMoreLinkContainer>
-        ... <Sans size="2">{children}</Sans>
+        ...{" "}
```
alloy: Aww, prettier

src/Styleguide/Components/MarketInsights.tsx
```diff
@@ -1,19 +1,11 @@
...
-  display: block;
-
-  @media ${themeGet("mediaQueries.xs")} {
-    display: flex;
-    flex-direction: column;
-    margin-bottom: 10px;
-  }
-`
+const wrapper = xs => props =>
+  xs ? <Flex flexDirection="column" mb={3} {...props} /> : <Box {...props} />
```
alloy: One thing I’ve started wondering with the approach of rendering different components at different breakpoints is if this is working against React’s reconciliation (i.e. being unable to re-use existing DOM elements/iOS views but adjusting them)

zephraph: Well, on web both of these things result in `div` elements. The classes and attributes are different so there would be some updates, but the reconciliation process would work the same. On native, this flag wouldn't change so the returned result should always be the same.

PR #804, opened 2018-06-14
Add key to iterated elements

zephraph:
I accidentally left off the key prop when iterating over some elements.


alloy:
LGTM 👍 

Merge on green.

PR #809, opened 2018-06-15
adding partner info to sidebar

oxaudo:
This adds partner part:
screen shot 2018-06-14 at 11 13 19 pm


src/Assets/Icons/Location.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from "react"
+
+export class Location extends Component<null, null> {
```
oxaudo: Copied the pattern on Component per Icon from other Icons in here.

src/Styleguide/Pages/Artwork/Sidebar.tsx
```diff
@@ -50,6 +60,8 @@ export class Sidebar extends Component<ArtworkSidebarProps> {
         <ArtworkMetadata artwork={artwork} />
         <Separator />
         <Commercial artwork={artwork} />
+        <PartnerInfo partner={artwork.partner} artwork={artwork} />
```
oxaudo: Debated for a while if this component should take just partner and always display partner info. In this case check for `artwork.collecting_institution` would have to happen here and somehow that did not feel correct... Open for suggestions:)

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
+import React from "react"
+import { Serif } from "@artsy/palette"
+import styled from "styled-components"
+import { space, SpaceProps } from "styled-system"
+import { Flex } from "Styleguide/Elements/Flex"
+import { Location } from "../../../../Assets/Icons/Location"
+
+export interface PartnerInfoProps {
+  artwork?: {
```
oxaudo: This part is also not cleat - maybe this component should take artwork  and partner is embedded in artwork? Not sure....

src/Assets/Icons/Location.tsx
alloy: FYI This is a prime example of where you could use a ‘stateless functional component’, rather than creating a class.

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
alloy: Yup, that’s how I would model it, because that’s also how we’d do it with a Relay fragment container.

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <a href={partner.href}>{partner.name}</a>
+      </Serif>
+    ) : (
+      <Serif size="5t" display="inline-block">
+        {partner.name}
+      </Serif>
+    )
+  }
+  renderCollectingInstitution(artwork) {
+    return <Serif size="3">{artwork.collecting_institution}</Serif>
```
alloy: Seeing as this is a method, I’d just access `this.props.artwork` here, because it will automatically give you the right typing from the ‘props’ interface, whereas passing it as a parameter means you’d need to type it again.

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+const PartnerInfoContainer = styled.div.attrs<SpaceProps>({})`
+  ${space};
+`
+const LocationsContainer = styled.div``
+
+export class PartnerInfo extends React.Component<PartnerInfoProps> {
+  renderPartnerName(partner) {
```
alloy: Same as below.

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+    ) : (
+      <Serif size="5t" display="inline-block">
+        {partner.name}
+      </Serif>
+    )
+  }
+  renderCollectingInstitution(artwork) {
+    return <Serif size="3">{artwork.collecting_institution}</Serif>
+  }
+  renderLocations(locations) {
```
alloy: You could decide to type this here `renderLocations(locations: PartnerInfoProps["partner"]["locations"])` but that’s more syntax than just accessing `this.props.partner.locations` inside the method.

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const locationCities = locations.map((location, index) => {
+      return location.city
+    })
+    const filteredForDuplicates = locationCities.filter((city, pos) => {
+      return locationCities.indexOf(city) === pos
+    })
+    return (
+      <Serif size="2" display="inline-block" pl={3} pt={1}>
```
zephraph: We updated the spacing to be inline with design spacing. You'll want to update this to be `pl={2}` `pt={0.5}`. That's padding-left 20px and padding top 5px. The design system has units matching to 10px intervals.

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+        {filteredForDuplicates.join(", ")}
+      </Serif>
+    )
+  }
+
+  render() {
+    const { artwork, partner } = this.props
+    return (
+      <PartnerInfoContainer pb={4}>
```
zephraph: `pb={4}` -> `pb={2}` (padding-bottom: 20px)

src/Styleguide/Pages/Artwork/Sidebar/PartnerInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+        {artwork && artwork.collecting_institution
+          ? this.renderCollectingInstitution(artwork)
+          : this.renderPartnerName(partner)}
+        {partner.locations.length > 0 && (
+          <LocationsContainer>
+            <Flex width="100%" pt={3}>
```
zephraph: `pt={3}` -> `pt={1}` (padding-top: 10px)

PR #810, opened 2018-06-15
@mzikherman => [Artist] Make page responsive

damassi:
Makes page responsive via proper column setup and additionally adds a new `<Spacer />` primitive for explicitly spacing content out. 

responsive


src/Styleguide/Elements/Spacer.tsx
```diff
@@ -0,0 +1,7 @@
+import React from "react"
+import { Box } from "Styleguide/Elements/Box"
+import { SpaceProps } from "styled-system"
+
+export const Spacer: React.SFC<SpaceProps> = props => {
```
oxaudo: @damassi - I'm wondering if it makes sense to combine this with `Separator` that I've added at some point. We can make it take a property weather to draw a line or not. What do you think?

damassi: I think we should keep them separate; both of these items qualify as primitives.

PR #811, opened 2018-06-15
Add tabs to artwork, refactor

zephraph:
Adds tabs to the artwork page.

Also goes through a refactor of the tabs api to be a bit more declaritive


src/Styleguide/Pages/Artist/index.js
```diff
@@ -42,145 +42,132 @@ export class Artist extends React.Component {
...
-                                  "Identity Politics",
-                                  "Racial and Ethnic Identity",
-                                  "Allegory",
-                                  "Paper Cut-outs",
-                                  "Sex",
-                                ].map((gene, index, list) => {
-                                  const geneDivider =
-                                    index < list.length - 1 ? "," : ""
+            <Tabs>
+              <Tab name="Overview">
```
damassi: 👍

PR #815, opened 2018-06-15
@xtina-starr => Rename redirectUrl to redirectTo

eessex:
Follows up on https://github.com/artsy/force/pull/2620

- Uses `options` object in props for FormSwitcher
- Renames `redirectUrl` to `redirectTo`
- Adds `destination` to  ModalOptions type


l2succes:
👍

PR #818, opened 2018-06-16
[Styleguide] Move fixtures to /Fixtures directory 

damassi:
Was noticing when trying to bring pages into force that it doesn't like anything to come from `__storybooks__` folders. Moved fixtures out (which were referenced in a few places in our code) so that we can start presenting stuff in force. 

cc @oxaudo, @alloy, @zephraph, @mzikherman

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -84,6 +84,7 @@ module.exports = (baseConfig, env) => {
       extensions: [".mjs", ".js", ".jsx", ".ts", ".tsx"],
       alias: {
         sharify: sharifyPath.replace(/\.js$/, ""),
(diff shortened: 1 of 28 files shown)


.storybook/webpack.config.js
```diff
@@ -84,6 +84,7 @@ module.exports = (baseConfig, env) => {
       extensions: [".mjs", ".js", ".jsx", ".ts", ".tsx"],
       alias: {
         sharify: sharifyPath.replace(/\.js$/, ""),
+        "styled-components": path.resolve("./node_modules/styled-components"),
```
alloy: Why is this needed?

damassi: It’s a multiple copies of lib error due to yarn link

PR #820, opened 2018-06-18
 adding ExtraLinks component to sidebar

oxaudo:
I think this is the last piece that makes sidebar for regular artwork look more or less like design.

I need to go back and double check the fonts, but otherwise I think it looks pretty close.

### DIFF
--- a/src/Styleguide/Pages/Artwork/Sidebar/ExtraLinks.tsx
+++ b/src/Styleguide/Pages/Artwork/Sidebar/ExtraLinks.tsx
@@ -1,0 +1,45 @@
+import React from "react"
+import styled from "styled-components"
+import { Sans } from "@artsy/palette"
+import { space, SpaceProps } from "styled-system"
+
+export interface ExtraLinksProps {
+  artwork: {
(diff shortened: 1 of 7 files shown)


src/Styleguide/Pages/Artwork/Sidebar/ExtraLinks.tsx
```diff
@@ -0,0 +1,45 @@
...
+  render() {
+    const isForSaleArtwork = this.props.artwork.is_for_sale
+    const consignableArtists = this.props.artwork.artists.filter(
+      artist => artist.is_consignable
+    )
+    if (!isForSaleArtwork && consignableArtists.length === 0) {
+      return null
+    }
+    return (
+      <ExtraLinksContainer pb={2}>
```
zephraph: If you just need a container you can use `Box`. It's a generic div w/ some styled-system hooks.

oxaudo:
Ok @zephraph - I think this is good to go. I will follow up with some fonts updates and will replace `LinksContainer` with more generic `Box` there as well

PR #821, opened 2018-06-18
Palette type fixes

zephraph:
Fixes several issues introduced by https://github.com/artsy/palette/pull/23. That PR needs to be merged and deployed before this can move forward.

### DIFF
--- a/src/Styleguide/Components/__stories__/ReadMore.story.tsx
+++ b/src/Styleguide/Components/__stories__/ReadMore.story.tsx
@@ -10,7 +10,7 @@ storiesOf("Styleguide/Components", module).add("ReadMore", () => {
     <React.Fragment>
       <Section title="Max lines">
         <Item>
-          <Sans>
+          <Sans size="3">
             <ReadMore maxLineCount={2}>
               Donald Judd, widely regarded as one of the most significant
               American artists of the post-war period, is perhaps best-known for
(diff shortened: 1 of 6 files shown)


src/Styleguide/Components/__stories__/ReadMore.story.tsx
```diff
@@ -10,7 +10,7 @@ storiesOf("Styleguide/Components", module).add("ReadMore", () => {
     <React.Fragment>
       <Section title="Max lines">
         <Item>
-          <Sans>
+          <Sans size="3">
```
oxaudo: Wondering if there is still a value in allowing one not to specify the size and in this case just inherit the value from a parent. That might come handy when nesting one Sans within the other to just update the font color or something along those lines. 

That would allow us to save a bit of definition in cases like that for example: https://github.com/artsy/reaction/blob/master/src/Styleguide/Pages/Artwork/Sidebar/TitleInfo.tsx#L24

Not that it's really hard to repeat couple of sizes:)

src/Styleguide/Elements/FollowIcon.tsx
```diff
@@ -54,15 +69,16 @@ export class FollowIcon extends React.Component<FollowIconProps> {
...
+        {isFollowed ? (
+          <FollowingHover>
+            <Following />
+            <Unfollow />
+          </FollowingHover>
+        ) : (
+          <FollowHover>
+            <Follow />
+          </FollowHover>
+        )}
```
oxaudo: 👍 A prove that there is always about 10 ways to accomplish any given task!

oxaudo:
I'm going to merge that to get the errors out.

PR #824, opened 2018-06-19
[PATCH types] Fix type errors

damassi:
Fixes type errors from recent Palette release.

### DIFF
--- a/src/Styleguide/Elements/__stories__/Typography.story.tsx
+++ b/src/Styleguide/Elements/__stories__/Typography.story.tsx
@@ -22,7 +22,7 @@ stories.add("Typography", () => {
         <Flex flexDirection="column" p={2}>
           {getTypeSizes("sans").map((size, key) => {
             return (
-              <Sans size={size} key={key}>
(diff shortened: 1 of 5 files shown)


src/Styleguide/Elements/__stories__/Typography.story.tsx
```diff
@@ -22,7 +22,7 @@ stories.add("Typography", () => {
         <Flex flexDirection="column" p={2}>
           {getTypeSizes("sans").map((size, key) => {
             return (
-              <Sans size={size} key={key}>
+              <Sans size={size as any} key={key}>
```
damassi: @alloy - I'm not sure how to coerce the items from the array into what Sans expects, since I can't cast it as a string (its expecting something concrete). Do I import `TypeSizes` from the package? And if so should we move those types top-level into `palette.d.ts` and / or export them from the root export?

alloy:
Looks like this may be a duplicate of https://github.com/artsy/reaction/pull/821

zephraph:
Yep. I've resolved these in 821.

PR #825, opened 2018-06-19
updates to use proper fonts based on specs

oxaudo:
Also refactoring based on @zephraph suggestion to not create new components for container but rather use generic `Box` component.


src/Styleguide/Pages/Artwork/Sidebar/Artists.tsx
```diff
@@ -14,18 +14,16 @@ export interface ArtistsProps {
   }>
 }
 
-const ArtistsContainer = styled.div.attrs<SpaceProps>({})`
-  ${space};
-`
+const ArtistsContainer = styled(Box)``
```
zephraph: You actually don't need to do `styled(Box)`` `. You can just use `ArtistsContainer = Box` being that `Box` is already a styled component.

oxaudo:
@zephraph - added some more rearrangements here. Notably:
- pulled the actual prod data for some artwork of multiple configurations
- added a specs to demonstrate how this data looks in a current sidebar
- noticed that I'm not properly rendering gallery location if gallery location does not have a city specified (it's messed up legacy data when partners were allowed to enter location without city names for online only locations) and corrected that.

alloy:
Nice 👌 

Feel free to merge after addressing @zephraph’s comment about using `Box` directly.

oxaudo:
Ok. Couple of more changes here: with @zephraph help extended Box to accept color and align attributes and removed the rest of specific class instantiations for containers in the Sidebar. 

Also added an example for collecting_institution artwork.

zephraph:
LGTM. Merging.

PR #826, opened 2018-06-19
Fix mobile spacing for some responsive components

zephraph:
Fixes some of the responsive components that require their mobile versions to fill the space that they contain. 

Also a slight update to market insights to have the name of the story match the name of the component


oxaudo:
👍

PR #828, opened 2018-06-19
@zephraph => Remove --copy-files from `yarn compile`

damassi:
Noticed that our source files were being copied over to `dist` leading to errors in force, so removed `--copy-files` from `yarn compile`. Not sure why that was there to begin with?

Also bumps palette to latest.


zephraph:
LGTM

PR #831, opened 2018-06-20
[PATCH redesign] @zephraph => Update pages for display in force

damassi:
Related: https://github.com/artsy/force/pull/2621

First pass and getting pages to play nice in Force. Will annotate things inline.


src/Styleguide/Components/Footer.tsx
```diff
@@ -33,7 +33,7 @@ const FooterContainer: React.SFC<FooterContainerProps> = props => {
         <Sans size="2" weight="medium">
           Buy
         </Sans>
-        <Serif size="2">
+        <Serif size="3t">
```
damassi: I _think_ all Serif sizes are off by 1.

src/Styleguide/Pages/Artist/index.js
```diff
@@ -1,62 +1,74 @@
...
     return (
-      <Grid fluid>
-        <Row>
-          <Col>
-            <ArtistHeader />
-          </Col>
-        </Row>
+      <Theme>
+        <GlobalStyles>
+          <ResponsiveProvider breakpoints={themeProps.mediaQueries}>
```
damassi: For stuff under root `/Pages` I've pulled in our providers since these pages will likely be rendered wholesale like this in Force. See [here](https://github.com/artsy/force/pull/2621/files#diff-efaf83a3a5ab87f1c6b9501f8e3dbb10R12) for an example. (Note that we still need to start defining overall routing structure, too. I think I'm going to work on that next)

src/Styleguide/Pages/__stories__/Pages.story.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react"
-import { storiesOf } from "storybook/storiesOf"
+import { storiesOf } from "@storybook/react"
```
damassi: Swapped this with normal `storiesOf` since `/Pages` will handle their own providers.

PR #833, opened 2018-06-20
minor: re-add sidebar for artwork page

oxaudo:
Not sure how that got messed up, but adding it back.


src/Styleguide/Pages/Artwork/Sidebar/Artists.tsx
```diff
@@ -41,7 +41,7 @@ export class Artists extends React.Component<ArtistsProps> {
     const { artists } = this.props
     return artists.map((artist, index) => {
       return (
-        <React.Fragment>
+        <React.Fragment key={artist.__id}>
```
oxaudo: Not having a key here was giving a js warning.

oxaudo:
Self merging this to bring sidebar back to the party!

alloy:
We had a bunch of merge issues in this file so that’s probably the cause 👍

PR #835, opened 2018-06-20
Starting to differentiate Sidebar for Auction Artwork

oxaudo:
This brings the easy part that is going to vary. Still missing the main bidding info section here.

I'll make several comments inline.


src/Styleguide/Pages/Artwork/Sidebar.tsx
```diff
@@ -71,10 +75,29 @@ export class Sidebar extends Component<ArtworkSidebarProps> {
     return (
       <SidebarContainer>
         <Artists artists={artwork.artists} />
+
+        {artwork.is_biddable &&
```
oxaudo: Was not sure how much of the checking for auction artwork to do here vs in the sidebar elements. There is no exact rule here, just kinda went with what feels right.

In this case - we wanted to prepend the lot number (very simple string) if the artwork is in the auction and the auction is in progress. Figured that testing for `artwork.is_biddable` should do the trick here.

src/Styleguide/Pages/Artwork/Sidebar.tsx
```diff
@@ -71,10 +75,29 @@ export class Sidebar extends Component<ArtworkSidebarProps> {
...
+          artwork.sale_artwork &&
```
oxaudo: Not sure if we really need those additional tests since in real life there should not be the case where artwork `is_biddable` but does not have a sale_artwork association. But that does not hurt anything and makes sure the stories are less likely to crash.

src/Styleguide/Pages/Artwork/Sidebar.tsx
```diff
@@ -71,10 +75,29 @@ export class Sidebar extends Component<ArtworkSidebarProps> {
...
+            <Serif size="2" weight="semibold" color="black100">
+              Lot {artwork.sale_artwork.lot_label}
+            </Serif>
+          )}
         <ArtworkMetadata artwork={artwork} />
-        <Separator />
-        <Commercial artwork={artwork} />
-        <PartnerInfo artwork={artwork} />
+
+        {!artwork.is_in_auction && (
```
oxaudo: Instead of the Commertial info artwork from the Auctions should include Bidding Activity component (or whatever we decide to name it). The tricky thing is that in case of regular artwork this commercial info comes before partner info and in case of auction artwork Bidding activity info comes after Auction Partner info.

So doing this check separately here

src/Styleguide/Pages/Artwork/Sidebar.tsx
```diff
@@ -71,10 +75,29 @@ export class Sidebar extends Component<ArtworkSidebarProps> {
...
+            <Commercial artwork={artwork} />
+          </React.Fragment>
+        )}
+
+        {artwork.is_in_auction ? (
+          <AuctionPartnerInfo artwork={artwork} />
+        ) : (
+          <PartnerInfo artwork={artwork} />
+        )}
+
```
oxaudo: TODO: That is where the hard part for the Auction artwork will go!

src/Styleguide/Pages/Artwork/Sidebar.tsx
```diff
@@ -71,10 +75,29 @@ export class Sidebar extends Component<ArtworkSidebarProps> {
...
-        <PartnerInfo artwork={artwork} />
+
+        {!artwork.is_in_auction && (
+          <React.Fragment>
+            <Separator />
+            <Commercial artwork={artwork} />
+          </React.Fragment>
+        )}
+
+        {artwork.is_in_auction ? (
```
oxaudo: This will include closed auctions as well.

src/Styleguide/Pages/Artwork/Sidebar/AuctionPartnerInfo.tsx
```diff
@@ -0,0 +1,51 @@
...
+    readonly sale_artwork?: {
+      readonly estimate?: string
+    }
+    readonly sale?: {
+      readonly is_with_buyers_premium?: boolean
+    }
+  }
+}
+
+const AuctionPartnerInfoContainer = Box
```
damassi: This is nice, I've started using this pattern too

src/Styleguide/Pages/Artwork/__stories__/Sidebar/AuctionPartnerInfo.story.tsx
```diff
@@ -0,0 +1,66 @@
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { Section } from "../../../../Utils/Section"
```
damassi: Can use absolute paths: `Styleguide/Utils/Section`

src/Styleguide/Pages/Artwork/__stories__/Sidebar/AuctionPartnerInfo.story.tsx
```diff
@@ -0,0 +1,66 @@
...
+    __id: "phillips",
+    name: "Phillips",
+  },
+  sale_artwork: {
+    estimate: "$300 - $500",
+  },
+  sale: {
+    is_with_buyers_premium: true,
+  },
+}
```
damassi: Can you move these items to `Styleguide/Pages/Fixtures/Artist`?

src/Styleguide/Pages/Artwork/Sidebar/AuctionPartnerInfo.tsx
oxaudo: Yea... I think it increases readability in code but lame part that it does not seem to prepend this name when exploring things in inspector in the browser. Need to figure out if there is a way to force that. Makes debugging nicer:)

oxaudo:
@damassi - updated here, will merge when green:)

PR #837, opened 2018-06-20
Create artwork grid responsive wrapper

zephraph:
This creates a wrapper for the ArtworkGrid to make it easier to use in the Styleguide for now. Also wires it up on the Artwork page.


src/Styleguide/Components/ArtworkGrid.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import { Responsive } from "../Utils/Responsive"
+import LegacyArtworkGrid from "Components/ArtworkGrid"
+
+export const ArtworkGrid = props => (
+  <Responsive>
+    {({ xs, sm }) => (
+      <LegacyArtworkGrid columnCount={xs ? 2 : sm ? 3 : 4} {...props} />
```
damassi: While this nested ternary reads pretty well right here would you mind breaking this out a bit into a normal if / else conditional or switch? These are hard to unpack when returning to code after being away for a while.

zephraph:
Doesn't look like it. Artwork page `n - 1`.

zephraph:
I updated the columns in the overview section of the artist page. Updated the ArtworkGrid placeholder component to have a `Full` version. Not super happy with that, but it does get us a little closer on the artwork page.

damassi:
👍

PR #842, opened 2018-06-21
Used definitely typed styles, fix type errors

zephraph:
Upgrades palette, pulls in styled-systems types from definitely typed, and fixes all the type errors that occurred.


package.json
```diff
@@ -86,6 +86,7 @@
     "@types/react-transition-group": "^2.0.7",
     "@types/relay-runtime": "^1.3.2",
     "@types/storybook__react": "^3.0.5",
+    "@types/styled-system": "^2.3.0",
```
damassi: 💯

damassi:
@zephraph - really awesome walking this all the way through `@types`. Def a good blog post topic ;) 

Danger, merge on green 👍

PR #846, opened 2018-06-21
[PATCH lib] Fix CI by adding react-select (?)

damassi:
Fixes storybook deployment issue that I saw when SSH'ing into CI: 

screen shot 2018-06-21 at 1 40 59 pm


package.json
```diff
@@ -119,6 +119,7 @@
     "react-addons-test-utils": "^15.5.1",
     "react-docgen-typescript-loader": "^2.1.0",
     "react-dom": "^16.2.0",
+    "react-select": "^1.2.1",
```
damassi: This isn't the proudest fix, but lets see if it does the trick? Not sure what was up with that missing dep issue.

PR #849, opened 2018-06-21
Authentication Documentation

kanaabe:
Thanks for the mob documentation session, Grow! Refined some of the wording a bit 😄. Also, how do we feel about this being in Reaction considering how high-level some of this is? I personally prefer it in GH rather than Notion since it will encourage more activity.


docs/authentication.md
```diff
@@ -0,0 +1,91 @@
...
+### Embed
+Similar to the above, authentication forms can be embedded anywhere on the site. (TODO: Add Artist CTA example once implemented)
+
+## Implementation
+### Basics
+We use [Formik](https://github.com/jaredpalmer/formik) to handle form entries and [Yup](https://github.com/jquense/yup) for form validation. Mobile uses the `Wizard` component to handle the multi-step flow. The form submission is handled in Force, and the authentication backend is handled in [artsy-passport](https://github.com/artsy/artsy-passport).
+
+### Component Responsibilities
+**Desktop**
+- `ModalContainer` (Force): Listens for changes to Force's mediator and triggers opens to ModalManager. It is also responsible for setting Cookies like `destination` and `afterAuthAction`.
```
eessex: Can we link to these components directly?

damassi: It would also be good to link to the associated StoryBooks entry

l2succes:
I agree it should be in reaction but since we're consolidating company-wide docs in Notion we should have a copy/link in there as well

docs/authentication.md
kanaabe: Just linked the Reaction stuff for now since we're going to swap auth2 -> auth soon in Force. I'll link it once that work is done!

kanaabe:
Cool, gonna merge this! I'll link this to Notion after 👍

PR #853, opened 2018-06-22
adding the bid info section for Auction artwork sidebar. Missing me related bid info

oxaudo:
Another baby step in the auction sidebar. A bit ashamed to admit how long it took me to get the logic just for those:
screen shot 2018-06-21 at 10 37 26 pm

I still want to chat with @sweir27 at some point (probably early next week) to verify that my assumptions here make sense and get some more info in certain questions:
- Is this view applicable to auction in `is_preview` state?
- nothing will differ here for auctions that are currently in `is_live_open` state?
- how would the option to `buy now` interacts with this bids info, namely is it possible for an artwork to be in the open auction, have bids but also being sold already? Should this view display `Sold` in this case?
- there is quite a lot of data manipulation going on in this component because MP does not seem to return the display strings that we want based on newer designs(https://app.zeplin.io/project/5b113f413ede4bb10dbfdc42/screen/5b114000413156335e285483). Would not it be better to have this logic in MP?


damassi:
@oxaudo - we've been moving a lot of logic into MP so this might be a good candidate for a refactor pass as long as tests. It looks good to merge now, though 👍

PR #856, opened 2018-06-22
@mzikherman => Start exploring routing

damassi:
1st pass, basically just moving some stuff around and experimenting; will continue in second PR. Opening up this PR because i moved our init stuff into `<Boot>` on /Artist and /Artworks and don't want to create any conflicts.


damassi:
(Merging per offline convo)

PR #859, opened 2018-06-22
@eessex => Rename ResetPassword

kanaabe:
- Renames `ResetPassword` to `ForgotPassword`
- Fixes Authentication Storybook

### DIFF
--- a/src/Components/Authentication/__tests__/Desktop/ModalManager.test.tsx
+++ b/src/Components/Authentication/__tests__/Desktop/ModalManager.test.tsx
@@ -8,8 +8,8 @@ const getWrapper = (): ReactWrapper<ModalManagerProps> => {
     <ModalManager
       submitUrls={{
         login: "/login",
-        signup: "/sign_up",
-        forgot_password: "/forgot_password",
+        signup: "/signup",
+        forgot: "/forgot",
       }}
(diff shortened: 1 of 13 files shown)


src/Components/Authentication/__tests__/Desktop/ModalManager.test.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { ModalManager, ModalManagerProps } from "../../Desktop/ModalManager"
 
 const getWrapper = (): ReactWrapper<ModalManagerProps> => {
   const wrapper = mount(
     <ModalManager
       submitUrls={{
         login: "/login",
         signup: "/sign_up",
-        reset_password: "/reset_password",
+        forgot_password: "/forgot_password",
```
eessex: I believe the route is '/forgot' in Force, not sure about submission but worth double-checking

kanaabe: Ah yeah, this is just a test and we grab the url from Force side but thanks for pointing out the inconsistency! Will update.

kanaabe: Ohhh but you know what, Force uses the option `forgot` which will need to be changed throughout!

PR #862, opened 2018-06-22
Enable tab preloading

zephraph:
Update the `Tabs` component to allow route functionality via the preload link.

### DIFF
--- a/src/Styleguide/Components/Tabs.tsx
+++ b/src/Styleguide/Components/Tabs.tsx
@@ -1,5 +1,7 @@
 import { Sans } from "@artsy/palette"
 import React from "react"
+import { PreloadLink } from "Router"
+import { PreloadLinkProps } from "Router/types"
 import styled from "styled-components"
 import { borders, themeGet, WidthProps } from "styled-system"
 import { Box } from "Styleguide/Elements/Box"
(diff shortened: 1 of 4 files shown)


src/Styleguide/Components/Tabs.tsx
```diff
@@ -76,7 +83,10 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
 interface TabProps {
   name: string
 }
-export class Tab extends React.Component<TabProps> {
+interface PreloadLinkTabProps extends TabProps, PreloadLinkProps {
+  children: never
```
zephraph: This doesn't work. @alloy have any advice here? It seems that typescript isn't using this to check the type of the children.

src/Styleguide/Components/Tabs.tsx
```diff
@@ -49,12 +51,17 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
   }
 
   renderTab = (tab, index) => {
-    const { name } = tab.props
+    const { name, children, ...preloadProps } = tab.props
+    const buttonContent = Object.keys(preloadProps).length ? (
```
zephraph: Maybe could use more robust logic to see if it's a preload or not.

damassi:
LGTM 👍 - gonna merge and we can iterate.

src/Styleguide/Components/Tabs.tsx
damassi: Can you add a `FIXME` here with this comment? At some point we can check the shape of `preloadProps` against the interface.

zephraph:
This version doesn't work... integrating everything into one component is a bit too much.

damassi:
I've got a naive CSS version going in my branch that piggybacks on router features, when ready i'll go ahead and add back here where appropriate

PR #874, opened 2018-06-24
@mzikherman => Add <Slider> component 

damassi:
Add slider in. Mostly done but need to work out some kinks in mobile: 

slider


damassi:
Nothing controversial here, just basic Slick slider config. Going to merge this to get a Force staging deploy going.

package.json
```diff
@@ -168,7 +168,7 @@
     "react-responsive-decorator": "^0.0.1",
     "react-router": "^4.2.0",
     "react-sizeme": "^2.4.4",
-    "react-slick": "^0.14.11",
+    "react-slick": "^0.23.1",
```
kanaabe: Looks like this version change is currently affecting our ad slideshow component due to [this issue](https://github.com/akiran/react-slick/pull/1242). Would it be ok if I reverted the version back until we can fix it @damassi? We have a few other priorities atm but this is affecting our display partners.

kanaabe: It looks like the version marked as "Latest Release" in GH is working for us as well so we don't have to go back _too_ far. [0.21.0](https://github.com/akiran/react-slick/releases/tag/0.21.0).

damassi: sure thing! Sorry about that -- i clicked around storybooks and everything seemed to be ok.

PR #880, opened 2018-06-25
@kanaabe: [WIP] Auth gdpr properties

xtina-starr:
This addresses the last two items in the original [Desktop Modal QA Ticket](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-702). It updates the gdpr prop to use the [correct name](https://github.com/artsy/force/blob/master/src/desktop/models/logged_out_user.coffee#L80) and change the font for signup/login to `sans 3t med`.


src/Components/Authentication/Desktop/SignUpForm.tsx
```diff
@@ -92,14 +94,16 @@ export const SignUpForm: FormComponentType = props => {
...
-              name="acceptedTermsOfService"
+              name="accepted_terms_of_service"
               onChange={handleChange}
               onBlur={handleBlur}
             />
             {status && !status.success && <Error show>{status.error}</Error>}
-            <SignUpButton disabled={isSubmitting || hasErrors || !dirty}>
-              Sign Up
-            </SignUpButton>
+            <Sans size="3t">
```
xtina-starr: I wanted to use palette here but I'm finding I would have to duplicate it in a lot of places (i.e. sign up for mobile, login for mobile and desktop as well as forgot password). Ideally I'd like for this to be set in the `<SignUpButton>` styled div. I saw [this approach](https://github.com/artsy/reaction/blob/master/src/Styleguide/Elements/FollowIcon.tsx#L18-L22) but not sure how that might work with the way we are setting the `attr` on this element.

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
```diff
@@ -54,12 +46,8 @@ export const ForgotPasswordForm: FormComponentType = props => {
...
             />
             {status && !status.success && <Error show>{status.error}</Error>}
-            <SubmitButton
-              block
-              type="submit"
-              disabled={isSubmitting || hasErrors}
-            >
-              Send Reset Instructions
+            <SubmitButton disabled={isSubmitting || hasErrors}>
+              Send reset instructions
```
kanaabe: Ah so clean :)

src/Components/Authentication/commonElements.tsx
```diff
@@ -1,6 +1,8 @@
+import { Sans } from "@artsy/palette"
```
damassi: Yessss 👍

PR #888, opened 2018-06-26
[Router] Add initial route config option

damassi:
Adds a new `initialRoute` config option which is good for building router apps in storybooks:

```jsx
<StorybooksRouter routes={routes} initialRoute="/andy-warhol/cv" />
```


src/Router/__test__/buildClientApp.test.tsx
```diff
@@ -3,13 +3,18 @@ import React from "react"
 import { buildClientApp } from "../buildClientApp"
 
 describe("buildClientApp", () => {
-  const getWrapper = async (props = {}) => {
+  const getWrapper = async (initialRoute = "/") => {
     const { ClientApp } = await buildClientApp({
       historyProtocol: "memory",
+      initialRoute: initialRoute || "/",
```
alloy: Would be slightly nicer if a test didn’t replicate behaviour of the actual code.

PR #898, opened 2018-06-26
@damassi => Small updates for Positron

eessex:
Necessary to bump Reaction in Positron:
- In `Text` component, moves require for Relay `LinkWithTooltip` so it is only imported if needed.
- Updates path for imported `ErrorBoundary` in `Truncator`, which was not resolving correctly.

### DIFF
--- a/src/Components/Publishing/Sections/Text.tsx
+++ b/src/Components/Publishing/Sections/Text.tsx
@@ -76,7 +76,7 @@ export class Text extends Component<Props, State> {
   transformNode = (node, index) => {
     // Dont include relay components unless necessary
     // To avoid 'regeneratorRuntime' error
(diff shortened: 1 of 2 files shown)


damassi:
👍

src/Components/Publishing/Sections/Text.tsx
```diff
@@ -75,6 +74,10 @@ export class Text extends Component<Props, State> {
   }
 
   transformNode = (node, index) => {
+    // Dont include relay components unless necessary
+    // To avoid 'regeneratorRuntime' error
+    const { LinkWithTooltip } = require("../ToolTip/LinkWithTooltip")
```
damassi: This seems fine for now but I think we're going to have to live with the regenerator issue for now, so might be good to account for it in various clients which consume reaction.

PR #902, opened 2018-06-26
 add the user bid info to CurrentBidInfo

oxaudo:
Not sure what is the best way to add this me data to the Sidebar component itself. But for now created all this different datapoints on CurrentBidInfo itself by just passing a separate `me` property.

@alloy - let me know if this approach seems ok to you. I'm basically planning to do the same for the last BidActions component that I'm adding next. And if we can somehow reuse the AuctionClock from emission - I think the static version of this sidebar will be pretty much done. Hope to accomplish that tomorrow. Hooking it up to the real data - not sure I'll have time to accomplish that:(

Variations added by this pr:
screen shot 2018-06-26 at 6 08 57 pm


src/Styleguide/Pages/Artwork/Components/ArtworkSidebar/ArtworkSidebarExtraLinks.tsx
```diff
@@ -52,12 +54,16 @@ export const ArtworkSidebarExtraLinksFragmentContainer = createFragmentContainer
...
       __id
-      is_biddable
+      is_in_auction
       is_for_sale
       artists {
         __id
         is_consignable
       }
+      sale {
+        __id
```
alloy: I haven’t paid close attention to the `__id` selections before. What is this being selected for?

oxaudo: Actually don't need it here. Only needed it when getting an array of objects to be able to put keys for the elements.

But we might need it later for links I think. Will remove for now though

src/Styleguide/Pages/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
```diff
@@ -6,16 +6,36 @@ import { Box } from "Styleguide/Elements/Box"
 import { Flex } from "Styleguide/Elements/Flex"
 
 import { ArtworkSidebarCurrentBidInfo_artwork } from "__generated__/ArtworkSidebarCurrentBidInfo_artwork.graphql"
+import { LoosingBid } from "Assets/Icons/LoosingBid"
+import { WinningBid } from "Assets/Icons/WinningBid"
 
 export interface ArtworkSidebarCurrentBidInfoProps {
   artwork: ArtworkSidebarCurrentBidInfo_artwork
+  readonly me?: {
```
oxaudo: Not sure it's valid to combine those like this. But seems to not complain for now. Just using this to render static elements. Definitely need to change.

oxaudo:
@alloy - I tried to merge with the latest changes here. All looks good in the code and the storybook parts work for me locally. The whole artwork/artist pages don't load locally for me though:(

Maybe you'll have a sec tomorrow to help me look at that?

alloy:
Yeah absolutely! 👍

src/Styleguide/Pages/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -65,10 +67,14 @@ export const ArtworkSidebarFragmentContainer = createFragmentContainer(
       sale_artwork {
         lot_label
       }
+      sale {
+        is_closed
+      }
```
alloy: This doesn’t get used inside this file, so it shouldn’t be fetching the data here.

alloy:
Looks good, I’ll remove that unnecessary fetch when replacing the `me` access with MP schema changes.

src/Styleguide/Pages/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx
```diff
@@ -0,0 +1,119 @@
+// import { Serif } from "@artsy/palette"
+// import { Sans } from "@artsy/palette"
```
alloy: Oh, stray imports

PR #907, opened 2018-06-27
[Router] Refactor PreloadLink and add global state pattern

damassi:
Cleans up / documents `<PreloadLink>` and lay out potential global state pattern via [`unstated`](https://github.com/jamiebuilds/unstated). 

Also adds transition hooks for tabs, which connects to PreloadLinkState: 

loader


damassi:
This is a big PR but i'm gonna merge and keep going. We can iterate on this.

src/Router/PreloadLink.tsx
```diff
@@ -1,177 +1,281 @@
...
+ *       return (
+ *         <PreloadLink to='/some-path'>
+ *           Fetching: {isFetching}
+ *         </PreloadLink>
+ *       )
+ *     }}
+ *   </Subscribe>
+ * )
+ */
+export class State extends Container<PreloadLinkState> {
```
damassi: This export pattern lets us take state that we're interested and "subscribe" to it anywhere in the tree. Keeps things nice and decoupled and great for lib code.

src/Styleguide/Pages/Artist/Components/LoadingArea.tsx
```diff
@@ -0,0 +1,46 @@
...
+        const loaderClass = isFetching ? "loading" : ""
+
+        return (
+          <React.Fragment>
+            <SpinnerToggle className={loaderClass} />
+            <Container className={loaderClass}>{props.children}</Container>
+          </React.Fragment>
+        )
+      }}
+    </Subscribe>
```
damassi: For example here. Import `PreloadLinkState` and subscribe, and now we've connected our main content area to the current state of the route request which enables some nice controlled behavior.

src/Styleguide/Pages/Artist/state.tsx
```diff
@@ -0,0 +1,23 @@
...
+      isLoggedIn: true,
+    })
+  }
+
+  logOut() {
+    this.setState({
+      isLoggedIn: false,
+    })
+  }
+}
```
damassi: Not being used ATM, just a sketch pad

alloy: How about we try to use local schema extension for this specific state?

src/Styleguide/Pages/Artist/Components/LoadingArea.tsx
alloy: Doesn’t that tie this component to its parent? Couldn’t this be done as well by taking a `isLoading` prop and having the parent pass in its `isFetching` state?

src/Router/PreloadLink.tsx
alloy: I feel like I’m missing something, the example being used in this change (and that I commented about below) is probably too contrived for me to see the bigger picture.

src/Styleguide/Pages/Artist/state.tsx
damassi: Oh yeah, I'm curious about this, lets talk after meeting.

src/Styleguide/Pages/Artist/Components/LoadingArea.tsx
damassi: The issue is that we're working across distant areas in the tree which make this sort of passing impractical and rigid. Will elaborate in Slack.

PR #912, opened 2018-06-27
Allow Tabs to have user-specified data and justification 

anandaroop:
Reaction noob in the house!

This is the first example of consuming a design system component in CMS, and then customizing it  here per our needs in the design spec.

I'm open to the possibility that everything here is done wrong 😛 , but this is where I've gotten so far.

Example usage below…

gif

H/t @orta for the suggestion to start documenting the props interfaces, and seeing that flow through to the Volt side in VS Code.

cc @jonallured

### DIFF
--- a/src/Styleguide/Components/Tabs.tsx
+++ b/src/Styleguide/Components/Tabs.tsx
@@ -22,7 +22,7 @@ export interface TabInfo {
   tabIndex: number
 
   /** Data associated with the newly selected Tab */
-  data: object
+  data: any
 }
 
 export interface TabsProps extends WidthProps, JustifyContentProps {
@@ -100,7 +100,7 @@ interface TabProps {
    *
    * Will be passed to the parent <Tabs>'s onChange handler.
    */
-  data?: object
+  data?: any
 }
 
 export class Tab extends React.Component<TabProps> {
--- a/src/Styleguide/Components/Tabs.tsx
+++ b/src/Styleguide/Components/Tabs.tsx
@@ -1,7 +1,12 @@
 import { Sans } from "@artsy/palette"
 import React from "react"
 import styled, { css } from "styled-components"
-import { borders, themeGet, WidthProps } from "styled-system"
+import {
+  borders,
+  JustifyContentProps,
+  themeGet,
+  WidthProps,
+} from "styled-system"
 import { Box } from "Styleguide/Elements/Box"
 import { Flex } from "Styleguide/Elements/Flex"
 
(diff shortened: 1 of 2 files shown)


src/Styleguide/Components/__stories__/Tabs.story.tsx
```diff
@@ -37,6 +37,25 @@ storiesOf("Styleguide/Components", module).add("Tabs", () => {
           <Tab name="You?">You?</Tab>
         </Tabs>
       </Section>
+      <Section title="With arbitrary data and onChange handler">
+        <Tabs
+          onChange={activeTab => {
+            // tslint:disable-next-line
+            console.log(activeTab.data)
+          }}
```
damassi: 👍

damassi:
Looks great @anandaroop, also welcome to TypeScript land -- it really makes things easier with time even if only a minimal subset of features is used.

src/Styleguide/Components/Tabs.tsx
```diff
@@ -74,8 +92,17 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
...
 interface TabProps {
+  /** Display name of the Tab */
   name: string
+
+  /**
+   * Arbitrary data that can be associated with a Tab.
+   *
+   * Will be passed to the parent <Tabs>'s onChange handler.
+   */
+  data?: object
```
alloy: If this should truly be any type of data, then maybe this needs to be `any` so scalars like numbers are supported too.

anandaroop: Good call, fixed in fdc609ce252ea185a3771e5160bf5257eda87f76

anandaroop:
Thanks @damassi I am definitely on board 😄 

I'm taking the WIP off this, and I re-wrote the commit messages based on my understanding of how semver works in this repo.

src/Styleguide/Components/Tabs.tsx
```diff
@@ -60,11 +78,11 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
   }
 
   render() {
-    const { children = [] } = this.props
+    const { children = [], justifyContent = "left" } = this.props
```
zephraph: Instead of giving justifyContent a default here, how about we add it as a part of the class. This way if we ever integrate a tool to read in the default props it'll be picked up. 

```typescript
export class Tabs extends React.Component<TabsProps, TabsState> {
 public static defaultProps: Partial<TabsProps> = {
        justifyContent: "left"
    }
}
```

zephraph:
One comment, but otherwise LGTM.

src/Styleguide/Components/Tabs.tsx
anandaroop: Yes, that's better! Fixed in ee239365589aab88ef5c4837ba308cc12b2f1d88

Good to know about `Partial` too — thanks for the tip 😃

PR #913, opened 2018-06-27
add disabled option to button

oxaudo:
Added it as a property that overwrites both loading and variant.

### DIFF
--- a/src/Styleguide/Elements/Button.tsx
+++ b/src/Styleguide/Elements/Button.tsx
@@ -165,6 +165,7 @@ export interface ButtonBaseProps
     HeightProps {
   buttonSize?: any // FIXME
   loading?: boolean
+  disabled?: boolean
   onClick?: () => void
   variantStyles?: any // FIXME
 }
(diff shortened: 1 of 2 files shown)


src/Styleguide/Elements/Button.tsx
```diff
@@ -176,12 +177,13 @@ export class ButtonBase extends Component<ButtonBaseProps> {
...
   render() {
-    const { children, loading, ...rest } = this.props
+    const { children, loading, disabled, ...rest } = this.props
     const textProps = pick(["color", "size", "weight"], rest)
     const loadingClass = loading ? "loading" : ""
+    const disabledClass = disabled ? "disabled" : ""
 
     return (
-      <Container {...rest} className={loadingClass}>
+      <Container {...rest} className={[loadingClass, disabledClass].join(" ")}>
```
damassi: 👍

damassi: I've typically used `classNames` lib for joining this stuff but i since classes are pretty minimal in Reaction at this point this seems fine

PR #919, opened 2018-06-28
[PATCH qa] QA Issues 

damassi:
- Make JSON serialization safer (**TODO**: figure out what's causing circular ref issues *only* on staging)
- Pass in initial breakpoint for responsive SSR rendering
- Fix Current event square
- Fix Overview grid filter layout

### DIFF
--- a/src/Router/AppShell.tsx
+++ b/src/Router/AppShell.tsx
@@ -6,6 +6,13 @@ import { AppShellProps } from "./types"
 export const AppShell: SFC<AppShellProps> = props => {
   const { loadableState, data, children, provide } = props
 
+  let hydrationData = {}
+  try {
+    hydrationData = JSON.stringify(data)
+  } catch (error) {
+    console.error("reaction/Router/AppShell Error serializing data:", error)
+  }
+
   return (
     <ContextProvider {...provide}>
       <div>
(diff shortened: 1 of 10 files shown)


src/Styleguide/Pages/Boot.tsx
```diff
@@ -12,10 +13,19 @@ injectGlobalCSS(`
   }
 `)
 
-export const Boot = ({ children }) => {
+export interface BootProps {
+  initialState?: {
+    breakpoint: Breakpoint
+  }
```
damassi: This was the best i could come up with right now in the 11th hour :/

src/Router/AppShell.tsx
```diff
@@ -4,7 +4,15 @@ import { ContextProvider } from "../Components/Artsy"
 import { AppShellProps } from "./types"
 
 export const AppShell: SFC<AppShellProps> = props => {
-  const { loadableState, data, children, provide } = props
+  const { loadableState, data = {}, children, provide } = props
+
+  let hydrationData
+  try {
+    hydrationData = serialize(data, { isJSON: true })
```
zephraph: 👍

damassi: this file is so ugly... really need to refactor all of this

zephraph: We'll get there.

PR #930, opened 2018-06-29
@eessex => Remove `shouldComponentUpdate` on Modal

kanaabe:
We should be allowed to update on any changes to the modal such as type changes in `FormSwitcher`


kanaabe:
Gonna merge as per discussion!

PR #932, opened 2018-06-29
Resolve error causing boot component to fail

zephraph:
Resolves an issue that was preventing all the component stories to fail to render to an error in the boot component.


damassi:
Thanks for fixing this! 

LGTM 👍 

merge on green

damassi:
@zephraph - can you update the commit message so that deploys

zephraph:
Will do.

PR #950, opened 2018-07-06
@zephraph => Universal modal component

eessex:
Addresses [Grow-703](https://artsyproduct.atlassian.net/browse/GROW-703) and the Reaction-side of [Grow-704](https://artsyproduct.atlassian.net/browse/GROW-704).

Refactors Modal component to match [design system specs](https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5b1ac304a2d566315992711f). 
Updates Auth components to use new Modal attributes.
Removes `DesktopHeader` and unused `ModalHeader` component.

This will require a small update in Force to use new `ModalHeader` rather than removed `DesktopHeader` component.

screen shot 2018-07-06 at 2 37 45 pm
screen shot 2018-07-06 at 2 37 56 pm


artsyit:

Built with commit d043287414ab98a89b11e419e09fa21921351f2d

https://deploy-preview-950--artsy-reaction.netlify.com

zephraph:
Eeee! It's so exciting 😁. I'll dig in shortly.

src/Components/Modal/Modal.tsx
```diff
@@ -143,4 +211,28 @@ export const ModalOverlay = styled.div`
   background: rgba(200, 200, 200, 0.5);
 `
 
+export const CloseButton = styled(Icon).attrs({
+  color: Colors.graySemibold,
```
zephraph: Instead of using the colors here, I recommend grabbing them from palette.

https://github.com/artsy/palette/blob/master/src/Theme.tsx#L37

You can use it by...

```typescript
import { color } from "@artsy/palette"

color("black60");
```

If your interested in why the colors are named the way that they are, checkout the design system doc in notion. 

https://www.notion.so/artsy/Color-a0c24896daf8433d9409aee2146ac267

src/Components/Authentication/__tests__/Desktop/Components/DesktopModal.test.tsx
```diff
@@ -1,9 +1,7 @@
 import { mount } from "enzyme"
 import React from "react"
-import {
-  CloseButton,
-  DesktopModal,
-} from "../../../Desktop/Components/DesktopModal"
+import Icon from "../../../../Icon"
+import { DesktopModal } from "../../../Desktop/Components/DesktopModal"
```
zephraph: We have the `src` directory set in jest's [moduleDirectories](https://jestjs.io/docs/en/configuration#moduledirectories-array-string) setting, so we shouldn't have to do relative imports of these components. For icon it could be 

```typescript
Import icon from "Components/Icon"
```

src/Components/Modal/ModalCta.tsx
```diff
@@ -0,0 +1,57 @@
...
+    border-top: 1px solid ${Colors.grayRegular};
+  `};
+
+  ${media.sm`
+    padding-bottom: 20px;
+    ${props =>
+      props.isFixed &&
+      `
+      right: 20px;
+      left: 20px
```
zephraph: Something you _could_ use in some of these places in lieu of hardcoding pixel values is to convert them to [spacing units](https://www.notion.so/artsy/Spacing-93eeaed9fdf9480099fec7094fd1b9f3#3e3c921525b7468f93cd2c2c1a662b6d). Like colors, Palette has first class support for spacing units. 

```typescript
import { space } from "@artsy/palette"

`
right: ${space(2)}px
left: ${space(2)}px
`
```

It might seem a little unnecessary from the outset, but over time it'll help us unify how we handle space between things. With that said, I don't think it needs to be tackled now or in this PR. It is an option though.

zephraph:
This looks great! None of my comments are blockers. Free free to merge at your discretion.

src/Components/Authentication/__tests__/Desktop/Components/DesktopModal.test.tsx
eessex: Oh cool -- had some issues in the past with absolute paths in jest, but seems to run ok now.  Vscode still gives me errors, maybe there's a local config I'm missing?

PR #952, opened 2018-07-08
Add the max bid line.

oxaudo:
Still need to add dropdown with increments.

@alloy - I think this is mergeable but still needs to have a dropdown with increments listed. I added the data but still missing the actual element. Need to look up how we do dropdowns within the styleguide.


artsyit:

Built with commit eba6b734c79bf11d0824652c6b9c1a268c56978f

https://deploy-preview-952--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarArtists.tsx
```diff
@@ -10,7 +10,7 @@ export interface ArtistsProps {
   artwork: ArtworkSidebarArtists_artwork
 }
 
-type Artist = ArtworkSidebarArtists_artwork["artists"][0]
+export type Artist = ArtworkSidebarArtists_artwork["artists"][0]
```
alloy: Why s this being exported? I don’t see it being used anywhere, but also these types should generally stay local to the module that they are used in.

alloy:
Besides that export this looks good, merge after fixing 👍

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarArtists.tsx
alloy: How was it complaining?

oxaudo: It was red underlined below and telling that we are using the type that is not exported. @alloy

oxaudo: btw... The actual warn is: `[ts] Parameter 'artist' of public method from exported class has or is using private name 'Artist'`. This actually makes sense since some functions on this component take Artist as a parameter.

alloy: Ohh, I see. Hmm, maybe we should mark the method as private, because it isn’t even meant to be used outside of the component.

PR #964, opened 2018-07-10
[PATCH authentication] add missing analytics properties for social auth

xtina-starr:
Addresses https://artsyproduct.atlassian.net/browse/GROW-737 and https://artsyproduct.atlassian.net/browse/GROW-780 by adding support for tracking social auth analytics properties. We are explicitly passing the service provider the user authenticated with as well fixing the 'create account' and 'successfully logged in' events.


l2succes:
👍

PR #965, opened 2018-07-10
First pass at update circle config

zephraph:
First pass at circle ci update for reaction. 

Somewhat kinda crazy. Just doing this because these build times are killin' me >_>


artsyit:

Built with commit 672c40f1365b0d8e17263d768881143254732053

https://deploy-preview-965--artsy-reaction.netlify.com

zephraph:
If someone merges this, please do a squash merge 😆

verify-node-version.js
```diff
@@ -1,6 +1,12 @@
-const requiredVersion = require("fs").readFileSync(".nvmrc", { encoding: "utf8" }).trim()
+const requiredVersion = require("fs")
+  .readFileSync(".nvmrc", { encoding: "utf8" })
+  .trim()
 
-if (process.version !== requiredVersion) {
-  console.error(`[!] This project requires Node.js ${requiredVersion}, current version is ${process.version}`)
+if (process.version.split(".")[0] !== requiredVersion.split(".")[0]) {
```
zephraph: Updates our version check script to reflect what's specified in the `engines` field of the package.json. Currently we're only requiring `8.x.x`. Loosening this restriction allows us to use the latest version of the docker image which includes `yarn`. That removes the need for us to install `nvm` and download `node` or `yarn`.

.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
+version: 2
+jobs:
+  relay:
+    working_directory: ~/artsy/reaction
+    docker:
+    - image: circleci/node:8
```
zephraph: The image field here is what specifies what version of node we'll be running the tests against. Currently just specified as the latest version of Node 8.

.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+    steps: 
+      - checkout
+      - restore_cache:
+          keys:
+          - yarn-deps-{{ checksum "yarn.lock" }}
+          - yarn-deps-
```
zephraph: The cache keys listed here are important. Circle treats cache as an immutable store. Our strategy is to default to a cache key that's tied to the contents of the `yarn.lock` file. That means so long as the `yarn.lock` doesn't update, we pull down the same cache. If the `yarn.lock` is updated but isn't yet cached, it'll fall back and do a partial match against `yarn-deps-`... meaning it grabs the last good version. That'll give us at least a partial cache hit.

.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+      - run: yarn --check-files
```
zephraph: Because we're caching `node_modules` and there's a potential that our dependencies won't be present or up-to-date, we call yarn with the `--check-files` flag. It does what it says... looks through the node_modules folder to see if anything is missing. If nothing is missing, it quits as per usual, else it installs the missing stuff.

.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+      - run: yarn test -w 4
```
zephraph: We were running jest tests inband (serially) before. The tests were pretty slow so I tried removing that restriction. Turns out, they eat a lot of memory and when too many run in parallel it starves out the CI. Restricting it to 4 seems to be a good balance of speed and conservative resource consumption.

.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+    - checkout
+    - restore_cache:
+        keys:
+        - yarn-deps-{{ checksum "yarn.lock" }}
+        - yarn-deps-
+    - run: yarn --check-files
+    - save_cache:
+        key: yarn-deps-{{ checksum "yarn.lock" }}
+        paths:
+        - ./node_modules
```
zephraph: This task is interesting... its sole purpose is to ensure that the cache gets populated. Because this job runs in parallel to all the other jobs we have a worse case performance scenario where all jobs end up installing their dependencies. This is the only job that stories them in CI though. The cache persists across multiple workflows and is immutable, so the worse case can ever only happen once. By the nature of CircleCI's cache fallback mechanism, we should *never* hit the worse case (unless they remove the cache out from under us). The other outlier scenario is when a dependency updates. That'll trigger a new cache write, but all the jobs will have to install those missing dependencies on that first run. All subsequent runs should be 🙆‍♂️ .

zephraph:
I copied over parts of this config to Palette and merged that in. The release happened as expected so we know the deploy step works as intended. 

https://github.com/artsy/palette/pull/43

jonallured:
Impressive work here, thanks for hanging in there @zephraph!! I know it can be a real slough to get through this stuff, but the end result is pretty damn sweet. I'm good to merge this if you're ready.

zephraph:
I'm good to go

.circleci/config.yml
starsirius: Nice! I learned the partial cache hit here! 👍

.circleci/config.yml
starsirius: On a separate note, `yarn test -w 4` seems to blindly append the argument `-w 4` to the actual command in npm script, which is currently `node verify-node-version.js && jest`. _(I was surprised for a few seconds by how smart yarn is to apply the argument to the right command!)_ I wonder how we can avoid future mistakes when `jest`is no longer the last command.

.circleci/config.yml
```diff
@@ -0,0 +1,92 @@
...
+      - <<: *save_cache
+workflows:
+  version: 2
+  build_and_verify:
+    jobs:
+      - update-cache
+      - relay
+      - lint
+      - type-check
+      - test
```
starsirius: It's worth noting again that as we move more projects to Circle Workflows, we might need to optimize the container use (or increase the number of containers) to avoid PRs blocking each other.

Open source projects got 3 more containers, so the parallel jobs here will take 2 containers from our org. I am not super worried here because each of the jobs will finish in seconds, but something to watch for more expensive jobs from other projects.

zephraph: If there are multiple long lived jobs, you definitely would want to gate on the shorter ones and have those fail before kicking off a longer process. 

According to the circle config we have 7 containers. It's not clear to me if that's per project or if it's for the entire organization.

starsirius: The 7 containers are shared among all projects in the organization, and you can observe the queues at https://circleci.com/gh/artsy.

Our current approach is that expensive builds like Volt and Gravity are configured to take 6 containers and that opens up at least 1 container for smaller builds.

starsirius: Well, looks like we actually have 8 containers (7 paid) because the first one is free. Open source projects get 3 more, so that's why it showed Reaction can have up to 11.

PR #970, opened 2018-07-11
@l2succes => Scroll analytics

eessex:
Adds impression and click events for `RelatedArticlesCanvas`, `RelatedArticlesPanel` and `ReadMoreButton`


artsyit:

Built with commit 2f271b090b06cd46ec48e5d9280b8533c43e0744

https://deploy-preview-970--artsy-reaction.netlify.com

src/Components/Publishing/ReadMore/ReadMoreButton.tsx
```diff
@@ -1,20 +1,34 @@
...
-export class ReadMore extends React.Component<any, any> {
-  constructor(props) {
-    super(props)
-    this.onClick = this.onClick.bind(this)
+interface ReadMoreProps {
+  onClick: () => void
+  tracking?: any
+}
+
+export class ReadMore extends React.Component<ReadMoreProps, any> {
```
l2succes: you can just do `React.Component<ReadMoreProps>` without including any for State

src/Components/Publishing/ReadMore/ReadMoreButton.tsx
```diff
@@ -1,20 +1,34 @@
...
+  onClick = () => {
+    const { onClick, tracking } = this.props
+
+    tracking.trackEvent({
+      action: "Clicked read more",
+    })
+    onClick()
   }
 
-  @track({ action: "Clicked read more" })
```
l2succes: Can you remind me what we the issues we were getting with using decorators aside from it using the context API and overriding any parent contexts?

src/Components/Publishing/RelatedArticles/RelatedArticlesPanel.tsx
```diff
@@ -1,52 +1,87 @@
...
-              <ArticleImage src={articleImageSrc} />
-              <ArticleTitle>{article.thumbnail_title}</ArticleTitle>
-            </ArticleLink>
-          )
-        })}
-      </Collection>
-    </RelatedArticlesContainer>
-  )
-}
+export class RelatedArticlesPanel extends React.Component<
```
l2succes: see above ^

src/Components/Publishing/ReadMore/ReadMoreButton.tsx
eessex: That was the issue I had problems with in the past, not sure if there are other bugs on different teams-- it will overwrite context, so anything outside of tracking is not shared with children - https://github.com/NYTimes/react-tracking/blob/master/src/withTrackingComponentDecorator.js#L37

l2succes: Got it, because if that's the only problem it may make more sense to fix it upstream in `react-tracking` in a fork and send them a PR instead of removing all decorators as we're using them extensively already.

eessex: Agreed that this should be fixed upstream, but also find the props method to be more clear and easier to implement, so leaning that way for any new/updated implementations.

PR #994, opened 2018-07-13
Add react-tracking analytics to Artist

alloy:
All the files of interest with regards to schema/tracking: https://github.com/artsy/reaction/tree/0d053387cbf84af6be5c792203fd602de9249b76/src/Analytics

Saves and follows events should remain exactly the same as they currently are, because these work through the mediator (e.g. save/follow _after_ authentication) and so we’re not touching those for now.

These are functional examples:

### Consign

```json
{
    "context_page": "Artist",
    "context_page_owner_id": "4d8b92b34eb68a1b2c0003f4",
    "context_page_owner_slug": "andy-warhol",
    "context_page_owner_type": "Artist",
    "context_module": "ArtistOverview",
    "action_type": "Click",
    "subject": "Learn more about consignment",
    "destination_path": "/artist/andy-warhol"
}
```

### Bio readmore

```json
{
    "context_page": "Artist",
    "context_page_owner_id": "4d8b92b34eb68a1b2c0003f4",
    "context_page_owner_slug": "andy-warhol",
    "context_page_owner_type": "Artist",
    "context_module": "ArtistBio",
    "action_type": "Click",
    "subject": "Read more"
}
```

### Tab

```json
{
    "context_page": "Artist",
    "context_page_owner_id": "4d8b92b34eb68a1b2c0003f4",
    "context_page_owner_slug": "andy-warhol",
    "context_page_owner_type": "Artist",
    "context_module": "NavigationTabs",
    "action_type": "Click",
    "subject": "Related artists",
    "destination_path": "/artist2/andy-warhol/related-artists"
}
```

### Carousel

```json
{
    "context_page": "Artist",
    "context_page_owner_id": "4d8b92b34eb68a1b2c0003f4",
    "context_page_owner_slug": "andy-warhol",
    "context_page_owner_type": "Artist",
    "context_module": "ArtistHeader",
    "action_type": "Click",
    "subject": "carouselSlide",
    "destination_path": "/show/the-whitworth-artist-rooms-andy-warhol"
}
```

### In auction

```json
{
    "context_page": "Artist",
    "context_page_owner_id": "4d8b92b34eb68a1b2c0003f4",
    "context_page_owner_slug": "andy-warhol",
    "context_page_owner_type": "Artist",
    "context_module": "ArtistOverview",
    "action_type": "Click",
    "subject": "In featured show",
    "destination_path": "/show/james-fuentes-downtown-art-ephemera-1970s-1990s-curated-by-marc-h-miller"
}
```


src/Styleguide/Components/ArtistBio.tsx
```diff
@@ -1,15 +1,25 @@
...
 interface Props {
   bio: ArtistBio_bio
 }
 
+@track({ context_module: "Artist bio" })
 export class ArtistBio extends React.Component<Props> {
+  @track({
+    action_type: Schema.ActionTypes.Click,
+    action_name: Schema.ActionNames.ReadMoreExpanded,
+  })
```
alloy: TODO: This should really move into the ReadMore component, or remain here if we need to keep using the artist specific `label`

src/Analytics/track.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+  if (action_type) properties.action = action_type
+  if (action_name) properties.label = action_name
+  Events.postEvent(properties)
+}
+
+/**
+ * Use this interface to augment the `track` function with props, state, or custom tracking-info schema.
+ *
+ * @example
```
mzikherman: Cool, I love all the examples here.

src/Analytics/track.ts
```diff
@@ -0,0 +1,215 @@
...
+ *      ```
+ */
+export const track: Track = _track
+
+/**
+ * A typed page view decorator for the top level component for your screen. This is the
+ * function you must use at the root of your component tree, otherwise your track calls
+ * will do nothing.
+ *
+ * For the majority of Emission code, this should only be used inside the AppRegistry,
```
mzikherman: 🍝 s/emission/reaction  :))

artsyit:

Built with commit 1e847332edcaf18aef2099786bdf1459770f1026

https://deploy-preview-994--artsy-reaction.netlify.com

src/Analytics/Schema/Values.ts
```diff
@@ -0,0 +1,112 @@
...
+  BidFlowPlaceBid = "placeBid",
+  BidFlowSaveBillingAddress = "saveBillingAddress",
+
+  /**
+   * Generic
+   */
+  ReadMoreExpanded = "readMoreExpanded", // TODO: This differs from old event
+  InSale = "In current auction", // TODO: Old schema
+  InShow = "In featured show", // TODO: Old schema
+}
```
l2succes: @eessex Once this PR is merged in, we can add article related events here as well

anipetrov:
looks good to me! cc @cavvia

alloy:
@mzikherman Alright, this should be it.

alloy:
@anipetrov Good point, saves and follows events should remain exactly the same as they currently are, because these work through the mediator (e.g. save/follow after authentication) and so we’re not touching those for now.

src/Router/Boot.tsx
```diff
@@ -1,48 +1,55 @@
...
+import Events from "Utils/Events"
 import { ResponsiveProvider } from "Utils/Responsive"
 import { ContextProvider } from "../Components/Artsy"
 import { AppState } from "./state"
 import { BootProps } from "./types"
 
-export const Boot: React.SFC<BootProps> = ({ children, ...props }) => {
-  const appState = new AppState(props)
+// TODO: Do we want to let Force explicitly inject the analytics code?
+@track(null, { dispatch: data => Events.postEvent(data) })
```
alloy: This is were all the track events end up bubbling-up to and get reported to the host app (Force) through `Events.postEvent()`.

alloy:
For a little more detail on react-tracking, event payloads are composed by merging all `@track(…)` ancestor calls with the one being sent. Take for example the `ReadMore` component, it itself [doesn’t hold any context](https://github.com/artsy/reaction/blob/0d053387cbf84af6be5c792203fd602de9249b76/src/Styleguide/Components/ReadMore.tsx#L38-L41) of where or on what page it is shown. However, when it triggers it merges with [the tracking info](https://github.com/artsy/reaction/blob/0d053387cbf84af6be5c792203fd602de9249b76/src/Styleguide/Components/ArtistBio.tsx#L14) on the `ArtistBio` component and then [the tracking info](https://github.com/artsy/reaction/blob/0d053387cbf84af6be5c792203fd602de9249b76/src/Apps/Artist/ArtistApp.tsx#L26-L31) of the page (‘app’ in our code-base). Resulting in the final payload shown in the examples above.

alloy:
I should note that this implementation ended up differing quite a bit from Emission’s, but that’s because there was more cleanup to be done to the analytics schema and during doing so @anipetrov and I realised what modelling was missing, which resulted in this PR. So in the future we’ll have to backpport this to Emission.

cc @ashfurrow @orta

mzikherman:
Here's the analytics logged out thru Force:

screen shot 2018-07-23 at 8 08 26 pm

Everything looks amazing!! and I'm definitely going to merge this, but note the `undefined` as that first param there. I believe (currently), its 'Click', and things like that, so `analytics.track('Click', {.....})`.

Not sure if that's ok to drop that first param now (since we provide a richer/consistent schema with that second object), just wanting to point out there _might_ be some tiny thing still off.

But, we can totally merge this and check it out in staging/prod 😎 


Thank you!!

src/Analytics/Schema/ContextModule.ts
```diff
@@ -0,0 +1,8 @@
+export interface ContextModule {
```
damassi: @alloy - interesting, so the idea is to think of `/Schema` as a collection of interfaces?

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -20,176 +22,220 @@ interface Props {
...
 
-export const ArtistHeader: SFC<Props> = props => {
-  return (
-    <Subscribe to={[AppState]}>
-      {({ state }) => {
-        const {
-          mediator,
-          system: { currentUser },
-        } = state
+type Image = Props["artist"]["carousel"]["images"][0]
```
damassi: Wasn't aware that access like this was possible 👍

src/Router/Boot.tsx
damassi: 👍

src/Analytics/Schema/ContextModule.ts
alloy: Yep! And what I ended up doing is defining the possible schema defs as a union (Trackables), meaning that as you add properties to the object TS narrows it down to a single schema def in the union (type narrowing).

src/Apps/Artist/Components/ArtistHeader.tsx
alloy: In the case of arrays it gets a little less nice, but you can still do it: `foo[0]`. I guess that in the case of the array being a tuple you can access the exact typing.

alloy:
@mzikherman Ah yes, good catch! I was handling that before the last schema change. Will send a fix.

cavvia:
Hey @alloy - Is this ready to QA on staging yet? Artist follows and artwork saves are still not logging analytics events on the new artist page there.

alloy:
@cavvia Not yet, plus we want the follows/saves to keep working the way they worked previously (as they go through the mediator and also need to work post-authentication), so this work lies in Force. I’m going to finish the Force side right now and will let you know once it’s in staging.

src/Apps/Artist/Components/ArtistHeader.tsx
orta: This is also new to me, nice 👍

PR #998, opened 2018-07-16
minor name change based on MP tix

oxaudo:
This should be merged in tandem with https://github.com/artsy/metaphysics/pull/1173


alloy:
Thanks!

PR #1003, opened 2018-07-16
[Router] Refactor API to allow host app to interact with state

alloy:
The host app can now inject initial state (e.g. filter state parsed from URL) and subscribe to state changes (e.g. update URL based on filter state changes).

Will comment inline.

----

Force needs to be updated before deploying this code.

https://github.com/artsy/force/blob/fbefcd1a13a0ed2c3e6f4eb376001579adf38a38/src/desktop/apps/artist2/client.js:

```tsx
buildClientApp({
  routes,
  user: sd.CURRENT_USER,
  initialAppState: { mediator },
  subscribe: {
    to: [FilterState],
    onChange(filter => console.log(filter.state)),
  },
}).then(({ ClientApp }) => {
    ReactDOM.hydrate(
      <Container>
        <ClientApp />{/* mediator prop moved up to initialAppState */}
      </Container>,
      document.getElementById('react-root')
    )
  })
```

### DIFF
--- a/src/Apps/__stories__/Apps.story.tsx
+++ b/src/Apps/__stories__/Apps.story.tsx
@@ -3,6 +3,7 @@ import { routes as artistRoutes } from "Apps/Artist/routes"
 import { routes as artworkRoutes } from "Apps/Artwork/routes"
 import React from "react"
 import { StorybooksRouter } from "Router/StorybooksRouter"
+import { FilterState } from "../Artist/Routes/Overview/state"
 
 storiesOf("Apps", module)
   .add("Artwork Page", () => {
@@ -14,11 +15,16 @@ storiesOf("Apps", module)
     )
   })
   .add("Artist Page", () => {
+    const filterState = new FilterState()
+    // TODO: This currently only changes the UI, the first fetch isn’t taking
+    //       this state into account.
+    filterState.state.for_sale = true
     return (
       <StorybooksRouter
         routes={artistRoutes}
         initialRoute="/artist2/andy-warhol"
         initialAppState={{ mediator: { trigger: x => x } }}
+        initialState={[filterState]}
       />
     )
   })
--- a/src/Apps/__stories__/Apps.story.tsx
+++ b/src/Apps/__stories__/Apps.story.tsx
@@ -25,6 +25,12 @@ storiesOf("Apps", module)
         initialRoute="/artist2/andy-warhol"
         initialAppState={{ mediator: { trigger: x => x } }}
         initialState={[filterState]}
-      />
+        subscribeTo={[FilterState]}
+      >
+        {x => {
+          // tslint:disable-next-line:no-console
+          console.log(x.state)
+        }}
+      </StorybooksRouter>
     )
   })
--- a/src/Apps/__stories__/Apps.story.tsx
+++ b/src/Apps/__stories__/Apps.story.tsx
@@ -25,12 +25,13 @@ storiesOf("Apps", module)
         initialRoute="/artist2/andy-warhol"
         initialAppState={{ mediator: { trigger: x => x } }}
         initialState={[filterState]}
-        subscribeTo={[FilterState]}
-      >
-        {x => {
-          // tslint:disable-next-line:no-console
-          console.log(x.state)
+        subscribe={{
+          to: [FilterState],
+          onChange: x => {
+            // tslint:disable-next-line:no-console
+            console.log(x.state)
+          },
         }}
-      </StorybooksRouter>
+      />
     )
   })
--- a/src/Apps/__stories__/Apps.story.tsx
+++ b/src/Apps/__stories__/Apps.story.tsx
@@ -18,7 +18,7 @@ storiesOf("Apps", module)
       <StorybooksRouter
         routes={artistRoutes}
         initialRoute="/artist2/andy-warhol"
-        initialState={{ mediator: { trigger: x => x } }}
+        initialAppState={{ mediator: { trigger: x => x } }}
       />
     )
   })
--- a/src/Router/Boot.tsx
+++ b/src/Router/Boot.tsx
@@ -3,40 +3,35 @@ import React from "react"
 import { GridThemeProvider } from "styled-bootstrap-grid"
 import { GlobalStyles } from "Styleguide/Elements/GlobalStyles"
 import { Grid } from "Styleguide/Elements/Grid"
-import { Provider as StateProvider } from "unstated"
+import { Container, Provider as StateProvider } from "unstated"
 import { ResponsiveProvider } from "Utils/Responsive"
+import { Breakpoint } from "Utils/Responsive"
 import { ContextProvider } from "../Components/Artsy"
-import { AppState } from "./state"
-import { BootProps } from "./types"
+import { AppStateContainer } from "./types"
 
-export const Boot: React.SFC<BootProps> = ({ children, ...props }) => {
-  const appState = new AppState(props)
+export interface BootProps extends AppStateContainer {
+  initialBreakpoint?: Breakpoint
+  initialState: Array<Container<any>>
+}
 
-  return (
-    <StateProvider inject={[appState]}>
-      <ContextProvider
-        relayEnvironment={props.relayEnvironment}
-        currentUser={props.currentUser}
+export const Boot: React.SFC<BootProps> = ({ children, ...props }) => (
+  <StateProvider inject={props.initialState}>
+    <ContextProvider
+      relayEnvironment={props.system.relayEnvironment}
+      currentUser={props.system.currentUser}
+    >
+      <ResponsiveProvider
+        initialBreakpoint={props.initialBreakpoint}
+        breakpoints={themeProps.mediaQueries}
       >
-        <ResponsiveProvider
-          initialBreakpoint={props.initialBreakpoint}
-          breakpoints={themeProps.mediaQueries}
-        >
-          <GlobalStyles>
-            <Theme>
-              <GridThemeProvider gridTheme={themeProps.grid}>
-                <Grid fluid>{children}</Grid>
-              </GridThemeProvider>
-            </Theme>
-          </GlobalStyles>
-        </ResponsiveProvider>
-      </ContextProvider>
-    </StateProvider>
-  )
-}
-
-Boot.defaultProps = {
-  initialBreakpoint: null,
-  relayEnvironment: null,
-  currentUser: null,
-}
+        <GlobalStyles>
+          <Theme>
+            <GridThemeProvider gridTheme={themeProps.grid}>
+              <Grid fluid>{children}</Grid>
+            </GridThemeProvider>
+          </Theme>
+        </GlobalStyles>
+      </ResponsiveProvider>
+    </ContextProvider>
+  </StateProvider>
+)
--- a/src/Router/StorybooksRouter.tsx
+++ b/src/Router/StorybooksRouter.tsx
@@ -1,10 +1,12 @@
 import React from "react"
 import { buildClientApp } from "Router"
+import { Container } from "unstated"
 
 interface Props {
   routes: Array<object>
   initialRoute?: string
-  initialState?: object
+  initialState?: Array<Container<any>>
+  initialAppState?: object
 }
 
 export class StorybooksRouter extends React.Component<Props> {
@@ -22,6 +24,8 @@ export class StorybooksRouter extends React.Component<Props> {
         routes: this.props.routes,
         historyProtocol: "memory",
         initialRoute: this.props.initialRoute,
+        initialAppState: this.props.initialAppState,
+        initialState: this.props.initialState,
       })
 
       this.setState({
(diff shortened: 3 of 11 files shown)


artsyit:

Built with commit f937162b289c9ff5910dcdc5bda5d3c120034be0

https://deploy-preview-1003--artsy-reaction.netlify.com

src/Apps/__stories__/Apps.story.tsx
```diff
@@ -14,11 +15,22 @@ storiesOf("Apps", module)
     )
   })
   .add("Artist Page", () => {
+    const filterState = new FilterState()
+    // TODO: This currently only changes the UI, the first fetch isn’t taking
+    //       this state into account.
+    filterState.state.for_sale = true
```
alloy: Setting up some initial state.

See comment about the initial state not yet adjusting fetch. Although I now realise that the initial state should pretty mirror what the server has already loaded, so this probably isn’t a real problem.

src/Apps/__stories__/Apps.story.tsx
```diff
@@ -14,11 +15,22 @@ storiesOf("Apps", module)
...
-        initialState={{ mediator: { trigger: x => x } }}
-      />
+        initialAppState={{ mediator: { trigger: x => x } }}
+        initialState={[filterState]}
+        subscribeTo={[FilterState]}
+      >
+        {x => {
+          // tslint:disable-next-line:no-console
+          console.log(x.state)
+        }}
```
alloy: Just like with the `Subscribe` API, pass a render prop as `children` to get state updates.

One thing to note is that it currently also invokes the callback on the first render, something which we probably don’t need or even want?

src/Router/Boot.tsx
```diff
@@ -3,40 +3,35 @@ import React from "react"
...
+import { AppStateContainer } from "./types"
 
-export const Boot: React.SFC<BootProps> = ({ children, ...props }) => {
-  const appState = new AppState(props)
-
-  return (
-    <StateProvider inject={[appState]}>
-      <ContextProvider
-        relayEnvironment={props.relayEnvironment}
-        currentUser={props.currentUser}
```
alloy: Neither of these props were being passed in, they were only inside the `props.system` object.

src/Router/StorybooksRouter.tsx
```diff
@@ -22,6 +25,8 @@ export class StorybooksRouter extends React.Component<Props> {
         routes: this.props.routes,
         historyProtocol: "memory",
         initialRoute: this.props.initialRoute,
+        initialAppState: this.props.initialAppState,
+        initialState: this.props.initialState,
```
alloy: Rather than passing some properties here and some to the returned ‘app’ component, pass them all here.

src/Router/StorybooksRouter.tsx
```diff
@@ -33,11 +38,15 @@ export class StorybooksRouter extends React.Component<Props> {
...
-    const { ClientApp } = this.state
+    const ClientApp = this.state && this.state.ClientApp
 
     return (
       <React.Fragment>
-        {ClientApp && <ClientApp {...this.props.initialState} />}
+        {ClientApp && (
+          <ClientApp subscribeTo={this.props.subscribeTo}>
+            {this.props.children}
+          </ClientApp>
```
alloy: Having said the above (about passing all props in one location), I guess these could move there too.

src/Router/buildClientApp.tsx
```diff
@@ -61,25 +66,38 @@ export function buildClientApp(config: AppConfig): Promise<ClientResolveProps> {
...
+
+      const ClientApp: App = props => {
         return (
-          <Boot {...bootProps} {...props}>
+          <Boot
+            system={system}
+            initialBreakpoint={initialBreakpoint}
+            initialState={[
+              new AppState({ ...initialAppState, system }),
+              ...initialState,
```
alloy: We need to move `system` and `mediator` out of the state and into the `Artsy` context, which is static. Once that’s done the `initialAppState` prop and `AppState` can be removed.

src/Router/buildServerApp.tsx
```diff
@@ -44,17 +53,28 @@ export function buildServerApp(config: AppConfig): Promise<ServerResolveProps> {
...
-            <AppShell
-              data={props.relayData}
-              loadableState={props.loadableState}
-            >
+          <Boot
+            system={system}
+            initialBreakpoint={initialBreakpoint}
+            initialState={[
+              new AppState({ ...initialAppState, system }),
+              ...initialState,
```
alloy: Same as above

src/Router/buildServerApp.tsx
```diff
@@ -44,17 +53,29 @@ export function buildServerApp(config: AppConfig): Promise<ServerResolveProps> {
...
+              config.subscribe.to &&
+              config.subscribe.onChange && (
+                <Subscribe to={config.subscribe.to}>
+                  {(...args) => {
+                    config.subscribe.onChange(...args)
+                    return null
+                  }}
+                </Subscribe>
+              )}
+            <AppShell data={props.data} loadableState={props.loadableState}>
```
alloy: The `data` was being passed as `props.relayData`, which doesn’t exist https://github.com/artsy/reaction/pull/1003/files#diff-078cfdc29c36b8e0e6e77f70fba6ae7bR89. This was leading to no data being hydrated for the client side and thus on the client side an additional identical fetch was being made.

src/Apps/__stories__/Apps.story.tsx
```diff
@@ -14,11 +15,23 @@ storiesOf("Apps", module)
...
         initialRoute="/artist2/andy-warhol"
-        initialState={{ mediator: { trigger: x => x } }}
+        initialAppState={{ mediator: { trigger: x => x } }}
+        initialState={[filterState]}
+        subscribe={{
+          to: [FilterState],
+          onChange: x => {
+            // tslint:disable-next-line:no-console
+            console.log(x.state)
+          },
```
alloy: This currently also triggers on the first render. Not sure if we want that, we’ll have to see if the user (host app) ends up ignoring it every time.

alloy:
@mzikherman I’d suggest you also rename the modules that use this code in Force to TS files.

damassi:
It feels like all of the ‘initial’ props being passed in (including subscribe) is muddying the api a bit by increasing the surface area unnecessarily. What was the reason for these changes? Being able to arbitrarily pass in props that are then “subscribable” in terms of unstated keeps things decoupled and independent of the framework. If the problem is rehydration I think we should fix that first within existing framework code and then maybe loop back and modify overall api only if necessary. Stuff like mediator and so on shouldnt need to exist in context; mediator as it was being passed is just a ref, and could be anything. By encoding dependencies we’re increasing complexity, and initialState / initialAppState feels confusing.

I’m flying back to the states on Thursday night and will be at a comp fri afternoon; if possible could we wait to merge this until I can look more closely?

alloy:
This change is only related to rehydration insofar that there were seemingly a few bugs that came to light during this work that would break rehydration.

The main change is what’s described in the PR post. Which is being able to set initial state for the app and/or register to changes. I ended up choosing to expose the unstated API, rather than very explicit with passing props down the tree, because passing them through the router wasn’t a trivial change either, nobody really likes passing props all the way down the tree, and stuff that isn’t state (mediator, user, relay env) was already being passed around the app using unstated, so following the same example to interact with actual state didn’t seem like a stretch.

I will maintain that anything that you would normally pass as props to a component and is essential to the workings of the app overall should not be mixed with a system that’s intended to manage state, but should very much be encoded into a stable API and have proper typings. There are now 2 places where the user and relay env etc are being stored and that’s not making things more clear. If the reason to add it to the state was because of the API being nicer (HOC vs render prop) then let’s just fix that instead of mixing.

alloy:
This change is only related to rehydration insofar that there were seemingly a few bugs that came to light during this work that would break rehydration.

The main change is what’s described in the PR post. Which is being able to set initial state for the app and/or register to changes. I ended up choosing to expose the unstated API, rather than very explicit with passing props down the tree, because passing them through the router wasn’t a trivial change either, nobody really likes passing props all the way down the tree, and stuff that isn’t state (mediator, user, relay env) was already being passed around the app using unstated, so following the same example to interact with actual state didn’t seem like a stretch.

I will maintain that anything that you would normally pass as props to a component and is essential to the workings of the app overall should not be mixed with a system that’s intended to manage state, but should very much be encoded into a stable API and have proper typings. There are now 2 places where the user and relay env etc are being stored and that’s not making things more clear. If the reason to add it to the state was because of the API being nicer (HOC vs render prop) then let’s just fix that and keep things clearly separated.

Regarding `initialAppState` it’s only there to deal with the current situation of the rest of the app’s code already expecting `AppState` to exist and contain the user and relay env etc. The idea is for it to be removed as soon as there has been time to cleanup and from then on there would only be `initialState` leaving setting that up entirely to the user.

To conclude, I will say that from my perspective the complexity has gone down by being a bit more explicit and the API was more muddy before by passing through props in a couple of places, which was making it hard to track either by prop name or by types.

alloy:
(As a short clarification to what, in my mind, constitutes state vs props in this case: state is stuff that changes, props are constant and shouldn’t have an API that allows it to change because that could allow one component to break others.)

alloy:
The props that need to be passed are state parsed from e.g. URL parameters and getting a callback when  state changes to update the URL. These are host app concerns, not Reaction concerns, so the host should be able to communicate those in some way.

alloy:
I will defer to @mzikherman for merging now vs holding off, but I will note that this PR largely just:

* improves typings
* is internally more explicit and easier to follow
* fixes a few bugs

The external API increasing is a very tiny part of the PR and very limited in scope; and no real changes have been made around separating the constants from state other than adding the `initialAppState` property.

alloy:
Btw, good point about `initialAppState` vs `initialState` being confusing. I think that that property could/should be renamed to something that reflects what it holds, which is the plumbing that the app needs to work (e.g. your choice for something like ‘system’ is a better one). How/if we then re-architect the internals of Reaction to deal with those system essential ones vs state won’t further change the API in the future.

src/Router/types.ts
```diff
@@ -1,53 +1,53 @@
...
   redirect?: string
   status?: string
 }
 
-export interface AppShellProps {
-  loadableState?: {
-    getScriptTag: () => string
-  }
-  data?: Array<object>
-}
```
alloy: Colocating typings with the one place they’re being used increases readability and refactor-ability of the codebase. A ‘types’ module is I think mostly helpful when e.g. typing external data that comes into the system and various parts of the system need to use those typings.

alloy:
@damassi After thinking about your thoughts some more, could it be that the tension for what the ‘Router’ code is supposed to do comes from the `Boot` component being included in there? The `Boot` component doesn’t really seem to do anything regarding routing, but to get data into into the app it has to pass through the ‘Router’ code right now. i.e., what if configuring routing and setting up app props/state would be separated, something like:

```tsx
/* This really only wires up the right routing with found and fetches data through found-relay. Same goes for `buildServerApp` */
buildClientApp({ routes, url }).then(({ ClientApp }) => {
  ReactDOM.hydrate(
    <Boot /* Passes on most details to the `Artsy` context provider and otherwise sets up the theme/responsive stuff */
      currentUser={sd.CURRENT_USER}
      mediator={mediator}
      subscribe={{
        to: [FilterState],
        onChange(filter => console.log(filter.state)),
      }}
    >
      <ClientApp />
    </Boot>,
    document.getElementById('react-root')
  )
})
```

damassi:
Since I don’t have access to a full computer it’s hard to review properly but I think the tension is more around solving an issue framework side vs sub-app side (I’m envisioning this code as slightly more than just a router now that boot is in and all subsequent app sections will utilize it). Its also around general approach; the problems being solved don’t require modifications on the framework side and can, in current form, be handled without modification to our API or exposing underlying libraries.

Again, it’s hard to review properly from a phone and your additions could be (and are likely) right on, but I would very much appreciate a hold on this until I get back and can look over closely. The nice thing about our router / framework in its current form is there’s nothing to it, with very few wires necessary to accomplish quite a lot. Since this will be used going forward for future sections I think a solid code review is warranted just to be sure there are no misunderstandings.

alloy:
@damassi FYI We’re holding off and working around things for now. Enjoy the remainder of your vacation!

damassi:
@alloy - I've had a chance to look at this closely and I think this will work, though I think it can be simplified in the sense that 

- We no longer need `initialState` or remap exposure of unstated via the `subscribe` config prop since we're handling this via the router directly and it seems to add some unnecessary surface area.
- I like how we're moving `initialAppState` values into the config; that makes more sense. The original implementation was written somewhat blindly late at night before my trip and I'm unsure why I was passing it in as a prop.
- I don't think we should move any additional values into Artsy Context. Being able to pass values into `initialAppState` and then `<Subscribe to=[AppState]>` from wherever in the tree is super flexible and doesn't lock us into concrete values that must be maintained over time across app boundaries. Being able to quickly pluck `system` out of there and see whats going on under the hood quickly is also nice. 
- Regarding AppState as a storehouse of static values, it's common in larger app frameworks like Redux et al to store data in state so that its accessible globally. IMO "state" as a language construct doesn't _require_ changes over time. Unstated is designed to mirror the new Context API, but with additional functionality to **update** the "context" should one need to. It's like React Context + 😄, as described in the unstated docs. 

Thoughts on my above points? Again, so much of this was worked on under heavy time constraints and so there's certainly room for improvement / change before we start building out further sections. This PR indeed doesn't change the API that much. 

Thanks for looking into this and handling cleanup and sorry I couldn't give it a more thorough look when I was OOO. Let me know if you'd like any help with the rebase / refactor / anything I can do otherwise to get this PR merged and over the line. Since we only have one other app (Checkout Flow) currently in development I think things should be fairly easy to update.

PR #1024, opened 2018-07-19
@xtina-starr => Refactor Article Header

eessex:
Related to Article/Display margins project, also addresses long-standing tech debts:
- Stop using children for editing in headers, now uses props (Last components to be converted)
- Breaks out feature header into components by type (previously feature covered 4 layouts and was tricky to understand)
- Headers no longer care about `props.isMobile`
- Feature header margins are determined by article data `is_super_article` and `seriesArticle` rather than margin-specific props passed from Force
- Move classic byline info into `Byline` directory
- Fixes [GROW-679](https://artsyproduct.atlassian.net/browse/GROW-679)

This is currently not yet implemented in articles and lives in `/Header2` to make it easier to review.  Changes are fairly consolidated by commit for review as well.
Check out the new components in netlify storybooks under `/Header2`
Will follow up after this is merged with PR to remove margin props in articles and swap headers.


xtina-starr:
Went over this IRL with @eessex and this looks good but seeing that it's a pretty significant change I'll hold off from merging so that @l2succes can take a look as well.

src/Components/Publishing/Header2/Header.tsx
```diff
@@ -0,0 +1,64 @@
+import React from "react"
+import { ArticleData } from "../Typings"
+import { ClassicHeader } from "./Layouts/ClassicHeader"
+import { FeatureHeader } from "./Layouts/FeatureHeader"
+import { StandardHeader } from "./Layouts/StandardHeader"
+
+interface HeaderProps {
+  article: ArticleData
+  date?: string
+  editDeck?: any
```
l2succes: Update types

src/Components/Publishing/Header2/Layouts/Components/FeatureBasicHeader.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+interface State {
+  isPlaying: boolean
+}
+
+export class FeatureBasicHeader extends React.Component<
+  FeatureHeaderProps & { tracking?: any },
+  State
+> {
+  trackVideoPlay = () => {
```
l2succes: Based on our discussion about analytics and types. We should at least use the Tracking schema here.
https://github.com/artsy/reaction/pull/994/files#diff-b7e2851428e5c7793760ae30da591015

src/Components/Publishing/Header2/Layouts/Components/FeatureFullscreenHeader.tsx
```diff
@@ -0,0 +1,115 @@
...
+    article: { hero_section, is_super_article, seriesArticle, super_article },
+    editImage,
+  } = props
+  const url = (hero_section && hero_section.url) || ""
+  const isVideo = url.includes("mp4")
+  const src = !isVideo && url.length && resize(url, { width: 1600 })
+
+  return (
+    <FeatureHeaderContainer hasNav={seriesArticle && true}>
+      <FeatureAssetContainer src={src ? src : undefined}>
```
l2succes: why not just `src={src}`?

src/Components/Publishing/Header2/Layouts/FeatureHeader.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react"
+import styled from "styled-components"
+import FeatureBasicHeader from "./Components/FeatureBasicHeader"
+import { FeatureFullscreenHeader } from "./Components/FeatureFullscreenHeader"
+import { FeatureSplitHeader } from "./Components/FeatureSplitHeader"
+import { FeatureTextHeader } from "./Components/FeatureTextHeader"
+
+export interface FeatureHeaderProps {
+  article?: any
```
l2succes: Types

src/Components/Publishing/Header2/Layouts/Components/FeatureFullscreenHeader.tsx
eessex: This prop determines the background image, so undefined prevents a background image from being applied, otherwise it gets `background-image: url('')`

PR #1031, opened 2018-07-23
Scrollable tabs, body doesn't scroll

zephraph:
Fixes [DISCO-104](https://artsyproduct.atlassian.net/browse/DISCO-104).

Makes entire page not scrollable. Does this in a few ways. 

1. Makes the grid have `overflow: hidden`. It'll never be scrollable because of that (for better or worse)
2. Updates the tabs to be scrollable on xs. This is the main thing that was forcing the site to be too wide on xs.


artsyit:

Built with commit 74e2807ff9897ef039960397b5fcb6a66f49c484

https://deploy-preview-1031--artsy-reaction.netlify.com

PR #1033, opened 2018-07-23
@l2succes => Article can render classic layouts

eessex:
Experimental -- adds 'ClassicLayout' component, and adds ability to render 'classic' layout articles from top level article component. 

This is not yet used in production, but will allow us to view classic articles in storybooks.

This also gives us ability to spike on killing Backbone article pages, or using stitch to render this component rather than backbone views where `/mobile/components/article` is still used.


artsyit:

Built with commit e5df7546ffd6364ee517d9d0710e73cf7523986f

https://deploy-preview-1033--artsy-reaction.netlify.com

PR #1036, opened 2018-07-24
[test] Remove tsc from pipeline and set precedent for testing analytics

alloy:
cc @eessex @l2succes @mzikherman @damassi @zephraph @yuki24 @xtina-starr


artsyit:

Built with commit 548364fd685f7c699dc953378c1bd55a38de7323

https://deploy-preview-1036--artsy-reaction.netlify.com

src/Analytics/track.ts
```diff
@@ -107,4 +107,80 @@ export interface Track<PP = {}, SS = null, AA extends Array<any> = Array<any>> {
...
+export function mockTracking<P>(
+  Subject: React.ComponentType<P>
+): {
+  Component: React.ComponentType<P>
+  dispatch: jest.Mock<(trackedData: Trackables) => void>
+} {
+  const dispatch = jest.fn() as jest.Mock<() => void>
+  const Component = _track({}, { dispatch })(Subject)
+  return { Component, dispatch }
+}
```
orta: This is nice, did you consider the `__mocks` file-system approach? I think I did that in Emission

orta: Or not, must have debated doing that a some point.

alloy: Hmm, I’m not sure how you’d imagine that to work? In this case we’re not actually mocking and decorating the component being tested with the right context for `react-tracking` to work and capture the tracked data.

src/Components/Publishing/Header/__tests__/__snapshots__/ClassicHeader.test.tsx.snap
```diff
@@ -382,7 +382,6 @@ exports[`Classic Header renders classic header with children properly 1`] = `
     >
       <div
         className="c3 c4"
-        color={undefined}
```
alloy: Not sure why, but looks like with tsc we were capturing all the `undefined` props as well 🤷‍♂️

alloy:
Alright, going for the merge 👍

src/Analytics/track.ts
orta: Mentally I model the fs `__mocks` idea to be an "export API" compatible version of the mocked module that also contains introspection tools / doesn't run work outside the boundaries of that module. Then `track` would not send out to force, etc. 

This technique is different from the HOC wrapper this function actually provides though. 

IMO your technique ends up with less type faff in the test code though, because it's shipping as a part of the normal export. Example: [module](https://github.com/danger/peril/blob/master/source/db/getDB.ts) - [mock](https://github.com/danger/peril/blob/master/source/db/__mocks__/getDB.ts) -  [test faff](/Users/orta/dev/projects/danger/hazmat/peril/source/scheduler/_tests/_startScheduler.test.ts)

alloy: I like that the types are immediately settled, but the main thing that needs done is setup that context and I don’t easily see a way to do that with a `__mocks__` mock 🤔

alloy:
@orta Is that Peril error supposed to show up? ^

orta:
the fail is legit, conceptually.

Peril ran on the pr closed web hook, by which point our auto-branch deleter had removed the branch - so peril can't read the dangerfile on that deleted branch

orta:
Ideally, I figure a way to make peril not fail in that either by saying that some dangerfile's can be nullable, or have a fallback to master if the PR has been merged somehow (or have it work from the commit instead)

alloy:
Ahhh, I see. Thanks for the info 👍

PR #1043, opened 2018-07-25
Shipping layout

zephraph:
This goes a few steps further in laying out the shopping experience. 

- Tweaks the layout spacings to be in line with designs.
- Adds placeholders for each shipping field

### DIFF
--- a/src/Apps/Order/Components/TwoColumnLayout.tsx
+++ b/src/Apps/Order/Components/TwoColumnLayout.tsx
@@ -1,6 +1,36 @@
 import React from "react"
+import { Box } from "Styleguide/Elements/Box"
+import { Flex } from "Styleguide/Elements/Flex"
 import { Col, Row } from "Styleguide/Elements/Grid"
+import { Spacer } from "Styleguide/Elements/Spacer"
 import { Responsive } from "Utils/Responsive"
+
+const CONTENT_SPAN = 7
+const SIDEBAR_SPAN = 4
+const ROW_SPACE = 2
+const COL_SPACE = 2
+
+const columnRatioWidth = (spans, size) => Math.round((spans / size) * 100) + "%"
+
+export const TwoColumnSplit = ({ children, ...props }) => {
+  let [firstColumn, secondColumn] = React.Children.toArray(children)
+
+  return (
+    <Responsive>
+      {({ xs }) => (
+        <Flex flexDirection={xs ? "column" : "row"} {...props}>
+          <Box width={xs ? "100%" : columnRatioWidth(4, CONTENT_SPAN)}>
+            {firstColumn}
+          </Box>
+          <Spacer mr={xs ? null : COL_SPACE} mb={xs ? ROW_SPACE : null} />
+          <Box width={xs ? "100%" : columnRatioWidth(3, CONTENT_SPAN)}>
+            {secondColumn}
+          </Box>
+        </Flex>
+      )}
+    </Responsive>
+  )
+}
 
 export const TwoColumnLayout = props => (
   <Responsive>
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit e209ba8c14595dc5195aba1d70dd04fdef644c3d

https://deploy-preview-1043--artsy-reaction.netlify.com

src/Apps/Order/Components/TwoColumnLayout.tsx
```diff
@@ -1,7 +1,37 @@
...
+      {({ xs }) => (
+        <Flex flexDirection={xs ? "column" : "row"} {...props}>
+          <Box width={xs ? "100%" : columnRatioWidth(4, CONTENT_SPAN)}>
+            {firstColumn}
+          </Box>
+          <Spacer mr={xs ? null : COL_SPACE} mb={xs ? ROW_SPACE : null} />
+          <Box width={xs ? "100%" : columnRatioWidth(3, CONTENT_SPAN)}>
+            {secondColumn}
+          </Box>
+        </Flex>
```
ashfurrow: I almost wonder if, instead of the ternary checks for `xs`, we just had an `if`/`else` statement? It might clarify this and make it easier to read. Something like:

```jsx
if (xs) {
  return (
    <Flex flexDirection={"column"} {...props}>
      <Box width={"100%"}>{firstColumn}</Box>
      ...
    </Flex>
  )
} else {
  return (
    <Flex flexDirection={"row"} {...props}>
      <Box width={columnRationWidth(4, CONTENT_SPAN)}>{firstColumn}</Box>
      ...
    </Flex>
  )
}
```

Not sure what the conventions here are.

zephraph: Yeah, that's a good point. I'm switching on enough logic here that it probably makes the most sense to split it out.

ds300: I'm normally a fan of turning lots of inline ternaries into one big ternary/if statement like this. Although an interesting approach for this responsive component in particular might be an api like this:

```jsx
<Responsive
  xs={{ columnWidth: "100%", flexDirection: "column" ... }}
  md={{ columnWidth: columnRatioWidth(4, CONTENT_SPAN), flexDirection: "row" ...}}
/>
  {({ columnWidth, flexDirection ... }) => (
    <Flex flexDirection={flexDirection} {...props}>
      <Box width={columnWidth}>{firstColumn}</Box>
      ...
    </Flex>
  )}
</Responsive>
```

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -1,8 +1,13 @@
 import { Shipping_order } from "__generated__/Shipping_order.graphql"
 import React, { Component } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { Join } from "Styleguide/Elements/Join"
```
damassi: @zephraph - what does `Join` do?

ds300: It inserts a separator between all of its direct children. In FP circles this might be called 'interpose'. I suppose in JS land the closest thing is Array.join, which does optionally insert a separator, but also concatenates the result into a string at the end.

damassi: Ahh, makes sense 👍

src/Apps/Order/Components/TwoColumnLayout.tsx
zephraph: I'm going to tackle this in a separate PR. It's brilliant though. I love it, heh.

src/Apps/Order/Routes/Shipping/index.tsx
zephraph: Yep. It wasn't one of my more creative naming days.

PR #1044, opened 2018-07-25
Orders Helper component

ashfurrow:
This PR creates a new `Summary` component that will server as the whole sidebar for orders (containing item summary, price summary, and a helper). The helper is also added but not yet functional.

This is a WIP because it needs the following:

- [x] Confirm this component organization makes sense.
- [x] Confirm the use of styled-components.
- [x] Hook up the links to modals (these don't appear to live in Reaction, might present a problem...)
- [x] Tests (?)


artsyit:

Built with commit 171c5b4751912e4f9971912d0cfa17ef52df7f79

https://deploy-preview-1044--artsy-reaction.netlify.com

src/Apps/Order/Components/Summary.tsx
```diff
@@ -0,0 +1,37 @@
+import Colors from "Assets/Colors"
+import { unica } from "Assets/Fonts"
```
zephraph: So instead of these you'll want to use the components from Palette.... which I need to write some good documentation for, ha. 

```typescript
import { Sans } from "@artsy/palette"
```
So the `Sans` component maps to the values in our [typography design specs](https://www.notion.so/artsy/Typography-d1f9f6731f3d47c78003d6d016c30221). 

Instead of building the `Text` component use...

```jsx
<Sans size="2" color="black60">
...
</Sans>
```

If you're using Zeplin, I wrote an [extension](https://github.com/artsy/palette-zeplin-extension) to help pick out the right typography component.

zephraph:
Awesomeeee! LGTM! 🎉

ashfurrow:
Thanks! I opened the following ticket to capture the work we need to do in Force: 
https://artsyproduct.atlassian.net/browse/PURCHASE-316

PR #1056, opened 2018-07-27
Add font smoothing to global styles

zephraph:
@ds300 found this. 

Fonts in storybook looked thicker than the fonts in Zeplin. Turns out that force is enabling some extra CSS properties that makes fonts look a bit thinner on OSX. 

We added these to global styles so that it'll carry forward on pages without base styling from force.


artsyit:

Built with commit 24693eff70c500e841eb900b1d63c48ab68adb21

https://deploy-preview-1056--artsy-reaction.netlify.com

PR #1060, opened 2018-07-30
Add Message component

ds300:
[jira](https://artsyproduct.atlassian.net/browse/PURCHASE-310)

image

This was originally called 'Messenger' in the designs, but @zephraph @katarinabatina and I discussed renaming it to sound less like an interactive chat thing. I'm proposing 'Message', but definitely open to other suggestions.

### DIFF
--- a/src/Styleguide/Elements/Message.tsx
+++ b/src/Styleguide/Elements/Message.tsx
@@ -5,6 +5,7 @@ import { Flex, FlexProps } from "Styleguide/Elements/Flex"
 
 export const StyledFlex = Flex.extend`
   background-color: ${color("black5")};
+  border-radius: 2px;
 `
 
 interface MessageProps extends FlexProps {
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit dd198071d67c021248fbf58f739c9c059ab1ac38

https://deploy-preview-1060--artsy-reaction.netlify.com

src/Styleguide/Elements/Message.tsx
```diff
@@ -0,0 +1,22 @@
+import { color } from "@artsy/palette"
+import { Sans } from "@artsy/palette"
+import React from "react"
+import { Flex, FlexProps } from "Styleguide/Elements/Flex"
+
+export const StyledFlex = Flex.extend`
+  background-color: ${color("black5")};
```
zephraph: this should also have `border-radius: 2px`

ds300: Good catch! 2px is so subtle it didn't even register with me. I still struggle to see it now unless I zoom in :D

PR #1063, opened 2018-07-30
@mzikherman => [ArtworkGrid] Add support for dynamic image resizing

damassi:
Updates the `GridItem` component to support dynamic image sizing via Gemini. 

**Before** (arbitrary browser size): 
screen shot 2018-07-30 at 2 37 42 pm

**After**:
screen shot 2018-07-30 at 2 36 58 pm

### DIFF
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -1,5 +1,6 @@
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { data as sd } from "sharify"
 import styled from "styled-components"
 import { Responsive } from "Utils/Responsive"
 import colors from "../../Assets/Colors"
@@ -28,9 +29,53 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
   }
 }
 
-class ArtworkGridItemContainer extends React.Component<Props, null> {
+interface State {
+  width: number
+  height: number
+}
+
+const IMAGE_QUALITY = 80
+
+class ArtworkGridItemContainer extends React.Component<Props, State> {
   static defaultProps = {
     useRelay: true,
+  }
+
+  private image: HTMLImageElement = null
+
+  state = {
+    width: 0,
+    height: 0,
+  }
+
+  componentDidMount() {
+    const scale = window.devicePixelRatio
+    const width = this.image.width * scale
+    const height =
+      (this.image.width / this.props.artwork.image.aspect_ratio) * scale
+
+    this.setState({
+      width,
+      height,
+    })
+  }
+
+  get imageURL() {
+    const imageURL = this.props.artwork.image.url
+    if (imageURL) {
+      // Either scale or crop, based on if an aspect ratio is available.
+      const type = this.props.artwork.image.aspect_ratio ? "fit" : "fill"
+      const width = String(this.state.width)
+      const height = String(this.state.height)
+      // tslint:disable-next-line:max-line-length
+      return `${
+        sd.GEMINI_CLOUDFRONT_URL
+      }/?resize_to=${type}&width=${width}&height=${height}&quality=${IMAGE_QUALITY}&src=${encodeURIComponent(
+        imageURL
+      )}`
+    } else {
+      return null
+    }
   }
 
   render() {
@@ -46,7 +91,7 @@ class ArtworkGridItemContainer extends React.Component<Props, null> {
       <div className={className} style={style}>
         <Placeholder style={{ paddingBottom: artwork.image.placeholder }}>
           <a href={artwork.href}>
-            <Image src={artwork.image.url} />
+            <Image src={this.imageURL} innerRef={img => (this.image = img)} />
           </a>
           <Responsive>
             {({ hover }) =>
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit 828bc1a9749467f9cb37634ff6ec55dace06dc23

https://deploy-preview-1063--artsy-reaction.netlify.com

src/Components/Artwork/GridItem.tsx
```diff
@@ -28,11 +28,52 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
   }
 }
 
-class ArtworkGridItemContainer extends React.Component<Props, null> {
+interface State {
+  width: number
+  height: number
+}
+
+const GEMINI_HOST = "d7hftxdivxxvm.cloudfront.net"
```
mzikherman: We should probably be reading this from ENV, maybe that can be a follow up `FIXME`?

src/Components/Artwork/GridItem.tsx
```diff
@@ -28,11 +28,52 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
...
+const IMAGE_QUALITY = 80
+
+class ArtworkGridItemContainer extends React.Component<Props, State> {
   static defaultProps = {
     useRelay: true,
   }
 
+  image: HTMLImageElement = null
```
mzikherman: I'm a bit unsure exactly what this is, is this just a private instance variable?

damassi: Yup just an instance variable, though not private; will prefix.

src/Components/Artwork/GridItem.tsx
```diff
@@ -28,11 +28,52 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
...
+      width,
+      height,
+    })
+  }
+
+  get imageURL() {
+    const imageURL = this.props.artwork.image.url
+    if (imageURL) {
+      // Either scale or crop, based on if an aspect ratio is available.
+      const type = this.props.artwork.image.aspect_ratio ? "width" : "fill"
```
mzikherman: This line in Emission https://github.com/artsy/emission/blob/27eb8294373be344e4fc2b28675cece224efda93/src/lib/Components/OpaqueImageView.tsx#L85 uses `fit` instead of `width`. Intentional?

damassi: Yup, width scales the image up a bit to account for vertically oriented images. At `fit` the quality is quite a bit worse with no real gain.

mzikherman:
Looks great, merge when ready!

src/Components/Artwork/GridItem.tsx
```diff
@@ -28,11 +28,52 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
...
+    height: 0,
+  }
+
+  componentDidMount() {
+    const { width } = this.image
+    const height = this.image.width * this.props.artwork.image.aspect_ratio
+
+    this.setState({
+      width,
+      height,
```
alloy: Multiply them both by `window.devicePixelRatio`.

src/Components/Artwork/GridItem.tsx
damassi: @alloy - the best way to expand is to compare two images: 

**Fill**: 
screen shot 2018-07-30 at 4 14 15 pm

**Width**: 
screen shot 2018-07-30 at 3 42 59 pm

Width returns the "correct" resolution out of all of the presets [listed here](https://github.com/artsy/gemini/blob/7c483f05007c725ac71013a28e47000ce17b8154/app/util/image_resizing.rb#L9)

alloy: Hrmm. @mzikherman do you know what’s happening here?

src/Components/Artwork/GridItem.tsx
damassi: This is small, but it adds up: 

screen shot 2018-07-30 at 4 47 34 pm

vs

screen shot 2018-07-30 at 2 36 58 pm

Given the overall small dimensions of the grid items, is it worth it to double the image size for a slight increase in resolution? At a larger size (for example, the center image on `/artwork/id`) it's a bit more noticable.

damassi:
@mzikherman - feedback has been addressed, though @alloy has a comment about the image resize setting. If its good to go feel free to merge and I'll send out a deploy, otherwise let me know and I can play with the setting some more.

orta:
Nice work!

src/Components/Artwork/GridItem.tsx
```diff
@@ -28,11 +29,55 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
...
+  state = {
+    width: 0,
+    height: 0,
+  }
+
+  componentDidMount() {
+    const scale = window.devicePixelRatio
+    const width = this.image.width * scale
+    const height =
+      this.image.width * this.props.artwork.image.aspect_ratio * scale
```
mzikherman: This is what's causing the 'fit' resizing to not work properly, (although 'width' resizing will work with these).

Since aspect ratio is generally defined as ratio of width to height (aka Width divided by Height), if you have the width and want to derive the height, you divide by the aspect ratio rather than multiply.

So, `height = width / aspect`. Then you can use 'fit' rather than 'width' in the resizing.

mzikherman:
Looks great! 😎  One point about the aspect ratio and fit vs width thing, but should be good to go after thats addressed!

src/Components/Artwork/GridItem.tsx
mzikherman: Ah, addressed in https://github.com/artsy/reaction/pull/1063#discussion_r206525766

PR #1064, opened 2018-07-30
small fix for the radio group field

sweir27:
This was causing an issue when rendering the app in force (I'm not exactly sure why it works here and not there, but since we don't pass a `name` into the radio button from the radio group, we were getting a "`name not defined`" error.


artsyit:

Built with commit 9bcad059781344116e191ff29c74b45721ac3846

https://deploy-preview-1064--artsy-reaction.netlify.com

PR #1066, opened 2018-07-30
step summary item

ds300:
[jira](https://artsyproduct.atlassian.net/browse/PURCHASE-307)

image

### DIFF
--- a/src/Styleguide/Components/StepSummaryItem.tsx
+++ b/src/Styleguide/Components/StepSummaryItem.tsx
@@ -28,7 +28,7 @@ export function StepSummaryItem({
   ...others
 }: StepSummaryItemProps) {
   return (
-    <Wrapper flexDirection="column" {...others}>
+    <Wrapper flexDirection="column" width="100%" maxWidth={542} {...others}>
       <Flex justifyContent="space-between" alignItems="baseline" mb={1}>
         <Serif size="3t" weight="semibold" color="black100">
           {title}
--- a/src/Styleguide/Components/StepSummaryItem.tsx
+++ b/src/Styleguide/Components/StepSummaryItem.tsx
@@ -1,9 +1,9 @@
 import { Sans, Serif, space } from "@artsy/palette"
 import React from "react"
+import { media } from "styled-bootstrap-grid"
 import styled from "styled-components"
 import { BorderBox } from "Styleguide/Elements/Box"
 import { Flex, FlexProps } from "Styleguide/Elements/Flex"
-import { media } from "Styleguide/Elements/Grid"
 
 interface StepSummaryItemProps extends FlexProps {
   title: string
@@ -27,7 +27,7 @@ export const StepSummaryItem: React.SFC<StepSummaryItemProps> = ({
   ...others
 }) => {
   return (
-    <Wrapper flexDirection="column" width="100%" maxWidth={542} {...others}>
+    <Wrapper flexDirection="column" {...others}>
       <Flex justifyContent="space-between" alignItems="baseline" mb={1}>
         <Serif size="3t" weight="semibold" color="black100">
           {title}
--- a/src/Styleguide/Components/StepSummaryItem.tsx
+++ b/src/Styleguide/Components/StepSummaryItem.tsx
@@ -1,0 +1,45 @@
+import { Sans, Serif, space } from "@artsy/palette"
+import React from "react"
+import styled from "styled-components"
+import { BorderBox } from "Styleguide/Elements/Box"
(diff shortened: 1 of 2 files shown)


src/Styleguide/Components/StepSummaryItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+`
+
+export function StepSummaryItem({
+  title,
+  onChange,
+  children,
+  ...others
+}: StepSummaryItemProps) {
+  return (
+    <Wrapper flexDirection="column" width="100%" maxWidth={542} {...others}>
```
ds300: Should this maxWidth come from somewhere else? It's not really a breakpoint but it feels like it should be part of the design system somehow.

artsyit:

Built with commit ab322d660b29c1c517ddb02b3850430b1108aa9a

https://deploy-preview-1066--artsy-reaction.netlify.com

src/Styleguide/Components/StepSummaryItem.tsx
zephraph: Where'd this value come from?

src/Styleguide/Components/__stories__/StepSummaryItem.story.tsx
```diff
@@ -0,0 +1,50 @@
...
+            within 2 business days to handle pickup logistics.
+          </Sans>
+        </StepSummaryItem>
+      </Section>
+      <Section title="Multiple StepSummaryItems">
+        <StepSummaryItem
+          title="Pick up (New York, NY)"
+          onChange={() => console.warn("change")}
+        >
+          <Sans size="2" color="black60" maxWidth={391}>
```
zephraph: This one too?

src/Styleguide/Components/StepSummaryItem.tsx
ds300: Zeplin

src/Styleguide/Components/__stories__/StepSummaryItem.story.tsx
ds300: Also Zeplin

src/Styleguide/Components/StepSummaryItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+import React from "react"
+import styled from "styled-components"
+import { BorderBox } from "Styleguide/Elements/Box"
+import { Flex, FlexProps } from "Styleguide/Elements/Flex"
+import { media } from "Styleguide/Elements/Grid"
+
+interface StepSummaryItemProps extends FlexProps {
+  title: string
+  onChange: () => void
+  children: React.ReactNode | null
```
alloy: Just out of curiosity, this only removes the optional typing, right? Why is that needed/helpful?

src/Styleguide/Components/StepSummaryItem.tsx
zephraph: Ahhh. Be careful with that. Zeplin can be, and often is, wrong. It’s very likely that instead this is intended to align with the grid in some way.

src/Styleguide/Components/StepSummaryItem.tsx
ds300: `children` isn't in FlexProps, and `StepSummaryItem` is a SFC so doesn't get `children` automatically. If I delete this:

no children

alloy: Ah I see. The issue is that the SFC is not typed accordingly. If you were to do this then I think it should work:

```tsx
export const StepSummaryItem: React.SFC<Props> = (...) => {...}
```

ds300: Done :+1: This is a neat pattern.

src/Styleguide/Components/StepSummaryItem.tsx
ds300: Ok for now I've moved the width constraints out of this component, such that the dimensions are controlled by the consumer. I left the widths hard-coded in the story file. They will presumably be constrained by the grid system when in actual use.

src/Styleguide/Components/__stories__/StepSummaryItem.story.tsx
ds300: I left this hard-coded in the story file for now. It's not clear to me that this constraint should come from the grid system, because it's not about alignment so much as it is about upholding the legibility of copy that spans multiple lines.

My guess is that it should probably be defined in palette based on the font size? But overridable for situations where having longer lines is the desired behaviour, of course.

PR #1070, opened 2018-07-31
Update the height of the <LargeSelect> component to always be 40px

yuki24:
In the new Buy Now flow we have a use case for the existing `<LargeSelect>`component. Currently, the height of this component is `46px`, which makes it slightly larger than other existing components used in the form. This PR updates the component so the component of the height will always be `40px`. This component is used in [the auction results page](https://staging.artsy.net/artist/yayoi-kusama/auction-results). In zeplin, the height of the select box is also `40px`, so this will actually make the height consistent across design mockups and the actual implementation.

Here are some screenshots of what it's going to look like after this PR. As you can see, it makes the height consistent and there will be no significant visual issue in the auction results page.

## Shipping address form the purchase team is working on

shipping-address-form

## Auction results page

action-results


zephraph:
We’d need to run this one by Brian first.

PR #1071, opened 2018-07-31
@l2succes => Lock @babel versions

eessex:
We were using `^` carrot version numbers of `@babel` packages, which causes mismatched versions to be used in Yarn lock.  This led to conflicts during build when used in projects-- specifically, Positron was unable to build when using `yarn link` from current master of both, although it worked fine when installing the package from NPM. 

This locks all `@babel` packages to `7.0.0-beta.54` except `@babel/preset-es2015`, which ends at `7.0.0-beta.53`.

Confirmed to compile successfully in Force and Positron when using `yarn link`.


artsyit:

Built with commit 66faea7e171459ede6dde64678d2870bb769b287

https://deploy-preview-1071--artsy-reaction.netlify.com

PR #1078, opened 2018-07-31
[ArtworkGrid] Limit image size by breakpoint

damassi:
Caps column image widths to 400 and 300, depending on breakpoint size. 

```js
        switch (match) {
          case "xs":
          case "sm":
          case "md":
            return 400
          case "lg":
          case "xl":
          default:
            return 300
        }
```

### DIFF
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -30,8 +30,7 @@ interface Props extends RelayProps, React.HTMLProps<ArtworkGridItemContainer> {
 }
 
 interface State {
-  width: number
-  height: number
+  isMounted: boolean
 }
 
 const IMAGE_QUALITY = 80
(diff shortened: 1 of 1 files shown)


src/Components/Artwork/GridItem.tsx
```diff
@@ -88,32 +102,38 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
       currentUserSpread = { currentUser }
     }
     return (
-      <div className={className} style={style}>
-        <Placeholder style={{ paddingBottom: artwork.image.placeholder }}>
-          <a href={artwork.href}>
-            <Image src={this.imageURL} innerRef={img => (this.image = img)} />
-          </a>
-          <Responsive>
```
damassi: @alloy - Should we maybe hoist `<Responsive>` up a level into ArtworkGrid? Seems like we could pass values into `GridItem` instead and save a lot of operations.

artsyit:

Built with commit c923a6de3f305eace6277c6fe410a6930e06a54d

https://deploy-preview-1078--artsy-reaction.netlify.com

mzikherman:
We should set the `GEMINI_ENDPOINT` in the netlify env (I think there's a UI for it in Netlify).

src/Components/Artwork/GridItem.tsx
```diff
@@ -44,37 +43,52 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
+          case "lg":
+          case "xl":
+          default:
+            return 300
+        }
+      }
+
+      const scale = window.devicePixelRatio
+      const width = String(getWidth() * scale)
+      const height = String(
```
mzikherman: should we maybe use `Math.floor` here so we don't have long decimals in the generated urls? 
screen shot 2018-07-31 at 7 56 56 pm

src/Components/Artwork/GridItem.tsx
```diff
@@ -44,37 +43,52 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
+          case "md":
+            return 400
+          case "lg":
+          case "xl":
+          default:
+            return 300
+        }
+      }
+
+      const scale = window.devicePixelRatio
```
mzikherman: can we drop the pixel ratio thing for now? it can result in a resize request of 800px (the 400px case * 2). The large images themselves are only 640px max so this is actually worse, plus upscales the image.

i dont really notice a difference / am unfamiliar with this setting?

src/Components/Artwork/GridItem.tsx
```diff
@@ -44,37 +43,52 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
+      const width = String(getWidth() * scale)
+      const height = String(
+        (this.image.width / this.props.artwork.image.aspect_ratio) * scale
```
mzikherman: This shouldn't be `this.image.width` anymore, but the width from above (ie- the 'max' width). We can drop the `innerRef` stuff then as well.

Also, realizing we don't need `String(...)` around this either since we're using string interpolation anyway.

I think we should replace these two lines with:

```
const width = Math.floor(getWidth())
const height = Math.floor(width / this.props.artwork.image.aspect_ratio)
```

mzikherman:
Actually I just added a commit rather than nitting :)

src/Components/Artwork/GridItem.tsx
alloy: The pixel ratio defines the number of pixels that make up a “point”. A “retina” screen has twice the amount of pixels per axis.

I guess if images get scaled up you won’t notice any difference indeed, so that requires a little more thinking, but I wouldn’t throw out the baby with the water just yet. Often devices with larger pixel ratios have better connections. We’re also all about showing images and the difference _should_ be noticeable. In my opinion we should talk to Brian.

mzikherman: We should, but for now I'm comfortable with dropping it. This whole approach probably deserves some thinking (and thus prioritization and time allocation within our sprints), but for now, this basically matches existing Force behavior, which should obviously be fine to roll with (for now!).

alloy: Sounds good 👍 (Be sure to file a ticket.)

PR #1083, opened 2018-08-06
fix radio group disabled and hover states

ds300:
I noticed that the Radio Group component had an extra hover state in the design system zeplin project, which I hadn't seen when working on this component originally. https://zpl.io/VqE4M6G

I also noticed that the `cursor: pointer` rule was applied for the label component even when the radio group was disabled, so I fixed that too.


artsyit:

Built with commit 59469c7fd6f19ac9e904b0647e9cb49069dfa013

https://deploy-preview-1083--artsy-reaction.netlify.com

PR #1084, opened 2018-08-06
Adds superscript to tabs

sepans:
screen shot 2018-08-06 at 2 30 54 pm

By passing `superscriptSelector`  function to `<tabs />`, you can select witch property of `<tab />` `data` prop to be displayed as superscript. (any other suggestion for the function name?)

```
<Tabs superscriptSelector={d => d.count}>
  <Tab data={{ count: 2 }} name="Open" />
  <Tab data={{ count: 3 }} name="Ready to ship" />
  <Tab data={{ count: 20 }} name="Complete" />
</Tabs>
```

PR #1087, opened 2018-08-06
Adds superscript to tabs (2)

sepans:
screen shot 2018-08-06 at 6 39 52 pm

As per @damassi and @zephraph feedback closing https://github.com/artsy/reaction/pull/1084 in favor of this one which doesn't require new component API 🎉 . 

Usage:
```
<Tabs justifyContent="center">
  <Tab
    name={
      <React.Fragment>
        Open<TabSuperscript>100</TabSuperscript>
      </React.Fragment>
    }
  />
  <Tab
    name={
      <React.Fragment>
        Ready to ship<TabSuperscript>4</TabSuperscript>
      </React.Fragment>
    }
  />
  <Tab
    name={
      <React.Fragment>
        Complete<TabSuperscript />
      </React.Fragment>
    }
  />
</Tabs>
```

### DIFF
--- a/src/Styleguide/Components/Tabs.tsx
+++ b/src/Styleguide/Components/Tabs.tsx
@@ -11,9 +11,10 @@ export interface TabLike extends JSX.Element {
   props: TabProps
 }
 
+type TabNameType = string | JSX.Element
 export interface TabInfo {
   /** Display name of the newly selected Tab */
-  name: string
+  name: TabNameType
 
   /** Index of the newly selected Tab */
   tabIndex: number
@@ -109,8 +110,7 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
 
 interface TabProps {
   /** Display name of the Tab */
-  name: string
-
+  name: TabNameType
   /**
    * Arbitrary data that can be associated with a Tab.
    *
@@ -185,3 +185,15 @@ const TabContainer = styled.div`
 const ActiveTabContainer = styled.div`
   ${styles.activeTabContainer};
 `
+
+const TabSuperScriptWrapper = styled.sup`
+  margin-left: 2px;
+`
+
+export const TabSuperscript = ({ children }: { children?: any }) => (
+  <TabSuperScriptWrapper>
+    <Sans size="1" weight="medium" display="inline">
+      {children}
+    </Sans>
+  </TabSuperScriptWrapper>
+)
--- a/src/Styleguide/Components/__stories__/Tabs.story.tsx
+++ b/src/Styleguide/Components/__stories__/Tabs.story.tsx
@@ -2,7 +2,7 @@ import React from "react"
 import { storiesOf } from "storybook/storiesOf"
 import { Box } from "Styleguide/Elements/Box"
 import { Section } from "Styleguide/Utils/Section"
-import { Tab, Tabs } from "../Tabs"
+import { Tab, Tabs, TabSuperscript } from "../Tabs"
 
 storiesOf("Styleguide/Components", module).add("Tabs (Simple)", () => {
   return (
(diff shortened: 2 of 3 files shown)


artsyit:

Built with commit 0e87470a0f41920459abb31c2be4990078de5cae

https://deploy-preview-1087--artsy-reaction.netlify.com

src/Styleguide/Components/__stories__/Tabs.story.tsx
```diff
@@ -58,12 +58,37 @@ storiesOf("Styleguide/Components", module).add("Tabs (Simple)", () => {
...
+            name={
+              <React.Fragment>
+                Ready to ship<TabSuperscript>4</TabSuperscript>
+              </React.Fragment>
+            }
+          />
+          <Tab
+            name={
+              <React.Fragment>
+                Complete<TabSuperscript />
```
damassi: Can `<Tabsuperscript>` be removed here?

src/Styleguide/Components/Tabs.tsx
```diff
@@ -185,3 +185,15 @@ const TabContainer = styled.div`
 const ActiveTabContainer = styled.div`
   ${styles.activeTabContainer};
 `
+
+const TabSuperScriptWrapper = styled.sup`
+  margin-left: 2px;
+`
+
+export const TabSuperscript = ({ children }: { children?: any }) => (
```
damassi: It might be nicer if this was exported as `<Sup>` as a play on real html tag `<sup>`. Then if we ever need a sub tab we could add `<Sub>`.

damassi: Also, `any` can be typed as `JSX.Element`

ds300: Another option to type the children would be like:

```ts
export const TabSuperscript: React.SFC<{}> = ({ children }) => ( ... )
```

src/Styleguide/Components/__stories__/Tabs.story.tsx
sepans: I left it intentionally as an example of having a tab without visible superscript and having the alignments perfect. I added a comment there.

src/Styleguide/Components/__stories__/Tabs.story.tsx
```diff
@@ -58,12 +58,38 @@ storiesOf("Styleguide/Components", module).add("Tabs (Simple)", () => {
...
           <Tab name="Pricing" />
           <Tab name="Condition" />
         </Tabs>
       </Section>
+      <Section title="With superscript">
+        <Tabs justifyContent="center">
+          <Tab
+            name={
+              <React.Fragment>
+                Open<Sup>100</Sup>
```
damassi: There we go :)

sepans:
👍 thanks for catching that!

src/Styleguide/Components/__stories__/Tabs.story.tsx
```diff
@@ -58,12 +58,38 @@ storiesOf("Styleguide/Components", module).add("Tabs (Simple)", () => {
...
+              </React.Fragment>
+            }
+          />
+          <Tab
+            name={
+              <React.Fragment>
```
zephraph: Instead of writing `<React.Fragment>` you can use `<></>`. It's a shorthand for fragment. https://reactjs.org/docs/fragments.html#short-syntax

src/Styleguide/Components/Tabs.tsx
```diff
@@ -185,3 +185,15 @@ const TabContainer = styled.div`
 const ActiveTabContainer = styled.div`
   ${styles.activeTabContainer};
 `
+
+const SupWrapper = styled.sup`
+  margin-left: 2px;
+`
+
+export const Sup: React.SFC<{}> = ({ children }) => (
```
zephraph: Is `Tabs` the best place for this component? It seems it might be something that should stand on its own.

sepans: Good point. It was originally named `TabSuperscript` and I renamed it to `Sup` but its implementation is very specific to this context so it shouldn't be its own thing. I can rename it back to something like `TabSup` if it is confusing, and/or differ the decision about if we want e general purpose `Sup` component  for now.

PR #1091, opened 2018-08-07
Price summary/review

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-305 https://artsyproduct.atlassian.net/browse/PURCHASE-308

https://zpl.io/2ymooBy


artsyit:

Built with commit 41d99675ef3fb28a699cd5227cebbada22927d04

https://deploy-preview-1091--artsy-reaction.netlify.com

src/Styleguide/Components/PriceSummary.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Spacer } from "Styleguide/Elements/Spacer"
+
+interface PriceSummaryProps extends FlexProps {
+  price: string
+  shipping: string
+  tax: string
+  total: string
+}
+
+export const PriceSummary: React.SFC<PriceSummaryProps> = ({
```
ds300: Hmmm maybe this component should live in src/Apps/Order/Components ?

ds300: ok done

src/Styleguide/Elements/Box.tsx
```diff
@@ -34,33 +35,51 @@ import {
...
+      :not(:first-child) {
+        border-top-left-radius: 0;
+        border-top-right-radius: 0;
+      }
+      :not(:last-child) {
+        border-bottom: 0;
+        border-bottom-left-radius: 0;
+        border-bottom-right-radius: 0;
+      }
+    `};
```
zephraph: Could you talk me through this? It seems like this is a specific implementation detail of a certain use case that's being added to a more generic component. 

The `hasSibilings` prop here is expecting that the borderbox appears in a stack with one item on top of another. That isn't necessarily the only usecase for this. I'm also uncertain of the responsibility of the `responsive` prop and why it would be used.

src/Styleguide/Elements/Box.tsx
```diff
@@ -34,33 +35,51 @@ import {
...
   ${maxWidth};
   ${width};
   ${height};
+  ${({ hover }) =>
+    hover &&
+    css`
+      :hover {
+        border-color: ${color("black60")};
+      }
+    `};
```
zephraph: :+1:

src/Apps/Order/Components/PriceSummary.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const PriceSummary: React.SFC<PriceSummaryProps> = ({
+  price,
+  shipping,
+  tax,
+  total,
+  ...others
+}) => {
+  return (
+    <BorderBox hasSiblings responsive flexDirection="column" {...others}>
```
damassi: What is the `hasSiblings` prop? And `responsive`?

src/Styleguide/Elements/Box.tsx
damassi: I was also curious. This feels a bit like API creep around something that can naturally be handled without explicit values being passed.

ds300: Yeah these are options to extend the behaviour of BorderBox. `hasSiblings` makes the borders seamless between sibling BorderBoxes, and `responsive` adds responsive padding.

I was taking the same approach as the `hover` prop, assuming that was the idiomatic way to add optional behaviour, but it sounds like you were both surprised to see these kinds of options being added. What would be the more idiomatic way to do this? Make a new component?

src/Apps/Order/Components/PriceSummary.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Entry label="Price" value={price} />
+      <Entry label="Shipping" value={shipping} />
+      <Entry label="Tax" value={tax} />
+      <Spacer height={space(2)} />
```
damassi: Does `<Spacer>`'s `height` prop not read directly into the theme and grab from `space`? If so that seems like a bug in styled-system 🤔

src/Styleguide/Elements/Box.tsx
```diff
@@ -34,33 +35,51 @@ import {
...
   border-radius: 2px;
-  ${props => props.hover && hover};
+  padding: ${space(2)}px;
   ${styledSpace};
   ${maxWidth};
   ${width};
   ${height};
+  ${({ hover }) =>
+    hover &&
+    css`
```
damassi: Is the `css` func necessary here? I (believe) that whenever a string is interpolated within the scope of a styled components template tag function that it automatically applies `css` under the hood.

src/Apps/Order/Components/PriceSummary.tsx
zephraph: It's not. It only grabs those values for the `space` properties which are limited to `p` and `m` values. I don't think it's a bug of system system so much as it is a nuance of how we do sizing.

src/Styleguide/Elements/Box.tsx
damassi: Ah, so `hasSiblings` allows things to be stackable, so that the bottom border on one component doesn't double up with the top border on a box below, yah? Rather than adding a new API method It might be more flexible to piggy back on styled-system and make the border more generic; e.g., 

```jsx
<BorderBox borderBottom={0}>
```

Since `BorderBox` is a primitive, I feel like adding `responsive` as a configuration prop might be some unnecessary overhead in terms of what `BorderBox` is responsible for. Two possible approaches would be to configure padding from the outside using the `<Responsive>` component, e.g., 

```js
<Responsive>
  {({xs} => 
     <BorderBox p={xs ? 2 : 0}>...</BorderBox>
  }}
<Responsive>
```

Or worry about padding from within the content of the `BorderBox`, like 

```jsx
<BorderBox>
   <SomeLocalStyledComponentWithResponsivePadding />
</BorderBox>
```

src/Apps/Order/Components/PriceSummary.tsx
damassi: I see. Would it reduce complexity to follow existing patterns and piggy back on existing props that accept space, like 

```jsx
<Spacer mb={2} />
```

Then we could remove the import and thus some indirection.

src/Styleguide/Elements/Box.tsx
ds300: You're right it's not strictly necessary in this instance. I would still include it though. Heres why:

e.g. if I was to change this to 

```js
  ${({ hover }) =>
    hover &&
    css`
      :hover {
        border-color: ${color("black60")};
      }
      ${media.xs`
        padding: 43px;
      `}
    `};
```

Then removing the `css` would break it because `media.xs` returns an array rather than an interpolated string. So it would be immediately rendered as `[object Object]` or something which is not correct CSS :)

I've been bitten by that a few times already, so now I use `css` whenever I interpolate non-primitive values. Think of it as a prophylactic.

src/Apps/Order/Components/PriceSummary.tsx
ds300: I changed this. Good spot!

src/Styleguide/Elements/Box.tsx
damassi: Makes sense 👍

src/Styleguide/Elements/Box.tsx
zephraph: In general, I think we should design for responsiveness. A parent shouldn't need to worry if a child it's consuming is responsive or not, unless it's implicitly implied (i.e SmallComponent, LargeComponent). By default we should be able to consume something like `BorderBox` in our given context and assume it to work correctly. The fact that this component changes its padding on mobile makes me wonder if `borderbox` currently has invalid padding on mobile. I think these two should probably behave inherently the same.

Ultimately, if the component generally responds a certain way on different breakpoints that should be encoded in the component directly, not passed down on every usage of the component. The parent should only handle responsive behavior when it has special needs. 

Having a separate primitive for a group of connecting border-box components does feel right. I think once we sort out the difference in padding behavior between this component and border-box this component will be a nice edition. At that point I'd propose just renaming it to `StackableBorderBox`. 

Works as is for now though. @ds300 thanks for your patience through this! 😃

damassi: Re `StackableBorderBox` -- one thought is that it could also exist as an export from `BorderBox` rather than an additional file.

PR #1097, opened 2018-08-07
@eessex: Mobile signup cta for articles

xtina-starr:
This addresses https://artsyproduct.atlassian.net/browse/GROW-705.

- It implements the CTA on the article page.
- Updates the MinimalCtaBanner to include animations per the spec.
- Allows the user to dismiss the CTA Banner .

signup cta

### DIFF
--- a/src/Components/MinimalCtaBanner.tsx
+++ b/src/Components/MinimalCtaBanner.tsx
@@ -1,55 +1,89 @@
 import React from "react"
 import styled from "styled-components"
 import { unica } from "../Assets/Fonts"
+import SlideTransition from "./Animation/SlideTransition"
 import Icon from "./Icon"
 
 export interface MinimalCtaBannerProps extends React.Props<HTMLDivElement> {
   backgroundColor?: string
   copy?: string
+  height?: string
   href?: string
   position: "top" | "bottom"
   textColor?: string
+  showCtaBanner?: boolean
 }
 
-const MinimalCtaBanner: React.SFC<MinimalCtaBannerProps> = props => (
-  <BannerContainer
-    position={props.position}
-    backgroundColor={props.backgroundColor}
-  >
-    <Banner href={props.href} textColor={props.textColor}>
-      <p>{props.copy}</p>
-      <IconContainer>
-        <Icon name="chevron-right" color={props.textColor} fontSize="12px" />
-      </IconContainer>
-    </Banner>
-  </BannerContainer>
-)
+export interface State {
+  showCta: boolean
+}
+
+export class MinimalCtaBanner extends React.Component<
+  MinimalCtaBannerProps,
+  State
+> {
+  state = {
+    showCta: this.props.showCtaBanner,
+  }
+
+  dismissCta = () => {
+    const showCta = false
+    this.setState({ showCta })
+  }
+
+  render() {
+    const ctaBanner = (
+      <SlideTransition
+        in={this.props.showCtaBanner}
+        timeout={{ enter: 10, exit: 250 }}
+        height={this.props.height}
+      >
+        <BannerContainer
+          position={this.props.position}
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit 79a3fc68f257af7e8997e803f598038bc9898832

https://deploy-preview-1097--artsy-reaction.netlify.com

src/Components/Publishing/Article.tsx
```diff
@@ -33,6 +36,17 @@ export interface ArticleProps {
...
 }
 
+export interface State {
+  showCtaBanner: boolean
+}
+
+export enum CtaCopy {
+  news = "Sign up for the best in art world news",
+  standard = "Sign up to get our best stories everyday",
+  feature = "Sign up to get our best stories everyday",
```
eessex: Since news is the only one that differs from the default, we probably don't need to declare standard and feature here.

src/Components/Publishing/Article.tsx
```diff
@@ -45,7 +59,42 @@ export interface ArticleProps {
...
+      this.setState({ showCtaBanner })
+    }
+    this.lastScrollPosition = newScrollPosition
+  }
+
+  componentDidMount() {
+    if (window) {
+      window.addEventListener("scroll", e => {
+        this.handleScroll()
+      })
```
eessex: Might be a good idea to debounce `handleScroll` here so we aren't calling every pixel.

src/Components/Publishing/Article.tsx
```diff
@@ -68,7 +117,28 @@ export class Article extends React.Component<ArticleProps> {
     }
   }
 
+  shouldRenderSignUpCta = () => {
+    const { article, isLoggedIn, isMobile } = this.props
+
+    return isMobile && article.layout !== "series" && !isLoggedIn
```
eessex: Worth double checking w @owendodd here if we should exclude from 'video' layouts as well -- they are also on a black background so it may not look right or need a different BG color.

src/Components/Publishing/__tests__/Article.test.tsx
```diff
@@ -43,3 +44,42 @@ it("renders news articles in news layout", () => {
   const article = mount(<Article article={NewsArticle} />)
   expect(article.find(NewsLayout).length).toBe(1)
 })
+
+it("renders mobile MinimalCtaBanner for standard article layouts", () => {
+  const article = shallow(
```
eessex: Why are we checking for standard layout, but not other layouts? Should at least check that it isn't appearing on series.

src/Components/MinimalCtaBanner.tsx
```diff
@@ -1,55 +1,88 @@
...
+  dismissCta = () => {
+    this.setState({ dismissed: true })
+  }
+
+  render() {
+    const ctaBanner = (
+      <SlideTransition
+        in={this.props.show}
+        timeout={{ enter: 10, exit: 250 }}
+        height={this.props.height}
```
eessex: I wonder if there is a way to default the `SlideTransition`'s height to `min-content` or `inherit` and avoid this prop altogether

src/Components/Publishing/Article.tsx
owendodd: Good catch, thanks @eessex. Would be great to invert the banner color in these instances. So it would be a white banner with black text.

src/Components/MinimalCtaBanner.tsx
xtina-starr: I chatted about this IRL with @l2succes and @yuki24. They thought it might be fine to have a specified height for the time being. Let me know if you have strong opinions about this because I would totally be open to more dialogue.

eessex: Ok I think worth revisiting when the artist CTA is implemented.

PR #1106, opened 2018-08-09
@zephraph => Update Payoff page to have layout

sweir27:
This updates the payoff screen to have the correct layout.

Pretty standard except I did end up moving (after talking to @zephraph) the stepper component into its own `BuyNowStepper` which is rendered only on the pages that have a stepper (meaning, not on the payoff page).

Open to other ideas about naming, etc.!

image

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -1,9 +1,11 @@
 import { Payment_order } from "__generated__/Payment_order.graphql"
+import { BuyNowStepper } from "Apps/Order/Components/BuyNowStepper"
 import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
 import React, { Component } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { Link } from "Router"
 import { Button } from "Styleguide/Elements/Button"
+import { Col, Row } from "Styleguide/Elements/Grid"
 import { Join } from "Styleguide/Elements/Join"
 import { Spacer } from "Styleguide/Elements/Spacer"
 import { Placeholder } from "Styleguide/Utils/Placeholder"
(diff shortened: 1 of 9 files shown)


src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -21,42 +23,52 @@ export class PaymentRoute extends Component<PaymentProps> {
...
-                  {!xs && (
-                    <Link to={`/order2/${order.id}/review`}>
-                      <Button size="large" width="100%">
-                        Continue
-                      </Button>
-                    </Link>
+      <>
+        <Row>
+          <Col>
+            <BuyNowStepper currentStep={"payment"} />
```
sweir27: I left this outside of the `Responsive` component since that's how it was before, but am happy to change if necessary.

src/Apps/Order/Routes/Submission/index.tsx
```diff
@@ -1,24 +1,58 @@
...
+          <TwoColumnLayout
+            Content={
+              <>
+                <Join separator={<Spacer mb={3} />}>
+                  <>
+                    <Serif size="6" weight="regular" color="black100">
+                      Your order has been submitted
+                    </Serif>
+                    <Sans size="2" weight="regular" color="black60">
+                      Order #{order.code}
```
sweir27: We use the `code` (more human-readable) here instead of the ID.

artsyit:

Built with commit 387ebe7e01dffd2bafaad83b788474cdd692e381

https://deploy-preview-1106--artsy-reaction.netlify.com

src/Apps/Order/Routes/Payment/index.tsx
zephraph: It's fine because the stepper is responsive itself

src/Apps/Order/Routes/Submission/index.tsx
zephraph: Would you update the mock in the storybook definition to give this an example value? `graphql-tools` is infering a value based off of a type, so it's rendering...

image

Good that it renders something! 😅 Would be nice to have a more relealistic value coming through though.

zephraph: https://github.com/artsy/reaction/blob/387ebe7e01dffd2bafaad83b788474cdd692e381/src/Apps/__stories__/OrderApp.story.tsx#L14

sweir27: Oh!! I actually was attempting to track down where that `"Hello World"` was coming from-- good to know it's graphql-tools. I'll add a default.

src/Apps/__stories__/OrderApp.story.tsx
```diff
@@ -12,6 +12,7 @@ const mock = {
   Order: (_, { id }) => {
     return {
       id,
+      code: "abcdefg",
```
sweir27: @zephraph updated!

PR #1107, opened 2018-08-09
Removes 'deg' from chevron transform rotation

sepans:
@jonallured found out that unlike [css transform rotation](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate) svg [transform](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#Rotate) doesn't work with `deg`. This PR fixes that.

### DIFF
--- a/src/Assets/Icons/ChevronIcon.tsx
+++ b/src/Assets/Icons/ChevronIcon.tsx
@@ -2,10 +2,10 @@ import { color } from "@artsy/palette"
 import React from "react"
 
 export enum Direction {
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 6b1f1982b7138f0295b2543ab2fe038180d02132

https://deploy-preview-1107--artsy-reaction.netlify.com

jonallured:
Thanks for fixing my mistake!! ❤️

PR #1109, opened 2018-08-09
Outline basic navigation logic

zephraph:
Begins addressing [PURCHASE-338](https://artsyproduct.atlassian.net/browse/PURCHASE-338). Currently this adds transition hooks to the router to give messaging when navigating away from the purchase flow. If you're navigating between parts of the purchase flow the transition continues, otherwise it'll give appropriate messaging about losing saved changes. 

The route transition living in the order app isn't ideal because it'll trigger even on non-form routes. I'll refactor this a bit once we have a better idea of how we access previous state data when moving back through the flow or potentially on refresh.

I'll link the related force PR to correctly configure the router on that side.


artsyit:

Built with commit 3e2f69e5c145d3551db6294ebc169b85f849281d

https://deploy-preview-1109--artsy-reaction.netlify.com

src/Apps/Order/OrderApp.tsx
```diff
@@ -18,26 +18,55 @@ export interface OrderAppProps {
...
+    // FIXME: This logic shouldn't be invoked on a non-form based route
+    // Refresh
+    if (window && location.pathname === window.location.pathname) {
+      // Most browsers will ignore this and supply their own messaging for refresh
+      return "Are you sure you want to refresh? Your changes will not be saved."
+    }
+
+    // https://regexper.com/#%5C%2Forder%5Cd%3F%5C%2F.%2B%5C%2F%28%5Cw%2B%29
+    // Attempting to navigate to another route
+    if (location.pathname.match(/\/order\d?\/.+\/(\w+)/)) {
```
damassi: is it possible to scope this directly from the `router` object rather than regex?

zephraph: That is an _excellent_ suggestion. Thanks Chris! It felt ~a little~ dirty, but I've been staring at this forever, ha. I'll update it!

sweir27:
Ditto to Chris's comment + looks like it needs a rebase... but this looks great! Ended up being way simpler than I was imagining.

Probably out-of-scope of this PR but as you're thinking about test examples for reaction, I hope that we can eventually have robust routing tests as this logic grows 😄.

damassi:
Re tests this might be helpful at some point: https://github.com/4Catalyzer/found/blob/master/test/makeRouteConfig.test.js

zephraph:
Yeah, that'll be my next step.

src/Apps/Order/OrderApp.tsx
```diff
@@ -8,13 +16,38 @@ export interface OrderAppProps {
...
+  componentWillUnmount() {
+    this.removeTransitionHook()
+  }
+
+  onTransition = location => {
+    const { routes, routeIndices } = this.props
+    let route = findRoute(routes, routeIndices)
+
+    if (route.onTransition) {
+      return route.onTransition(location)
```
zephraph: A transition will only be called for a route that has specified an `onTransition` parameter.

src/Apps/Order/routes.tsx
```diff
@@ -22,6 +22,29 @@ import { ShippingProps } from "./Routes/Shipping"
 // @ts-ignore
 import { SubmissionProps } from "./Routes/Submission"
 
+/**
+ * Used to match against child route locations. Supports order or order2.
+ * Captures the child route name in the first capture group.
+ *
+ * https://regexper.com/#%5C%2Forder%5Cd%3F%5C%2F.%2B%5C%2F%28%5Cw%2B%29
+ */
+const ORDER_REGEX = /\/order\d?\/.+\/(\w+)/
```
zephraph: The regex is still around.. the location is the only thing that I'm aware of that exposes the full path. I haven't figured out how to get the location of the individual routes without manually building it. As that requires knowledge of the route structure, it would ultimately result in the same thing.

src/Apps/Order/OrderApp.tsx
damassi: I like this pattern 👍

zephraph:
@damassi @alloy I've re-arranged the routing logic a bit. I like this setup better from a logical organization perspective. Also, now only the forms where a user might lose information give the transition messages. The logic for the transition still isn't great. I played with logically building the routes and trying to get the router to do by bidding, but it's a wily one. 

I'm going to start adding tests, but feedback on the current approach would be appreciated.

src/Apps/Order/OrderApp.tsx
```diff
@@ -8,13 +16,38 @@ export interface OrderAppProps {
...
+    )
+  }
+
+  componentWillUnmount() {
+    this.removeTransitionHook()
+  }
+
+  onTransition = location => {
+    const { routes, routeIndices } = this.props
+    let route = findRoute(routes, routeIndices)
```
damassi: Can replace with `const` if `route` isn't going to change

damassi:
@zephraph - Looking through the [Found docs](https://github.com/4Catalyzer/found#blocking-navigation) this seems like exactly the way to do it, so LGTM 👍. 

I don't want to hold this up unless @alloy has other comments but a couple things: 

One thought about a possible future reusable refactor would be to encode this into a declarative component similar to how React Router 4 [does it](https://reacttraining.com/react-router/web/example/preventing-transitions). The component could take a validation function and wire up the hooks inside of its own `componentDidMount` lifecycle method. The current route could be be matched against the validator and block a transition using pattens found in this PR. 

Another thing to note is that all of the attributes of the route can be extracted from the router itself, with no need to rely on `window` as [seen here](https://github.com/artsy/reaction/pull/1109/files#diff-4de6c2335984a96e9a30d2366b343e24R35). When a route is rendered from a route definition a lot of attributes are spread into props: 

screen shot 2018-08-13 at 4 09 23 pm

Passing these along will make things safer for SSR.

Also note that in that screenshot there is a `match` object. This value is derived from [this exported util function](https://github.com/4Catalyzer/found/blob/master/test/Matcher.test.js) on found. It might be more solid than using the Regex.

src/Apps/Order/routes.tsx
```diff
@@ -22,6 +22,29 @@ import { ShippingProps } from "./Routes/Shipping"
...
+
+const confirmRouteExit = location => {
+  // Refresh
+  if (window && location.pathname === window.location.pathname) {
```
damassi: This should be `typeof window !== 'undefined'`; I believe `window` as a value will error out.

alloy:
I had a quick play with the netlify build and I _think_ we can use the Found Mather API like so:

```ts
export class OrderAppTransitionHandler extends React.Component<OrderAppProps> {
    removeTransitionHook: () => void
  
    constructor(props) {
      super(props)
  
      this.removeTransitionHook = props.router.addTransitionHook(
        this.onTransition
      )
    }
  
    componentWillUnmount() {
      this.removeTransitionHook()
    }
  
    onTransition = location => {
        const match = this.props.router.matcher.match(location)
        if (match) {
            const matchedRoutes = this.props.router.matcher.getRoutes(match)
            if (
                matchedRoutes &&
                matchedRoutes[0].Component === OrderAppTransitionHandler &&
                matchedRoutes[1].AllowTransitionWithoutAlert
            ) {
                return true
            }
        }
        return "Are you sure you want to leave? Your changes will not be saved."
    }
  
    render() {
      return <OrderApp {...this.props} />
    }
  }

export const routes = [
    {
      path: "/order2/:orderID",
      Component: OrderAppRouterRoot,
      children: [
        {
          path: "shipping",
          Component: ShippingRoute,
          AllowTransitionWithoutAlert: true,
        },
      ],
    }
]
```

Whenever we may need to re-use this in the future, it should be trivial to turn the `OrderAppTransitionHandler` into a generic HOC.

src/Apps/Order/OrderApp.tsx
zephraph: We should likely have a lint rule if that's something we're going to enforce

zephraph:
Oh, that's a good catch @alloy! I'll try that out, thanks!

zephraph:
Okaaaaaaayyyyyy. I think this is good to go. 🤞

src/Apps/Order/routes.tsx
```diff
@@ -22,27 +22,26 @@ import { ShippingProps } from "./Routes/Shipping"
...
-  // Attempting to navigate to another route
-  if (location.pathname.match(ORDER_REGEX)) {
-    return true
+  // Attempting to navigate to another route in the orders app
+  const match = router.matcher.match(newLocation)
+  if (match) {
+    const matchedRoutes = router.matcher.getRoutes(match)
+    if (matchedRoutes && matchedRoutes[0].Component === OrderApp) {
+      return true
+    }
```
damassi: 👍

src/Apps/Order/OrderApp.tsx
```diff
@@ -8,13 +16,39 @@ export interface OrderAppProps {
     orderID: string
   }
   location: any
+  routeIndices: number[]
+  routes: any
+  router: any
```
alloy: I think that Found actually comes with TS type defs, no?

damassi: https://github.com/4Catalyzer/found/blob/master/index.d.ts

zephraph: 🙀 

Okay. I'll take a look.

alloy: Looks like it could do with some improvement to the `Matcher` API, though. Let me know if you want to pair on a PR.

zephraph: This PR or the found Matcher types?

alloy: The Matcher types

zephraph:
Okay, I think this is **really** ready this time.

sweir27:
👏 👏 Thanks for taking this on @zephraph!

PR #1113, opened 2018-08-10
Add the ability to override the width for <Separator>

yuki24:
We have a case where we need to override the `width` in the new `<BorderedRadioGroup>`. This PR removes the hard-coded `width: 100%` and make it more flexible.


artsyit:

Built with commit 93af954b2ab85ef1987b569f3274452e61c30241

https://deploy-preview-1113--artsy-reaction.netlify.com

PR #1114, opened 2018-08-10
[FEATURE components] Add <Collapse> component

yuki24:
This adds a new `<Collapse>` component we are going to use in the new Buy Now flow:

collapse

### DIFF
--- a/src/Styleguide/Components/Collapse.tsx
+++ b/src/Styleguide/Components/Collapse.tsx
@@ -1,0 +1,50 @@
+import React from "react"
+import styled from "styled-components"
+
+export interface CollapseProps {
+  open: boolean
+}
+
+export interface CollapseState {
+  computedHeight: string
+}
+
+export class Collapse extends React.Component<CollapseProps, CollapseState> {
+  state = {
+    computedHeight: null,
+  }
+
+  private element: HTMLElement
+
+  handleRef = element => {
+    this.element = element
+  }
+
+  componentDidMount() {
+    if (this.state.computedHeight === null) {
+      const prevHeight = this.element.style.height
+      this.element.style.height = "auto"
+      const computedHeight = getComputedStyle(this.element).height
+      this.element.style.height = prevHeight
+
+      this.setState({ computedHeight })
+    }
+  }
+
+  render() {
+    const height = this.props.open ? this.state.computedHeight || "auto" : "0"
+
+    return (
+      <Collapseable
+        innerRef={this.handleRef}
+        {...this.props}
+        style={{ height }}
+      />
+    )
+  }
+}
+
+const Collapseable = styled.div`
+  overflow: hidden;
+  transition: height 0.25s ease-in-out;
+`
--- a/src/Styleguide/Components/__stories__/Collapse.story.tsx
+++ b/src/Styleguide/Components/__stories__/Collapse.story.tsx
@@ -1,0 +1,47 @@
+import { Sans } from "@artsy/palette"
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { Collapse } from "Styleguide/Components/Collapse"
+import { Button } from "Styleguide/Elements"
+import { Section } from "Styleguide/Utils/Section"
+
+class Collapseable extends React.Component {
+  state = {
+    open: true,
+  }
+
+  render() {
+    return (
+      <>
+        <Button onClick={() => this.setState({ open: !this.state.open })}>
+          Toggle
+        </Button>
+
+        <Collapse open={this.state.open}>{this.props.children}</Collapse>
+      </>
+    )
+  }
+}
+
+storiesOf("Styleguide/Components", module).add("Collapse", () => {
+  return (
+    <React.Fragment>
+      <Section title="Collapse">
+        <Collapseable>
+          <Sans mt={3} size="3">
+            The elegant spiral of the Nautilus shell, the sinuous pattern of the
+            banks of a river, or the swirling vortex street of clouds - patterns
+            exist on every level in nature. Along with fractals, chaos theory is
+            one of the essential, universal influences on patterns in nature. In
+            essence, the theory shows how systems of chaotic, apparent
+            randomness have an underlying pattern, or repetition.
+          </Sans>
+        </Collapseable>
+
+        <Sans mt={3} size="3">
+          This text should move up and down as the section above animates.
+        </Sans>
+      </Section>
+    </React.Fragment>
+  )
+})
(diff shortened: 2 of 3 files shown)


artsyit:

Built with commit aeb12a1a2c4ca40b5d451bb74a809008d83becfd

https://deploy-preview-1114--artsy-reaction.netlify.com

damassi:
@yuki24 - Since https://github.com/artsy/reaction/pull/1082 we've started writing tests for all UI components; would you mind adding one before merge?

src/Styleguide/Components/Collapse.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  componentDidMount() {
+    if (this.state.computedHeight === null) {
+      const prevHeight = this.element.style.height
+      this.element.style.height = "auto"
+      const computedHeight = getComputedStyle(this.element).height
+      this.element.style.height = prevHeight
+
+      this.setState({ computedHeight })
```
damassi: One thought is that elsewhere we've started using `react-spring`: https://github.com/drcmda/react-spring; that might be something to consider here.

yuki24: Nice, I did some research yesterday but totally missed `react-spring`! I'll look into that and see if we could take advantage of it.

zephraph: We're using `react-spring` and `react-transition-group`?

zephraph: Regardless, yeah, we really should look to standardize on animations... @eessex and @xtina-starr might have some insight too.

damassi: Yup yup, react spring is being used in various places around Reaction

zephraph: Even given that `react-spring` is around, I like that this component has no dependencies. I'm 👍 with it as is unless there's an attribute of the animation that needs to be tweaked that would call for using `react-spring`.

damassi: One thing tho is that spring reduces the complexity. No more computations like what is happening here, we simply say `height: 0` and it will collapse, but declaratively.

yuki24: Here is a proof-of-concept that I just got working. Code is definitely shorter, I just need to figure out how to change the animation duration.

```tsx
import React from "react"
import { Spring, animated } from 'react-spring'

export interface CollapseProps {
  isOpen: boolean
}

export class Collapse extends React.Component<CollapseProps> {
  render() {
    return (
      <Spring native force from={{ height: 0 }} to={{ height: this.props.isOpen ? 'auto' : 0 }}>
        {props => (
          <animated.div style={props}>
            {this.props.children}
          </animated.div>
        )}
      </Spring>
    )
  }
}
```

damassi: Nice! 

@eessex - the `react-transition-group` API is pretty complicated and the pattern is somewhat dated compared to more modern options; I wonder if there was something missing in terms of mounting / unmounting that would have fixed the problem?

l2succes: I believe `react-spring` has a nicer API than `react-transition-group` and would take us one step closer to being able to use the same animation library for web and native mobile in our universal components. @eessex the issue you're describing could have been fixed, I think it would be beneficial for us to give try it again and potentially using it for all our animation needs, we don't have to do it all at once but this collapse component is a great candidate to test it out again.

xtina-starr: I haven't used `react-spring` only `react-transition-group` to add a slide animation. Despite not being able to speak to `react-spring`, I'm in favor of being modesty with the dependencies we decide to add. Is there any compelling reason to use both?

l2succes: @xtina-starr see my answer above but just to restate. `react-transition-group` is a DOM specific animation library. `react-spring` allows us to use the same animations and components in reaction and emission for instance. Also, having a simpler API to deal with would be a win for us in terms of developer accessibility. 

The docs for `react-spring` has a good explanation for what problems it solves (https://github.com/drcmda/react-spring#why-do-we-need-yet-another-), You should check it out.

orta: Not to pour gasoline on the discussion, but do we have enough animations in our apps to warrant a dependency at all? Most of the cases I can think of (except maybe that "click here to inquire on an artwork" thing) are mostly just hiding/showing/fading things elegantly

eessex: @damassi @l2succes as far as i could tell there was no option in `react-spring` to not mount a component that shouldn't be in view, although this is build into `react-transition-group` -- IMO hiding something with CSS only kind of defeats the purpose of using a library, or using react with animations.  I guess we could add this functionality to `react-spring`, but there didn't seem to be a point of adding functions to a library we had no other uses of.

damassi: looks like mount / unmount this was added: https://github.com/drcmda/react-spring#mountunmount-transitions-demo

briansw: Libraries/dependencies aside – from a UX perspective, we want to be taking a more modern approach to interactions by making use of animation to clarify interactions, particularly complex interactions that change the page composition to ground the user in order to avoid confusion. Hope that's helpful.

alloy:
Great addition @ds300 👌 With regards to abstractions, is there a library that would have encouraged the performant option? Or is that hard to do and thus possibly meaning that we shouldn’t be using any off the shelf libs for our needs, especially because they are relatively little?

ds300:
I've heard great things about this: https://popmotion.io/pose/ it apparently errs on the side of performance and is quite flexible.

yuki24:
I ended up using `react-spring` since it has a universal component for React Native and it will make it easier to port this component when necessary. 

I'm not that concerned about performance, but just FYI, `react-spring` is slower than the plain CSS `transition` compoment. Here is the performance breakdown captured using Chrome's performance feture with 8 collapse animations running simultaneously and the settings of CPU throttling `6x slowdown`:

## Plain CSS `transition`

The lowest FPS was 18 with a rough average of 35 FPS:

screen shot 2018-08-14 at 12 38 29 pm

## `react-spring`

The lowest FPS was 9 with a rough average of 23 FPS:

screen shot 2018-08-14 at 12 46 03 pm

zephraph:
I don't think we've really had a consensus on how to tackle this, but in consideration of @yuki24's time, I'm going to merge this so he can move on. I think we're just going to need an RFC for animation strategy.

ds300:
Cool stuff Yuki! These benchmarks are mainly helpful relative to each
other. We'd need to benchmark the transition on a page with a more
representative quantity and arrangement of elements to get a better picture
of whether performance is good or not, since the cost associated with
animating `height` is in re-computing the layout of the page. If the entire
page is like 4 elements that's not going to take much time :)

I suspect we'll be OK with BNMO pages, but would be a little more weary of
using this on more busy pages like artist and the home page, speaking from
personal experience.

PR #1115, opened 2018-08-10
Add a new <BorderedRadioGroup> component

yuki24:
This PR basically changes three things:

 * Update the existing `<RadioGroup>` component to not render borders
 * Break down the monolith `<RadioGroup>` interface into a composable API
 * Add a new `<BorderedRadioGroup>` component

At this point, we don't use the `<RadioGroup>` component outside of Reaction, so this should be ok to change the interface. We may be able to use this component on the new Artist page (e.g. https://www.artsy.net/artist/andy-warhol) in the future.

 * [x] Add new tests
 * [x] Fix failing tests

## Screenshot

screen shot 2018-08-10 at 3 37 36 pm

### DIFF
--- a/src/Styleguide/Elements/RadioGroup.tsx
+++ b/src/Styleguide/Elements/RadioGroup.tsx
@@ -1,25 +1,21 @@
-import { color, space } from "@artsy/palette"
 import React from "react"
 import { BorderProps, SizeProps, SpaceProps } from "styled-system"
 
-import { css } from "styled-components"
+import { BorderBox } from "Styleguide/Elements/Box"
 import { Flex } from "Styleguide/Elements/Flex"
-import { Radio } from "Styleguide/Elements/Radio"
+import { Join } from "Styleguide/Elements/Join"
+import { RadioProps } from "Styleguide/Elements/Radio"
+import { Separator } from "Styleguide/Elements/Separator"
+
+export interface RadioGroupElementProps {
+  value?: string
+}
 
 export interface RadioGroupProps {
   disabled?: boolean
-  onSelect: (selectedOption: string) => void
+  onSelect?: (selectedOption: string) => void
   defaultValue?: string
-  options: Array<{ label: React.ReactNode | null; id: string }>
-  renderRadio?: (
-    props: {
-      id: string
-      label: React.ReactNode
-      selected: boolean
-      onSelect: () => void
-      disabled: boolean
-    }
-  ) => React.ReactNode | null
+  children: Array<React.ReactElement<RadioProps>>
 }
 
 export interface RadioGroupToggleProps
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit d95689c03253a4970f517cfed03b6943b05c8298

https://deploy-preview-1115--artsy-reaction.netlify.com

ds300:
:thinking: It seems like this could have been achieved without changing the API by simply overriding the `renderRadio` prop? I could try to put together a POC if you like.

Regarding the API change: could you explain in more detail why your changes are an improvement?

I'm always a bit weary of these APIs that modify children. They make a lot of assumptions about what kind of children you're passing, and there's no way to type that, so refactorings become more brittle. It's also less developer-friendly because a lot of the time (and definitely in this case) you need to be aware that children modification is happening, and exactly how, in order to use the component safely.

I think what we had before is a good deal more robust and grokkable. But I know I wrote this code originally so I'm biased, and it would be good to have some other peeps weigh in.

yuki24:
@ds300 thanks for your feedback!

I did spike on a proof-of-concept using the `renderRadio` prop and wasn't sure about that - In my experience, props that render parts of the component's children are tricky to use unless it renders a very few, preferably just one component with no nesting. In this particular case, what slightly concerned me was the necessity to declare e.g. `{id === "PICKUP" && (...)}` to display additional content for each particular radio button. Let's say there's [a radio group that has three buttons in it](https://zpl.io/VqE4M6G), each of which has additional content. In order to achieve it the `renderRadio` function would need a couple conditionals:

```jsx
renderRadio={({ ... }) => (
  <StyledRadio ...>
    {label}

    {id === "COUNTER_OFFER" && (
      <Collapse ...>
        Your offer
        <Input .../>
      </Collapse>
    )}

    {id === "DECLINE" && (
      <Collapse ...>
        <Sans>Declining ...</Sans>
      </Collapse>
    )}
  </StyledRadio>
)}
```

Which doesn't easily translate to the actual DOM tree. Then to avoid the conditionals, we may consider adding a new property to the `options` list prop, but then the question becomes, why couldn't we just simply declare them, even without `<RadioGroup>`?

```jsx
<BorderBox flexDirection="column">
  <Radio ...>
    Accept seller’s offer
  </Radio>

  <Separator ... />

  <Radio>
    Send a counteroffer
    <Collapse ...>
      Your offer
      <Input .../>
    </Collapse>
  </Radio>

  <Separator ... />

  <Radio>
    Decline seller’s offer
    <Collapse ...>
      <Sans>Declining ...</Sans>
    </Collapse>
  </Radio>
</BorderBox>
```

With regards to the assumptions the parent component makes, yes, these are all valid arguments. In order to make it safer, we could add a different component that has the same styles but takes a different set of props:

```jsx
<RadioGroup defaultValue="SHIP" disabled>
  <RadioGroup.Button value="SHIP">
    Provide shipping address
  </RadioGroup.Button>

  <RadioGroup.Button value="PICKUP">
    Arrange for pickup

    <Collapse ...>
      <Sans>
        After you place your order, ...
      </Sans>
    </Collapse>
  </RadioGroup.Button>
</RadioGroup>
```

Of course, it wouldn't remove all the assumptions but would make the usage very explicit and less error-prone.

ds300:
Ah I see the trouble with the first example. The solution is that you don't need to do the conditionals inside `renderRow` because the `label` can be any old JSX.

e.g.

```
<RadioGroup
  renderRadio={({ label, ...props }) => (
    <Radio {...props}>{label}</Radio>
  )}
  options={[
    {
      id: "ACCEPT",
      label: "Accept the seller's offer",
    },
    {
      id: "COUNTER_OFFER",
      label: (
        <>
          Set a counter offer
          <Collapse>
            Your offer
            <Input ... />
          </Collapse>
        </>
      ),
    },
    {
      id: "DECLINE",
      label: (
        <>
          Decline the seller's offer
          <Collapse>Declining...</Collapse>
        </>
      ),
    },
  ]}
/>
```

But now that I think some more, I'm not so invested in this API discussion. If you prefer the inline JSX way then I can get on board with that too 👍It's only a little bit of magic.

src/Styleguide/Elements/RadioGroup.tsx
```diff
@@ -1,25 +1,20 @@
...
-  renderRadio?: (
-    props: {
-      id: string
-      label: React.ReactNode
-      selected: boolean
-      onSelect: () => void
-      disabled: boolean
-    }
-  ) => React.ReactNode | null
+  children: Array<React.ReactElement<RadioGroupElementProps>>
```
yuki24: I added this type to make it slightly safer in terms of the `children` usage. Of course it wouldn't make it perfectly safe but at least we can get a warning at compile time:

screen shot 2018-08-14 at 3 10 48 pm

src/Styleguide/Elements/__tests__/RadioGroup.test.tsx
```diff
@@ -1,64 +1,144 @@
...
+    expect(
+      wrapper
+        .find("Radio")
+        .last()
+        .props().disabled
+    ).toBe(true)
   })
 
-  it("disables radio group", () => {
+  it("ignores the 'selected' prop on the Radio component", () => {
```
yuki24: @zephraph I don't think this is going to be extremely confusing but this `selected` prop would collide with the `defaultValue` prop on the parent `<RadioGroup>` component. Any thoughts?

yuki24:
@zephraph This PR is good for review.

src/Styleguide/Elements/__tests__/RadioGroup.test.tsx
zephraph: Yeah, that's a good point... I feel like this is indeterministic behavior. It should likely result in an error in `dev`.

zephraph:
🎉

PR #1123, opened 2018-08-13
Adds TransactionSummary

ashfurrow:
This PR addresses [PURCHASE-304](https://artsyproduct.atlassian.net/browse/PURCHASE-304), and is a follow-up to #1091. It looks like a Jira ticketing mishap had us duplicating work a bit – I've added [a card](https://trello.com/c/hFauaFPm/2-work-got-duplicated) to this sprint's retrospective board to discuss. 

I took @ds300's work on the PriceSummary and expanded it into a general TransactionSummary, which has designs here: [zpl.io/a8K5qYQ](zpl.io/a8K5qYQ). I had to change the use of `StackableResponsiveBorderBox` because the separator needs to go full-width. Let me know if you have a better idea. This work is a bit outside my wheelhouse, so please give your unsalted opinion.

Here it is on desktop:

screen shot 2018-08-13 at 4 22 26 pm

And on mobile:

screen shot 2018-08-13 at 4 34 05 pm

And finally, in context of the shipping form:

screen shot 2018-08-13 at 4 22 38 pm

It's all stubbed data right now of course 😅 Can't wait to turn this into fragments.


artsyit:

Built with commit 87088290dbd953d650577ab181a0d2c8ec0bfab8

https://deploy-preview-1123--artsy-reaction.netlify.com

src/Apps/Order/Components/TransactionSummary.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Flex, FlexProps } from "Styleguide/Elements/Flex"
+import { Image } from "Styleguide/Elements/Image"
+import { Separator } from "Styleguide/Elements/Separator"
+import { Spacer } from "Styleguide/Elements/Spacer"
+import styled from "../../../../node_modules/styled-components"
+
+const SummaryContainer = styled(Flex)`
+  margin: ${space(2)}px;
+  ${media.sm`
+  margin: ${space(3)}px;
```
ds300: Hmm there's a formatting issue here. I guess prettier doesn't respect the `media.*` pattern?

src/Apps/Order/Components/TransactionSummary.tsx
```diff
@@ -0,0 +1,100 @@
...
+  total,
+  artistName,
+  artworkName,
+  sellerName,
+  artworkLocation,
+  imageURL,
+  ...others
+}) => {
+  return (
+    <BorderBox p={0} flexDirection="column" {...others}>
```
ds300: Just noticed your comment about the separator width and StackableBorderBox. This kind of thing was intended to be achieved was like this:

```js
<Flex flexDirection="column">
  <StackableBorderBox>{...}</StackableBorderBox>
  <StackableBorderBox>{...}</StackableBorderBox>
</Flex>
```

That way you don't need explicit separators. The 'Stackable' aspect of the box means that the borders blend into one seamlessly. This would also take care of the padding, so you wouldn't need the SummaryContainer element.

What you've done works fine though, I see no reason to change it right now except for the sake of consistency with other components.

ashfurrow: Oh! That makes _so_ much more sense, thank you! I think it's worth it to be consistent with the other components – I'll update the PR.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -119,11 +119,13 @@ export class ShippingRoute extends Component<ShippingProps> {
 
                   <Spacer mb={3} />
 
-                  <Link to={`/order2/${order.id}/payment`}>
-                    <Button size="large" width="100%">
-                      Continue
-                    </Button>
-                  </Link>
+                  {!xs && (
```
ashfurrow: See my comment here: https://github.com/artsy/reaction/pull/1125#pullrequestreview-146079061

ashfurrow:
Okay! This is 🍏

PR #1128, opened 2018-08-14
Add ItemReview component

ds300:
[Jira](https://artsyproduct.atlassian.net/browse/PURCHASE-309)
[Zeplin](https://zpl.io/bzWx4KE)

This component is currently only used on desktop.

image

### DIFF
--- a/src/Apps/Order/Components/__stories__/ItemReview.story.tsx
+++ b/src/Apps/Order/Components/__stories__/ItemReview.story.tsx
@@ -1,0 +1,60 @@
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { Flex } from "Styleguide/Elements/Flex"
+import { Section } from "Styleguide/Utils/Section"
+import { ItemReview } from "../ItemReview"
+
+storiesOf("Styleguide/Components", module).add("ItemReview", () => {
+  return (
+    <>
+      <Section title="Item Review">
+        <Flex width={542} flexDirection="column">
+          <ItemReview
+            artist="Francesca DiMattio"
+            title="The Fox and the Hound"
+            date="2018"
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 99042fec1eb83c0ec1d22d8d7177de3e28832313

https://deploy-preview-1128--artsy-reaction.netlify.com

src/Apps/Order/Components/__stories__/ItemReview.story.tsx
```diff
@@ -0,0 +1,60 @@
...
+            title="The Fox and the Hound"
+            date="2018"
+            medium="Oil on canvas"
+            dimensions={{
+              in: "96 × 79 in",
+              cm: "243.8 × 200.7 cm",
+            }}
+            attributionClassDescription="This is a unique work"
+            imageURL="https://d32dm0rphc51dk.cloudfront.net/XIlup0jb5BESj71JI7ZHpQ/larger.jpg"
+          />
```
damassi: @ds300 - one thing we continually noticed while building out the Artist page from static comps was that defining prop names ahead of time always had to be unwound once the Metaphysics / Relay side of things was introduced. (If you're certain that these are the fields that will eventually flow in all good tho 👍)

damassi:
LGTM 👍

src/Apps/Order/Components/__stories__/ItemReview.story.tsx
damassi: You can if you'd like (that's what we did while building out the new `/artwork page), but typically its better to leave the props-defining / wiring phase to a later stage when the component is actually being wired up into a view (meaning, wait until the component is plugged into relay). By keeping things totally static it increases flexibility during the design / comp phase.

PR #1130, opened 2018-08-14
@damassi => Adds basic error page

sweir27:
This PR adds a basic error page to reaction.

Some notes:
- It's just the error page _content_ right now (since it seems that the header is still rendered right now from force? I wasn't totally sure about this. If we need to render a custom header here, hopefully we can do something minimal with just the logo)
- This doesn't hook up the page rendering to anything, it's just the component which accepts `code`, `message`, and `detail`.
- I copied this verbatim from force, but tried to alter the styles within reason to match our design system

image

image

### DIFF
--- a/src/Components/ErrorPage.tsx
+++ b/src/Components/ErrorPage.tsx
@@ -1,0 +1,101 @@
+import { color, Serif } from "@artsy/palette"
+import { garamond } from "Assets/Fonts"
+import React from "react"
+import styled from "styled-components"
+import { Button } from "Styleguide/Elements/Button"
+import { Spacer } from "Styleguide/Elements/Spacer"
+
+interface ErrorPageProps {
+  code: number
+  message?: string
+  detail?: string
+}
+
+interface ErrorCodeBackgroundProps extends React.HTMLProps<HTMLDivElement> {
+  code: Number
+}
+
+export class ErrorPage extends React.Component<ErrorPageProps, null> {
+  render() {
+    const { code, message, detail } = this.props
+    const defaultMessage =
+      code === 404
+        ? "Sorry, the page you were looking for doesn’t exist at this URL."
+        : "Internal Error"
+
+    const detailMessage = message ? `Error Message: ${message}` : detail
+
+    return (
+      <>
+        <StyledErrorCodeBackground code={code}>
+          <StyledErrorDefaultMessage>
+            {defaultMessage}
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit aa4686e91d747c716e1fff7b2c6f57d284cdd142

https://deploy-preview-1130--artsy-reaction.netlify.com

src/Components/ErrorPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const defaultMessage =
+      code === 404
+        ? "Sorry, the page you were looking for doesn’t exist at this URL."
+        : "Internal Error"
+
+    const detailMessage = message ? `Error Message: ${message}` : detail
+
+    return (
+      <>
+        <StyledErrorCodeBackground code={code}>
```
damassi: I think these components can be simplfied to `ErrorCodeBackground`, `ErrorDefaultMessage`, and so on. The idiom for using a `Style` prefix is when needing to wrap (override) an existing component; e.g., 

```js
import styled from 'styled-components'
import { SomeComponent } from 'path/to/SomeComponent'

const StyledSomeComponent = styled(SomeComponent)`
  ...
`

<StyledSomeComponent />
```

src/Components/ErrorPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+  max-width: 75%;
+  font-family: "Menlo", "Monaco", "Andale Mono", "lucida console", "Courier New",
+    "monospace";
+  overflow-x: auto;
+  padding: 5px 9px;
+  text-align: left;
+  word-break: break-word;
+`
+
+const StyledErrorCodeBackground = styled<ErrorCodeBackgroundProps, any>("div")`
```
damassi: ooo I didn't realize this TS syntax was possible. Much nicer than previous way

sweir27: I found it somewhere towards the bottom of this issue: https://github.com/styled-components/styled-components/issues/630 (always nice to see friendly faces when reading through these 😄 cc @alloy )

src/Components/ErrorPage.tsx
sweir27: Oh! I can get behind that (and wasn't sure about the convention).

yuki24:
This has been deployed to staging by https://github.com/artsy/force/pull/2800

PR #1132, opened 2018-08-14
@l2succes => Display video bugs

eessex:
Addresses
https://artsyproduct.atlassian.net/browse/GROW-851
https://artsyproduct.atlassian.net/browse/GROW-841

I think this may have originated from our switch to typed analytics -- we now use the `react-tracking` library rather than a deprecated version that was in utils.  The tracking event was expecting an additional arg of state, which caused an error when clicking the video link.


artsyit:

Built with commit 9abfce370fe419e193198eab79037a7aaa439de2

https://deploy-preview-1132--artsy-reaction.netlify.com

eessex:
@l2succes bumping

PR #1139, opened 2018-08-15
Shipping form commits mutation

ashfurrow:
This pull request adds the first mutation to the new order components (🎉) and address [PURCHASE-329](https://artsyproduct.atlassian.net/browse/PURCHASE-329). I've left a few TODOs with associated tickets to make sure they don't get dropped; the most immediate is moving the address entry fields into their own component so we can re-use it on the billing address screen. We'll almost certainly want to use Formik, as @damassi suggested, so I've included that in the note as well.

Since Relay will update its store with the new order state, we can tie into using that in our routing logic once this PR is merged. /cc @zephraph.

Here it is in action:

screen shot 2018-08-15 at 4 23 39 pm

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -1,14 +1,18 @@
 import { Sans, Serif } from "@artsy/palette"
 import { Shipping_order } from "__generated__/Shipping_order.graphql"
 import React, { Component } from "react"
-import { createFragmentContainer, graphql } from "react-relay"
+import {
+  commitMutation,
+  createFragmentContainer,
+  graphql,
+  RelayProp,
+} from "react-relay"
 
-import Input from "Components/Input"
-import { Link } from "Router"
 import { Collapse } from "Styleguide/Components/Collapse"
 import { CountrySelect } from "Styleguide/Components/CountrySelect"
 import { Button } from "Styleguide/Elements/Button"
 import { Flex } from "Styleguide/Elements/Flex"
+import { Col, Row } from "Styleguide/Elements/Grid"
 import { Join } from "Styleguide/Elements/Join"
 import { Radio } from "Styleguide/Elements/Radio"
 import { BorderedRadioGroup } from "Styleguide/Elements/RadioGroup"
@@ -18,7 +22,7 @@ import { Responsive } from "Utils/Responsive"
 import { SummaryFragmentContainer as Summary } from "../../Components/Summary"
 
 import { BuyNowStepper } from "Apps/Order/Components/BuyNowStepper"
-import { Col, Row } from "Styleguide/Elements/Grid"
+import { Input } from "Components/Input"
 import {
   TwoColumnLayout,
   TwoColumnSplit,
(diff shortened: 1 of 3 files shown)


src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  it("commits the mutation with the orderId", () => {
+    const component = getWrapper(props)
+    component
+      .find(Radio)
+      .last()
+      .simulate("click")
+    const mockCommitMutation = commitMutation as jest.Mock<any>
+    mockCommitMutation.mockImplementationOnce((_environment, config) => {
+      expect(config.variables.input.orderId).toBe("1234")
+    })
```
damassi: nice 👍

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -0,0 +1,83 @@
+import { mount } from "enzyme"
+import React from "react"
+import { commitMutation, RelayProp } from "react-relay"
+
+import { Button } from "Styleguide/Elements/Button"
+import { Radio } from "Styleguide/Elements/Radio"
+import { Provider } from "../../../../../node_modules/unstated"
```
damassi: I suspect this was from VSCode auto-import; should be 
```
import { Provider } from "unstated"
```

(I wonder if this is a bug in VSCode or a setting that needs to be adjusted -- this has happened quite a bit to me too)

ashfurrow: Ha, yeah it's definitely an auto-import issue. Not sure if it's a bug or config issue.

artsyit:

Built with commit edd1484d97be23d5a87d5a09848c6810157c9246

https://deploy-preview-1139--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
orta: Hrm, it looks like this _should_ be fixed with TS 3.0 - but maybe it isn't? 
https://github.com/Microsoft/TypeScript/issues/23230#issuecomment-404330449

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
zephraph: Seriously! This is awesome @ashfurrow!

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -29,15 +33,91 @@ export interface ShippingProps {
...
+  onUpdateName = e => {
+    this.setState({ name: e.target.value })
+  }
+  onUpdateAddressLine1 = e => {
+    this.setState({ addressLine1: e.target.value })
+  }
+  onUpdateAddressLine2 = e => {
+    this.setState({ addressLine2: e.target.value })
+  }
+  onUpdateCity = e => {
```
orta: the `{` and `}` are optional here, unless you *really* care about not returning the same value as this.setState

ashfurrow: Oh yeah, those are left over from an earlier draft of moving the address form into its own component, before I decided to just get the Relay mutation working first. I’ll take them out.

alloy:
This reminds me that I need to update the typings for `commitMutation` so that we can actually use the typings mutation related typings that are emitted by our TS plugin.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -29,15 +33,91 @@ export interface ShippingProps {
   mediator?: {
     trigger: (action: string, config: object) => void
   }
+  relay?: RelayProp
+}
+
+// TODO: When the todo for abstracting the address is done and we have an Address component, we won't need this here, so the wonky state generic on ShippingRoute is fine for now.
+export interface Address {
+  // TODO: Shipping name is not in Exchange's schema yet.
```
sweir27: Helpful `TODO`s!

sweir27:
This is an awesome start. Once you feel good about the little cleanup, I'd say we should merge!

Some follow-ups related to abstracting out the `Address` component which may be embedded in other stories are:
- We may want different validation logic depending on whether you're entering a shipping or billing address (or maybe not. In any case we'll likely want to add some validation around the presence of fields, which is hopefully easy to do with a library like formik).
- The shipping form will need to handle certain extra logic-- described here https://artsyproduct.atlassian.net/browse/PURCHASE-350 to send mutations/update the shipping/tax based on values in the form. We don't need to do this now by any means, just mentioning it in case it affects how that generic component is architected.

src/Apps/Order/Routes/Shipping/index.tsx
ashfurrow: Thanks!

ashfurrow:
Thanks @sweir27! I've added a link to those constraints to PURCHASE-375 to abstract on the component, to make sure we have those constraints in mind. I'll wrap this up momentarily.

PR #1143, opened 2018-08-16
@sweir27 => Connect TransactionSummary and refactor Summary

ds300:
Sorry about the big PR 😬

I refactored the `Summary` component which was previously `TransactionReview` + a private component `Helper`. Made `Helper` public and used them both inline where previously Summary was used alone.

image

I did that because I need to use TransactionReview on it's own in this PR, and in the next PR I need to make this happen:

image

Which uses the `Helper` but not the `TransactionReview`

So it just felt cleaner to separate these components out and reassemble them in main page components.


artsyit:

Built with commit 970e71e2c901f7703c4768eed7eeaf7db491ab1b

https://deploy-preview-1143--artsy-reaction.netlify.com

src/Apps/Order/Components/Helper.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const Helper: React.SFC<HelperProps> = ({ artworkId }) => (
+  <WithMediator>
+    {mediator => (
+      <>
+        <Sans size="2" color="black60">
+          Have a question?{" "}
+          <Link
+            onClick={() => {
+              mediator && mediator.trigger("openOrdersBuyerFAQModal")
```
damassi: Can `WithMediator` supply default props so that guards are no longer necessary? 

(One thing to note is that we will likely be migrating away from `<Subscribe>` soon and instead update ArtsyContext (located in `Components/Artsty`) to support the renderProps pattern; at that point `WithMediator` will become unnecessary, and `<Subscribe>`, if necessary, would be used exclusively for shared state changes over time involving components spread across a tree -- i.e., clicking a login button three levels down then updates a status in some header, etc.)

ds300: > Can WithMediator supply default props so that guards are no longer necessary?

Good idea. Done :+1:

src/Apps/Order/Components/__tests__/Helper.test.tsx
```diff
@@ -1,20 +1,21 @@
...
+import { Helper } from "../Helper"
 
 describe("Order summary", () => {
   const mediatorMock = {
     trigger: jest.fn(),
   }
-  it("renders the summary properly", () => {
+  it("renders the helper properly", () => {
     const appState = new AppState({ mediator: mediatorMock })
+    console.dir(appState)
```
sweir27: Do we need this?

src/Router/WithMediator.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import { AppState } from "Router/state"
+import { Subscribe } from "unstated"
+
+type Mediator = {
+  trigger: (action: string, config?: object) => void
+}
+
+export const WithMediator: React.SFC<{
```
sweir27: 🙏

src/Apps/Order/Components/__tests__/Helper.test.tsx
ds300: whoops 😬nope, deleted.

src/Router/WithMediator.tsx
```diff
@@ -0,0 +1,23 @@
...
+  children: (mediator: Mediator) => React.ReactNode
+}> = ({ children }) => (
+  <Subscribe to={[AppState]}>
+    {({ state: { mediator } }) =>
+      children(
+        mediator || {
+          trigger() {
+            console.error("mediator is not present in AppState")
```
damassi: 👍 Nice. A good habit that has helped us time and again across the Reaction / Force / Sentry divide is to log errors with a little file-location metadata, like `console.error("Router/WithMediator mediator is not present in AppState")`. Not strictly necessary here but just a future thought.

damassi: Yeah our source-map story is still broken in force :( though we do have uploads working -- what's missing is the associated sourcemap-to-release id, as well as needing to replace our very old in-house [fingerprinting lib](https://www.npmjs.com/package/bucket-assets) with webpack. Sourcemaps are 404 in prod since webpack isn't aware of bucket-assets rewriting filenames.)

ds300:
This should be 🍏 – had to make a couple of last-minute changes so will require another looksie :eyes:.

src/Apps/__stories__/OrderApp.story.tsx
```diff
@@ -11,33 +12,9 @@ const mock = {
   }),
   Order: (_, { id, ...others }) => {
     return {
+      ...IncompleteOrder,
```
sweir27: #minor `Incomplete` is a little ambiguous-- does this refer to an order with no shipping? No payment details? Or just one that hasn't been submitted yet?

sweir27:
Made a minor comment but it's not blocking. This looks good to go! Thanks for following up on `shippingOrigin`.

PR #1149, opened 2018-08-17
@alloy => [Router] Refactor / Cleanup / Replace deprecated React context with React.createContext

damassi:
Refactors router around a few items that have shaken out during use:
- [x] Update `buildClient/ServerRouter` config to be more intuitive 
- [x] For static config (`system`) items, move out of `<Subscribe>` and into `ArtsyContext` and swap
- [x] Add (additional) RenderProps version of ArtsyContext
- [x] Move `/Analytics` and `/Router` and `/Relay` to `/Artsy` folder
- [x] **Squash WIP commits**
- [x] **Update Force consuming apps**
  - [x] `/artwork2`
  - [x] `/artist2`
  - [x] `/collect2`
  - [x] `/order2`
  - Corresponding PR here: https://github.com/artsy/force/pull/2823

Any other changes / requests? cc @zephraph / @sweir27 / @l2succes / @mzikherman / @ds300 (since y'all have all used it) 

**Note:** In a few places the new React Context is not compatable with enzyme and snapshots. A Fix for this will be released in React 16.4.5. See https://github.com/facebook/react/issues/13150#issuecomment-411134477 for more info. 

#### Example

```jsx
const { ClientApp } = await buildClientRouter({
  context: {
    user,
    mediator,
    etc
  },
  history: {
    protocol: 'browser',
    options: {
      ...
    }
  },
  routes: [
    {
      path: '/',
      Component: () => {
        return (
          <Artsy.Consumer>
            {({ user, mediator }) => {
              ...
            }} 
          </Artsy.Consumer>
        )
      }
    }
  ]
})
```


artsyit:

Built with commit f9e27280b90da492df3b28b8ebed9ba610649f73

https://deploy-preview-1149--artsy-reaction.netlify.com

alloy:
Thanks for starting this 👍

damassi:
Np -- it feels good to do some cleanup :) (I might need to ping for some typing help once this is a bit further along but not quite yet)

src/Components/Artsy2.tsx
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Creates a new Context.Consumer and validates that a Context.Provider has
+ * been created.
+ */
+export const ContextConsumer: SFC<ArtsyConsumerProps> = ({ children }) => {
+  return (
+    <Context.Consumer>
+      {props => {
+        // FIXME: Find way to exclude _isNested from check
+        if (!(props as any)._isNestedInProvider) {
```
damassi: @alloy - re typing, specifically this prop. I noticed in the other `Artsy` component that there's a [validation map](https://github.com/damassi/reaction/blob/02208759b96be69311e6ac3934c69cefb92532f5/src/Components/Artsy.tsx#L44-L45) that is implemented on the Class but not too sure how that would work here (if indeed that's what's allowing this prop to pass transparently through).

src/Components/Artsy2.tsx
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Creates a new Context.Consumer and validates that a Context.Provider has
+ * been created.
+ */
+export const ContextConsumer: SFC<ArtsyConsumerProps> = ({ children }) => {
+  return (
+    <Context.Consumer>
+      {contextValues => {
+        // FIXME: Find way to exclude _isNested from check
+        if (!(contextValues as any)._isNestedInProvider) {
```
damassi: @alloy - re typing, specifically this prop. I noticed in the other `Artsy` component that there's a [validation map](https://github.com/damassi/reaction/blob/02208759b96be69311e6ac3934c69cefb92532f5/src/Components/Artsy.tsx#L44-L45) that is implemented on the Class but not too sure how that would work here (if indeed that's what's allowing this prop to pass transparently through).

damassi: I'll validate that a React warning is thrown and if so will do 👍

src/Components/Artsy2.tsx
```diff
@@ -0,0 +1,66 @@
...
+          return children(contextValues)
+        }
+      }}
+    </Context.Consumer>
+  )
+}
+
+/**
+ * A HOC utility function for injecting renderProps into a component.
+ */
```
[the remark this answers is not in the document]
damassi: This is for backwards compatibility so that our older code doesn't break. Cool?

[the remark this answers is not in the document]
alloy: Ah, but it doesn’t follow the exact same API, which is why I didn’t pick up on that.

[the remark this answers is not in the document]
damassi: 👍 All of the original tests passed just fine with me only replacing `react-test-renderer` with Enzyme, so all good.

src/Router/__test__/buildClientApp.test.tsx
```diff
@@ -19,26 +22,101 @@ describe("buildClientApp", () => {
...
+            expect(Object.keys(system)).toEqual([
+              "history",
+              "routes",
+              "context",
+              "relayEnvironment",
+              "resolver",
+              "currentUser",
+            ])
+            expect(Object.keys(system.context)).toEqual(["mediator", "user"])
+            done()
```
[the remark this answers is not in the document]
damassi: Like, 

```
it("injects default renderProps", done => {
    render(
      <Artsy.ContextProvider>
        <Artsy.ContextConsumer>
          {props => {
            setImmediate(() => {
              expect(Object.keys(props).sort()).toEqual([
                "_isNestedInProvider",
                "currentUser",
                "relayEnvironment",
              ])
              done()
            })
            return <div />
          }}
        </Artsy.ContextConsumer>
      </Artsy.ContextProvider>
    )
  })
```
?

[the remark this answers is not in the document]
alloy: I was thinking `setImmediate(done)`, but this probably works too.

My reasoning is this; you were telling the test framework that you were done,  when in fact you were _still_ inside the current frame (the function scope) and caller sites may still decide to do work when control returns to them. I think following the semantic reasoning that you call done _outside_ the ongoing call execution isn’t only semantically more correct, it may by chance even save you from annoying bugs. (Annoying in that they are hard to debug as they leak into other concurrent code and avoiding them altogether is always a win.)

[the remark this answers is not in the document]
damassi: This makes sense, will update 👍Better to be safe.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -28,13 +28,8 @@ export class ArtistAuctionResultItem extends Component<Props> {
...
     return (
-      <Subscribe to={[AppState]}>
-        {({ state }) => {
-          const {
-            mediator,
-            system: { currentUser },
-          } = state
-
+      <ContextConsumer>
+        {({ currentUser, mediator }) => {
```
damassi: @ds300 - I removed `WithMediator` in this PR since `<ContextConsumer>` will be the new form, and things passed into `context` upon init will be accessible here

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -5,13 +5,12 @@ import * as Schema from "Analytics/Schema"
 import FollowArtistButton from "Components/FollowButton/FollowArtistButton"
 import React, { Component, Fragment } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
-import { AppState } from "Router/state"
+import { ContextConsumer } from "Router/Artsy2"
```
damassi: @alloy - my idea is to adjust reactions folder structure a bit so that things will look like 

```
/Artsy
  /Analytics
    ...
  /Router
    ...
  ArtsyContext.tsx
  index.tsx
```
Thoughts? It seems like placing this file under `/Components` might be a little less intuitive at this point in time.

src/Router/index.ts
```diff
@@ -1,5 +1,50 @@
...
+  initialRoute?: string
+  /**
+   * Array of routes to be passed to Found
+   */
+  routes: RouteConfig
+  /**
+   * URL passed from server
+   */
+  url?: string
+}
```
damassi: `Router/index.tsx` seemed the best place to put this interface

typings/found.d.ts
```diff
@@ -11,3 +11,11 @@ declare module "found" {
...
   const ResolverUtils: ResolverUtils
 }
+
+declare module "farce" {
+  type HistoryProtocol = "browser" | "hash" | "memory"
+
+  interface HistoryOptions {
+    useBeforeUnload?: boolean
+  }
+}
```
damassi: Since these two modules go together I kept them in the same file, but can break out into `farce.d.ts` if this seems weird.

src/Utils/getUser.ts
```diff
@@ -0,0 +1,22 @@
+export function getUser(user: User): User {
+  let currentUser = user
+
+  if (currentUser === undefined) {
```
alloy: Seeing as this is valid, you should type the function parameter appropriately (as optional).

src/Utils/getUser.ts
```diff
@@ -0,0 +1,22 @@
+export function getUser(user: User): User {
```
alloy: Is this function never called for unauthenticated users? Because I would expect it to return `null` in such a case.

src/Router/Artsy2.tsx
```diff
@@ -0,0 +1,48 @@
+import { ContextProps as _ContextProps } from "Components/Artsy"
+import React, { SFC } from "react"
+import { getUser } from "Utils/getUser"
+import { createEnvironment } from "../Relay/createEnvironment"
+
+// FIXME: Move from Artsy to Artsy2
+export interface ContextProps extends _ContextProps {
+  // FIXME: Type properly
+  mediator?: any
+  system?: any
```
alloy: What is this?

src/Router/Artsy2.tsx
```diff
@@ -0,0 +1,48 @@
+import { ContextProps as _ContextProps } from "Components/Artsy"
+import React, { SFC } from "react"
+import { getUser } from "Utils/getUser"
+import { createEnvironment } from "../Relay/createEnvironment"
+
+// FIXME: Move from Artsy to Artsy2
+export interface ContextProps extends _ContextProps {
+  // FIXME: Type properly
+  mediator?: any
```
alloy: Please do type this appropriately.

alloy: And is it really optional? Seems like it’s something all apps depend on, no?

src/Router/Artsy2.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const providerValues = {
+    ...props,
+    currentUser,
+    relayEnvironment,
+  }
+
+  return <Context.Provider value={providerValues}>{children}</Context.Provider>
+}
+
+export const ContextConsumer = Context.Consumer
```
alloy: Is this missing typings for the callsite of the consumer or do they come in on line 13?

src/Router/Artsy2.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return props => {
+    return (
+      <ContextConsumer>
+        {contextValues => {
+          return <Component {...contextValues} {...props} />
+        }}
+      </ContextConsumer>
+    )
+  }
+}
```
alloy: I think the types for this context need to be adjusted such that some of them are optional for users of the provider but not for users of the consumer. E.g. a relay env can optionally be passed to the provider, but a consumer will always have one.

alloy:
Looking great thus far 👏

Requested changes are mostly around types. Let me know if you want to pair on any of that.

src/Router/Artsy2.tsx
damassi: This is passed through fr line 13: 

screen shot 2018-08-23 at 12 56 40 pm

src/Router/Artsy2.tsx
damassi: In the future we should be able to phase it out, but for the time being it does seem necessary since thats how we trigger logins and so on. Will update.

src/Router/Artsy2.tsx
damassi: Removed

src/Router/index.ts
```diff
@@ -1,5 +1,50 @@
...
+export interface RouterConfig {
+  /**
+   * Context values to be passed to ArtsyContext
+   */
+  context?: {
+    user?: User
+    relayNetwork?: RelayNetwork
+    initialMatchingMediaQueries?: MatchingMediaQueries
+    [key: string]: any
+  }
```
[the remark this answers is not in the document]
damassi: @alloy - this actually is correct -- we're type-hinting values that can optionally (or will likely) be passed while also allowing for the user to pass whatever they need for their app, e.g., 

```jsx
const { ServerApp } = await buildServerApp({
  ...
  context: {
    someRandomComponent
  }
})

File: src/Router/Artsy2.tsx (hunk 14 above)
Comment (damassi) [reply src/Router/Artsy2.tsx]: How should I express this?

Comment (alloy) [reply src/Router/Artsy2.tsx]: Looking at the context typings you’ve shown above:

screen shot 2018-08-23 at 12 56 40 pm

It seems like you’ll need to separately use the `Provider` and the `Consumer` types. E.g.

```tsx
const Context = React.createContext({})

export const Provider: React.Provider<{ mediator: Mediator, relayEnvironment?: Relay.Environment }> = Context.Provider

export const Consumer: React.Consumer<{ mediator: Mediator, relayEnvironment: Relay.Environment }> = Context.Consumer
```

----

Seeing as repeating the list could get cumbersome if they are identical except for some of them being optional in the provider interface, you could do something like the following to only make a subset optional:

```ts
type X = "foo" | "bar"

interface Foo {
    foo: string
    bar: string
    baz: string
}

type PartialSubset<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>> & Partial<Pick<T, K>>

type Y = PartialSubset<Foo, "foo" | "bar">

const y: Y = { baz: "baz" }
```

src/Router/index.ts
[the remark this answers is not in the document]
alloy: I don’t think we should do that, as it doesn’t help clearly expressing those individual app’s needs in the public interface. Instead we should probably add some generics to it for this purpose.

Do we have an actual example right now that we could model? If not, let’s just wait until we do, but remove the `any` part.

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -330,13 +330,13 @@ class Filter extends Component<Props> {
...
                               // Desktop
-                              <Sidebar width="30%" mr={2}>
+                              <Sidebar width="25%" mr={2}>
                                 <Filters />
                               </Sidebar>
                             )}
 
                             {/* Main Artwork Grid */}
-                            <Box width={xs ? "100%" : "70%"}>
+                            <Box width={xs ? "100%" : "75%"}>
```
alloy: Why are these changes being made here? To me they don’t really seem a part of what this PR is about.

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -95,7 +98,7 @@ export function buildServerApp(config: RouterConfig): Promise<Resolve> {
         status,
       })
     } catch (error) {
-      console.error("[Reaction Router/buildServerApp] Error:", error)
+      console.error("[Artsy/Router/buildServerApp] Error:", error)
```
alloy: At some point we should look at just having webpack encode the values for us based on `__filename`, so that we don’t need to think about it too much and can just move it around.

src/Utils/getUser.ts
alloy: Nit: It’s more idiomatic for parameters to use `?` to indicate optionality. E.g.

```ts
export function getUser(user?: User | null): User | null {
 // …
}
```

Btw, is it expected that when passing in `null` as `user` you get back `null`? Or was the intent to have `currentUser === undefined || currentUser === null` ?

alloy:
Looking good! The only remaining two things to address imo are:

* 
Removing the `any` part from the `context` parameter to `build*App` https://github.com/artsy/reaction/pull/1149#discussion_r212573511
* 
Typing the context provider and consumer different based on optional inputs vs non-optional outputs https://github.com/artsy/reaction/pull/1149#discussion_r212572499

src/Artsy/Router/buildServerApp.tsx
damassi: Ahh this is really interesting, haven't thought about how we can leverage this at build time and inject things right in.

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
damassi: This is a little thing Brian and I were chatting about in Slack but your right, will remove 👍

typings/found.d.ts
alloy: Seems fine to me 👍

Just try to PR upstream ASAP!

damassi:
Corresponding PR located here: https://github.com/artsy/force/pull/2823

I'm going to merge and follow up with a second PR around adding the generic type for context.

PR #1151, opened 2018-08-17
hook up ShippingAndPaymentDetails

ds300:
I bumped metaphysics to the latest version and synced the schema because I needed the `Order#shippingName` field introduced in https://github.com/artsy/metaphysics/pull/1227

I also added a missing state for the Shipping details component. The 'pick up' state:
image

We should get Brian to do a copy review on Monday.

### DIFF
--- a/src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
@@ -56,7 +56,7 @@ export const ShippingAndPaymentDetails: React.SFC<
           <Serif size="3" weight="semibold" color="black100">
             Pick up
           </Serif>
-          <Serif size="3" color="black100">
+          <Serif size="3t" color="black100">
             You’ll be appointed an Artsy specialist within 2 business days to
             handle pickup logistics.
           </Serif>
@@ -66,7 +66,7 @@ export const ShippingAndPaymentDetails: React.SFC<
           <Serif size="3" weight="semibold" color="black100">
             Ship to
           </Serif>
-          <Serif size="3" color="black100" style={{ whiteSpace: "pre-wrap" }}>
+          <Serif size="3t" color="black100" style={{ whiteSpace: "pre-wrap" }}>
             {renderAddress(addressProps)}
           </Serif>
         </>
--- a/src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
@@ -35,7 +35,7 @@ const renderAddress = ({
     cityLine,
     shippingRegion,
   ]
-    .filter(x => x)
+    .filter(Boolean)
     .join("\n")
 }
 
--- a/src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
@@ -45,6 +45,7 @@ export const ShippingAndPaymentDetails: React.SFC<
   order: {
     fulfillmentType,
     creditCard: { brand, last_digits, expiration_year, expiration_month },
+    lineItems,
     ...addressProps
   },
   ...others
@@ -54,7 +55,7 @@ export const ShippingAndPaymentDetails: React.SFC<
       {fulfillmentType === "PICKUP" ? (
         <>
           <Serif size="3" weight="semibold" color="black100">
-            Pick up
+            Pick up ({lineItems.edges[0].node.artwork.shippingOrigin})
           </Serif>
           <Serif size="3t" color="black100">
             You’ll be appointed an Artsy specialist within 2 business days to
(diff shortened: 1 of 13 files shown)


src/Apps/__test__/Fixtures/Order.ts
```diff
@@ -1,4 +1,4 @@
-export const IncompleteOrder = {
+export const UntouchedOrder = {
```
ds300: Renamed this based on a suggestion from @sweir27 https://github.com/artsy/reaction/pull/1143#discussion_r210924336

artsyit:

Built with commit 9511abb951fb3ce1d7ee0e9d167d78cb40f1046e

https://deploy-preview-1151--artsy-reaction.netlify.com

src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
```diff
@@ -1,38 +1,107 @@
...
+  shippingCity,
+  shippingPostalCode,
+  shippingRegion,
+}: {
+  shippingName: string
+  shippingAddressLine1: string
+  shippingAddressLine2: string | null
+  shippingCity: string
+  shippingPostalCode: string | null
+  shippingRegion: string | null
```
damassi: Might these types be better served as an interface?

src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
```diff
@@ -1,38 +1,107 @@
...
+  const cityLine =
+    shippingCity + (shippingPostalCode ? ` ${shippingPostalCode}` : "")
+  return [
+    shippingName,
+    shippingAddressLine1,
+    shippingAddressLine2,
+    cityLine,
+    shippingRegion,
+  ]
+    .filter(x => x)
```
damassi: Could also do `.filter(Boolean)`

src/Styleguide/Elements/icons/CreditCardIcon.tsx
```diff
@@ -1,28 +1,29 @@
 import React from "react"
 
+// https://stripe.com/docs/api#card_object-brand
```
damassi: 👍

damassi:
This code looks really solid and idiomatic -- I'm impressed by how fast you've dove in and figured out our stack 💯

src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
alloy: I wonder if maybe the `Order` type should have a `fulfillment` field which returns a union of a `Pickup` or a `Shipping` type. That way the `Order` type doesn’t get overloaded with all these `shippingFoo` fields _and_ when e.g. the type is `Pickup` you don’t end up with all these `shippingFoo` fields in the response data.

ashkan18: interesting point, im not too familiar with union type in GraphQL, MP work here is pretty minimal and mostly copies Exchange's schema, if we want to do this i'd suggest lets do it in Exchange and we can use/copy it here in MP too.

alloy: (This does not block this PR btw.)

src/Apps/Order/Routes/Submission/index.tsx
```diff
@@ -50,11 +50,7 @@ export class SubmissionRoute extends Component<SubmissionProps> {
             }
             Sidebar={
               <Flex flexDirection="column">
-                <Placeholder
-                  height="180px"
-                  name="Shipping and payment details"
-                  mb={xs ? 2 : 3}
-                />
+                <ShippingAndPaymentDetails order={order} mb={xs ? 2 : 3} />
```
ds300: By actually reading the docs I discovered that these styled-system props can be specified with arrays to get responsive styles. So `xs ? 2 : 3` would be just `[2, 3]` I think I'll fix all of these up in a separate PR.

src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
```diff
@@ -1,38 +1,117 @@
...
-interface ShippingAndPaymentDetailsProps {
-  address: string
-  creditCardLast4digits: string
-  creditCardtype: CreditCardType
-  creditCardExpiry: string
+interface ShippingAndPaymentDetailsProps extends FlexProps {
+  order: ShippingAndPaymentDetails_order
+}
+
+const renderAddress = ({
```
sweir27: This can come in a separate PR, but it would be nice to add some tests for this screen (especially with the logic around rendering the address/conditionally showing the pickup option).

sweir27:
This is awesome @ds300! 🎉

src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
ashkan18: https://github.com/artsy/exchange/pull/119

PR #1152, opened 2018-08-17
Abstract out the shipping form fields into an address form component

zephraph:
@erikdstock and I are working on the payment step so we'd like to pull this out of shipping so that we can reuse it for billing address. The types aren't finalized, but we can update that in another PR.


artsyit:

Built with commit 339d80727926b31e820889c817431b31de157a7e

https://deploy-preview-1152--artsy-reaction.netlify.com

zephraph:
Certainly open for refactoring as we move forward. Right now we just want to get the forms displaying on the payment page. It should be a little easier to refactor with the form in one place.

ashfurrow:
Sounds good!

PR #1155, opened 2018-08-20
RadioGroup fixes

ds300:
We refactored the RadioGroup last week and accidentally broke some stuff. Here's what I did:

1. Fixed button–text alignment 

image

2. Removed outer padding, which should be provided by parents in most cases.

2. Fixed hit area for bordered radio
3. Added back the hover state for bordered radio

Before:
bad

After:
good

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -12,8 +12,8 @@ import { Collapse } from "Styleguide/Components/Collapse"
 import { Button } from "Styleguide/Elements/Button"
 import { Flex } from "Styleguide/Elements/Flex"
 import { Col, Row } from "Styleguide/Elements/Grid"
-import { Radio } from "Styleguide/Elements/Radio"
-import { BorderedRadioGroup } from "Styleguide/Elements/RadioGroup"
+import { BorderedRadio } from "Styleguide/Elements/Radio"
+import { RadioGroup } from "Styleguide/Elements/RadioGroup"
 import { Spacer } from "Styleguide/Elements/Spacer"
 import { Responsive } from "Utils/Responsive"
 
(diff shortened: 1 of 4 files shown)


src/Styleguide/Elements/RadioGroup.tsx
```diff
@@ -1,25 +1,15 @@
...
 
-export interface RadioGroupProps {
+export interface RadioGroupProps extends FlexProps {
   disabled?: boolean
   onSelect?: (selectedOption: string) => void
   defaultValue?: string
   children: Array<React.ReactElement<RadioProps>>
 }
 
-export interface RadioGroupToggleProps
```
ds300: Deleted this because it's not used anywhere and doesn't seem to be useful for anything.

src/Styleguide/Elements/RadioGroup.tsx
```diff
@@ -63,22 +53,11 @@ export class RadioGroup extends React.Component<
   }
 
   render() {
+    const { disabled, onSelect, defaultValue, children, ...others } = this.props
```
ashfurrow: Just to verify I understand this, we have assigned `others` to the values in `this.props` except for `disabled`, `onSelect`, etc? This is a really cool pattern! Does it have a name?

ds300:
I'd be happy to!

src/Styleguide/Elements/RadioGroup.tsx
damassi: 😄

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -115,15 +115,17 @@ export class ShippingRoute extends Component<
...
                 <>
-                  <BorderedRadioGroup
+                  <RadioGroup
                     onSelect={shippingOption =>
                       this.setState({ shippingOption })
                     }
                     defaultValue="SHIP"
                   >
-                    <Radio value="SHIP">Provide shipping address</Radio>
+                    <BorderedRadio value="SHIP">
```
damassi: 👍

PR #1156, opened 2018-08-20
adds v1 of review page with dynamic components

sweir27:
This starts to add components to the "review" page.

No major changes except the addition of a `TermsOfServiceCheckbox` which is different enough from the current one that we use in the auth modal that I thought it deserved its own.

image

I left the `Step Summary Item` placeholders there until https://artsyproduct.atlassian.net/browse/PURCHASE-384 is done!

### DIFF
--- a/src/Apps/Order/Components/ItemReview.tsx
+++ b/src/Apps/Order/Components/ItemReview.tsx
@@ -4,7 +4,7 @@ import { Serif } from "@artsy/palette"
 import { ItemReview_artwork } from "__generated__/ItemReview_artwork.graphql"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
-import { StackableBorderBox } from "Styleguide/Elements/Box"
+import { BorderBox } from "Styleguide/Elements/Box"
 import { Flex } from "Styleguide/Elements/Flex"
 
 interface ItemReviewProps {
@@ -34,7 +34,7 @@ export const ItemReview: React.SFC<ItemReviewProps> = ({
     },
   },
 }) => (
-  <StackableBorderBox>
+  <BorderBox>
     <Flex flexGrow={1} flexDirection="column">
       <Serif size="2" weight="semibold" color="black60">
         {artist_names}
@@ -62,7 +62,7 @@ export const ItemReview: React.SFC<ItemReviewProps> = ({
     <ImageBox>
       <img alt={`${title} by ${artist_names}`} src={url} />
     </ImageBox>
-  </StackableBorderBox>
+  </BorderBox>
 )
 
 export const ItemReviewFragmentContainer = createFragmentContainer(
--- a/src/Apps/Order/Components/ItemReview.tsx
+++ b/src/Apps/Order/Components/ItemReview.tsx
@@ -34,7 +34,7 @@ export const ItemReview: React.SFC<ItemReviewProps> = ({
     },
   },
 }) => (
-  <BorderBox>
+  <BorderBox p={[2, 3]}>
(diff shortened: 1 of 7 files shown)


src/Apps/Order/Components/ItemReview.tsx
```diff
@@ -34,7 +34,7 @@ export const ItemReview: React.SFC<ItemReviewProps> = ({
     },
   },
 }) => (
-  <StackableBorderBox>
+  <BorderBox>
```
sweir27: This is #minor but I am pretty sure every place we show the `ItemReview` it will be encased in a border, so we don't need the `StackableBorderBox`.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -16,11 +18,34 @@ import { TwoColumnLayout } from "../../Components/TwoColumnLayout"
...
+
+  updateTermsCheckbox() {
+    const { termsCheckboxSelected } = this.state
+    this.setState({
+      termsCheckboxSelected: !termsCheckboxSelected,
+    })
+  }
+
+  onOrderSubmitted() {
+    this.props.router.push(`/order2/${this.props.order.id}/submission`)
```
sweir27: cc @ashfurrow I think you can add something similar to this in your mutation handler to advance to the next step.

artsyit:

Built with commit b0c6a0676dea164cbeedb33e082dcb86237f417a

https://deploy-preview-1156--artsy-reaction.netlify.com

src/Apps/Order/Components/ItemReview.tsx
ds300: Makes sense :+1: but BorderBox doesn't have responsive padding by default, so let's add it back with `p={[2, 3]}`

src/Apps/Order/Components/TermsOfServiceCheckbox.tsx
```diff
@@ -0,0 +1,22 @@
...
+          Terms & Conditions
+        </A>
+        {". All sales are final."}
+      </Serif>
+    </Checkbox>
+  )
+}
+
+const A = styled.a`
+  color: ${props => props.color};
```
ds300: This doesn't seem to be used?

src/Apps/Order/Components/TermsOfServiceCheckbox.tsx
```diff
@@ -0,0 +1,22 @@
+import { color, Serif } from "@artsy/palette"
+import React from "react"
+import styled from "styled-components"
+import { Checkbox } from "Styleguide/Elements/Checkbox"
+
+export const TermsOfServiceCheckbox = props => {
+  return (
+    <Checkbox {...props}>
+      <Serif size="3" color={color("black60")}>
```
ds300: I think you don't need the inner `color` wrapping here. It can just be `color="black60"`.

src/Apps/Order/Routes/__tests__/Review.test.tsx
```diff
@@ -0,0 +1,47 @@
+import { mount } from "enzyme"
+import React from "react"
+
+import { IncompleteOrder } from "Apps/__test__/Fixtures/Order"
```
ds300: This was renamed to `UntouchedOrder` in master, you might need to rebase or merge.

src/Apps/Order/Components/ItemReview.tsx
sweir27: Oooh! Didn't realize that, thanks! 🙌

sweir27:
I think I've addressed everything! Thanks for the comments 🎉

PR #1157, opened 2018-08-20
@sweir27 => Use StepSummaryItem in ShippingAndPaymentDetails

ds300:
I added a kind of 'Review' mode for ShippingAndPaymentDetails so it can be used on the Review page as well as the Status page.

image

The main thing I'm worried about is that the spacing below the title is slightly larger in StepSummaryItem, but I reckon that looks better and they should probably be the same anyway? @briansw what do you think?

image

i.e. we're consolidating the title margin from these two very similar components: 
image

Also I'm not crazy about the resulting API, it's not very explicit, but it's also not too odd. I wouldn't feel bad about it being merged, but seems like there should be a better solution. Let me know if you think of anything.

Happy to treat this PR as WIP if these issues need to be resolved.


artsyit:

Built with commit 9dc9111d74e0ed566ce2894d638dd8e4bfb5ea59

https://deploy-preview-1157--artsy-reaction.netlify.com

src/Apps/Order/Components/ShippingAndPaymentDetails.tsx
```diff
@@ -51,38 +55,37 @@ export const ShippingAndPaymentDetails: React.SFC<
...
+    ) : (
+      <StepSummaryItem onChange={onShippingChange} title="Ship to">
+        <div style={{ whiteSpace: "pre-wrap" }}>
+          {renderAddress(addressProps)}
+        </div>
+      </StepSummaryItem>
+    )}
+    <StepSummaryItem
+      onChange={onPaymentChange}
+      title={onPaymentChange ? "Payment method" : null}
```
sweir27: It feels a little weird to infer whether or not to show the title based on the presence of `onPaymentChange`. Maybe we can have a separate prop for hiding the title so we are explicit? `hidePaymentTitle` or something like that?

ds300:
@sweir27 More differences cropped up (sans vs serif, review copy, line spacings) and I felt really gross adding all these exceptions so I split the component into two and the API seems a lot cleaner to me now. Let me know what you think.

ds300:
OK done, I think. The PR got a bit bigger than originally intended :D

sweir27:
👍 this looks good! Looks like it needs a rebase but otherwise feel free to merge.

src/Apps/Order/Components/ShippingAndPaymentReview.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ShippingAndPaymentReview,
+  graphql`
+    fragment ShippingAndPaymentReview_order on Order {
+      fulfillmentType
+      shippingName
+      shippingAddressLine1
+      shippingAddressLine2
+      shippingCity
+      shippingPostalCode
+      shippingRegion
```
alloy: Didn’t this part of the schema get updated? cc @ashkan18

ds300: yep seems like we haven't updated reaction in response to the metaphysics shipping refactor yet.

PR #1163, opened 2018-08-21
[Artwork] Add fallback for when users don’t use a Responsive provider.

alloy:
@mzikherman noticed that the save button wasn’t showing up on e.g. https://www.artsy.net/gene/animals, which turns out to be because the new pattern of using the `Responsive` component requires wrapping the tree in a `ResponsiveProvider`, but that’s not something that’s done in the existing `Artsy` wrapper component.

 Ideally all places in Force that use components from Reaction the required setup would be done, but that’s something that requires a bit more thinking first. /cc @damassi


src/Components/Artwork/GridItem.tsx
```diff
@@ -150,7 +150,10 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
                   <Image src={this.getImageUrl(breakpoints)} />
                 </a>
                 {hasBuyNowLabFeature && this.renderArtworkBadge(artwork)}
-                {hover && (
+                {/* The undefined check is a fallback for Force code that uses
+                    Reaction code without wrapping the tree in a Responsive
+                    provider component. */}
+                {(hover === undefined || hover) && (
```
alloy: It would be `true` or `false` otherwise.

artsyit:

Built with commit c78a80b91dc86f906477db9a672dd225e2cc3b7d

https://deploy-preview-1163--artsy-reaction.netlify.com

damassi:
@alloy - Indeed, all places that use our new component work should be using `Boot` or some other respective provider. I'll raise this tomorrow in #dev so that everyone is on the same page.

PR #1170, opened 2018-08-22
add display of artwork condition info

oxaudo:
Add condition info to storybook. Still need to add other additional info there. 

The idea is that we'll have ArtworkDetails that will hold tabs with all the data we are interested in.

For now to make it easier I was thinking maybe just put ArtworkDetailsConditionInfo directly into force?


oxaudo:
Looks like that:
screen shot 2018-08-22 at 11 58 21 pm

artsyit:

Built with commit aa06c7d302cd1aa5d16a9f040d08e86f492cc53c

https://deploy-preview-1170--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsConditionInfo.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export interface ArtworkDetailsConditionInfoProps {
+  artwork: ArtworkDetailsConditionInfo_artwork
+}
+
+export class ArtworkDetailsConditionInfo extends React.Component<
+  ArtworkDetailsConditionInfoProps
+> {
+  renderRow(label: string, details: string) {
+    if ((!label || label.length < 1) && (!details || details.length < 1)) {
```
mzikherman: I think empty strings would be not truthy, so you can probably leave off the `.length` check.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsConditionInfo.tsx
```diff
@@ -0,0 +1,92 @@
...
+      return null
+    }
+    return (
+      <Box>
+        <Box display="inline" mr={1}>
+          <Checkmark stroke={colors.black} />
```
mzikherman: ✔️

mzikherman:
LGTM! 😎 
Merge at will.

Re: the tabs, are you thinking to rebuild that section here in Reaction, and then (probably behind a lab feature), swap it into the existing artwork page?

oxaudo:
@mzikherman - I think I'll do just about the artwork tab content as a separate component and try to put this in and then deal with the rest.

oxaudo:
Latest:
screen shot 2018-08-23 at 8 30 04 pm

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -22,22 +22,17 @@ describe("Shipping", () => {
     )
   }
 
-  let props: ShippingProps
+  let testProps: ShippingProps
```
oxaudo: @mzikherman, @damassi - Lint would not let me commit because of the shadowed props error. Was wondering if it was supposed to be repressed locally but ended up renaming vars in test since I think it actually improves test readability. Let me know if you think otherwise.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWork.tsx
```diff
@@ -0,0 +1,35 @@
+import { Serif } from "@artsy/palette"
```
mzikherman: If this is already in an `ArtworkDetails` folder, it can probably just be `AboutTheWork.tsx` (ie- drop the prefix).

mzikherman:
This LGTM, with the little nit about the naming (redundant `ArtworkDetails` prefixes). 😎 

Re: the lint rule, I'm not sure! I would defer to @damassi on the desired behavior there. It's possible we weren't running lint in test before or something?

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWork.tsx
oxaudo: I *believe* we specifically added the full prefixes when @alloy was doing data for ArtworkSidebar.  This way it is easier to correctly understand which data fragment goes with which component. Though I might be misremembering something. @alloy - could you please confirm the proper naming here?

alloy: Correct 👍 https://artsy.slack.com/archives/CAZG82X4G/p1530019149000063

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
damassi: Shadowed vars oftentimes use an `_` but I'm not particular about this in tests.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -69,8 +69,8 @@ export const ArtworkApp: React.SFC<Props> = props => {
           </Tabs>
         </Col>
       </Row>
-      <Row mb={6}>
-        <Col xl={8} lg={8} md={8} sm={12} xs={12}>
+      <Row>
+        <Col xl={8} lg={8} md={8} sm={8} xs={12}>
```
damassi: @oxaudo - Idiomatically, breakpoints like this should go from smallest to largest (left to right) and note that by default it's 12, and a breakpoint set differently will expand up. This is hard to explain in words, but by example it should make sense and can be simplified to just one prop: 

`<Col sm={8} />` 

Does this make sense?

oxaudo: Cool. Did not realize that. This code was there before, but what you are saying makes total sense. Will update with next change.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
oxaudo: Happy to update that as well:)

damassi: Oh no that's ok! This is mostly applies to one-off stuff, but i see you updated in a few places. Its all good 😄

src/Apps/Artwork/ArtworkApp.tsx
damassi: 👍

PR #1179, opened 2018-08-24
Update shipping mutation

ashfurrow:
This PR updates the mutation to match the [recently changed schema](https://github.com/artsy/metaphysics/pull/1239), as well as updating the local schema itself. /cc @ashkan18


artsyit:

Built with commit 04c2a17dbb5240ead4cb17438e052ded0152c93b

https://deploy-preview-1179--artsy-reaction.netlify.com

ashfurrow:
Ah! Forgot to update tests 🙈

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -93,13 +94,15 @@ export class ShippingRoute extends Component<
...
               fulfillmentType: this.state.shippingOption,
-              shippingName: this.state.name || "",
-              shippingAddressLine1: this.state.addressLine1 || "",
-              shippingAddressLine2: this.state.addressLine2 || "",
-              shippingCity: this.state.city || "",
-              shippingRegion: this.state.region || "",
-              shippingCountry: this.state.country || "",
-              shippingPostalCode: this.state.postalCode || "",
+              shipping: {
+                name: this.state.name || "",
```
ashkan18: is it possible to pass `nil` instead of `""` im worried that this will actually set those fields to empty string, i know this was there from before.

orta: Yeah, this would be a better semantic improvement - they're nullable

ashfurrow: Yeah good idea.

ashfurrow: I've added this, but had to keep country pending a purchase team ticket. There's some form validation that's related to this, too, but we'll get there when we get there 😄

PR #1182, opened 2018-08-24
[PATCH refactor] Move relay-mock-network-layer to deps

damassi:
Move `relay-mock-network-layer` to deps to ungate `<StorybooksRouter>` for future use.


artsyit:

Built with commit 392b0731c8cbdcbdf7b4ccd53d01bd2c5c6b84f2

https://deploy-preview-1182--artsy-reaction.netlify.com

alloy:
Are we actually using `StorybooksRouter` anywhere else yet? Because this dependency is now being forced upon consumers, which makes adoption for new projects, like Volt, slightly more cumbersome.

For the future, in general I think that changes like these should get a second opinion rather than being self-merged.

damassi:
@alloy - Its not about Storybooks -- and StorybooksRouter needs to be renamed to something like `ClientRouter` -- it's about the ability to inject data into Relay components without needing https://github.com/artsy/reaction/blob/master/src/Utils/RelayStubProvider.ts, which throws warnings and has perf implications.  This should remain a dep rather than a peerDep -- its part of our framework.

alloy:
The issue was that the only breadcrumbs we could find was this PR, which only mentions storybook, and so it seems like it’s a dev dependency rather than a runtime one. You can sense the confusion mostly in this commit https://github.com/artsy/volt/commit/e8b504ecc9f28ddbfdea9897257624a2d9aaf814

damassi:
Ah I see, the extended commit message. 

`relay-mock-network-layer` is fairly small but it pulls in 

```js
const { graphql, printSchema, buildClientSchema } = require('graphql');
```

anandaroop:
@damassi there were a couple of issues…

1. A failed Webpack build that led me to believe we were importing test-related dependencies from Reaction.

```
13:58:53 webpack.1 | ERROR in ./node_modules/graphql-tools/dist/schemaGenerator.js
13:58:53 webpack.1 | Module not found: Error: Can't resolve 'graphql' in '/Users/roop/Projects/volt/node_modules/graphql-tools/dist'
13:58:53 webpack.1 | resolve 'graphql' in '/Users/roop/Projects/volt/node_modules/graphql-tools/dist'

[...]

13:58:53 webpack.1 |  @ ./node_modules/graphql-tools/dist/mock.js 3:16-34 4:16-34
13:58:53 webpack.1 |  @ ./node_modules/graphql-tools/dist/index.js
13:58:53 webpack.1 |  @ ./node_modules/relay-mock-network-layer/index.js
13:58:53 webpack.1 |  @ ./node_modules/@artsy/reaction/dist/Artsy/Relay/createMockNetworkLayer.js
13:58:53 webpack.1 |  @ ./node_modules/@artsy/reaction/dist/Artsy/Router/Components/StorybooksRouter.js
13:58:53 webpack.1 |  @ ./node_modules/@artsy/reaction/dist/Artsy/Router/index.js
13:58:53 webpack.1 |  @ ./node_modules/@artsy/reaction/dist/Styleguide/Components/RouteTabs.js
13:58:53 webpack.1 |  @ ./node_modules/@artsy/reaction/dist/Styleguide/Components/index.js
13:58:53 webpack.1 |  @ ./node_modules/@artsy/reaction/dist/Styleguide/Components/Stepper.js
13:58:53 webpack.1 |  @ ./app/javascript/components/order_page/OrderPage/index.jsx
13:58:53 webpack.1 |  @ ./app/javascript/packs/order_page.jsx
```

Seeing that, I was confused that the `Stepper` implementation (which we use in Volt) was requiring us to import test-related code as well (as suggested by seeing terms like 'mock' and 'story' in the stacktrace).

I gather from your comment that `StorybooksRouter` name might be a misnomer, which I didn't realize at the time.

---

2. Ultimately, moving `relay-mock-network-layer` to a peerDependency didn't actually solve our problem, because other libraries inside of Reaction also needed `graphql` as a peerDependency. In the end adding that `graphql` dep explicitly to Volt was the expedient route.

Bottom line: if it's deemed necessary to undo #1266, Volt should be fine with that.

damassi:
As we're now getting failing builds in Force I think an undo is needed, though waiting for @alloy's feedback in #frontend.

PR #1183, opened 2018-08-24
Run certain logos and ads through resizer

ansor4:
This addresses GROW-806 and also adds some tests for mockTracking


src/Components/Publishing/Display/Canvas/CanvasText.tsx
```diff
@@ -2,6 +2,7 @@ import { garamond } from "Assets/Fonts"
 import { compact, map } from "lodash"
 import React from "react"
 import styled, { StyledFunction } from "styled-components"
+import { resize } from "../../../../Utils/resizer"
```
damassi: Can use absolute paths here (`Utils/resizer`)

artsyit:

Built with commit 5a2e514ba2be9bebe9042b5471e884c4dbb8d2be

https://deploy-preview-1183--artsy-reaction.netlify.com

src/Components/Publishing/Display/Canvas/CanvasText.tsx
```diff
@@ -16,13 +17,18 @@ interface DivProps extends React.HTMLProps<HTMLDivElement> {
 
 export const CanvasText: React.SFC<CanvasTextProps> = props => {
   const { disclaimer, unit } = props
+  let resized_logo = resize(unit.logo, {
```
damassi: If this variable will never change then best to use `const`. `let` is typically used to signify reassignment at some later point, while `const` will throw an error in the compiler as well as the linter if reassignment is attempted.

artsyit:

Built with commit ae01fdaa23c1a23fb68f34ad68ca147f381fdbdd

https://deploy-preview-1183--artsy-reaction.netlify.com

src/Components/Publishing/Partner/PartnerInline.tsx
```diff
@@ -19,42 +18,38 @@ interface DivProps extends React.HTMLProps<HTMLDivElement> {
...
-          <a href={url} target="_blank" onClick={this.onPartnerClick}>
-            <img src={logo} />
+        {resized_logo && <IconPlus color={color} />}
+        {resized_logo && (
+          <a
+            href={url}
+            target="_blank"
+            onClick={this.onPartnerClick.bind(this)}
+          >
+            <img src={resized_logo} />
```
damassi: A slight optimization is to use a React fragment, like 
```jsx
{resized_logo &&
  <>
    <IconPlus color={color} />
    <a href='..'>...</a>
  </>
}
```

src/Components/Publishing/Display/Canvas/CanvasText.tsx
eessex: We've had a few issues in staging where unfinished display ads are published by the content team, might be worth checking if the unit.logo exists before calling `resize` on it -- not sure what it does when we pass it undefined.

src/Components/Publishing/Display/Canvas/CanvasText.tsx
ansor4: 👍 Are absolute paths always preferable? I can change the rest of the import paths for these files

src/Components/Publishing/Partner/PartnerBlock.tsx
```diff
@@ -1,9 +1,9 @@
 import { unica } from "Assets/Fonts"
+import { pMedia } from "Components/Helpers"
 import React from "react"
 import track from "react-tracking"
```
alloy: If the tracks in this module adhere to the unified schema we’ve been trying to define, then you need to make sure to import our typed version of react-tracking from.

CC @eessex

eessex: I don't think this is blocking for this PR -- we'd need to add the values for these actions into the schema, maybe best to ticket updates to analytics for another date since @ansor4 is no longer on Grow.

eessex:
Looks like the snapshots need to be updated (`yarn jest --u`) to run green.

src/Components/Publishing/Partner/PartnerBlock.tsx
eessex: Also heads up @alloy added a follow up ticket to update analytics https://artsyproduct.atlassian.net/browse/GROW-874

eessex:
Hmm conflicts... want to try pulling the latest from master and re-running jest?

src/Components/Publishing/Partner/PartnerBlock.tsx
alloy: Thanks!

PR #1186, opened 2018-08-27
User can see the payment step

zephraph:
Shows the payment details step with most of the visual styling for the form


.storybook/preview-head.html
```diff
@@ -2,3 +2,4 @@
 <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.6.0/slick.min.css" />
 <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.6.0/slick-theme.min.css"
 />
+<script id="stripe-js" src="https://js.stripe.com/v3/" async></script>
```
zephraph: We'll need to make sure this dependency is being correctly pulled into force.

src/Apps/Order/OrderApp.tsx
```diff
@@ -33,6 +45,21 @@ export class OrderApp extends React.Component<OrderAppProps> {
     )
   }
 
+  componentDidMount() {
+    if (window.Stripe) {
+      this.setState({
+        stripe: window.Stripe("pk_test_BGUg8FPmcBs1ISbN25iCp2Ga"),
```
sweir27: We should use an ENV var for this (or `sd`, whatever the current pattern is). If you don't want to add it to this PR please add a TODO 😄

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -1,24 +1,54 @@
...
   order: Payment_order
 }
 
-export class PaymentRoute extends Component<PaymentProps> {
+interface PaymentState {
+  country: string
+  hideBillingAddress: boolean
+}
+
+const ContinueButton = ({ order }) => (
```
sweir27: This is OK for now but we might not need this abstraction-- once we're sending mutations, we'll get rid of the `Link` element anyways and likely handle the transition in an `onClick` handler.

sweir27:
Cool! This looks great. I just had one minor comment about extracting out a variable.

.storybook/preview-head.html
ds300: Might be possible to use react-helmet for that, depending on SSR needs/compatibility.

src/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const StyledCardElement = styled(CardElement)`
+  width: 100%;
+`
+
+interface StyledBorderBox {
+  hasError: boolean
+}
+const StyledBorderBox = styled(BorderBox).attrs<StyledBorderBox>({})`
+  ${border};
```
ds300: is this used?

src/Apps/Order/OrderApp.tsx
```diff
@@ -22,7 +29,12 @@ export interface OrderAppProps {
   router: Router
 }
 
-export class OrderApp extends React.Component<OrderAppProps> {
+interface OrderAppState {
+  stripe: stripe.Stripe
```
ds300: Not a problem with this PR: kind of a bummer that the stripe typings are exposed through a global namespace rather than through the module system 😕

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -31,46 +61,57 @@ export class PaymentRoute extends Component<PaymentProps> {
...
+                        />
+                      </Collapse>
+                      {!xs && <ContinueButton order={order} />}
+                    </Join>
+                    <Spacer mb={3} />
+                  </>
+                }
+                Sidebar={
+                  <Flex flexDirection="column">
+                    <TransactionSummary order={order} mb={xs ? 2 : 3} />
```
ds300: `xs ? 2 : 3` can be replaced with `[2, 3]` – although we're planning to fix all instances of this pattern in a later PR so feel free to leave this as-is.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -1,24 +1,54 @@
...
+      Continue
+    </Button>
+  </Link>
+)
+
+export class PaymentRoute extends Component<
+  PaymentProps,
+  PaymentState & Address
+> {
+  state = { country: "USA", hideBillingAddress: true }
```
ds300: this should be `"US"` rather than `"USA"`

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -31,46 +61,57 @@ export class PaymentRoute extends Component<PaymentProps> {
...
+                        onSelect={hideBillingAddress =>
+                          this.setState({ hideBillingAddress })
+                        }
+                      >
+                        Billing and shipping addresses are the same
+                      </Checkbox>
+                      <Collapse open={!this.state.hideBillingAddress}>
+                        <AddressForm
+                          country={this.state.country}
+                          onUpdateName={this.onUpdateName}
```
ds300: non-blocking: I hadn't seen this component before, and would like to take the opportunity to +1 ash's suggestion from #1152 to collapse these individual props into a single one.

src/Apps/Order/OrderApp.tsx
orta: I guess it's meant to reflect the way it's imported

ds300: Ah, true, it's defined on `window`. Still, a global namespace assumes that all typescript files in one's project will be running in a browser env that has stripe available. Unlikely to cause problems though.

orta: yeah, jest's types have the same issues, I think you can start writing `expect` anywhere:

screen shot 2018-08-28 at 8 04 33 am

src/Apps/Order/Routes/Payment/index.tsx
orta: linter rule?

PR #1191, opened 2018-08-28
@zephraph => [Refactor] Add root exports to /Styleguide

damassi:
This allows us to import components / elements like

```js
import { ReadMore, Slider, ... } from 'Styleguide/Components'
import { Box, Flex, ... } from 'Styleguide/Elements'
```

I kept the `/Components` and `/Elements` division in the import (though technically things could be imported from just `/Styleguide`) because I don't think we've quite settled on the design-system lingo.


src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -8,21 +8,24 @@ import FollowArtistButton from "Components/FollowButton/FollowArtistButton"
...
+  Box,
+  Button,
+  Checkbox,
+  Flex,
+  Message,
+  Radio,
+  Separator,
+  SmallSelect,
+  Spacer,
+} from "Styleguide/Elements"
```
zephraph: Much, much better.

artsyit:

Built with commit a71755bef3c1b99fce1167a4f4f1006c973d9a46

https://deploy-preview-1191--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -1,17 +1,16 @@
 import { Serif } from "@artsy/palette"
 import { artistResponse } from "Apps/__test__/Fixtures/MarketInsights"
 import React from "react"
-import { ArtistBio } from "Styleguide/Components/ArtistBio"
+import { Box } from "Styleguide/Elements/Box"
```
zephraph: I think this is meant to be sans `Box` at the end there.

src/Styleguide/Components/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+export * from "./ReadMore"
+export * from "./RecentlyViewed"
+export * from "./RouteTabs"
+export * from "./SelectedExhibitions"
+export * from "./Slider"
+export * from "./Stepper"
+export * from "./StepSummaryItem"
+export * from "./Tabs"
+export * from "./Toggle"
+export * from "./Lightbox"
```
zephraph: I wonder if there's a lint rule that exists to ensure every file is included in this export list...

PR #1193, opened 2018-08-28
Updates metaphysics schema

ashfurrow:
Hey, so these two PRs had breaking changes to the schema used by Reaction:

- https://github.com/artsy/metaphysics/pull/1244
- https://github.com/artsy/metaphysics/pull/1241

This PR updates the schema with `yarn sync-schema` and attempts to modify the components/tests/etc to reference the new schema. **However** I can't get it to work with storybooks or embedded within Force. @zephraph maybe we can pair on the remainder of this, since I think the problem is in our use of the router in `OrderApp.story.tsx`, or if this is blocking anyone, feel free to pick up from here on your own.

### DIFF
--- a/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
@@ -8,7 +8,19 @@ import { CreditCardDetails } from "./CreditCardDetails"
 import { ShippingAddress } from "./ShippingAddress"
 
 export const ShippingAndPaymentReview = ({
-  order: { fulfillmentType, lineItems, creditCard, ...address },
+  order: {
+    requestedFulfillment: {
+      fulfillmentType,
+      name,
+      addressLine1,
+      addressLine2,
+      city,
+      postalCode,
+      region,
+    },
+    lineItems,
+    creditCard,
+  },
   onChangePayment,
   onChangeShipping,
   ...others
@@ -30,7 +42,14 @@ export const ShippingAndPaymentReview = ({
       </StepSummaryItem>
     ) : (
       <StepSummaryItem onChange={onChangeShipping} title="Shipping address">
-        <ShippingAddress {...address} />
+        <ShippingAddress
+          shippingName={name}
+          shippingAddressLine1={addressLine1}
+          shippingAddressLine2={addressLine2}
+          shippingCity={city}
+          shippingPostalCode={postalCode}
+          shippingRegion={region}
+        />
       </StepSummaryItem>
     )}
     <StepSummaryItem onChange={onChangePayment} title="Payment method">
@@ -43,13 +62,19 @@ export const ShippingAndPaymentReviewFragmentContainer = createFragmentContainer
   ShippingAndPaymentReview,
   graphql`
     fragment ShippingAndPaymentReview_order on Order {
-      fulfillmentType
-      shippingName
-      shippingAddressLine1
-      shippingAddressLine2
-      shippingCity
-      shippingPostalCode
-      shippingRegion
+      requestedFulfillment {
+        ... on Pickup {
+          fulfillmentType
+        }
+        ... on Ship {
+          name
+          addressLine1
+          addressLine2
+          city
+          postalCode
+          region
+        }
+      }
       lineItems {
         edges {
           node {
--- a/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
@@ -5,19 +5,11 @@ import { createFragmentContainer, graphql } from "react-relay"
 import { StepSummaryItem } from "Styleguide/Components/StepSummaryItem"
 import { Flex, FlexProps } from "Styleguide/Elements/Flex"
 import { CreditCardDetails } from "./CreditCardDetails"
-import { ShippingAddress } from "./ShippingAddress"
+import { ShippingAddress, ShippingAddressProps } from "./ShippingAddress"
 
 export const ShippingAndPaymentReview = ({
   order: {
-    requestedFulfillment: {
-      fulfillmentType,
-      name,
-      addressLine1,
-      addressLine2,
-      city,
-      postalCode,
-      region,
-    },
+    requestedFulfillment: { fulfillmentType, ...address },
     lineItems,
     creditCard,
   },
@@ -42,14 +34,7 @@ export const ShippingAndPaymentReview = ({
       </StepSummaryItem>
     ) : (
       <StepSummaryItem onChange={onChangeShipping} title="Shipping address">
-        <ShippingAddress
-          shippingName={name}
-          shippingAddressLine1={addressLine1}
-          shippingAddressLine2={addressLine2}
-          shippingCity={city}
-          shippingPostalCode={postalCode}
-          shippingRegion={region}
-        />
+        <ShippingAddress {...address as ShippingAddressProps} />
       </StepSummaryItem>
     )}
     <StepSummaryItem onChange={onChangePayment} title="Payment method">
--- a/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
@@ -9,7 +9,7 @@ import { ShippingAddress, ShippingAddressProps } from "./ShippingAddress"
 
 export const ShippingAndPaymentReview = ({
   order: {
-    requestedFulfillment: { fulfillmentType, ...address },
+    requestedFulfillment: { __typename, ...address },
     lineItems,
     creditCard,
   },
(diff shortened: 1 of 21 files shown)


artsyit:

Built with commit 616478003e9fc08626429507080af32bcd08ef4c

https://deploy-preview-1193--artsy-reaction.netlify.com

src/Apps/Order/Components/ShippingAndPaymentReview.tsx
```diff
@@ -43,13 +62,19 @@ export const ShippingAndPaymentReviewFragmentContainer = createFragmentContainer
...
-      fulfillmentType
-      shippingName
-      shippingAddressLine1
-      shippingAddressLine2
-      shippingCity
-      shippingPostalCode
-      shippingRegion
+      requestedFulfillment {
+        ... on Pickup {
+          fulfillmentType
```
alloy: Rather than query this field, query for `__typename` of the union (I.e. hoist one level up). That way you can always use its value and the generated types will reflect that. It’s also more standard GraphQL usage.

@ashkan18 this field should imo not even be exposed in the schema, only on the input type to work around the lack of union input types.

src/Apps/Order/Components/ShippingAndPaymentSummary.tsx
```diff
@@ -38,13 +57,19 @@ export const ShippingAndPaymentSummaryFragmentContainer = createFragmentContaine
...
-      shippingName
-      shippingAddressLine1
-      shippingAddressLine2
-      shippingCity
-      shippingPostalCode
-      shippingRegion
+      requestedFulfillment {
+        ... on Pickup {
+          fulfillmentType
+        }
```
alloy: Ditto

src/__generated__/ShippingAndPaymentReview_order.graphql.ts
```diff
@@ -1,17 +1,18 @@
...
 export type ShippingAndPaymentReview_order$ref = typeof _ShippingAndPaymentReview_order$ref;
 export type ShippingAndPaymentReview_order = {
-    readonly fulfillmentType: OrderFulfillmentType | null;
-    readonly shippingName: string | null;
-    readonly shippingAddressLine1: string | null;
-    readonly shippingAddressLine2: string | null;
-    readonly shippingCity: string | null;
-    readonly shippingPostalCode: string | null;
-    readonly shippingRegion: string | null;
+    readonly requestedFulfillment: ({
```
ds300: @alloy would it be possible to get named types that represent either side of the union here? If you bring in `__typename` automatically for unions it would even be possible to generate predicate functions in this file too, e.g. `isShip` and `isPickup`

ds300:
Should this also update the `externals/metaphysics` submodule?

src/__generated__/ShippingAndPaymentReview_order.graphql.ts
alloy: That is somewhat how it’s supposed to work and why [I requested](https://github.com/artsy/reaction/pull/1193#discussion_r213549330) the change to select `__typename` instead.

For instance, take [this union selection](https://github.com/artsy/emission/blob/b743779dff6480deedfe61a51dfcbc7fa9324a9b/src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx#L189-L208) as an example. The [emitted typings](https://github.com/artsy/emission/blob/master/src/__generated__/ConversationSnippet_conversation.graphql.ts#L13-L35) for it are a union with each their own `__typename: "TypeName"` field. This is done so that TS can automatically narrow down the type based on the `__typename` value, which you can see in action [here](https://github.com/artsy/emission/blob/b743779dff6480deedfe61a51dfcbc7fa9324a9b/src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx#L145-L149) where TS knows that if `__typename === "Artwork"` the `item` can only be the `Artwork` part of the union etc.

src/Apps/Order/Components/ShippingAndPaymentReview.tsx
alloy: See [here](
https://github.com/artsy/reaction/pull/1193/files/bd97d6d100cec49e6e11ec71dcd185505749ec5b#r21363642)4 for some details about why we want to encourage selection for `__typename` and thus removing these unnecessary fields from the schema completely would be preferable to me.

src/__generated__/ShippingAndPaymentReview_order.graphql.ts
alloy: Yes we do, this is the repo https://github.com/relay-tools/relay-compiler-language-typescript. I actually filed an issue just this morning about emitting named types for selections https://github.com/relay-tools/relay-compiler-language-typescript/issues/64.

Could you expand a bit on why we’d need it in this case?

ds300: to be able to have sub-components or methods that don't care about the union type and don't need to specify their input types independently of the generated types. e.g.

```
class ShippingInformation {
  renderAddress({name, addressLine1, ...}: Ship) {
     return ...
  }
  render() {
    const {requestedFulfilment, shippingOrigin} = this.props.order

    return isShip(requestedFulfilment)
       ? renderAddress(requestedFulfilment)
       : "Pick up (${shippingOrigin})"
  }
}

Comment (alloy) [reply src/__generated__/ShippingAndPaymentReview_order.graphql.ts]: Gotcha.

FWIW Sub-components can work _today_ if you’d give them their own container and fragment.

Comment (alloy) [reply src/__generated__/ShippingAndPaymentReview_order.graphql.ts]: Or, as you [pointed out](
https://github.com/relay-tools/relay-compiler-language-typescript/issues/64#issuecomment-41692849)3, by using the `Extract` type 👍

File: src/Apps/Order/Components/ShippingAndPaymentReview.tsx (hunk 1 above)
Comment (ashfurrow) [reply src/Apps/Order/Components/ShippingAndPaymentReview.tsx]: Okay, nearly got this sorted. I think that `fulfillmentType` might _have_ to be included, actually. We need to spread over `Pickup` here in order for Relay to recognize `__typename` might be `Pickup` and not just `Ship`, but in order to spread over a type, it needs to have at least one field. At least that's my working understanding.

**Comment by ashfurrow** [comment]
Okay, this is working locally. Just need to update a test. However, the problem with our Stripe integration blocked me from testing until I commented it out, @sweir27 [described the issue on Slack](https://artsy.slack.com/archives/C9YNS4X32/p1535499665000100). Going to follow up @zephraph about it.

**Comment by ashfurrow** [comment]
🍏

File: src/Apps/Order/Components/ShippingAndPaymentReview.tsx [hunk 4]
```diff
@@ -30,7 +34,7 @@ export const ShippingAndPaymentReview = ({
       </StepSummaryItem>
     ) : (
       <StepSummaryItem onChange={onChangeShipping} title="Shipping address">
-        <ShippingAddress {...address} />
+        <ShippingAddress {...address as ShippingAddressProps} />
```
ds300: hmm I'd prefer not to see this cast, but I'll leave it for now because I'm planning to refactor the types here some more after this is merged.

ashfurrow: Yeah, the issue was that `ShippingAddressProps` are all non-null, but come as nullable from the schema. I think changing the `ShippingAddress` component would make sense.

src/Apps/Order/Components/ShippingAndPaymentReview.tsx
alloy: I took a quick check on master and it seems to work fine when I remove the `Pickup` inline fragment, you just need to make sure that you change the guard to check for `Ship` rather than `Pickup` https://gist.github.com/alloy/1140d14e4b1f487fb207e997b9ec9283

ashfurrow: I'll take a look – I was getting a type check failure.

alloy: Ah I see. Here’s a new version of the patch that changes the story to use the sentinel value `%other` which stands for “any other union member than the types being queried for” https://gist.github.com/alloy/1140d14e4b1f487fb207e997b9ec9283#file-shippingandpaymentreview-tsx-diff-L49. (Relay will always include that one by default to be forward compatible for when new union member types are added that existing clients don’t know how to handle yet.)

> I realize that _technically_ this is data might not be needed, but it would also arguably be weird to have requested fulfillment type be one of two things, but developers can only check against one of them.

Not sure I understand. The `Pickup` type doesn’t have any useful data of its own yet, when it does you can query for it. Until then there’s no real point in doing so, it only leads to more data being fetched than would otherwise happen (a field like `__typename` will already implicitly be included by Relay by default on unions, so adding an explicit selection for it doesn’t actually fetch any more data).

ashfurrow: Ah, that makes sense. Thanks. I'll send a follow-up PR. 

To clarify my other point, it just feels weird to have a type that's "Pickup" or "Ship", but the type-checker only allows it to be compared to "Ship", that's all. It's totally fine, just weird!

ashfurrow: Opened #1198 and assigned to Justin to spread this around a bit.

PR #1194, opened 2018-08-28
@l2succes: adds analytics for mobile auth cta

xtina-starr:
Addresses https://artsyproduct.atlassian.net/browse/GROW-880.

Adds analytics events:
- Auth impression when the cta banner is displayed
- Click event for when the user dismiss the cta banner
- Passes analytic properties to auth sign up modal

### DIFF
--- a/src/Artsy/Analytics/Schema/Interaction.ts
+++ b/src/Artsy/Analytics/Schema/Interaction.ts
@@ -8,7 +8,7 @@ import { ActionName, ActionType } from "./Values"
  * Some actions lead to results, such as following an artist, in which case the
  * action name is used to tie the interaction and result events together.
  */
-export interface BaseInteraction {
+export interface Interaction {
   /**
    * The type of interaction that this event represents. E.g. `Click`.
    *
@@ -36,7 +36,7 @@ export interface BaseInteraction {
   destination_path?: string
 }
 
-export interface AuthInteraction {
+export interface AuthenticationInteraction extends Interaction {
   /*
   * The action taken that prompted user to signup or login.
   */
@@ -52,5 +52,3 @@ export interface AuthInteraction {
   */
   trigger?: string
 }
-
-export type Interaction = BaseInteraction & AuthInteraction
--- a/src/Artsy/Analytics/Schema/Interaction.ts
+++ b/src/Artsy/Analytics/Schema/Interaction.ts
@@ -8,7 +8,7 @@ import { ActionName, ActionType } from "./Values"
  * Some actions lead to results, such as following an artist, in which case the
  * action name is used to tie the interaction and result events together.
  */
-export interface Interaction {
+export interface BaseInteraction {
   /**
    * The type of interaction that this event represents. E.g. `Click`.
    *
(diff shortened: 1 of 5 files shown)


eessex:
@alloy yeah good point -- we should be adding new events especially to the analytics schema, and converting old ones as we touch them.  Happy to pair on this if you have q's.

src/Components/MinimalCtaBanner.tsx
```diff
@@ -1,6 +1,8 @@
 import { Sans } from "@artsy/palette"
 import React from "react"
+import track from "react-tracking"
```
l2succes: We should use our typed version of `track` here
https://github.com/artsy/reaction/blob/master/src/Artsy/Analytics/track.ts

and add approriate values to the schema (https://github.com/artsy/reaction/blob/master/src/Artsy/Analytics/Schema/Values.ts#L70) if needed

src/Components/MinimalCtaBanner.tsx
```diff
@@ -26,7 +30,27 @@ export class MinimalCtaBanner extends React.Component<
     dismissed: false,
   }
 
+  componentDidMount() {
+    if (this.props.tracking) {
+      this.props.tracking.trackEvent({
```
l2succes: use `@track` decorator

xtina-starr:
I made some changes based on the feedback provided but I did have questions regarding the difference in schema that I was provided in the ticket and the way we are approaching this in reaction. I left a comment in the [ticket](https://artsyproduct.atlassian.net/browse/GROW-880) itself so feel free to chime in.

src/Artsy/Analytics/Schema/Interaction.ts
```diff
@@ -35,3 +35,22 @@ export interface Interaction {
...
+  */
+  flow?: string
+
+  /*
+  *  The type of action that triggered the modal (eg: click, timed)
+  */
+  trigger?: string
+}
+
+export type Interaction = BaseInteraction & AuthInteraction
```
alloy: I don’t think this is right, as it effectively adds these 3 optional properties to all places where an `Interaction` is used. I think you should export both, have `AuthInteraction` extend `BaseInteraction`, and in `src/Artsy/Analytics/Schema/index.ts` add `AuthInteraction` to the `Trackables` union.

I’m also not a big fan of naming things “base”, I would prefer `BaseInteraction` were named after what it’s used for. (The same applies to `AuthInteraction` btw, which I _think_ may be an interaction that led to an authentication prompt?)

I also wonder if indeed `intent` should really be optional or if you only had to do that to make type-checking work, which would presumably fail on all places where `Interaction` was already being used without `intent`? Once you make `AuthInteraction` its own thing (that extends `BaseInteraction`), this problem would go away and allow you to have required properties.

xtina-starr: @l2succes and I actually discussed this approach. I can't remember why we decided to combined the types though. I do think it makes sense to export `AuthenticationInteraction` separately and add it to the `Trackables` union. Luc do you have any objection?

`intent` is used just about everywhere except for when the CTA Banner is dismissed. For that reason I think we would want to make it optional.

xtina-starr:
@l2succes @l2succes wondering if there is any additional feedback for this PR?

l2succes:
@xtina-starr Just left a comment about the signup URL string building. Everything else looks good

PR #1198, opened 2018-08-29
Removes extraneous fragment spreading

ashfurrow:
As discussed in [this thread](
https://github.com/artsy/reaction/pull/1193#discussion_r213549330), we've hit kind of a weird edge of Relay/GraphQL schemas/tsc where we can only compare `__typename` to the names of types we spread over, even though the `Pickup` type has no useful information. That leads to two interesting tradeoffs:

- include the spread and transfer extra data from Metaphysics, but get type-checking working as expected, or
- not include it and only compare to `"Ship"` (or `"%other"`, which is a sentinel for non-spread-over value types). 

#1193 chose the first option, but this PR changes it to the second one. If you see weird type errors, let me know.

### DIFF
--- a/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
+++ b/src/Apps/Order/Components/ShippingAndPaymentReview.tsx
@@ -22,7 +22,11 @@ export const ShippingAndPaymentReview = ({
   onChangeShipping(): void
 } & FlexProps) => (
   <Flex flexDirection="column" {...others}>
-    {__typename === "Pickup" ? (
+    {__typename === "Ship" ? (
+      <StepSummaryItem onChange={onChangeShipping} title="Shipping address">
+        <ShippingAddress {...address as ShippingAddressProps} />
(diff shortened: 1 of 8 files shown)


artsyit:

Built with commit c53a76643e5925f83f3ac825e3722ce33ddd8737

https://deploy-preview-1198--artsy-reaction.netlify.com

ds300:
This is a bit better, but still an unfortunate tradeoff in terms of code clarity (at least for the mock data).

Is there something preventing relay from figuring out the possible values of __typename regardless of which types you select fields from? Aside from that not being implemented yet, i mean.

zephraph:
I'm trying to catch up with all the conversations around this, but I'm having a little difficulty wrapping my head around what's going on here.

alloy:
I don’t think we should spend too much time on the mocking of data, because it’s looking more likely that we’ll make more use of a mocked schema and actually exercising the Relay containers.

PR #1204, opened 2018-08-30
Update to latest Mutation response coming from MP

ashkan18:
# Problem
`setShipping` mutation has been changed to match latest style in https://github.com/artsy/metaphysics/pull/1247 We need to follow that here

# Solution
Update to match ☝️ , looks like in current code we don't check for errors or read the response yet so basically had to only change the mutation.


artsyit:

Built with commit 28ad921cbc17dbe745e6bfab364b3ec9dba0fa1c

https://deploy-preview-1204--artsy-reaction.netlify.com

PR #1205, opened 2018-08-31
Execute submit_order mutation from Review page

ds300:
- bumped metaphysics to include schema changes from artsy/metaphysics#1247
- execute submitOrder mutation when clicking the submit button on the review page

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -34,7 +34,7 @@ input ApproveOrderInput {
 }
 
 type ApproveOrderPayload {
-  result: OrderReturnType
+  orderOrError: OrderOrFailureUnionType
   clientMutationId: String
 }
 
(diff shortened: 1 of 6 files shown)


sweir27:
This looks good!!

At this stage, maybe we could add a test (something that mocks the mutation similar to what's in https://github.com/artsy/reaction/blob/master/src/Apps/Order/Routes/__tests__/Shipping.test.tsx) to verify that when the mutation is successful, it advances to the `status` page, and when unsuccessful it remains on this page.

ds300:
🍏

sweir27:
grrr what's up with the netlify thing?

ds300:
seems to be failing type checking but only when building storybook 😕 almost as if storybook is using an older version of typescript.

ds300:
Ok pretty sure this is 🍏now

artsyit:

Built with commit 39427dc968baf4e7f612d66ece7759884d15f6b6

https://deploy-preview-1205--artsy-reaction.netlify.com

PR #1206, opened 2018-08-31
User sees a full country name in shipping details views

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-421

This PR adds a new constant called `COUNTRY_CODE_TO_COUNTRY_NAME` , which is a country-code-to-full-name map based on the same data used in the `<CountrySelect>` component. The name sounds ok but a bit too long, so open to suggestions.

I also noticed that in the storybook we had `US` in the `region` property, which has been fixed as part of this PR.

## Screenshot

screen_shot_2018-08-31_at_9_34_53_am

### DIFF
--- a/src/Apps/Order/Components/ShippingAddress.tsx
+++ b/src/Apps/Order/Components/ShippingAddress.tsx
@@ -1,14 +1,16 @@
+import { Serif } from "@artsy/palette"
 import React from "react"
 
-import { Serif } from "@artsy/palette"
+import { COUNTRY_CODE_TO_COUNTRY_NAME } from "Styleguide/Components"
 
 export interface ShippingAddressProps {
   name: string
   addressLine1: string
-  addressLine2: string | null
+  addressLine2?: string | null
   city: string
   postalCode: string
   region: string
+  country: string
 }
 
 export const ShippingAddress = ({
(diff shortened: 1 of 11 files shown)


artsyit:

Built with commit 263b8ebdc54e13a87c9f7b3c34cd1e3808cb2f3d

https://deploy-preview-1206--artsy-reaction.netlify.com

src/Apps/Order/Components/ShippingAddress.tsx
```diff
@@ -18,13 +20,20 @@ export const ShippingAddress = ({
...
   return (
-    <Serif size="3t" style={{ whiteSpace: "pre-wrap" }}>
-      {[name, addressLine1, addressLine2, cityLine, region]
-        .filter(Boolean)
-        .join("\n")}
-    </Serif>
+    <>
+      <Serif size="3t">{name}</Serif>
+      <Serif size="3t">
+        {[addressLine1, (addressLine2 || "").trim()].filter(Boolean).join(", ")}
```
ds300: Hmm should these two lines be joined with a comma? I would expect them to be on separate lines in the view.

src/Apps/Order/Components/ShippingAddress.tsx
```diff
@@ -18,13 +20,20 @@ export const ShippingAddress = ({
...
+      </Serif>
+      <Serif size="3t">
+        {city}, {region}, {postalCode}
+      </Serif>
+      <Serif size="3t">
+        {COUNTRY_CODE_TO_COUNTRY_NAME[country] || country}
```
ds300: You could move this logic into a function `getCountryName(countryCode)` rather than using this big object directly everwhere?

src/Apps/Order/Components/ShippingAddress.tsx
```diff
@@ -18,13 +20,20 @@ export const ShippingAddress = ({
...
-        .filter(Boolean)
-        .join("\n")}
-    </Serif>
+    <>
+      <Serif size="3t">{name}</Serif>
+      <Serif size="3t">
+        {[addressLine1, (addressLine2 || "").trim()].filter(Boolean).join(", ")}
+      </Serif>
+      <Serif size="3t">
+        {city}, {region}, {postalCode}
```
ds300: I think there shouldn't be a comma between region and postalCode here

src/Apps/Order/Components/ShippingAddress.tsx
yuki24: I think a hash lookup is simple enough for now.

src/Apps/Order/Components/ShippingAddress.tsx
yuki24: Maybe it's only stantard in the U.S, but this feels more natural to me. Also, [the mockup also uses a comma](https://zpl.io/bzWx4KE).

yuki24:
Thanks for reviewing @ds300!

sweir27:
🎉 🎉 Nice!

PR #1217, opened 2018-09-06
Add phone number, refactor address form

ds300:
- bumped metaphysics to include changes from artsy/metaphysics#1265
- Added phone number field to AddressForm, while refactoring it to not have individual change handlers for each field, as discussed in https://github.com/artsy/reaction/pull/1152#pullrequestreview-147370701 and https://github.com/artsy/reaction/pull/1186#discussion_r213277528

https://artsyproduct.atlassian.net/browse/PURCHASE-422
https://artsyproduct.atlassian.net/browse/PURCHASE-397

force screenshots of phone number appearing in:
- review address
    image
- summary address
    image
- shipping address form
    image

### DIFF
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -47,6 +47,7 @@ export class AddressForm extends React.Component<
   changeEventHandler = (key: keyof Address) => (
     ev: React.FormEvent<HTMLInputElement>
   ) => {
+    console.log("FECK", ev.currentTarget.value)
     this.onChangeValue(key, ev.currentTarget.value)
   }
 
@@ -65,6 +66,7 @@ export class AddressForm extends React.Component<
       <Join separator={<Spacer mb={2} />}>
         <Flex flexDirection="column">
           <Input
+            id="AddressForm_name"
             placeholder="Add full name"
             title="Full name"
             defaultValue={this.props.defaultValue.name}
@@ -86,6 +88,7 @@ export class AddressForm extends React.Component<
 
           <Flex flexDirection="column">
             <Input
+              id="AddressForm_addressLine2"
               placeholder="Add postal code"
               title="Postal code"
               onChange={this.changeEventHandler("postalCode")}
@@ -96,6 +99,7 @@ export class AddressForm extends React.Component<
         <TwoColumnSplit>
           <Flex flexDirection="column">
             <Input
+              id="AddressForm_addressLine1"
               placeholder="Add street address"
               title="Address line 1"
               onChange={this.changeEventHandler("addressLine1")}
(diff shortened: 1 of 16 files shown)


artsyit:

Built with commit 53c3de65cf8feaa1b8bebaf7000ba9fd939db54d

https://deploy-preview-1217--artsy-reaction.netlify.com

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -5,26 +5,70 @@ import { CountrySelect } from "Styleguide/Components"
...
+  postalCode: string
+  addressLine1: string
+  addressLine2: string
+  city: string
+  region: string
+  phoneNumber: string
 }
 
-export class AddressForm extends React.Component<AddressFormProps> {
+export const emptyAddress: Address = Object.freeze({
```
zephraph: Are we polyfilling Object.freeze? I remember seeing Core in our bundle payloads, so I think we probably are, but it'd be good to double check.

ds300: 🤔 do we need to? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze#Browser_compatibility

ds300:
Alright this is 🍏

src/Apps/Order/Components/AddressForm.tsx
zephraph: You make a valid point 😄

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -14,17 +14,20 @@ import { Col, Row } from "Styleguide/Elements/Grid"
...
 import { TransactionSummaryFragmentContainer as TransactionSummary } from "../../Components/TransactionSummary"
-import { Address } from "../Shipping"
 
 export interface PaymentProps {
   order: Payment_order
 }
 
 interface PaymentState {
-  country: string
+  address: Partial<Address>
```
ds300: argh, this should not be `Partial`

note to self: fix

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -5,26 +5,72 @@ import { CountrySelect } from "Styleguide/Components"
...
+  AddressFormState
+> {
+  state = {
+    address: { ...emptyAddress, ...this.props.defaultValue },
+  }
+
+  changeEventHandler = (key: keyof Address) => (
+    ev: React.FormEvent<HTMLInputElement>
+  ) => {
+    console.log("FECK", ev.currentTarget.value)
```
ashfurrow: Ah, missed this.

ds300: :o 

Fixed in #1222

PR #1218, opened 2018-09-06
[PATCH artwork] Artwork details implementation

oxaudo:
This is mostly super minor changes to Artwork details. Also added a test just to illustrate how input with special characters will look in ReadMore. Does not do anything about it for now but wanted to have an example.

I'll self merge that for faster deploy but please let me know of any feedback here.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
+++ b/src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
@@ -42,6 +42,11 @@ export class ArtworkDetailsAdditionalInfo extends React.Component<
     return (
       <StackableBorderBox pb={3}>
         <Box>
+          <Box>
+            <Sans size="2" weight="medium">
+              Additional info
+            </Sans>
+          </Box>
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit a7b3b3266e08957272806ef78dc29b9e8728b7da

https://deploy-preview-1218--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
```diff
@@ -42,6 +42,11 @@ export class ArtworkDetailsAdditionalInfo extends React.Component<
     return (
       <StackableBorderBox pb={3}>
         <Box>
+          <Box>
+            <Sans size="2" weight="medium">
+              Additional info
+            </Sans>
+          </Box>
```
damassi: Any reason this has to be wrapped in a box?

oxaudo: Without box - it was trying to put it inline. I could have given Sans a display property but thought that it is more descriptive to put it as a separate box to make it easier to move around and mark that it's a separate header element.

PR #1221, opened 2018-09-06
User can submit payment details

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-331

I was having trouble writing tests due to the Stripe elements requiring `context`. I'll follow up and add more test coverage once I solve the issue in mocking `context`.

## Screenshot

(Here is [a high quality MP4](https://artsy.slack.com/archives/C9YNS4X32/p1536183131000100) in case the screenshot below isn't helpful.)

out

PR #1223, opened 2018-09-07
Backfill tests for <PaymentRoute> component

yuki24:
This is a follow-up PR on https://github.com/artsy/reaction/pull/1221 that backfills test coverage for the `<PaymentRoute>` component. I've managed to test it but it's questionable how we could test a component that depends on the React Stripe elements. Let me know what you think!


artsyit:

Built with commit 860f8ca81b6be7071c663de176038dc51dad6897

https://deploy-preview-1223--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const mountPaymentRoute = props => {
+    const providerComponent = shallow(<PaymentRoute {...props} />)
+    const responsiveComponent = shallow(
+      providerComponent
+        .childAt(2)
+        .props()
+        .children({ xs: false })
+    ) as any
+
+    return shallow(responsiveComponent.props().children({ xs: false }))
```
yuki24: This is the questionable function that shallow-mounts the component. We don't want to mount the component entirely otherwise Stripe components will fail due to the lack of `context` values that are supposed to be injected. The library's test does seem to [test them with context values injected properly](https://github.com/stripe/react-stripe-elements/blob/732dd3ae7c64f171218064af012fbc2b13842482/src/components/Element.test.js#L38), but I couldn't figure out how we could do the same in here.

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+    ) as any
+    const twoColumnLayout = shallow(
+      responsiveComponent.props().children({ xs: false })
+    )
+
+    return {
+      paymentRouteComponent,
+      responsiveComponent,
+      twoColumnLayout,
+    }
```
yuki24: This is the questionable function that shallow-mounts the component three times. We don't want to mount the component entirely otherwise Stripe components will fail due to the lack of `context` values that are supposed to be injected. The library's test itself does seem to [test them with context values injected properly](https://github.com/stripe/react-stripe-elements/blob/732dd3ae7c64f171218064af012fbc2b13842482/src/components/Element.test.js#L38), but I couldn't figure out why this was not working in our setup.

damassi: Have you tried an approach similar to https://github.com/artsy/reaction/blob/master/src/Utils/RelayStubProvider.ts?

src/Apps/__test__/Fixtures/MutationResults/createCreditCard.ts
```diff
@@ -0,0 +1,9 @@
+export const creatingCreditCardSuccess = {
+  createCreditCard: {
+    creditCardOrError: {
+      creditCard: {
+        id: "gravityCreditCardId",
+      },
+    },
+  },
+}
```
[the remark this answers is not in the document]
damassi: Or maybe inside of `/Apps/__test__/Fixtures/...`

src/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -39,9 +41,17 @@ export class CreditCardInput extends React.Component<
...
+  static getDerivedStateFromProps(props, state) {
+    return {
+      error: props.error,
+    }
+  }
+
   render() {
     const { message } = this.state.error || { message: null }
 
+    // console.log("CreditCardInput", message)
```
zephraph: whoops, stray console

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+    expect(thenMock.mock.calls.length).toBe(1)
+  })
+
+  it("tokenizes credit card information with a different billing address", () => {
+    const thenMock = jest.fn()
+    stripeMock.createToken.mockReturnValue({ then: thenMock })
+
+    const paymentRoute = mount(<PaymentRoute {...testProps} />)(
+      paymentRoute.find(Checkbox).props() as CheckboxProps
+    ).onSelect(false)
```
zephraph: I was pairing w/ @yuki24 on this... for whatever reason, `simulate('select', false)` wasn't working here. I've seen a few conversations about similar things online, but we weren't really able to figure out why it wasn't working. Does anyone else have any ideas?

ds300: Yeah. `simulate` behaves differently depending on whether you do shallow mount or regular old mount.

- With shallow mount, simulate just calls the prop that maps to the event name you provide. So `blah.simulate('foo', data)` calls `blah.props().onFoo(data)`.
- With regular old mount, simulate simulates actual dom events. It will find the outermost actual dom rendered by the component you call it on, then trigger an event on that node, merging whatever data you provide with a fully-fleshed out event object. So `blah.simulate('foo', data)` is something like `ReactDOM.findDomNode(blah).trigger('foo', Object.assign(new WhateverEvent(), data))`

so to fix this you need to narrow your selection to the actual dom node that react attaches the event handler to, or just use the props like you already did. A lot of people do the latter 🤷‍♂️

zephraph: Cool! Thanks for the information!

I guess we can leave it as is then.

yuki24:
Ok now this is good for review.

yuki24:
Well, the CI build is failing to run the test suite and I'm looking into that now.

yuki24:
Alright, CI is back to green.

PR #1224, opened 2018-09-07
User sees an error message for credit card input immediately

yuki24:
Right now users don't see an error message for the `<CreditCardInput>` as soon as they type, but this PR updates that so users will get immediate feedback. This also moves the message from the `<PaymentRoute>` to the `<CreditCardInput>` so it'll help implement the same, consistent behaviour across the site.

# Screenshot

kapture 2018-09-07 at 15 53 30

### DIFF
--- a/src/Apps/Order/Components/CreditCardInput.tsx
+++ b/src/Apps/Order/Components/CreditCardInput.tsx
@@ -1,4 +1,4 @@
-import { color } from "@artsy/palette"
+import { color, Sans } from "@artsy/palette"
 import { fontFamily } from "@artsy/palette/dist/platform/fonts"
 import {
   border as inputBorder,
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 138c9787d929730d2c5e95043b7fe6db9d7e30a3

https://deploy-preview-1224--artsy-reaction.netlify.com

src/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -18,37 +18,58 @@ const StyledBorderBox = styled(BorderBox).attrs<InputBorderProps>({})`
...
+  CreditCardInputState
+> {
   state = {
     focused: false,
-    error: false,
+    error: null,
   }
 
   render() {
+    const message = (this.state.error || this.state.error || {}).message
```
sweir27: `this.state.error || this.state.error`

yuki24: Maybe I should test some tests 😩

yuki24:
Ok, I added some test coverage and not this is good for review.

sweir27:
🙌 🙌 Nice test coverage!

PR #1226, opened 2018-09-07
[COORDINATED MERGE] Refactor phoneNumber & nest ShippingAddress fragment

ds300:
⚠️ Do not merge until all the related PRs are green: artsy/metaphysics#1272 , artsy/exchange#162 , and a forthcoming PR in potential.

This one in particular needs to be merged  **after** the metaphysics PR, because the metaphysics submodule in here needs to be bumped to master when that happens.

---

### What I did

- The phoneNumber refactor (moving a copy of `buyerPhoneNumber` into `Ship` as `phoneNumber`)
- Refactor ShippingAddress to be a nested fragment container on `Ship`. @alloy and I had been pairing on this before, and I wanted to make it work as a proof of concept, but we needed extra testing infrastructure to render lower-level components through relay in storybook (previously we only had it set up to render whole pages through relay). So I made a generic `StorybookRelay` wrapper which can inject a query into a component.

  Part of me likes the architectural flexibility of this, but part of me feels like it's really heavy and it makes fewer of our components presentational, e.g. `ShippingAndPaymentReview` and `ShippingAndPaymentSummary` in this case, since they now use a fragment container for `ShippingAddress` under the hood and can't be rendered without a relay environment.


artsyit:

Built with commit 3c6815152fa3001637583c8f903a76e77f6ac891

https://deploy-preview-1226--artsy-reaction.netlify.com

alloy:
This component was in the grey zone for if it makes sense to create a container for or not, but it was a good exercise for you and in reality I don’t expect these to be used outside of a Relay context at all. If they need to after all, they can still be used without Relay, in practice.

src/Apps/Order/Components/__stories__/ShippingAndPaymentDetails.story.tsx
```diff
@@ -30,9 +40,17 @@ const order: ShippingAndPaymentReview_order &
...
     expiration_year: 21,
   },
-  buyerPhoneNumber: "555-8390344",
 }
 
+const orderQuery = graphql`
+  query ShippingAndPaymentDetailsQuery {
+    order(id: "foo") {
+      ...ShippingAndPaymentSummary_order
+      ...ShippingAndPaymentReview_order
```
alloy: Smart

src/Artsy/Router/Components/StorybookRelay.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react"
+import { QueryRenderer } from "react-relay"
+import { Environment, RecordSource, Store } from "relay-runtime"
+import { createMockNetworkLayer } from "../../Relay/createMockNetworkLayer"
+
+export const StorybookRelay = ({
```
alloy: I like it 👌 

Is it really only for storybooks, though? Seems like this can be used for tests as well, in which case the name could do with updating too.

src/Artsy/Router/Components/StorybookRelay.tsx
```diff
@@ -0,0 +1,37 @@
...
+  Component,
+  query,
+  mockResolvers,
+}: {
+  Component: React.ComponentClass | React.SFC
```
alloy: These two can be replaced by `React.ComponentType`

src/Artsy/Router/Components/StorybookRelay.tsx
```diff
@@ -0,0 +1,37 @@
...
+  query: any
```
alloy: `RelayRuntimeTypes.GraphQLTaggedNode` ?

src/Artsy/Router/Components/StorybookRelay.tsx
```diff
@@ -0,0 +1,37 @@
...
+  mockResolvers: { [typeName: string]: () => any }
```
alloy: `import { IResolvers } from "graphql-tools/dist/Interfaces"`

src/Artsy/Router/Components/StorybookRelay.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const source = new RecordSource()
+  const store = new Store(source)
+  const environment = new Environment({
+    network,
+    store,
+  })
+
+  return (
+    <QueryRenderer
+      // tslint:disable-next-line
```
alloy: What is this for? Can it be reduced to disabling a single named rule (and maybe just for 1 line)?

src/Apps/Order/Components/__stories__/ShippingAndPaymentDetails.story.tsx
```diff
@@ -41,40 +59,64 @@ storiesOf("Apps/Order Page/Components", module).add(
         <Section title="Shipping and Payment Summary">
           <h4>Delivery</h4>
           <Flex flexDirection="column" width={300}>
-            <ShippingAndPaymentSummary order={order} />
+            <StorybookRelay
+              Component={ShippingAndPaymentSummary}
+              mockResolvers={{ Order: () => order }}
+              query={orderQuery}
+            />
```
damassi: Nice 😎

damassi: @sweir27 has been pushing for a solid way to test bits of Relay code for a while and this feels right

src/Artsy/Router/Components/StorybookRelay.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react"
+import { QueryRenderer } from "react-relay"
+import { Environment, RecordSource, Store } from "relay-runtime"
+import { createMockNetworkLayer } from "../../Relay/createMockNetworkLayer"
```
damassi: Can this be updated to an absolute path?

src/Artsy/Router/Components/StorybookRelay.tsx
damassi: If this unnamed interface were exported it would automatically generate a `.d.ts` file on build that could be read by consumers of Reaction.

src/Artsy/Router/Components/StorybookRelay.tsx
ds300: Ah I meant to ask you about this! It's a lint rule you wrote:

https://github.com/artsy/reaction/blob/6b33396f3aaa2a11423455df2a294704c1019e4d/tslint/relayOperationGenericsRule.js#L64

Does it make sense to ignore it in this context as a one-off, or is the rule not doing what's intended?

alloy: Ah I see. Yeah ignoring it here makes sense, but narrow it down to just this rule and only for this line.

ds300:
I've addressed all the issues 🍏

alloy:
LGTM ✨

ds300:
merge on green

PR #1228, opened 2018-09-08
[WIP] Uses the schema SDL instead of the JSON

orta:
Would love some advice here - this won't render in storybooks. 

The `relay-mock-network-layer` will create the schema correctly when passed the SDL in that format, so everything runs fine in the tests.


src/Artsy/Relay/createMockNetworkLayer.ts
```diff
@@ -1,12 +1,16 @@
+import { readFileSync } from "fs"
 import { IMocks } from "graphql-tools/dist/Interfaces"
 import getNetworkLayer from "relay-mock-network-layer"
 import { Network } from "relay-runtime"
-import schema from "../../../data/schema.json"
+
+// TODO: This doesn't work in client side rendering
+const schemaDSL = readFileSync("data/schema.graphql", "utf8")
```
alloy: Ah yes, I bet it’s because you’re trying to read a file, which you can’t do from a browser. I now faintly remember this being a reason for me to choose the json file at the time.

I’m sure that webpack has solutions for embedding data.

alloy: Maybe there’s a more efficient solution, but I think that the simple ‘replace’ plugin could do the job by just replacing this call with a string literal that contains the GraphQL source.

ds300: It's not much hassle to do in webpack: https://github.com/webpack/docs/wiki/how-to-write-a-loader

Then we could just `require('<any-old-file>.graphql')`

ds300: in fact several already exist  e.g. https://github.com/samsarahq/graphql-loader

orta:
OK, I've started down the loader path

artsyit:

Built with commit 3db8da0cd8c64a0811ae68158471d3062eb405b7

https://deploy-preview-1228--artsy-reaction.netlify.com

tooling/webpack/graphqlLoader.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function(source) {
+  this.cacheable()
+  this.value = source
+  const sanitizedSource = source.replace(/`/g, "\\`")
+  return `export default \`${sanitizedSource}\``
+}
```
damassi: It would be worth investigating some other implementations: https://github.com/samsarahq/graphql-loader

orta:
argh, I need to update this

orta:
Deprecated by https://github.com/artsy/reaction/pull/1477

PR #1232, opened 2018-09-10
Shipping form fills in Relay data

ashfurrow:
This PR addresses [PURCHASE-376](https://artsyproduct.atlassian.net/browse/PURCHASE-376), whose goal it is to populate the shipping form with any details that the user has already entered for the order. There is some tricky state management with the Address form, default values, etc, so let me know if something looks fishy. I'll comment inline with a few things I'm unsure of.


ashfurrow:
Cool, you can see the shipping details filled in on the [storybook deploy](https://deploy-preview-1232--artsy-reaction.netlify.com/?selectedKind=Apps%2FOrder%20Page&selectedStory=Shipping&full=0&addons=1&stories=1&panelRight=0&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel):

screen shot 2018-09-10 at 4 31 44 pm

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -87,9 +97,10 @@ export class AddressForm extends React.Component<
 
           <Flex flexDirection="column">
             <Input
-              id="AddressForm_addressLine2"
+              id="AddressForm_postalCode"
               placeholder="Add postal code"
               title="Postal code"
+              defaultValue={this.props.defaultValue.postalCode}
```
ashfurrow: This was included already for name, but none of the others. Seems fine to add them, but let me know.

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -40,8 +40,18 @@ export class AddressForm extends React.Component<
   AddressFormProps,
   AddressFormState
 > {
+  // We need to pull out _only_ the values specified by the Address type,
+  // since our state will be used for Relay variables later on. The
+  // easiest way to do this is with the emptyAddress.
+  sanitizedPropsAddress: Partial<Address> = Object.keys(emptyAddress).reduce(
```
ds300: This operation is a little confusing without a higher-level context, since `defaultValue` is already a `Partial<Address>`. Can we move it a level up (e.g. into `Shipping/index`)?

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -52,13 +52,14 @@ export interface ShippingState {
 }
 
 export class ShippingRoute extends Component<ShippingProps, ShippingState> {
-  // TODO: Fill in with Relay data on load.
-  // See: https://artsyproduct.atlassian.net/browse/PURCHASE-376
   state = {
-    shippingOption: "SHIP" as OrderFulfillmentType,
+    shippingOption: ((this.props.order.requestedFulfillment &&
+      this.props.order.requestedFulfillment.__typename.toUpperCase()) ||
+      "SHIP") as OrderFulfillmentType,
```
ds300: not really related to this PR, but it might make sense to align capitalization across the stack?

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -52,13 +52,14 @@ export interface ShippingState {
...
     address: {
       ...emptyAddress,
       country: "US",
+      ...(this.props.order.requestedFulfillment as Partial<Address>),
```
ds300: re: moving sanitizedPropsAddress  a level up

I'm imagining simply replacing this expression with something like `_.pick(this.props.order.requestedFulfillment, Object.keys(emptyAddress))` or equivalent.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -114,4 +115,38 @@ describe("Shipping", () => {
...
+      expect((input.props() as InputProps).defaultValue).toBe(
+        testProps.order.requestedFulfillment.name
+      )
+    })
+
+    it("includes already-filled-in data in mutation if re-sent", () => {
+      const component = getWrapper(testProps)
+      const mockCommitMutation = commitMutation as jest.Mock<any>
+      mockCommitMutation.mockImplementationOnce((_environment, config) => {
+        expect(config.variables.input.shipping.name).toBe("Dr Collector")
```
ds300: :+1:

src/Apps/Order/Components/AddressForm.tsx
ashfurrow: Yeah, I waffled on that. I think moving it up makes sense 👍

src/Apps/Order/Routes/Shipping/index.tsx
ashfurrow: Yeah I was wondering that too. I think the problem is we have two types (`Ship` and `Pickup`) as well as two enums with the same name (`SHIP` and `PICKUP`), but types and enums have different capitalization conventions. It's fairly minor, but if we had to do it over again, I'd say that the enums should be more verbose; for example, `SHIP_FULFILLMENT` or something. But that might just be the Objective-C developer in me talking 😉 /cc @ashkan18

src/Apps/Order/Routes/Shipping/index.tsx
ashfurrow: Ah that's way better, yeah.

src/Apps/Order/Routes/Shipping/index.tsx
ashkan18: heheh im not sure here either. I think the recommendation was to use all caps for enums but i can be totally wrong :)
Here is where we create that enum -> https://github.com/artsy/exchange/blob/master/app/graphql/types/order_fulfillment_type_enum.rb#L2

ashfurrow:
Okay, made the changes outlined above and tested locally. Let's see what CI thinks...

src/Apps/Order/Routes/Shipping/index.tsx
orta: yeah, enums in GraphQL are all caps: https://graphql.org/learn/schema/#enumeration-types

Issue #1166, opened 2018-08-22
[RFC] Enable type checking for tests


ds300:
# Proposal
Delete this line in tsconfig.json:

![image](https://user-images.githubusercontent.com/1242537/44465733-52983c00-a616-11e8-87c3-e7c45777a7bd.png)

# Reasoning
- This will allow the typescript compiler to check for type errrors in test files, which is extra armour on top of the tests themselves.

- I can imagine API changes that don't cause a test to fail but might still cause the test's code to be an out-of-date usage example for that particular API.

- The vscode TS plugin currently doesn't work properly in test files because it has the default contextless TS config applied, rather than our own tsconfig.json file. This means that, e.g. absolute import paths show up as errors, but work fine when run under jest:

    ![image](https://user-images.githubusercontent.com/1242537/44465957-f681e780-a616-11e8-886e-ce3fd9e3c37f.png)

   And of course if you choose to write test code that would fail type checking, you get other red squiggly lines that just add noise.

    ![image](https://user-images.githubusercontent.com/1242537/44467548-0ac7e380-a61b-11e8-89d1-f304b1ba40ce.png)

# Exceptions
Sometimes we don't care about types, or they get in the way. I would therefore recommend use of `any` or `@ts-ignore` or even `.js` files if our setup supports that?

# How
Turning type-checking on for test files and running `tsc` introduces 243 errors. It would be a big old PR to fix them all in one go. So I'd recommend doing it in a staggered way:

1. Turn on type-checking for tests
2. Fix a handful of erroring test files
3. If there are still erroring test files left over then turn off type-checking for tests
4. Submit PR
5. If there are still erroring test files left over then go to 1.

orta:
> .js files if our setup supports that

They do, that can be fine 👍 (and the `// @ts-ignore`)

---

I think we did this intentionally because we were teaching web folk typescript from coffeescript, and then never went back to change the setting once everyone was comfortable.

damassi:
I've been running into this for a few days now! Very much in favor, but I think there's a significant amount of work that will need to take place in order to get the type-checking to pass which will involve introducing a lot of `any`'s. 

(I'm not in favor of turning on `noImplicitAny`, though, since writing solid types is a different kind of work than shipping features under deadlines. If there are other ways to make the compiler stricter I'm in favor. I'm also all for RFC'ing this.)

orta:
[...]

PR #1235, opened 2018-09-11
[type checking test files] Boot -> MockBoot

ds300:
### See #1166 for background and approach

In this exciting first edition, we:

- Replaced a bunch of type-unsafe usages of the `Boot` component with a more safe `MockBoot` component which has a smaller API designed specifically for testing, that can grow over time.
- Type-checked some mock relay data.
- Deleted some unused things.

🕺


alloy:
As mentioned previously, I think we should get rid of all manual data insertion into these containers, but for the time being (and whenever really necessary), you may want to try this approach that somebody shared https://github.com/relay-tools/relay-compiler-language-typescript/issues/29#issuecomment-417101505

src/Styleguide/Components/__tests__/ArtistCard.test.tsx
```diff
@@ -18,39 +19,39 @@ describe("ArtistCard", () => {
...
 
   beforeAll(() => {
-    window.matchMedia = undefined // Immediately set matching media query in Boot
+    window.matchMedia = undefined // Immediately set matching media query in MockBoot
   })
 
   it("is responsive", () => {
     const small = mount(
-      <Boot initialMatchingMediaQueries={["xs"]}>
+      <MockBoot breakpoint="xs">
```
damassi: `breakpoint` is a bit easier on the hands. At some point if we wanted to make it more correct I think we could add a `mediaQuery` prop and merge the two together as it's passed into Boot.

src/Styleguide/Elements/__tests__/Select.test.tsx
```diff
@@ -11,16 +11,16 @@ describe("Select", () => {
 
   it("is responsive", () => {
     const small = mount(
-      <Boot initialMatchingMediaQueries={["xs"]}>
+      <MockBoot breakpoint={"xs"}>
```
damassi: #superminor - For consistency with the above changes, maybe just `breakpoint='xs`?

src/Utils/MockBoot.tsx
```diff
@@ -0,0 +1,21 @@
+import { Breakpoint } from "@artsy/palette/dist/Theme"
```
damassi: Can import this from the root as all exports are now available there. 

`import { Breakpoint } from '@artsy/palette'`

src/Utils/MockBoot.tsx
```diff
@@ -0,0 +1,21 @@
...
+      context={null as any}
+      user={null as any}
+      relayEnvironment={null as any}
+      resolver={null as any}
+      routes={null as any}
+    >
+      {children}
+    </Boot>
+  )
+}
```
damassi: I dig it 😄 We can expand as needed with time but i like how for now its very explicitly constrained.

ds300:
Thanks for your feedback! I've addressed the issues.

Merge on green

PR #1236, opened 2018-09-11
[PATCH ArtworkDetails] Include tabs

oxaudo:
Not for merging yet.

I'd really like to pair for a bit with you tomorrow at some point @damassi to discuss this tabs here and if we need to do something more complicated here.


artsyit:

Built with commit 08590f94936cc55a13f274dd504f612ba194c920

https://deploy-preview-1236--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -19,10 +21,23 @@ export class ArtworkDetails extends Component<ArtworkDetailsProps> {
...
+          <Tab name="Articles">
+            <Articles artwork={artwork} />
+          </Tab>
+          {artwork.exhibition_history && (
+            <Tab name="Exhibition history">{artwork.exhibition_history}</Tab>
+          )}
+          {artwork.literature && (
+            <Tab name="Bibliography">{artwork.literature}</Tab>
+          )}
+        </Tabs>
```
damassi: @oxaudo - re

> I'd really like to pair for a bit with you tomorrow at some point @damassi to discuss this tabs here and if we need to do something more complicated here.

This working well for you? It seems exactly right

damassi: The only question is finding a way to check that Articles is indeed renderable.

oxaudo: Yes, the code above is indeed working for the last 2 tabs here. I am putting the last 2 tabs directly inline instead of creating a separate components for those though I think it's ok because both of those literally just displaying a text field per tab.

The first 2 tabs is where things get more complicated. And I think it makes sense to have components for each chunk of data separately but my Details parent component(index here) needs to know all the data to figure out what to render if anything at all. That is where things get tricky....

I tried to see what you guys did for Artist tabs and it seems like you and @mzikherman created the whole `ArtistStatuses` type in MP to hold the data for the tabs. You think it's a good approach or alternatively I could just query all data directly in `fragment ArtworkDetails_artwork on Artwork` instead of including child's fragments? All the data needs to be requested either way to know if tab should be rendered or not.

That is what I wanted to discuss:)
I might ask people around here for opinions as well meanwhile.

oxaudo: @damassi - per our discussion in slack moved all the data getting into main ArtworkDetails component here. Kept all the rendering still split up into smaller components just to keep this file more concise and readable.

Let me know what you think here.

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -3,9 +3,11 @@ import { ArtworkDetailsQuery } from "__generated__/ArtworkDetailsQuery.graphql"
 import { ContextConsumer } from "Artsy/Router"
 import React, { Component } from "react"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
-import { ArtworkDetailsAboutTheWorkFragmentContainer as AboutTheWork } from "./ArtworkDetailsAboutTheWork"
-import { ArtworkDetailsAdditionalInfoFragmentContainer as AdditionalInfo } from "./ArtworkDetailsAdditionalInfo"
-import { ArtworkDetailsChecklistFragmentContainer as Checklist } from "./ArtworkDetailsChecklist"
+import { Tab, Tabs } from "Styleguide/Components"
+import { ArtworkDetailsAboutTheWork as AboutTheWork } from "./ArtworkDetailsAboutTheWork"
```
damassi: 👍

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
damassi: Looks good!

oxaudo:
I still need to do some conditional logic to determine weather About the work tab and the whole details component should be rendered or not but I think this is mergeable for now unless you have some reservations so that we can put in in Force. It'll look closer to the design than what we currently have there.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsArticles.tsx
```diff
@@ -0,0 +1,34 @@
...
+  ArtworkDetailsArticlesProps
+> {
+  render() {
+    const { articles } = this.props.artwork
+    if (!articles || articles.length < 1) {
+      return null
+    }
+    return <Box>{articles.length}</Box>
+  }
+}
```
damassi: @oxaudo - can you elaborate on why all of these interface props are here? If this whole component is only used to display a number it might be best to merge it into its parent. Otherwise, its best to limit the prop interface to what is explicitly being used inside of the component.

damassi:
@oxaudo - Thinking this through a bit further, I think a better approach in keeping with proper relay encapsulation (fragmentContainers) is to keep things as they were with each component requesting its own data, and then duplicating that query in the container component to check whether or not something should render. This update looks good, but I think we should be more strict with component spread and either: 

a) Keep everything in one file and avoid passing props / data down; or
b) Stick to fragments (similar to previous commit) and duplicate data as necessary. It doesn't seem like that much duplication will be necessary to keep things by the book.

oxaudo:
Oh.. sorry - should have left a comment. This is not finished, yet. We need to put articles display here. This view will be for admins only for now so I’m just displaying the count as a placeholder.

oxaudo:
@damassi - Ok, updated here again putting back fragments for all the child components and only getting needed data on parent component.

Also adding Articles display for Artwork articles. I think I need to remove Article display component to not be under Artis routes. Maybe into Components? What do you think?

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -18,11 +20,49 @@ const ArtworkDetailsContainer = Box
...
+      artwork.description ||
+      artwork.framed ||
+      artwork.signatureInfo ||
+      artwork.conditionDescription ||
+      artwork.certificateOfAuthenticity ||
+      artwork.series ||
+      artwork.publisher ||
+      artwork.manufacturer ||
+      artwork.provenance ||
+      artwork.image_rights
```
damassi: This is def #minor but when dealing with a lot of props like this its best to alpha (highlight, then cmd + shift + p > sort in VSCode) so in the future it will be easier to see what's there if we need to add / remove conditions.

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -18,11 +20,49 @@ const ArtworkDetailsContainer = Box
 export class ArtworkDetails extends Component<ArtworkDetailsProps> {
   render() {
```
damassi: If this isn't using `state` or instance methods maybe make an SFC?

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsArticles.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import { Col, Row } from "Styleguide/Elements/Grid"
+
+import { ArtworkDetailsArticles_artwork } from "__generated__/ArtworkDetailsArticles_artwork.graphql"
+
+export interface ArtworkDetailsArticlesProps {
+  artwork: ArtworkDetailsArticles_artwork
+}
+
+export class ArtworkDetailsArticles extends React.Component<
```
damassi: Same here -- if not using `React.Class` aspects would be better if it was a `React.SFC`

PR #1240, opened 2018-09-13
Update shipping mutation to show error modal when an error occurs

sweir27:
This addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-438

This uses our generic error modal (in this case not piping in any specific titles or messages).

shipping-mutation

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -35,6 +35,7 @@ import {
   Spacer,
 } from "@artsy/palette"
 
+import { ErrorModal } from "Components/Modal/ErrorModal"
 import { Col, Row } from "Styleguide/Elements/Grid"
 
 export interface ShippingProps {
@@ -50,6 +51,7 @@ export interface ShippingState {
   address: Address
   shippingOption: OrderFulfillmentType
   isComittingMutation: boolean
+  isModalOpen: boolean
 }
 
 export class ShippingRoute extends Component<ShippingProps, ShippingState> {
@@ -66,6 +68,7 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
       ...pick(this.props.order.requestedFulfillment, Object.keys(emptyAddress)),
     },
     isComittingMutation: false,
+    isModalOpen: false,
   }
 
   onContinueButtonPressed = () => {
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit 3185296c402c4ff580743eb507c6a583a1f3af99

https://deploy-preview-1240--artsy-reaction.netlify.com

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -197,6 +220,10 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
             />
           )}
         </Responsive>
+        <ErrorModal
+          onClose={this.onCloseModal.bind(this)}
```
zephraph: Given that `onCloseModal` is defined as an arrow function I don't think that you have to bind it to the instance... right? (I'm not 100% sure about that, but that's how I thought it worked).

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -50,6 +51,7 @@ export interface ShippingState {
   address: Address
   shippingOption: OrderFulfillmentType
   isComittingMutation: boolean
+  isModalOpen: boolean
```
zephraph: This state variable is a bit generic. It's tied to the behavior of the Error modal but that really doesn't show based on the name. Maybe `isErrorModalOpen` or something?

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -84,12 +89,17 @@ describe("Shipping", () => {
...
+      commitMutation.mockReset()
+    })
+
     it("routes to payment screen after mutation completes", () => {
       const component = getWrapper(testProps)
       const mockCommitMutation = commitMutation as jest.Mock<any>
       mockCommitMutation.mockImplementationOnce(
         (_environment, { onCompleted }) => {
-          onCompleted()
+          onCompleted(settingOrderShipmentSuccess)
```
zephraph: @ds300, could `MockRelayRenderer` be used in place of mocking `createFragmentContainer`? I was thinking that it would be useful for these cases, but I'm not 100% certain.

sweir27:
Merging after talking to @zephraph last night.

PR #1241, opened 2018-09-13
Set title for order app

zephraph:
This adds the ability to update head properties via Reaction while enabling their rendering via SSR.


src/Apps/Order/OrderApp.tsx
```diff
@@ -81,9 +82,12 @@ export class OrderApp extends React.Component<OrderAppProps, OrderAppState> {
   render() {
     const { children } = this.props
     return (
-      <StripeProvider stripe={this.state.stripe}>
-        <Elements>{children}</Elements>
-      </StripeProvider>
+      <>
+        <Title>Checkout | Artsy</Title>
```
zephraph: Sets the default title for the order app.

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -25,6 +26,7 @@ export class StatusRoute extends Component<StatusProps> {
           <TwoColumnLayout
             Content={
               <>
+                <Title>Order status | Artsy</Title>
```
zephraph: Overrides the title on the status route

artsyit:

Built with commit a45d8588fe2be5c89e1ec6587d217bbe01a4278b

https://deploy-preview-1241--artsy-reaction.netlify.com

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -27,6 +28,7 @@ export function buildServerApp(config: RouterConfig): Promise<Resolve> {
       const historyMiddlewares = [queryMiddleware]
       const resolver = new Resolver(relayEnvironment)
       const render = createRender({})
+      const headTags = []
```
zephraph: `headTags` will hold any components that should be rendered in the head after SSR is complete. It's passed via boot into a context provider.

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -96,6 +99,7 @@ export function buildServerApp(config: RouterConfig): Promise<Resolve> {
           <App data={relayData} loadableState={loadableState} {...props} />
         ),
         status,
+        headTags,
```
zephraph: `headTags` here is return in the resolve of `buildServerApp` which can be used in `server.js` in force to actually provide the head markup via SSR. See https://github.com/artsy/force/pull/2881.

src/Apps/Order/Routes/Status/index.tsx
sweir27: Wow this library is cool.

package.json
```diff
@@ -162,6 +162,7 @@
     "qs": "^6.5.0",
     "rc-slider": "^8.6.2",
     "react-css-transition-replace": "^3.0.2",
+    "react-head": "^3.0.0",
```
damassi: @zephraph - `react-head` looks good, but wondering if you investigated any other possibilities? Did this one seem most useful long-term?

zephraph: I did. `react-helmet` is the most popular option. While it gives a lot of flexibility, it's pretty heavy and it's _not_ async friendly. There's `react-helmet-async` which is a variation of `react-helmet` but it doesn't use `react-side-effects`. It's also still listed as unstable and there are outstanding issues that causes it to be a problem. `react-head` uses a bit of a simpler approach then the others (mainly relying on portals). It's written by one of the tech leads at the NYT, it's well tested, and best of all it's very small.

damassi: Beautiful, ty 👍

PR #1248, opened 2018-09-14
Purchase419

ashleyjelks:
Added a flag to check if artwork is able to be picked up. If it is not UI does not display radio buttons with options to ship or pickup, rather it just displays a shipping address form.

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -3,7 +3,7 @@ import { BuyNowStepper } from "Apps/Order/Components/BuyNowStepper"
 import { Helper } from "Apps/Order/Components/Helper"
 import { TransactionSummaryFragmentContainer as TransactionSummary } from "Apps/Order/Components/TransactionSummary"
 import { Router } from "found"
-import { pick } from "lodash"
+import { get, pick } from "lodash"
 import React, { Component } from "react"
 import { Collapse } from "Styleguide/Components"
 import { Responsive } from "Utils/Responsive"
@@ -138,7 +138,10 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
   render() {
     const { order } = this.props
-    this.props.order.lineItems[0].edges.node.artwork.pickup_available
+    const isPickupAvailable = get(
+      this.props,
+      "order.lineItems.edges[0].node.artwork.pickup_available"
+    )
     return (
       <>
         <Row>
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit 641ce5cf2c3b7388712e3a0410e32c9f9d53fc8f

https://deploy-preview-1248--artsy-reaction.netlify.com

sweir27:
This is looking good!

Would you mind posting a screenshot of what the shipping form looks like in this case? 😄

Also, it looks like the commit history got a little messy with all of the `Revert`s. Can we update so your changes appear as a single (or few) commits on top of master?

ashleyjelks:
After clicking Buy on an artwork that does not have a pickup available option user only sees the Shipping Form, the radio buttons are no longer displayed...see the following screenshots:

screen shot 2018-09-17 at 3 14 00 pm
screen shot 2018-09-14 at 6 23 29 pm

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -138,6 +138,10 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
   render() {
     const { order } = this.props
+    const isPickupAvailable = get(
+      this.props,
+      "order.lineItems.edges[0].node.artwork.pickup_available"
```
sweir27: One #minor note that I _think_ we try to avoid using `get` because we lose the ability to type-check.

damassi: This is an edge-case example I think since its so many props deep :) 

I think for props that are like one or two or three dots in then it makes sense to check using like 

```js
const isPickupAvailable = order && order.lineItems && order.lineItems.edges[0]
```

but if we were to check an entire chain like this it would become less readable (i think).

damassi:
This PR looks really good! One little suggestion that makes browsing PRs easier for other devs over time is pasting the JIRA ticket link into the body of the description as well as include a descriptive title. (Whenever I start working in a new repo one of the first things I do is click the `Closed PRs` link and see what others have done and model after that.)

Here's the list for Reaction: https://github.com/artsy/reaction/pulls?q=is%3Apr+is%3Aclosed
And here's one for Force: https://github.com/artsy/force/pulls?q=is%3Apr+is%3Aclosed

(It's also a good idea to look at [individual commits], just to get a sense of the style (https://github.com/artsy/reaction/commits/master).) 

We're not overly strict about how titles or commits are formatted, but it's important to note that in Reaction formatting a title a specific way will trigger an [automatic deploy](https://github.com/artsy/reaction#commits-and-deployments) via [semantic release](https://github.com/semantic-release/semantic-release).

sweir27:
Nice work!

PR #1249, opened 2018-09-14
Shipping + Payment address validations

erikdstock:
This adds error validations to both address screens (Shipping and Payment).
Tracked in [PURCHASE-401](https://artsyproduct.atlassian.net/browse/PURCHASE-401) & [PURCHASE-403](https://artsyproduct.atlassian.net/browse/PURCHASE-403)

This gif shows the form failing to submit, then correctly submitting & setting its button state based on the api response:
ship


artsyit:

Built with commit 616313c57755ae1e300f56ce9437e06f14c63373

https://deploy-preview-1249--artsy-reaction.netlify.com

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -44,32 +45,59 @@ export const ContinueButton = props => (
...
+            })
+          } else {
+            this.createCreditCard({ token: token.id })
+          }
+        })
     })
   }
 
+  validateAddress = values => {
+    if (!this.state.hideBillingAddress) {
```
erikdstock: this should probably follow the same check as `getSelectedBillingAddress()` below that also factors in fulfillment type.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -83,79 +111,116 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
...
     )
   }
 
-  private getSelectedBillingAddress(): stripe.TokenOptions {
+  // Infer the billing address from the form or shipping address (maybe not needed)
+  private getSelectedBillingAddress(formAddress: Address): Address {
+    return (this.isPickup() || !this.state.hideBillingAddress
+      ? formAddress
+      : this.props.order.requestedFulfillment) as Address
+  }
```
erikdstock: maybe not needed separate from `getStripeBillingAddress()` below.

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -19,6 +19,7 @@ interface Resolve {
 }
 
 export function buildServerApp(config: RouterConfig): Promise<Resolve> {
+  console.log("start", Boot, Hydrator, config)
```
erikdstock: ❌

src/Components/Input.tsx
```diff
@@ -190,7 +190,7 @@ export class Input extends React.Component<InputProps, InputState> {
       <Container>
         {title && <Title>{title}</Title>}
         {description && <Description>{description}</Description>}
-        <StyledInput {...rest} />
+        <StyledInput hasError={!!error} {...rest} />
```
erikdstock: Missing prop?

.vscode/settings.json
```diff
@@ -17,5 +17,6 @@
     "editor.formatOnSave": false
   },
   "tslint.autoFixOnSave": true,
-  "typescript.tsdk": "./node_modules/typescript/lib"
+  "typescript.tsdk": "./node_modules/typescript/lib",
+  "cSpell.enabled": false
```
damassi: What is this prop?

src/Apps/Order/Components/Validators.tsx
```diff
@@ -0,0 +1,21 @@
...
+  startingErrors?: E
+) => E = (fields, values, startingErrors = {} as any) =>
+  reduce(
+    fields,
+    (acc, fieldName) =>
+      values[fieldName].length
+        ? acc
+        : Object.assign({}, acc, { [fieldName]: "Required" }),
+    startingErrors
+  )
```
damassi: Since there is a lot of data to unpack here (which in the future might need to be `console.log`'d) can you add brackets to each function? Implicit returns should never be used on code that exceeds one line as it makes inspectability more difficult as well as introduces the potential for subtle errors over time.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -22,6 +23,7 @@ import {
 } from "../../Components/AddressForm"
 import { Helper } from "../../Components/Helper"
 import { TransactionSummaryFragmentContainer as TransactionSummary } from "../../Components/TransactionSummary"
+import { requiredFields } from "../../Components/Validators"
```
damassi: Absolute paths are preferred

.vscode/settings.json
erikdstock: I was wondering about this and meant to ask before committing it. It seems like the artsy extension pack gave me a spell checker which is failing on `Formik.` I don't know why a spell checker is needed on ts files in the first place.

orta: I don't think it should be disabled, *it's really useful* - you should just add an exception for the word "formik" - it's with the lightbulb

src/Apps/Order/Routes/Payment/index.tsx
yuki24: yeah I think you can just inline it.

yuki24:
I feel like this PR would add a lot more complexity than [the previous work](https://github.com/artsy/emission/commit/e3716f1ec34417a262787fcceeea9756ac0395fa) we paired on would and I'm not actually sure formik is a good fit in this case. Is there any reason we can't wait until we absolutely need it?

PR #1251, opened 2018-09-17
[FIX display] don't call gemini with empty urls

eessex:
Addresses https://artsyproduct.atlassian.net/browse/GROW-889

Fixes a bug where gemini resize was called with empty strings, causing long timeouts and impacting article load performance.

### DIFF
--- a/src/Components/Publishing/Display/Canvas/CanvasVideo.tsx
+++ b/src/Components/Publishing/Display/Canvas/CanvasVideo.tsx
@@ -134,11 +134,14 @@ export class CanvasVideo extends Component<CanvasVideoProps, any> {
 
   render() {
     const { isPlaying } = this.state
+    const { coverUrl, src } = this.props
 
-    const resized_cover = resize(this.props.coverUrl, {
-      width: 760,
-      isDisplayAd: true,
-    })
+    const resized_cover =
+      coverUrl &&
+      resize(coverUrl, {
+        width: 760,
+        isDisplayAd: true,
+      })
 
     return (
       <VideoContainer onClick={this.onPlayVideo}>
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit f8efb7c6d0aaa0f796e61661f18d55b6f1eaa514

https://deploy-preview-1251--artsy-reaction.netlify.com

src/Components/Publishing/Display/Canvas/CanvasVideo.tsx
```diff
@@ -134,11 +134,14 @@ export class CanvasVideo extends Component<CanvasVideoProps, any> {
 
   render() {
     const { isPlaying } = this.state
+    const { coverUrl, src } = this.props
 
-    const resized_cover = resize(this.props.coverUrl, {
-      width: 760,
-      isDisplayAd: true,
-    })
+    const resized_cover =
```
l2succes: 👍 @eessex we should also add a check in `resize` to make sure we never send empty urls to gemini

l2succes: something as simple as

```javascript
if (!url) return null
```

Here https://github.com/artsy/reaction/blob/master/src/Utils/resizer.ts#L58

joeyAghion:
Thanks for this. I'll watch the metrics post-deploy!

This makes me think Gemini should _also_ fail much more gracefully than this. I'll ticket that and fix if straightforward.

src/Components/Publishing/Display/Canvas/CanvasVideo.tsx
eessex: Good call will add this

PR #1252, opened 2018-09-17
User sees error modal if there is an error submitting the mutation to submit the payment form

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-404

This is pretty much the same as https://github.com/artsy/reaction/pull/1240 but for the `<PaymentStep>`.

## Screenshot

kapture 2018-09-17 at 13 53 36

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -3,6 +3,7 @@ import { Payment_order } from "__generated__/Payment_order.graphql"
 import { BuyNowStepper } from "Apps/Order/Components/BuyNowStepper"
 import { CreditCardInput } from "Apps/Order/Components/CreditCardInput"
 import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
+import { ErrorModal } from "Components/Modal/ErrorModal"
 import { Router } from "found"
 import React, { Component } from "react"
 import {
@@ -34,6 +35,7 @@ interface PaymentState {
   hideBillingAddress: boolean
   error: stripe.Error
   isComittingMutation: boolean
+  isErrorModalOpen: boolean
 }
 
 export const ContinueButton = props => (
@@ -51,6 +53,7 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
     hideBillingAddress: true,
     error: null,
     isComittingMutation: false,
+    isErrorModalOpen: false,
   }
 
   onContinueButtonPressed = () => {
@@ -68,6 +71,10 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
         }
       })
     })
+  }
+
+  onCloseModal = () => {
+    this.setState({ isErrorModalOpen: false })
   }
 
   render() {
@@ -151,6 +158,11 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
             />
           )}
         </Responsive>
+
+        <ErrorModal
+          onClose={this.onCloseModal}
+          show={this.state.isErrorModalOpen}
+        />
       </>
     )
   }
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit 314dd71386100504224440ded3647dc578207b79

https://deploy-preview-1252--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -241,4 +249,74 @@ describe("Payment", () => {
...
+  it("shows an error modal when there is an error in CreateCreditCardPayload", () => {
+    stripeMock.createToken.mockReturnValue({
+      then: func => func({ token: { id: "tokenId" } }),
+    })
+
+    mutationMock.mockImplementationOnce((_, { onCompleted }) =>
+      onCompleted(creatingCreditCardFailed)
+    )
+
+    const component = mount(<PaymentRoute {...testProps} />)
```
sweir27: #superminor and others may have the opposite opinion, but it feels like there are too many line breaks in these tests

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -265,7 +283,8 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
   }
 
   private onMutationError(errors) {
-    console.error(errors)
+    console.error("Order/Routes/Payment/index.tsx", errors)
+    this.setState({ isComittingMutation: false, isErrorModalOpen: true })
```
sweir27: The original ticket had a requirement to also start displaying the formatted errors we're already getting back from metaphysics during the `createCreditCard` error (see [this file](https://github.com/artsy/emission/blob/c006850dc1383f4f05e8e25ef127af2c35f01066/src/lib/Components/Bidding/Screens/Registration.tsx#L164) for how we do it in emission).

If you think that is out-of-scope of this PR, we should create a sub-ticket to track that requirement. If you think it's small enough to add here that's 👍 too.

yuki24: I think I should be able to handle it as part of this story - I just missed the last sentence in the ticket.

src/Apps/Order/Routes/__tests__/Payment.test.tsx
yuki24: I just simply deleted the duplicate expectations since we don't have to test it in every test. We still have one test that checks this, but we may want to add a test dedicated for the modal behaviour rather than testing the integration with Modal/GraphQL.

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -241,4 +249,65 @@ describe("Payment", () => {
...
+
+    expect(component.find(ErrorModal).props().show).toBe(false)
+
+    component.find(ContinueButton).simulate("click")
+
+    expect(component.find(ErrorModal).props().show).toBe(true)
+    expect(component.find(ErrorModal).props().detailText).toBe(
+      "No such token: fake-token"
```
yuki24: Not sure if this is helpful enough from a user's perspective, but at least this will add a foundation that shows a message from the server and we should be able to change it easily.

yuki24:
Alright this is good for review.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -182,13 +197,20 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
...
           this.setOrderPayment({
-            creditCardId: data.createCreditCard.creditCardOrError.creditCard.id,
+            creditCardId: creditCardOrError.creditCard.id,
           })
         } else {
-          // TODO: Add error handling
-          console.error(errors)
+          this.onMutationError(
+            errors || creditCardOrError.mutationError,
+            creditCardOrError.mutationError &&
```
sweir27: Thanks for adding this handling!

PR #1257, opened 2018-09-18
@l2succes: [FIX onboarding] explicitly pass in a number value to onNextButtonPressed

xtina-starr:
Addresses https://artsyproduct.atlassian.net/browse/GROW-603

In the [Layout Component](https://github.com/artsy/reaction/blob/master/src/Components/Onboarding/Steps/Layout.tsx#L106), the `onClick` prop executes the `onNextButtonPressed` function. When we don't explicitly pass a number value, the `event` gets passed as `increaseBy` in the [Wizard](https://github.com/artsy/reaction/blob/master/src/Components/Onboarding/Wizard.tsx#L40-L42) and fails to increment to the next index (step) correctly.

### DIFF
--- a/src/Components/Onboarding/Steps/Genes/index.tsx
+++ b/src/Components/Onboarding/Steps/Genes/index.tsx
@@ -60,12 +60,17 @@ export default class Genes extends React.Component<StepProps, State> {
     })
   }
 
+  handleNextButtonClick() {
+    const increaseBy = 1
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit e926c22ca3ea8ff5cbb3fe73a371786ace42c080

https://deploy-preview-1257--artsy-reaction.netlify.com

PR #1265, opened 2018-09-19
Purch 401 payment + shipping address validations

erikdstock:
This was accidentally merged before it was complete- Here is the updated shipping and address routes with form validations, but no dependence on formik.

ship


src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -216,9 +234,9 @@ describe("Payment", () => {
       onCompleted(creatingCreditCardSuccess)
     )
 
-    mount(<PaymentRoute {...testProps} />)
-      .find(ContinueButton)
-      .simulate("click")
+    const paymentRoute = mount(<PaymentRoute {...testProps} />)
+    fillAddressForm(paymentRoute, validAddress)
```
yuki24: It seems like some of the `fillAddressForm` calls could be replaced with `paymentRoute.find(Checkbox).props().onSelect(false)` or you don't even need to call it given that the shipping address is already there.

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -310,4 +327,45 @@ describe("Payment", () => {
...
+      expect(input.props().error).toEqual("This field is required")
+    })
+
+    it("does not submit an empty form with billing address exposed", () => {
+      const paymentRoute = mount(<PaymentRoute {...testProps} />)
+      ;(paymentRoute.find(Checkbox).props() as CheckboxProps).onSelect(false)
+
+      paymentRoute.find(ContinueButton).simulate("click")
+      expect(commitMutation).not.toBeCalled()
+    })
```
yuki24: I'd remove this test since the test below also covers the same case.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -1,23 +1,46 @@
...
+  fillIn,
+  validAddress,
+} from "../testSupport/addressForm"
 
 jest.mock("react-relay", () => ({
   commitMutation: jest.fn(),
   createFragmentContainer: component => component,
 }))
 
+const fillAddressForm = (component: any, address: Address) => {
```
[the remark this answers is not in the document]
yuki24: I talked about this to @erikdstock but I'm not concerned about that at this point. There will be duplicate `fillIn` calls but it doesn't seem like a reasonable amount that can justify consolidation today.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -257,7 +307,10 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
           this.onMutationError(errors || orderOrError)
         }
       },
-      onError: this.onMutationError.bind(this),
+      onError: (error?: Error) => {
+        this.setState({ isComittingMutation: false })
+        this.onMutationError.bind(this)(error)
```
yuki24: Same here.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -134,27 +157,27 @@ describe("Shipping", () => {
 
     it("shows an error modal when there is an error from the server", () => {
       const component = getWrapper(testProps)
-      expect(component.find("ErrorModal").props().show).toBe(false)
+      expect(component.find("ErrorModal").prop("show")).toBe(false)
```
yuki24: We'd prefer `props().show` otherwise we'll lose all the TypeScript benefits.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -186,12 +210,66 @@ describe("Shipping", () => {
       const component = getWrapper(testProps)
       const mockCommitMutation = commitMutation as jest.Mock<any>
       mockCommitMutation.mockImplementationOnce((_environment, config) => {
+        component.update()
         expect(config.variables.input.shipping.name).toBe("Dr Collector")
```
yuki24: Let's just move the assertion out of this mock so you don't have to call `component.update()` and the test wouldn't look like a brain teaser.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -186,12 +210,66 @@ describe("Shipping", () => {
...
+      const component = getWrapper(shipOrderProps)
+      const { addressLine1, ...badAddress } = validAddress
+      component.find(Button).simulate("click")
+      expect(commitMutation).not.toBeCalled()
+    })
+
+    it("does not submit the mutation with an incomplete form for a SHIP order", () => {
+      const component = getWrapper(shipOrderProps)
+      const { addressLine1, ...badAddress } = validAddress
+      fillAddressForm(component, badAddress)
```
yuki24: Instead of using `fillAddressForm`, I think it may be better to just use `fillIn` so it'll be clearer what's provided and what's not and you don't have to deconstruct a `validAddress`.

yuki24:
Merge on green.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -136,8 +179,22 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+      address,
+      addressErrors: {
+        ...this.state.addressErrors,
+        [key]: this.validateAddress(address)[key],
+      },
+    })
+  }
+
   render() {
-    const { order } = this.props
```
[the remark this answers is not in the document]
erikdstock: I think it came up in resolving a merge conflict- a change someone else had made (or not made), and in this particular case removing that line turned out to be the path of least resistance. Will re-add.

PR #1266, opened 2018-09-20
[package] Make relay-mock-network-layer a peer dependency.

alloy:
Reverts the change in #1182 for this to be a hard dep.

This dep is useful for dev when projects want to make use of `StorybookRouter`, but that doesn’t (yet) apply to all consumers, hence it should be a peer dep.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "react": "^16.5.0",
     "react-dom": "^16.5.0",
     "react-router": "^4.2.0",
+    "relay-mock-network-layer": "^1.2.0",
     "styled-components": "^2 || ^3"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


anandaroop:
Thanks for cleaning this up! 

Alas this change alone was not sufficient I think. I still get an unmet `graphql` peer dependency from [this Reaction dep](https://github.com/artsy/reaction/blob/641eb2f629218e7867793d6ace0fd5f56211ac9c/package.json#L176) as well:

screen shot 2018-09-20 at 9 35 53 am

& 

screen shot 2018-09-20 at 9 36 21 am

(Curiously, when I `yarn link` local Reaction (off latest master) to Volt, I don't get the above Webpack errors. 🤷‍♂️)

How about if we just keep `graphql` in Volt's deps for now?

alloy:
Yeah that’s a good call, because this dep isn’t optional.

damassi:
https://github.com/artsy/reaction/pull/1182#issuecomment-423262615

PR #1268, opened 2018-09-20
@xtina-starr => Limit visible time filter options

eessex:
Limits time periods available in the filter options to a hard-coded list. 
https://artsyproduct.atlassian.net/browse/GROW-898


artsyit:

Built with commit 7181e364c3933ca4d6304228ff2347d80d024907

https://deploy-preview-1268--artsy-reaction.netlify.com

src/Apps/Collect/Components/ArtworkGrid/index.tsx
```diff
@@ -38,7 +39,12 @@ class Filter extends Component<Props> {
         ? filters.state.major_periods[0]
         : filters.state[category]
 
-    return counts.map((count, index) => {
+    const options =
+      category === "major_periods"
```
l2succes: @eessex We should also remove the periodAggregation from the relay query as it'll speed up the overall query.

Here, remove any references to `periodAggregation`
https://github.com/artsy/reaction/pull/1268/files#diff-d4b3365020edfd3839ec76ed28fd0a5eR130
```javascript
     const periodAggregation = aggregations.find(
       agg => agg.slice === "MAJOR_PERIOD"
     )

```

And here, remove the `MAJOR_PERIOD` string
https://github.com/artsy/reaction/pull/1268/files#diff-d4b3365020edfd3839ec76ed28fd0a5eL269

src/Apps/Collect/CollectApp.tsx
```diff
@@ -20,8 +19,7 @@ export interface CollectAppProps {
 export class CollectApp extends Component<CollectAppProps> {
   render() {
     return (
-      <>
-        <Title>Collect | Artsy</Title>
+      <div>
```
xtina-starr: I believe @l2succes intended for this to be a fragment instead of a `div`. Not sure if it even matters, but I wanted to point it out.

eessex: I was assuming the fragment was so we could add meta tags -- now that we have a meta template we can use a div with no problem. Is there any reason we shouldn't wrap a top level app in a div?  I find the `<>` a confusing/non-idomatic syntax.

l2succes: @eessex Fragments are actually the way forward in this particular use case. Prior to React 16.2, we've used divs to wrap a set of components because it was the only way to return multiple elements. https://reactjs.org/docs/fragments.html

Also our top level apps are injected inside our [`Boot`](https://github.com/artsy/reaction/blob/master/src/Artsy/Router/Components/Boot.tsx#L60) component which already includes a container grid. If you look at the other apps, they all use a top level fragment for that reason as well

eessex: Ok sounds good, what is the reason for using the syntax `<>` over `<Fragment>`?

xtina-starr: It’s the shorthand syntax. Perhaps we should you the regular syntax until people become more familiar with the pattern...? Because I had the same question.

orta: Yeah, the syntax is useful because they can provide logical JSX groupings without filling the HTML/RN DOM with filler elements

xtina-starr:
LGTM...merging.

PR #1270, opened 2018-09-20
[FEATURE BNMO] User sees generic error modal if there is an error when submitting an order

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-348

This is pretty much the same as #1252. Skipping screenshots since we've done this twice already but happy to for people who don't have context.


sweir27:
Seems pretty straightforward! 👍

zephraph:
This looks good! I'll merge if everyone is 👍

PR #1271, opened 2018-09-20
[FEATURE BNMO] Disable input fields (including radio buttons) when the UpdateShipping mutation is underway

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-436

I initially added `disabled` prop to the inputs but replaced them with a `pointer-events: none;` on the parent.

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -97,7 +97,10 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
           {({ xs }) => (
             <TwoColumnLayout
               Content={
-                <>
+                <Flex
+                  flexDirection="column"
+                  style={isComittingMutation ? { pointerEvents: "none" } : {}}
+                >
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -84,6 +87,7 @@ export class AddressForm extends React.Component<
             <CountrySelect
               selected={this.state.address.country}
               onSelect={this.changeValueHandler("country")}
+              disabled={this.props.disabled}
```
yuki24: The `< CountrySelect >` component doesn't disable itself even when `disabled` is true, but this is because the `<LargeSelect>`component that's in Palette. I'll address this on the palette side and when we upgrade palette we'll get the disabled state for free.

artsyit:

Built with commit 1674e8b8bc43afc3e55604cc3b2d9c7bffb28627

https://deploy-preview-1271--artsy-reaction.netlify.com

sweir27:
Cool! Looks good to me, thanks for updating to this scheme.

PR #1275, opened 2018-09-21
@oxaudo => Update <ArtistInfo /> component and add <EntityOverview />

damassi:
**Blocker:**
https://github.com/artsy/palette/pull/93

**Addresses:** 
- https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&projectKey=DISCO&view=planning&selectedIssue=DISCO-299

**Zeplin:** 
- https://zpl.io/an9njYY

TODO:
- [x] Update name to `EntityHeader` and link to Spec:  zpl.io/aNoYM6d

Will follow up with a second PR that wraps things in a QueryRenderer for Force. 

#### Updated `<ArtistInfo />`

screen shot 2018-09-21 at 9 29 35 pm

#### `<EntityOverline />`

screen shot 2018-09-21 at 9 32 36 pm

#### Original comps

screen shot 2018-09-20 at 9 44 23 pm

### DIFF
--- a/src/Apps/Artwork/Components/ArtistInfo.tsx
+++ b/src/Apps/Artwork/Components/ArtistInfo.tsx
@@ -1,7 +1,6 @@
-import { Box, Serif } from "@artsy/palette"
+import { Spacer, StackableBorderBox } from "@artsy/palette"
 import { artistResponse } from "Apps/__test__/Fixtures/MarketInsights"
 import React from "react"
-import { Responsive } from "Utils/Responsive"
 
 import {
   ArtistBio,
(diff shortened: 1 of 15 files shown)


src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -20,31 +20,39 @@ interface ArtistInfoProps
...
-      </React.Fragment>
+      <>
+        <StackableBorderBox p={2} flexDirection="column">
+          <EntityOverline {...props} />
+          <Spacer mb={1} />
+          <ArtistBio
+            maxChars={200}
+            bio={
+              {
+                biography_blurb: { text: this.props.bio, credit: "Gagosian" },
```
oxaudo: This `Gagosian` should also come from props, right?

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -10,6 +9,7 @@ import {
   SelectedExhibitions,
   SelectedExhibitionsProps,
 } from "Styleguide/Components"
+import { EntityOverline } from "Styleguide/Components/EntityOverline"
```
oxaudo: Is the idea of making it very generic here is to reuse it for Partner Info above in the design? The name seem to be quite non descriptive. Maybe if this component is repeated for several pages/sections we can chat with @design team on some naming convention here.

src/Styleguide/Components/ArtistCard.tsx
```diff
@@ -61,23 +64,13 @@ export const LargeArtistCard = (props: Props) => (
       <FollowArtistButton
         artist={props.artist}
         user={props.user}
-        useDeprecatedButtonStyle={false}
-        buttonProps={{
-          variant: "secondaryOutline",
-          size: "small",
-          width: space(9),
-        }}
-        onOpenAuthModal={() => {
```
oxaudo: Nice refactor on moving it into function!

src/Styleguide/Components/EntityOverline.tsx
```diff
@@ -0,0 +1,78 @@
+import { Avatar, Box, Flex, Sans, Serif } from "@artsy/palette"
+import FollowArtistButton from "Components/FollowButton/FollowArtistButton"
+import React, { Component } from "react"
+// import { AvatarInitials } from "Styleguide/Components/AvatarInitials"
```
oxaudo: This is moved as part of avatar now, right?

src/Apps/Artwork/Components/ArtistInfo.tsx
damassi: This is based on a spec that @briansw came up: zpl.io/VKl7LWE

src/Apps/Artwork/Components/ArtistInfo.tsx
damassi: Oops, will update! 🍝

src/Styleguide/Components/EntityOverline.tsx
damassi: Yup -- will remove comment

src/Apps/Artwork/Components/ArtistInfo.tsx
damassi: I think that the word "Overline" is a bit confusing outside of the design world but we can all learn 😄

oxaudo: I'm all totally for learning this! My question was more about the definition of this overline. Should it be ArtistOverline or more generic EntityOverline? Seems like in this case we need to pass the follow button action somehow. We can not assume that it's an artist Follow there.

But also so far seems like the same design is in PartnerProvidedInfo above and ArtistInfo. But there was a comment during the review that we might actually want to distinguish Partner from Artist more here. So I was wondering if this overline style will be reused.

src/Styleguide/Components/EntityOverline.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface EntityOverlineProps {
+  imageUrl?: string
+  initials?: string
+  meta: string
```
oxaudo: This meta can also be null I believe.

src/Styleguide/Components/EntityOverline.tsx
```diff
@@ -0,0 +1,78 @@
...
+            <Serif size="3" weight="semibold" color="black100">
+              {name}
+            </Serif>
+
+            <Flex>
+              <Sans size="2" color="black60">
+                {meta}
+              </Sans>
+
+              {showFollow && (
```
oxaudo: So if meta can be null - in this case we have to test for both to determine if we need to insert `•` between

src/Apps/Artwork/Components/ArtistInfo.tsx
damassi: I believe the idea is that this will be reused

artsyit:

Built with commit ed558712e09105303fc15fc7d9bb6841b133cc83

https://deploy-preview-1275--artsy-reaction.netlify.com

damassi:
@oxaudo - addressed feedback and will follow up in second PR

PR #1277, opened 2018-09-21
[BREAKING] Don’t mix build server and client app

alloy:
* `Artsy/Router/index` still exists, but it no longer exports server or client specific code, that has to be imported from either `Artsy/Router/client` or `Artsy/Router/server`.
* Renames `StorybooksRouter` to `ClientRouter`.
* Fixes custom Relay linter rule and makes lint green again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@
 
 Circle CI is set up to publish releases to NPM automatically via [semantic-release](https://github.com/semantic-release/semantic-release) following every successful merge to master.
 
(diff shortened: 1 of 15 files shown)


artsyit:

Built with commit 6bc3ad1bb01be849313179f5612241219f80e316

https://deploy-preview-1277--artsy-reaction.netlify.com

damassi:
These two commits will trigger two separate feature releases: 

screen shot 2018-09-21 at 10 15 28 am

Is that intended? If not can you reword so that only one major version change goes out? 

Otherwise LGTM 👍 merge when ready

PR #1278, opened 2018-09-21
@xtina-starr => SectionContainer handles blockquote width

eessex:
Related to new `Text` component changes...

We have been attaching a `section.layout` of `blockquote` to article sections that contain `<blockquote>` in order to achieve a wider width. However, this adds a lot of unnecessary complexity to writer for a simple style attribute, and is also not currently working in Force. 

This is a quick refactor of the `SectionContainer` component, which handles assigning width and mobile padding to article sections. It removes some complexity relating to `props.type` and adds a check to `text` sections that adds wider widths if a `<blockquote>` block is present.

screen shot 2018-09-21 at 10 07 32


artsyit:

Built with commit 538809d5f6e9b0855a737f427193791579bad60a

https://deploy-preview-1278--artsy-reaction.netlify.com

PR #1279, opened 2018-09-21
Fixed a bug where the shipping address form is not shown

yuki24:
There is an edge case where the shipping address form is not shown properly:

 * User creates an order and chooses to pick up but does not submit the order
 * Gallery changes the artwork's shipping option to shipping only
 * User comes back to the Buy Now flow to change the shipping option
 * The shipping address form is not shown

The expected behaviour is the form should be shown properly regardless of the user's selection since now the artwork can not be picked up.


artsyit:

Built with commit 4b84fbfdefb47ef018339862d2fbd0e60acb08a8

https://deploy-preview-1279--artsy-reaction.netlify.com

PR #1284, opened 2018-09-21
Update status page depending on order status

zephraph:
Fleshes out the implementation of the status page to make it update based on the state of the order. Also fixes some minor UI issues like aligning the sidebar with the message box. 

- ~Make canceled error message dynamic~ (not yet implemented on the MP/exchange side)
- [x] Confirm copy
- [x] Add tests

image
image


artsyit:

Built with commit 848cddb245a45cd13d4a8a80663de312efd4957f

https://deploy-preview-1284--artsy-reaction.netlify.com

artsyit:

Built with commit 95e28440f33546f3015d06e514fe358413099b02

https://deploy-preview-1284--artsy-reaction.netlify.com

artsyit:

Built with commit 65e441e2a9f28ef7e1c15dc4d7083b9d75747efa

https://deploy-preview-1284--artsy-reaction.netlify.com

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -17,42 +17,102 @@ export interface StatusProps {
...
+  messageCopy = () => {
+    const { order } = this.props
+    const { state, requestedFulfillment } = order
+    switch (state) {
+      case "SUBMITTED":
+        return (
+          <>
+            Thank you for your order. You’ll receive a confirmation email
+            shortly. If you have questions, please contact{" "}
+            <a href="#">orders@artsy.net</a>.
```
sweir27: I believe this should be a `mailto` link

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -17,42 +17,102 @@ export interface StatusProps {
 }
 
 export class StatusRoute extends Component<StatusProps> {
+  stateCopy = () => {
+    const { state, requestedFulfillment } = this.props.order
+    switch (state) {
+      case "SUBMITTED":
```
sweir27: It feels like we should have a default for these switch statements. Do you think `SUBMITTED` makes sense for that?

src/Apps/Order/Routes/Status/index.tsx
sweir27: Ah, actually based on the updated copy doc I don't think we need to show the email at all anymore 😮.

zephraph: Yep, It's removed now.

src/Apps/Order/Routes/Status/index.tsx
zephraph: I had a default before, but I'm not sure... I mean, if some case gets introduced and it's not _really_ in a submitted state it doesn't seem like that would make sense. Maybe showing an error state or something.

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -17,42 +17,123 @@ export interface StatusProps {
...
+              <>Estimated delivery: {fulfillment.estimatedDelivery}</>
+            )}
+          </>
+        ) : (
+          false
+        )
+      case "CANCELED":
+        return (
+          <>
+            The work is no longer available.
```
zephraph: I need to grab this reason from somewhere, but I'm not exactly sure where the data should come from.

src/Apps/Order/Routes/Status/index.tsx
sweir27: That's okay. We can also cross this bridge when we get there

src/Apps/Order/Routes/Status/index.tsx
sweir27: Is the intent there to show the specific reason? Or always this message (this seems generic enough)? 

If the former, I think we can get it from exchange but I'm not sure how human-readable it is.

joanna20:
Looks good @zephraph! Only change:
- fulfilled (pickup) and  canceled (pickup) can we remove the body text "You'll be appointed an Artsy specialist..." in the right sidebar? https://cl.ly/80899794d9b3  (but keep everything else there)

src/Apps/Order/Routes/Status/index.tsx
zephraph: Yeah, that's what notion says about the copy. 

```
<Reason>

Please allow 5–7 business days for the refund to appear on your bank statement.
```

I haven't seen a way to get it currently... unless it returns as an error or something?

zephraph: This isn't actually implemented yet. I'm going to note it in the ticket and we'll just need to create a follow to wire it up after the implementation is complete.

src/Apps/Order/Components/ShippingAndPaymentSummary.tsx
```diff
@@ -23,11 +30,15 @@ export const ShippingAndPaymentSummary = ({
           title={
             <>Pick up ({lineItems.edges[0].node.artwork.shippingOrigin})</>
           }
+          /* Fixes spacing issues with title when no pickup description copy is present */
+          mb={hidePickupCopy(state) ? -1 : undefined}
```
alloy: beep boop 😎

src/__generated__/ShippingAndPaymentSummary_order.graphql.ts
```diff
@@ -5,6 +5,7 @@ import { ShippingAddress_ship$ref } from "./ShippingAddress_ship.graphql";
 declare const _ShippingAndPaymentSummary_order$ref: unique symbol;
 export type ShippingAndPaymentSummary_order$ref = typeof _ShippingAndPaymentSummary_order$ref;
 export type ShippingAndPaymentSummary_order = {
+    readonly state: string | null;
```
alloy: I’m really surprised that [this isn’t an enum](https://github.com/artsy/metaphysics/blob/5e70fdf813d582c54d70c073ad9a25d8e9abc585/src/schema/ecommerce/types/order.ts#L27-L30). Isn’t the possible states finite and known? If so, I’d urge y’all to update that before this gets shipped.

src/Apps/Order/Components/ShippingAndPaymentSummary.tsx
```diff
@@ -23,11 +30,15 @@ export const ShippingAndPaymentSummary = ({
...
         >
-          <Serif size="3t">
-            You’ll be appointed an Artsy specialist within 2 business days to
-            handle pickup logistics.
-          </Serif>
+          {!hidePickupCopy(state) && (
```
alloy: Nit: maybe a tad much negation? How about `showPickupCopy(state) && (…)`

src/__generated__/ShippingAndPaymentSummary_order.graphql.ts
zephraph: I 1000% agree. It actually _is_ an [enum in exchange](https://github.com/artsy/exchange/blob/014227c476f3d78283495bcfa85ec2ebd8a00087/_schema.graphql#L260-L281). Way too many string references that aren't tied to types used in this code currently... makes me a bit nervous. 😅

src/Apps/__stories__/OrderApp.story.tsx
```diff
@@ -35,21 +34,56 @@ const Router = props => (
 storiesOf("Apps/Order Page", module)
   .add("Shipping", () => <Router initialRoute="/order2/123/shipping" />)
   .add("Review", () => <Router initialRoute="/order2/123/review" />)
-  .add("Status", () => <Router initialRoute="/order2/123/status" />)
+
+storiesOf("Apps/Order Page/Status", module)
+  .add("submitted", () => (
+    <Router
+      initialRoute="/order2/123/status"
+      mockResolvers={mock({ ...OrderWithShippingDetails, state: "SUBMITTED" })}
```
sweir27: Thanks for extracting this as a function!

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -17,42 +17,123 @@ export interface StatusProps {
...
+          </>
+        ) : (
+          <>
+            Thank you for your purchase. A specialist will contact you within 2
+            business days to coordinate pickup.
+          </>
+        )
+      case "FULFILLED":
+        const fulfillment =
+          order.lineItems.edges[0].node.fulfillments.edges[0].node
```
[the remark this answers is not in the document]
zephraph: I actually started to do that, yeah. Only problem is there's not really a good way to typecheck that... then again, it is safer from a results perspective.. torn.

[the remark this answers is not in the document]
sweir27: Yeah, brought this up with @damassi too in another case-- we lose type-checking but feels better to be safe (alternatively would have to have a bunch of `&&`s which would get unwieldy).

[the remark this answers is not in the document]
sweir27: In this case since we eventually check for the presence of `fulfillment`, seems like we shouldn't bork here.

zephraph:
@sweir27 this is finally good to go >_>

Tests are up, but they're not using the `MockRelayRenderer`. @alloy and I paired on it, but it's a tough nut to crack. We'll get there.

PR #1287, opened 2018-09-24
[collect] Switch router from /collect2 -> /collect

l2succes:
Switches collect app to `/collect` and also makes sure filter changes are reflected in the URL


xtina-starr:
LGTM!

PR #1295, opened 2018-09-25
Fixed an issue where the height of <Collapse> is always fixed

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-501

The issue is that the `<Collapse>` component always holds a fixed height and it doesn't change the height when the elements inside of it stretches vertically. I just replaced the implementation with `react-spring` that doesn't have this issue.

## Screenshot

kapture 2018-09-25 at 15 18 05

### DIFF
--- a/src/Styleguide/Components/Collapse.tsx
+++ b/src/Styleguide/Components/Collapse.tsx
@@ -1,50 +1,33 @@
 import React from "react"
-import styled from "styled-components"
+import { animated, Spring } from "react-spring"
 
 export interface CollapseProps {
   open: boolean
 }
 
-export interface CollapseState {
-  computedHeight: string
-}
-
-export class Collapse extends React.Component<CollapseProps, CollapseState> {
+export class Collapse extends React.Component<CollapseProps> {
   state = {
-    computedHeight: null,
-  }
-
-  private element: HTMLElement
-
-  handleRef = element => {
-    this.element = element
+    mounted: false,
   }
 
   componentDidMount() {
-    if (this.state.computedHeight === null) {
-      const prevHeight = this.element.style.height
-      this.element.style.height = "auto"
-      const computedHeight = getComputedStyle(this.element).height
-      this.element.style.height = prevHeight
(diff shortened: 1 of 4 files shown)


src/Styleguide/Components/Collapse.tsx
```diff
@@ -1,50 +1,24 @@
 import React from "react"
-import styled from "styled-components"
+import { animated, Spring } from "react-spring"
```
damassi: 😄

damassi: Since we had all of that back and forth in the other PR that introduced `<Collapse>`, it might be good to resurface that discussion.

sweir27: Yeah was going to ask about that too-- for reference it is here: https://github.com/artsy/reaction/pull/1114#discussion_r209354107

sweir27:
Looks like there are some CI failures.

Beyond that if we feel confident in `react-spring` here then this looks 👍 to me.

yuki24:
Right, I think this is good for review. I have spiked on all three options we've come up with, and I believe `react-spring` is the way to go.

 * We can easily switch to [the `react-transition-group`-based implementation](https://github.com/yuki24/reaction/blob/fix-collapse-component-2/src/Styleguide/Components/Collapse.tsx) when we decide to step away from `react-spring`
 * `react-spring`'s simpler, declarative API is very easy to understand
 * We may get React Native support for free

I understand the size of the dependency, but at least in the `<Collapse>` component the implementation details are encapsulated and we have a path to go back and replace with a smaller dependency. With that in mind, I believe the risk of `react-spring` possibly being obsolete in the future is reasonably mitigated.

That being said, I'm still open to suggestions and feedback. Let me know what you think!

damassi:
🎉

yuki24:
Sweet, I think we all came up to the same conclusion. Thanks everyone for bearing with me!

merge on green

PR #1297, opened 2018-09-25
Adds HTML attributes for iOS web views

ashfurrow:
Just noticed the TypeScript errors, will pick it up from there tomorrow.

### DIFF
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -77,6 +77,8 @@ export class AddressForm extends React.Component<
             id="AddressForm_name"
             placeholder="Add full name"
             title="Full name"
+            autoCapitalize="words"
+            autoCorrect="off"
             defaultValue={this.props.defaultValue.name}
             onChange={this.changeEventHandler("name")}
             error={this.props.errors && this.props.errors.name}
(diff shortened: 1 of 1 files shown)


ashfurrow:
Looks like React has already added attributes to take care of this, but are camel-cased, so I just need to use those instead.

artsyit:

Built with commit f0a4f0eac9d4499671f3b2924e1bd868febb3195

https://deploy-preview-1297--artsy-reaction.netlify.com

ashfurrow:
We should go through and QA these with Brian. I've made decisions based on what I think would be the best iOS UX, but it's always better to have a second opinion.

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -77,6 +77,8 @@ export class AddressForm extends React.Component<
             id="AddressForm_name"
             placeholder="Add full name"
             title="Full name"
+            autoCapitalize="words"
+            autoCorrect="off"
```
sweir27: 🙌

sweir27:
This looks pretty straightforward! Agreed on syncing with @briansw on this so we don't have to come back later, but other than that is good with me. 😄

ashfurrow:
Okay, I'm going to merge since getting this onto staging is the easiest way for Brian to see it. I'll take care of following up.

PR #1301, opened 2018-09-25
Show specific shipping error if there is an invalid address

sweir27:
This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-442

It's unlikely that we would _ever_ get these errors in this current iteration, since we have strong client-side validation on this form. However, if we change those requirements (especially around things like zip code), that could change. Additionally, this adds the ability to customize the header and detail message for the error modal on the shipping page, which we'll need for subsequent tickets.

image


artsyit:

Built with commit 1d4868a286871088fa331e956e260f8c5061aa91

https://deploy-preview-1301--artsy-reaction.netlify.com

zephraph:
Looks good!

PR #1302, opened 2018-09-25
[FEATURE util] Add `get` helper and update ArtworkDetails with EntityHeader

damassi:
Adds a new `Utils/get.ts` function for safe access to deeply nested properties. Examples:

```ts
const deepObject = { a: { b: { c: { d: { e: "hi!" } } } } }
expect(get(deepObject, p => p.a.b.c.d.e)).toEqual("hi!")
```

or 

```ts
const locations = get(partner, p => filterLocations(p.locations), []).split(', ')
```

Also updates `ArtworkDetails` to use new `EntityHeader` component: 

screen shot 2018-09-25 at 4 49 18 pm


artsyit:

Built with commit 2e57ee9002878bcde834c446eb18cae3a48ac51a

https://deploy-preview-1302--artsy-reaction.netlify.com

src/Utils/__test__/get.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  it("allows arbitrary chaining", () => {
+    expect(
+      get(deepObject, p => (p as any).a.b[10].baz, ["fallback", "value"]).join(
+        " "
+      )
+    ).toEqual("fallback value")
+  })
+
+  it("can pass types", () => {
```
damassi: Meant just as a usage example

src/Utils/get.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Thanks! https://github.com/claudepache/es-optional-chaining/issues/2#issuecomment-318680932
+ *
+ * @param obj                     Object to get deep property
+ * @param unsafeDataOperation     Function that returns the deep property
+ * @param valueIfFailOrUndefined  Value to return in case if there is no such property
+ */
+export function get<O, T>(
+  obj: O,
+  unsafeDataOperation: (x: O) => T,
+  valueIfFailOrUndefined?: any
```
alloy: Shouldn’t this be `T` instead of `any`?

src/Utils/get.ts
```diff
@@ -0,0 +1,26 @@
...
+): T {
```
alloy: Seeing as the default value for the fallback is `undefined` this return type should reflect that `T | undefined`

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -22,35 +24,24 @@ export class ArtworkDetailsAboutTheWorkFromPartner extends React.Component<
...
-      partner &&
-      partner.locations &&
-      partner.locations.length > 0 &&
-      filterLocations(partner.locations)
+
+    const locationNames = get(
+      partner,
+      p => filterLocations(p.locations),
+      []
+    ).join(", ")
```
alloy: I personally find this to be less readable than the previous version. Does it functionally protect against more?

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -22,35 +24,24 @@ export class ArtworkDetailsAboutTheWorkFromPartner extends React.Component<
...
+
+    const imageUrl = get(partner, p => p.profile.image.url, false)
```
alloy: This looks readable 👌

I am wondering about that fallback value, seems a bit weird to have a string or false.

src/Utils/get.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Type-safe access of deep property of an object
+ *
+ * Thanks! https://github.com/claudepache/es-optional-chaining/issues/2#issuecomment-318680932
+ *
+ * @param obj                     Object to get deep property
+ * @param unsafeDataOperation     Function that returns the deep property
+ * @param valueIfFailOrUndefined  Value to return in case if there is no such property
+ */
+export function get<O, T>(
+  obj: O,
```
alloy: I wonder if the input object is even really necessary? We really only care about `T` and can’t we have that by doing just `get(() => obj.deeply.nested.access ))`?

src/Styleguide/Components/EntityHeader.tsx
```diff
@@ -31,9 +31,11 @@ export class EntityHeader extends Component<EntityHeaderProps> {
             </Serif>
 
             <Flex>
-              <Sans size="2" color="black60">
-                {meta}
-              </Sans>
+              {!!meta && (
```
oxaudo: @damassi - curious why in the EntityHeader some properties are defined as optional, like `initials?: string` where meta is defined as `meta: string | null`. Why not `meta?: string`?

src/Utils/get.ts
zephraph: That's true I guess... it does somewhat enforce that you understand what object you're acting on though. A bit more explicit, but that's not always a bad thing.

src/Utils/get.ts
zephraph: It could be `T`... I suppose that would make sense. Then it would enforce that your fallback case matches what you expect the function to give...

Yeah, I'm 👍

damassi: Will play with it a bit, this was mostly copied over from the example (linked at the top) 👍

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
damassi: I think  prettier breaking it out onto multiple lines makes it a bit weird, but the previous version was quite unwieldy, with two blocks required to ensure that what's returned is valid in order to join: 

```js
const locationNames =
      partner &&
      partner.locations &&
      partner.locations.length > 0 &&
      filterLocations(partner.locations)

...

{locationNames &&
   locationNames.length > 0 && (
     <Serif size="3" color="black60">
       {locationNames.join(", ")}
     </Serif>
)}
```

which now becomes basically a one-liner: 

```js
const locationNames = get(partner, p => filterLocations(p.locations), []).join(", ")
```

src/Utils/get.ts
damassi: `get(() => obj.deeply.nested.access ))` is interesting but I think the input / output approach maps well with other common implementations.

src/Styleguide/Components/EntityHeader.tsx
damassi: This is carryover from your previous feedback here: https://github.com/artsy/reaction/pull/1275#discussion_r219944642. I can update though

src/Utils/get.ts
alloy: Sounds good 👍

zephraph:
I'm definitely 👍 on this. Sounds like there still might be some exploration around the failure return and `T` though. @damassi once you're done with that I think it looks good to go.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
alloy: I personally still prefer the former, but that doesn’t really matter for this one example. My main point is that I don’t feel like we should necessarily go ahead and replace all places with this, but make a good judgement on wether or not this runtime safety is really necessary.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
damassi: Updating `valueIfFailOrUndefined?: T` (from `any` to `T`) actually caught this bug nicely; updated. (Also good to note that we don't need a fallback value in that third spot.)

PR #1309, opened 2018-09-26
[FEATURE BNMO] disable country field when work is continental shipping only

javamonn:
Implements [PURCHASE-477](https://artsyproduct.atlassian.net/browse/PURCHASE-477) by disabling the country field on the Shipping panel when a work is continental US shipping only. A non-US billing address can still be selected.

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -1032,7 +1032,7 @@ type ArtworkContextFair {
     section: String
 
     # Sorts for shows in a fair
-    sort: ShowSort = null
+    sort: ShowSort
     after: String
     first: Int
     before: String
@@ -1964,7 +1964,7 @@ type Collection implements Node {
     before: String
     last: Int
     private: Boolean = false
-    sort: CollectionSorts = null
+    sort: CollectionSorts
   ): ArtworkConnection
   description: String!
   default: Boolean!
@@ -2705,7 +2705,7 @@ type Fair {
     section: String
 
     # Sorts for shows in a fair
-    sort: ShowSort = null
+    sort: ShowSort
     after: String
     first: Int
     before: String
@@ -3661,7 +3661,7 @@ type HomePageModuleContextFair {
     section: String
 
     # Sorts for shows in a fair
-    sort: ShowSort = null
+    sort: ShowSort
     after: String
     first: Int
     before: String
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -1,4 +1,4 @@
-import { Flex, Join, Serif, Spacer } from "@artsy/palette"
+import { Flex, Join, Sans, Serif, Spacer } from "@artsy/palette"
 import Input from "Components/Input"
 import React from "react"
 import { CountrySelect } from "Styleguide/Components"
@@ -35,6 +35,7 @@ export interface AddressFormProps {
   onChange: AddressChangeHandler
   defaultValue?: Partial<Address>
   billing?: boolean
+  continentalUsOnly?: boolean
   errors?: AddressErrors
 }
 
@@ -70,6 +71,7 @@ export class AddressForm extends React.Component<
   }
 
   render() {
+    const lockCountryToUS = !this.props.billing && this.props.continentalUsOnly
     return (
       <Join separator={<Spacer mb={2} />}>
         <Flex flexDirection="column">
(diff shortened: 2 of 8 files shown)


data/schema.graphql
```diff
@@ -810,6 +810,9 @@ type Artwork implements Node & Sellable {
   price: String
   price_currency: String
 
+  # Is this work available for shipping only within the Contenental US?
```
javamonn: Should these changes be here, or are they synced as part of a deploy process?

artsyit:

Built with commit c9e8f0813321389ffe9a666a8dd145902a37bece

https://deploy-preview-1309--artsy-reaction.netlify.com

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -92,9 +93,23 @@ export class AddressForm extends React.Component<
               Country
             </Serif>
             <CountrySelect
-              selected={this.state.address.country}
+              selected={
```
javamonn: I played with the idea of keeping the `AddressForm` agnostic about what it means to lock addresses to the continental US, but couldn't figure out a nice enough way to do it, I'm open to suggestions though. I was considering passing in objects of `disabled` and `description` similar to how we do for errors, so that the consumer could make the form continental US only by setting US as the default, disabling the field, and setting "Ships to continental US only" string as the description. It seemed relatively heavy though.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -213,6 +213,14 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
       this.props,
       "order.lineItems.edges[0].node.artwork.pickup_available"
     )
+    const isShippingContinentalUsOnly = get(
```
javamonn: I'll update tomorrow to use our new in-house `get`...

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -36,6 +36,7 @@ import {
 } from "react-relay"
 import { Collapse } from "Styleguide"
 import { Col, Row } from "Styleguide/Elements/Grid"
+import { get } from "Utils/get"
```
zephraph: 🎉

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -92,9 +93,23 @@ export class AddressForm extends React.Component<
...
+                !this.props.billing && this.props.continentalUsOnly
+                  ? "US"
+                  : this.state.address.country
+              }
               onSelect={this.changeValueHandler("country")}
+              disabled={!this.props.billing && this.props.continentalUsOnly}
             />
+            {!this.props.billing &&
```
erikdstock: This check happens 3 times- maybe we could move it to the top of `render()`?

javamonn:
Added a test here to make sure the `disabled` prop makes its way down correctly. Here's a screenshot of `Shipping` with a continental US only artwork:

screen shot 2018-10-01 at 5 14 15 pm

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -210,7 +211,15 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
-      "order.lineItems.edges[0].node.artwork.pickup_available"
+      props => order.lineItems.edges[0].node.artwork.pickup_available
+    )
+    const shipsToContinentalUSOnly = get(
+      this.props,
+      props => order.lineItems.edges[0].node.artwork.shipsToContinentalUSOnly
+    )
+    const artworkId = get(
+      this.props,
+      props => order.lineItems.edges[0].node.artwork.id
```
alloy: Looks like these could all be replaced with a single `get` for the artwork, no?

alloy:
Looks good to me! Just one minor comment about simplifying the amount of `get` usage.

alloy:
Btw, I don’t believe our semantic-release setup uses the format you’re using, it should probably be: `[FEATURE topic] Summary`

src/Apps/Order/Components/AddressForm.tsx
sweir27: I think logic-wise this is OK, but `disableCountrySelect` makes this read in a confusing way (since it's just just about putting a `disabled={true}` on the country field). Can we rename to something a little more descriptive? `lockCountryToUS`? (this is a hard one)

PR #1311, opened 2018-09-27
[FEATURE middleware] Add errorMiddleware

damassi:
Adds better logging in development around MP errors: 

screen shot 2018-09-26 at 10 23 35 pm

This should help surface common issues related to expired authTokens, 404s, etc. 

Also, turned off relay logging in prod.

### DIFF
--- a/src/Artsy/Relay/createEnvironment.ts
+++ b/src/Artsy/Relay/createEnvironment.ts
@@ -10,10 +10,22 @@ import { NetworkError } from "Utils/errors"
 
 import {
   cacheMiddleware,
+  errorMiddleware,
   loggerMiddleware,
   RelayNetworkLayer,
   urlMiddleware,
 } from "react-relay-network-modern"
+
+const isServer = typeof window === "undefined"
+const isDevelopment =
+  (isServer ? process.env.NODE_ENV : sd.NODE_ENV) === "development"
+
+// Only log on the client during development
+const loggingEnabled = isDevelopment && !isServer
+
+const METAPHYSICS_ENDPOINT = isServer
+  ? process.env.METAPHYSICS_ENDPOINT
+  : sd.METAPHYSICS_ENDPOINT
 
 const USER_AGENT = `Reaction/${ReactionVersion}`
 
@@ -30,7 +42,7 @@ interface RelayEnvironment extends Environment {
 
 export function createEnvironment(config: Config = {}) {
   const { cache = {}, checkStatus, user, relayNetwork } = config
-  const isServer = typeof window === "undefined"
+
   const relaySSRMiddleware = isServer
     ? new RelayServerSSR()
     : new RelayClientSSR(cache)
@@ -57,15 +69,11 @@ export function createEnvironment(config: Config = {}) {
     console.warn("Browser does not support i18n API, not setting TZ header.")
   }
 
-  const url = isServer
-    ? process.env.METAPHYSICS_ENDPOINT
-    : sd.METAPHYSICS_ENDPOINT
-
   const network =
     relayNetwork ||
     new RelayNetworkLayer([
       urlMiddleware({
-        url,
+        url: METAPHYSICS_ENDPOINT,
         headers: !!user
           ? {
               ...headers,
@@ -95,8 +103,8 @@ export function createEnvironment(config: Config = {}) {
           throw error
         }
       },
-      // TODO: Audit logging
-      !isServer && loggerMiddleware(),
+      loggingEnabled && loggerMiddleware(),
+      loggingEnabled && errorMiddleware(),
     ])
 
   const source = new RecordSource()
--- a/typings/sharify.d.ts
+++ b/typings/sharify.d.ts
@@ -19,6 +19,7 @@ declare module "sharify" {
       readonly FORCE_CLOUDFRONT_URL: string
       readonly GEMINI_CLOUDFRONT_URL: string
       readonly METAPHYSICS_ENDPOINT: string
+      readonly NODE_ENV: string
       readonly XAPP_TOKEN: string
     }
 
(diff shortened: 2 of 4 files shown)


artsyit:

Built with commit d1d1ec17c5309602daf27f301d728d22b6ee362c

https://deploy-preview-1311--artsy-reaction.netlify.com

src/Artsy/Relay/createEnvironment.ts
```diff
@@ -10,11 +10,20 @@ import { NetworkError } from "Utils/errors"
...
   urlMiddleware,
 } from "react-relay-network-modern"
 
+const isServer = typeof window === "undefined"
+const isDevelopment =
+  (isServer ? process.env.NODE_ENV : sd.NODE_ENV) === "development"
+
+const METAPHYSICS_ENDPOINT = isServer
+  ? process.env.METAPHYSICS_ENDPOINT
+  : sd.METAPHYSICS_ENDPOINT
```
damassi: Still need to set down a nicer pattern for retrieving client / server values so that these ternaries aren't necessary

src/Artsy/Relay/createEnvironment.ts
```diff
@@ -30,13 +39,21 @@ interface RelayEnvironment extends Environment {
...
 
   relaySSRMiddleware.debug = false
 
+  // Only log on the client during development
+  let loggingEnabled = isDevelopment && !isServer
+
+  // Add client window hook to debug logging in prod
+  if (!isServer) {
+    window._toggleRelayLogging = () => (loggingEnabled = !loggingEnabled)
+  }
```
damassi: @alloy figured its time to turn off logging in prod since things have settled a bit post /artist launch but wanted a way to toggle on if need be

src/Artsy/Relay/createEnvironment.ts
alloy: We already have that, you should only use the value from `sd` https://github.com/artsy/reaction/blob/8f6ac8b73b1c6cb8e79985a77ca86d3f6d06d342/.storybook/webpack.config.js#L33

and we should probably move all other `process.env` occurrences in the codebase to that too.

typings/sharify.d.ts
```diff
@@ -19,6 +19,7 @@ declare module "sharify" {
       readonly FORCE_CLOUDFRONT_URL: string
       readonly GEMINI_CLOUDFRONT_URL: string
       readonly METAPHYSICS_ENDPOINT: string
+      readonly NODE_ENV: string
```
alloy: 👍 Now the only thing missing is adding it to the `sharify` stub for development https://github.com/artsy/reaction/blob/8f6ac8b73b1c6cb8e79985a77ca86d3f6d06d342/.storybook/webpack.config.js#L31-L34

alloy:
Nice! 👍 

Only thing left is that there shouldn’t be any new `process.env` introductions to the codebase. If anything we need to start removing the ones that still exist.

src/Artsy/Relay/createEnvironment.ts
```diff
@@ -10,11 +10,23 @@ import { NetworkError } from "Utils/errors"
...
   RelayNetworkLayer,
   urlMiddleware,
 } from "react-relay-network-modern"
 
+const isServer = typeof window === "undefined"
+const isDevelopment =
+  (isServer ? process.env.NODE_ENV : sd.NODE_ENV) === "development"
+
+// Only log on the client during development
+const loggingEnabled = isDevelopment && !isServer
```
alloy: Why not on the server in development?

eessex: Agree, aren't there cases where logging would be of interest on the server side?

damassi: It's pretty noisy, especially when working in force. Since most of these logs are duplicated when we work in Storybooks on the client figured it was unnecessary and didn't feel too useful. I think a PR back to react-relay-network-modern that trims the amount of data being sent down the wire might help in this regard.

src/Artsy/Relay/createEnvironment.ts
damassi: Will also need to figure out how to pipe this sharify config setup through to our tests

damassi: @alloy - this is a larger task than expected due to tests as well as ensuring that things work properly over in force. Cool if I push this to a separate PR?

damassi:
@alloy - Updating `process.env` throughout codebase is a larger task that I would like to handle in a second PR if possible as there are implications for our tests -- cool? 

The last remaining question is around server-side logging in dev, which I haven't found to be useful as we do a majority of our development in Storybooks. We should revisit this once the new reaction-side router shell is built.

src/Artsy/Relay/createEnvironment.ts
alloy: Oh were you thinking of those cases where server request data is shared to the client via the import? Because global data like these can be considered immutable, they are not set/overwritten on each request.

alloy: (Hence the separate “Global” and readonly typing of these)

alloy:
Sounds good to me 👍

src/Artsy/Relay/createEnvironment.ts
damassi: Right right, i was just wondering about precident around removing all `process.env` across codebase and replacing with `sd`. What if we're dealing with user data and another dev uses `sd` in an incorrect way, leading to singleton issues? That's what i meant by finding a nice seamless way to handle each case where we don't have to keep these things in mind.

alloy: Yeah good point, we should have another think when we consider eventually deprecating Force. I think the we way we’ve done it in Reaction (see old LoyaltyApp code) was a good middle ground, as iirc it made sure you couldn’t use the wrong data in the wrong place.

PR #1321, opened 2018-09-28
Handle known errors while submitting an order

sweir27:
This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-443

We want to handle the following errors (aside from just showing the generic modal):
- `insufficient_inventory`--> show a unique message + link user back to the artist page if they hit "continue"
- `artwork_version_mismatch` --> show a unique message + link user back to the artwork page if they hit "continue"
- `failed_charge_authorize` --> show a unique message (including what we get back from Stripe)

screen shot 2018-09-28 at 9 35 18 am
screen shot 2018-09-28 at 9 36 02 am

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -277,7 +277,7 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
             this.onMutationError(
               errors || creditCardOrError.mutationError,
               creditCardOrError.mutationError &&
-                creditCardOrError.mutationError.message
+                creditCardOrError.mutationError.detail
             )
           }
         },
--- a/src/Apps/Order/Routes/Review/index.tsx
+++ b/src/Apps/Order/Routes/Review/index.tsx
@@ -5,7 +5,7 @@ import { BuyNowStepper } from "Apps/Order/Components/BuyNowStepper"
 import { ItemReviewFragmentContainer as ItemReview } from "Apps/Order/Components/ItemReview"
 import { ShippingAndPaymentReviewFragmentContainer as ShippingAndPaymentReview } from "Apps/Order/Components/ShippingAndPaymentReview"
 import { ErrorModal } from "Components/Modal/ErrorModal"
-import { Router } from "found"
+import { RouteConfig, Router } from "found"
 import React, { Component } from "react"
 import {
   commitMutation,
@@ -14,6 +14,7 @@ import {
   RelayProp,
 } from "react-relay"
 import { Col, Row } from "Styleguide/Elements/Grid"
+import { get } from "Utils/get"
 import { Responsive } from "Utils/Responsive"
 import { Helper } from "../../Components/Helper"
 import { TransactionSummaryFragmentContainer as TransactionSummary } from "../../Components/TransactionSummary"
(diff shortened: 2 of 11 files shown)


artsyit:

Built with commit 7f80fe898702423cf5e7231efb7bc2e4a9aca449

https://deploy-preview-1321--artsy-reaction.netlify.com

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -263,7 +263,7 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
             this.onMutationError(
               errors || creditCardOrError.mutationError,
               creditCardOrError.mutationError &&
-                creditCardOrError.mutationError.message
+                creditCardOrError.mutationError.detail
```
sweir27: This fixes an error I saw during the QA session yesterday-- the helpful message from Stripe actually lives in the `detail` in this case.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -80,12 +122,39 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
...
   }
 
-  private onMutationError(errors, errorModalMessage?) {
+  private routeToArtworkPage() {
+    const artworkId = get(
+      this.props.order,
+      o => o.lineItems.edges[0].node.artwork.id
+    )
+    // Don't confirm whether or not you want to leave the page
+    this.props.route.onTransition = () => null
```
sweir27: Faffed with this a little bit, but this was the cleanest way I could find to ignore the transition hook behavior just in this case.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -66,12 +71,49 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
...
+                submitOrder: { orderOrError },
+              } = result
+
+              const error = orderOrError.error
+              if (error) {
+                switch (error.code) {
+                  case "insufficient_inventory":
+                    const artistId = get(
+                      this.props.order,
+                      o => o.lineItems.edges[0].node.artwork.artist.id
```
ashkan18: what happens when there are multiple artists? https://github.com/artsy/metaphysics/blob/master/src/schema/artwork/index.ts#L78

src/Apps/Order/Routes/Review/index.tsx
zephraph: Interesting... We have a cleanup method in OrderApp, but it's not necessarily available here... hmm... 

https://github.com/artsy/reaction/blob/master/src/Apps/Order/OrderApp.tsx#L46

sweir27: Yeah :/ if you think there's a better pattern to adopt let me know. In their example they delegate all handling to the children, but that doesn't seem much better than what we're doing here.

src/Apps/Order/Routes/Review/index.tsx
sweir27: We'll have to choose one anyway-- `artist` grabs the first artist when it's returned from gravity, but I can explicitly query for `artists` here and grab the first one client-side to be more explicit.

src/Apps/Order/Routes/Review/index.tsx
zephraph: Nope, this is definitely good enough. I don't think it's worth the effort.

PR #1328, opened 2018-09-28
Updates the cancel copy to be more generic

zephraph:
This PR updates the cancellation message to be generic. This has evolved in a recent conversation between @briansw, @joanna20, and myself from where it was at previously. We had previously detailed a cancellation reason, but given the ambiguity and unknowns around the cases when an artwork is rejected by a seller, we've (at least temporarily) adopted this messaging. 

**NEW**
image

**OLD**
image

### DIFF
--- a/src/Apps/Order/Routes/Status/index.tsx
+++ b/src/Apps/Order/Routes/Status/index.tsx
@@ -91,11 +91,10 @@ export class StatusRoute extends Component<StatusProps> {
       case "CANCELED":
         return (
           <>
-            The work is no longer available.
-            <br />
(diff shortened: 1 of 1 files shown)


zephraph:
@erikdstock If this PR is accepted then we'll likely need to update the messaging in pulse to match.

artsyit:

Built with commit 1b2684f605d0cf6237e1c30d6dfaac72c5117b6f

https://deploy-preview-1328--artsy-reaction.netlify.com

sweir27:
Looks good to me!

PR #1331, opened 2018-09-28
[FIX ArtworkDetails] Need to use icon for partner profile instead of image

oxaudo:
Should fix `DISCO-331`

Just followed what is currently displayed in Force for this image and figured we are displaying icon instead on image.


artsyit:

Built with commit 82847c162b940d860a40af5cc8cdead80d778bb7

https://deploy-preview-1331--artsy-reaction.netlify.com

PR #1334, opened 2018-09-28
User sees appropriate error page when trying to access an order they don't have access to

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-351

We discussed showing the header or not with @briansw @zephraph and @sweir27 and decided this is ok for now.

# Screenshot

screen shot 2018-09-28 at 7 00 17 pm

### DIFF
--- a/src/Apps/Order/OrderApp.tsx
+++ b/src/Apps/Order/OrderApp.tsx
@@ -1,6 +1,7 @@
 import { Spacer } from "@artsy/palette"
 import { routes_OrderQueryResponse } from "__generated__/routes_OrderQuery.graphql"
 import { ContextConsumer } from "Artsy/SystemContext"
+import { ErrorPage } from "Components/ErrorPage"
 import { Location, RouteConfig, Router } from "found"
 import React from "react"
 import { Meta, Title } from "react-head"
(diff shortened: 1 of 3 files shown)


src/Apps/Order/OrderApp.tsx
```diff
@@ -82,13 +83,18 @@ export class OrderApp extends React.Component<OrderAppProps, OrderAppState> {
   render() {
     const { children, location, router, order, params } = this.props
 
+    if (!order) {
+      return <ErrorPage code={404} />
+    }
```
yuki24: @zephraph As we discussed we can't properly return `404` since we don't have access to the `errors`. When the user is not logged in [we can (will) handle that on the Force side](https://github.com/artsy/force/pull/2946), so I'm not extremely concerned about it at this point.

alloy: If you want to know for sure, you can add an `orderOrError` field.

In fact, I will work on this if nobody else does, because I want a real example for a conf talk I’m working on.

zephraph: 👍 Sounds like that can be a follow up item then.

zephraph:
There's a type failure happening that doesn't seem related to this ticket. I'll fix that in a separate PR before merging.

ashfurrow:
Looks like the same error that was found and fixed on master yesterday, rebasing should fix.

artsyit:

Built with commit 5eb1d1eee7f794e94de54e149d302ae9b528e9c8

https://deploy-preview-1334--artsy-reaction.netlify.com

PR #1335, opened 2018-10-01
 [FIX ArtworkDetails] Support proper tab selection when first tab is null

oxaudo:
Investigating this: https://artsyproduct.atlassian.net/browse/DISCO-339 figured out that there are 2 issues here:
- first we were crashing the tab when article did not have an image
- we were not supporting empty tabs in tabs arrangement.

This is an attempt to fix both:)


artsyit:

Built with commit c8fb1f8535b19bc3681d98b3c8d7c7f4cdac8e2e

https://deploy-preview-1335--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsArticles.tsx
```diff
@@ -22,10 +23,11 @@ export const ArtworkDetailsArticles: React.SFC<
       <Col>
         <Box>
           {articles.map((article, index) => {
+            const imageUrl = get(article, p => p.thumbnail_image.resized.url)
```
zephraph: 🙌 I love it so much. @damassi getting some good usage out of this `get` function.

src/Styleguide/Components/Tabs.tsx
```diff
@@ -55,7 +55,8 @@ export class Tabs extends React.Component<TabsProps, TabsState> {
   constructor(props) {
     super(props)
 
-    const activeTabIndex = props.initialTabIndex || 0
+    const activeTabIndex =
+      props.initialTabIndex || this.props.children.findIndex(tab => tab != null)
```
zephraph: Couldn't it be the case that `tab` is reported as `false` (in the case when you don't want to render a certain tab) and thereby causing this logic to report a `tab` that's not really a `tab`?

src/Styleguide/Components/__stories__/Tabs.story.tsx
```diff
@@ -90,6 +90,13 @@ storiesOf("Styleguide/Components", module).add("Tabs (Simple)", () => {
           />
         </Tabs>
       </Section>
+      <Section title="Supports empty(null) tabs by ignoring them">
+        <Tabs>
+          {null}
```
zephraph: i.e. if this was `false`... (see above comment)

zephraph:
Minor comment about the initialTabIndex calculation, but otherwise it looks pretty good. It would be good to write a test for it though.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsArticles.tsx
damassi: lol this `get` was a long time coming....

damassi:
Second @zephraph on adding a test -- it looks like we already have some going [here](https://github.com/artsy/reaction/blob/master/src/Styleguide/Components/__tests__/Tabs.test.tsx)

src/Styleguide/Components/__stories__/Tabs.story.tsx
oxaudo: Yea... I guess we should also check for false. Will update in a min.

Theoretically there might be the case where all the tabs are null/false and in this case selected is going to be -1, but I don't think it's a realistic scenario anyways.

PR #1345, opened 2018-10-02
[BREAKING DevTools] Adds tools for testing async trees, such as Relay ones.

alloy:
* Consolidate existing dev tools into single dir. Including the previously exported `ClientRouter`, because that uses our schema and would implicitly pull that into our JS bundles, so we should revisit that when we really need it somewhere at runtime.
* Add [`renderUntil`](https://github.com/artsy/reaction/pull/1345/files#diff-99ca552957c29b34c115ada9663e43b7) which is used to test an async tree.
* Add [`renderRelayTree`](
https://github.com/artsy/reaction/pull/1345/files#diff-672b34f9f1a8662fbbc556b68ef79a71) which is used to test Relay trees.


package.json
```diff
@@ -37,7 +37,7 @@
     "stylelint-project": " stylelint -- 'src/**/*.{ts,tsx}'",
     "prettier-project": "yarn run prettier-write 'src/**/*.{ts,tsx}'",
     "prettier-write": "yarn run prettier --write",
-    "relay": "relay-compiler --src ./src --schema data/schema.graphql --language typescript --artifactDirectory ./src/__generated__",
+    "relay": "relay-compiler --src ./src --schema data/schema.graphql --language typescript --artifactDirectory ./src/__generated__ --exclude '**/node_modules/****/__mocks__/**,**/__generated__/**'",
```
damassi: What does `****` and `**, **` do?

src/Apps/Order/Routes/__tests__/Status.test.tsx
```diff
@@ -1,32 +1,56 @@
...
-import { HeadProvider } from "react-head"
-import { Status_order } from "../../../../__generated__/Status_order.graphql"
+import { graphql } from "react-relay"
 import {
+  mockResolver,
   OrderWithShippingDetails,
   PickupOrder,
 } from "../../../../Apps/__test__/Fixtures/Order"
-import { StatusRoute } from "../Status"
+import { MockBoot, renderRelayTree, UntilCallback } from "../../../../DevTools"
```
damassi: Once TypeScript support is fixed in our tests these auto imports will be absolute

artsyit:

Built with commit af7b25ef5fc79c1646542073d4c08a3b66153fbc

https://deploy-preview-1345--artsy-reaction.netlify.com

package.json
alloy: Hmmmm, this is copy-🍝 from the `relay-compiler` source which was excluding tests, but I think I missed inserting a comma between `****`, these should be 3 exclude patterns.

Great catch! 👌

src/Apps/Order/Routes/__tests__/Status.test.tsx
```diff
@@ -1,32 +1,56 @@
...
+        }
+      `,
+      mockResolvers: mockResolver(order),
+      wrapper: renderer => (
+        <MockBoot breakpoint="xs" headTags={headTags}>
+          {renderer}
+        </MockBoot>
+      ),
+    })
+  }
```
damassi: 👍

src/Apps/Order/Routes/__tests__/Status.test.tsx
```diff
@@ -1,32 +1,56 @@
...
-  const getWrapper = (order: Status_order, headTags = []) =>
-    mount(
-      <HeadProvider headTags={headTags}>
-        <StatusRoute order={order} />
-      </HeadProvider>
-    )
+  const getWrapper = (
+    order: any,
+    headTags: Array<React.ReactElement<any>> = [],
+    until: UntilCallback = null
```
damassi: What does `UntilCallback` do?

src/DevTools/MockRelayRenderer.tsx
```diff
@@ -0,0 +1,225 @@
...
+      Component={Component}
+      mockResolvers={mockResolvers}
+      query={query}
+    />
+  )
+  return renderUntil<P, S, C>(
+    until || (tree => !tree.find(`.${LoadingClassName}`).length),
+    wrapper ? wrapper(renderer) : renderer
+  )
+}
```
alloy: I should move this function out into its own module.

src/setup_jest.ts
```diff
@@ -9,6 +9,12 @@ track.mockImplementation(y => x => x)
 
 jest.mock("react-sizeme", () => jest.fn(c => d => d))
 
+/**
+ * We want each test to have assertions, otherwise it’s too easy to write async
+ * tests that never end up making any, leading to false positives.
+ */
+beforeEach(() => expect.hasAssertions())
```
alloy: We should do this in all our test suites. /cc @orta

src/Styleguide/Components/__tests__/ArtistBio.test.tsx
```diff
@@ -19,7 +19,8 @@ describe("ArtistBio", () => {
     expect(wrapper.find("ReadMore").length).toBe(1)
   })
 
-  it("triggers callback when clicked", done => {
+  // TODO: Chris, this test needs finishing.
+  xit("triggers callback when clicked", done => {
```
alloy: cc @damassi

src/Styleguide/Components/__tests__/CountrySelect.test.tsx
```diff
@@ -3,7 +3,8 @@ import React from "react"
 import { CountrySelect } from "../CountrySelect"
 
 describe("CountrySelect", () => {
-  it("triggers callback on change", done => {
+  // TODO: Chris, this test needs finishing.
+  xit("triggers callback on change", done => {
```
alloy: cc @damassi

src/Apps/__stories__/Apps.story.tsx
```diff
@@ -8,15 +8,15 @@ import { routes as collectRoutes } from "../Collect/routes"
 storiesOf("Apps", module)
   .add("Artwork Page", () => {
     return (
-      <ClientRouter
+      <MockRouter
```
damassi: 👍 I guess if we need a quick client-side router its easy enough to use `buildClientRouter` directly

tslint.json
```diff
@@ -21,7 +21,8 @@
       { "artifactDirectory": "__generated__", "makeRelative": false }
     ],
     "trailing-comma": false,
-    "variable-name": [true, "ban-keywords", "allow-leading-underscore"]
+    "variable-name": [true, "ban-keywords", "allow-leading-underscore"],
+    "jsdoc-format": false
```
alloy: I’m excluding the asterisks in `@example` docs because of https://github.com/Microsoft/TypeScript/issues/23517

src/Apps/Artwork/Utils/__tests__/filterLocations.test.tsx
```diff
@@ -3,12 +3,12 @@ import { filterLocations } from "../filterLocations"
...
   it("returns null if no array of locations is given", () => {
     const filtered = filterLocations(null)
-    expect(filtered).toBeNull
+    expect(filtered).toBeNull()
   })
 
   it("returns null if empty array of locations is given", () => {
     const filtered = filterLocations([])
-    expect(filtered).toBeNull
+    expect(filtered).toBeNull()
```
alloy: Such Ruby! 😅

src/Apps/Order/Routes/__tests__/Status.test.tsx
alloy: See the doc in `renderUntil` https://github.com/artsy/reaction/pull/1345/files#diff-99ca552957c29b34c115ada9663e43b7R33

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -0,0 +1,57 @@
...
+      //   // TODO: Even though this code path is reached, the retry button keeps spinning. iirc it _should_ disappear when
+      //   //      `onRetry` on the instance is unset.
+      //   //
+      //   // This will re-use the native view first created in the renderFailure callback, which means it can
+      //   // continue its ‘retry’ animation.
+      //   return <LoadFailureView style={{ flex: 1 }} />
+      // } else {
+      //   retrying = true
+      //   return <LoadFailureView onRetry={retry} style={{ flex: 1 }} />
+      // }
```
alloy: @damassi cc

src/DevTools/__test__/MockRelayRenderer.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+      expect(tree.html()).toEqual(
+        renderToString(
+          <div>
+            <img src="http://test/image.jpg" />
+            <div>Mona Lisa</div>
+          </div>
+        )
+      )
+      done()
+    }, 10)
```
damassi: Will this ever unexpectedly fail in the future?

src/Styleguide/Components/RecentlyViewed.tsx
```diff
@@ -105,13 +106,7 @@ export const RecentlyViewedQueryRenderer = () => {
...
               }
             `}
-            render={({ props }) => {
-              if (props) {
-                return <RecentlyViewedFragmentContainer me={props.me} />
-              } else {
-                return null
-              }
-            }}
+            render={renderWithLoadProgress(RecentlyViewedFragmentContainer)}
```
damassi: Noticed a few spots lately that could use this

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -226,8 +226,6 @@ describe("Payment", () => {
     fillAddressForm(paymentRoute, validAddress)
 
     paymentRoute.find(ContinueButton).simulate("click")
-
-    expect.hasAssertions()
```
zephraph: Is this just some unrelated cleanup?

src/DevTools/MockRelayRenderer.tsx
```diff
@@ -0,0 +1,225 @@
...
+   ```
+ *
+ */
+export function renderRelayTree<
+  P = {},
+  S = {},
+  C extends React.Component = React.Component
+>(
+  params: MockRelayRendererProps & {
+    until?: UntilCallback<P, S, C>
```
damassi: `UntilCallback` usage makes total sense but I'm wondering if there might be a slightly more descriptive name for it? Since it **a)** involves a common situation around nested `QueryRenderer`'s and **b)** could be used in other ways.

src/Apps/__stories__/Apps.story.tsx
alloy: Probably yeah 👍 and likely best to do that once before we further abstract it

src/DevTools/__test__/MockRelayRenderer.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+const query = graphql`
+  query MockRelayRendererQuery {
+    artwork(id: "mona-lisa") {
+      ...MockRelayRenderer_artwork
+    }
+  }
+`
+
+function renderToString(element: React.ReactElement<any>) {
```
damassi: Thoughts on `React.ReactElement<any>` vs `JSX.Element` in common code? The former opens the doors to more safety later on and so seems better to use but was wondering your thoughts.

src/DevTools/__test__/MockRelayRenderer.test.tsx
alloy: Timers aren’t mocked by default: https://jestjs.io/docs/en/timer-mocks.html

What part of he code are you thinking of? I don’t really see a reason for this to fail (other than the code being broken).

src/Styleguide/Components/RecentlyViewed.tsx
alloy: Ah right, I had intended to cc you on me adding this to `RecentlyViewed`. Yeah we should use it in more places 👍

src/Apps/Order/Routes/__tests__/Payment.test.tsx
alloy: Yup, see this commit https://github.com/artsy/reaction/pull/1345/commits/1dec03116e2acce1092f644861245dd8f413bf49

I made our setup expect that every test should have an assertion, a few test bugs came out of that.

src/DevTools/__test__/MockRelayRenderer.test.tsx
damassi: Ah, was thinking about `jest.useFakeTimers()` and the seemingly arbitrary `10ms` use here.

src/Artsy/Relay/renderWithLoadProgress.tsx
zephraph: Are we leaving this for follow up?

src/Apps/Order/Routes/__tests__/Status.test.tsx
```diff
@@ -1,32 +1,56 @@
...
   it("should should have a title containing status", async () => {
-    const headTags = []
-    getWrapper(OrderWithShippingDetails, headTags)
-    // @ts-ignore
-    expect(render(headTags).text()).toBe("Order status | Artsy")
+    const headTags: Array<React.ReactElement<any>> = []
+    await getWrapper(
+      OrderWithShippingDetails,
+      headTags,
+      () => headTags.length > 0
```
alloy: @zephraph @damassi I now realise I need to dive into this a bit more. I had previously assumed that the title not showing up immediately was due to how `react-head` works, but am now thinking it’s because of our `Responsive` component. I thought that with an initial breakpoint set the `Responsive` component would immediately render, but maybe not?

src/Artsy/Relay/renderWithLoadProgress.tsx
alloy: Yeah. I probably should have linked to the comment above https://github.com/artsy/reaction/pull/1345/files/af7b25ef5fc79c1646542073d4c08a3b66153fbc#diff-d997111e63b21d4529fb81271cfe5c47R18

src/DevTools/MockRelayRenderer.tsx
alloy: Am I correct in understanding that you are suggesting to rename the parameters from `until` to `renderUntil`? And also rename the callback type from `UntilCallback` to `RenderUntilCallback`?

src/DevTools/__test__/MockRelayRenderer.test.tsx
alloy: Yeah that value is arbitrary, but shouldn’t break unless it is brought down to a very low value.

src/DevTools/__test__/MockRelayRenderer.test.tsx
alloy: Ah, a `ReactNode` is anything that React knows how to render, including strings, which doesn’t make sense for our case. Ok, going with `JSX.Element` 👍

src/DevTools/MockRelayRenderer.tsx
damassi: With TS this isn't _so_ much of an issue because one can hover and read description, but not everyone uses TS -- so was thinking of some way of better describe what's happening here. `until` -> `waitUntil` -> `renderUntil`. Haha not too sure... perhaps the variety of its uses warrants a simple name.

src/Apps/Order/Routes/__tests__/Status.test.tsx
damassi: There's a lot going on inside of react-head considering how it works. What makes you think the issue is with Responsive?

alloy: Yeah the new context API is built around Fiber, so while I haven’t traced it I’m pretty sure it’s async https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberNewContext.js

Hrmm 🤔 I think we should follow up in another PR, if we find a way to deal with that at all.

alloy: I’m gonna leave it at this for now https://twitter.com/alloy/status/1047199392581767168, we’ll see if we hear back or if we can come up with another solution after sleeping on it.

In any case, this shouldn’t block the PR.

src/DevTools/MockRelayRenderer.tsx
alloy: I didn’t go with `waitUntil` in the first place because this implementation is very tied to rendering React components and so didn’t want to be too ambiguous about that. I’ll rename things as per my above understanding for now.

src/Apps/Order/Routes/__tests__/Status.test.tsx
alloy: > What makes you think the issue is with Responsive?

I recall not seeing log statements showing up when I expected them to. I should probably just create a test that uses a context to verify.

src/DevTools/MockRelayRenderer.tsx
damassi: Sounds good -- i think `renderUntil` is a little more clear.

alloy:
@zephraph FYI This is complete from my pov.

PR #1350, opened 2018-10-02
[FEATURE BNMO] Display error modal on unsupported shipping location

javamonn:
Displays an error modal using the `unsupported_shipping_location` error validation introduced in https://github.com/artsy/exchange/pull/204.

screen shot 2018-10-02 at 3 20 03 pm

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -137,16 +137,29 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
               if (orderOrError.error) {
                 const errorCode = orderOrError.error.code
+                const errorData = get(
+                  orderOrError,
+                  o => JSON.parse(o.error.data),
+                  {}
+                )
                 if (
-                  errorCode &&
-                  (errorCode === "missing_region" ||
-                    errorCode === "missing_country" ||
-                    errorCode === "missing_postal_code")
+                  errorCode === "missing_region" ||
+                  errorCode === "missing_country" ||
+                  errorCode === "missing_postal_code"
                 ) {
                   this.onMutationError(
                     orderOrError.error,
                     "Invalid address",
                     "There was an error processing your address. Please review and try again."
+                  )
+                } else if (
+                  errorCode === "unsupported_shipping_location" &&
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 55169d1371ed6a7629cccb31699e84d746ea6404

https://deploy-preview-1350--artsy-reaction.netlify.com

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -148,6 +153,16 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
                     "Invalid address",
                     "There was an error processing your address. Please review and try again."
                   )
+                } else if (
+                  errorCode &&
```
ashfurrow: I see that this is the same as the other `if` statements, but is this line necessary? Checking `if errorCode && errorCode === "some string"` seems extraneous? I could be missing something.

javamonn: I think you're right! Updating...

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -137,17 +137,30 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
                     "Invalid address",
                     "There was an error processing your address. Please review and try again."
                   )
+                } else if (
+                  errorCode === "unsupported_shipping_location" &&
+                  errorData.failure_code === "domestic_shipping_only"
+                ) {
+                  this.onMutationError(
+                    orderOrError.error,
+                    "Can't ship to that address",
```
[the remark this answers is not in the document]
sweir27: There are only a couple of options here (with a lot of nested logic), so this looks 👍 to me

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -137,17 +137,30 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
               if (orderOrError.error) {
                 const errorCode = orderOrError.error.code
+                const errorData = get(
+                  orderOrError,
+                  o => JSON.parse(o.error.data),
```
sweir27: Oh cool, didn't know we could use `get` this way.

PR #1354, opened 2018-10-02
[FEATURE BNMO] Add responsive prop and styling to ModalWrapper

javamonn:
Implements [PURCHASE-471](https://github.com/artsy/reaction/compare/master...javamonn:purchase-471-responsive-error-modal?expand=1).

Adds an `isResponsive` prop to `ModalWrapper`  to implement the responsive modal styling seen [here](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5b6dca0f88ebbc1cd4521d7c). Right now, the `ErrorModal` is only user of this styling. The alternate way to approach this would've been to add something like a `ResponsiveModalWrapper`, but the behavior appears to be the same outside a few lines of style differences.

As seen in storybook:

screen shot 2018-10-02 at 5 04 30 pm

### DIFF
--- a/src/Components/Modal/ModalWrapper.tsx
+++ b/src/Components/Modal/ModalWrapper.tsx
@@ -10,6 +10,7 @@ export interface ModalWrapperProps extends React.HTMLProps<ModalWrapper> {
   cta?: CtaProps
   onClose?: () => void
   isWide?: boolean
+  fullscreenResponsiveModal?: boolean
   image?: string
   show?: boolean
 }
@@ -74,7 +75,7 @@ export class ModalWrapper extends React.Component<
   }
 
   render(): JSX.Element {
-    const { children, isWide, image } = this.props
+    const { children, isWide, fullscreenResponsiveModal, image } = this.props
     const { isShown, isAnimating } = this.state
 
     if (isShown) {
@@ -96,8 +97,14 @@ export class ModalWrapper extends React.Component<
             unmountOnExit
             timeout={{ enter: 10, exit: 200 }}
           >
-            <ModalContainer isWide={isWide} image={image}>
-              <ModalInner>{children}</ModalInner>
+            <ModalContainer
+              fullscreenResponsiveModal={fullscreenResponsiveModal}
+              isWide={isWide}
+              image={image}
+            >
+              <ModalInner fullscreenResponsiveModal={fullscreenResponsiveModal}>
+                {children}
+              </ModalInner>
             </ModalContainer>
           </FadeTransition>
         </Wrapper>
@@ -142,6 +149,7 @@ export const ModalOverlay = styled.div`
 
 export const ModalContainer = styled.div.attrs<{
   isWide?: boolean
+  fullscreenResponsiveModal?: boolean
   image?: string
 }>({})`
   position: fixed;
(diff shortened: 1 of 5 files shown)


src/Components/Modal/ModalWrapper.tsx
```diff
@@ -155,20 +161,28 @@ export const ModalContainer = styled.div.attrs<{
...
   animation: ${slideUp} 250ms linear;
 
-  ${media.sm`
-    width: 100%;
-    border-radius: 0;
-  `};
+  ${props =>
+    props.isResponsive
+      ? media.sm`
+          width: calc(100vw - 20px);
```
javamonn: `calc` always seems like a bit of a smell to me, but I think we need an explicit width here for centering that `ModalContainer` applies.

artsyit:

Built with commit be10baedcd086fd1ecb6faab2b4548702e6b7911

https://deploy-preview-1354--artsy-reaction.netlify.com

src/Components/Modal/ModalWrapper.tsx
erikdstock: Interesting- why is it a smell for you? In all cases or just css in js?

yuki24: I think this is actually a nice use of `calc` :+1:

javamonn: I find `calc` to be brittle over time, like here it's not clear _why_ `20px` is being subtracted, outside of the comment on this PR, though I guess I could / should add a comment inline here. Also, it's less of an issue now, but `calc` used to have slightly different implementations and constraints across browsers that often led to quirks. My preference for a modal layout like this would be to use flex box for the layout.

src/Components/Modal/ModalWrapper.tsx
```diff
@@ -96,8 +97,12 @@ export class ModalWrapper extends React.Component<
...
             timeout={{ enter: 10, exit: 200 }}
           >
-            <ModalContainer isWide={isWide} image={image}>
-              <ModalInner>{children}</ModalInner>
+            <ModalContainer
+              isResponsive={isResponsive}
+              isWide={isWide}
+              image={image}
+            >
+              <ModalInner isResponsive={isResponsive}>{children}</ModalInner>
```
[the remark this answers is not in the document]
javamonn: I think we have other components built off of `Modal` / `ModalWrapper` other than `ErrorModal` and was concerned that they would no longer look as they should, as the responsive styling is pretty different than than the current styling, i.e. it's no longer full height and  width. We could also do this the other way around maybe - make the responsive styling the new default, and update the other modals to pass something like a `nonResponsive` prop.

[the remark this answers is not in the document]
sweir27: Maybe this is another naming thing 😅since technically they're both responsive, just different designs. Should we use something like `fullscreenResponsiveModal={true}` for the existing one and `false` for the error one?

src/Components/Modal/__tests__/ErrorModal.test.tsx
```diff
@@ -51,6 +52,11 @@ describe("ErrorModal", () => {
       expect(text).toContain("OK")
     })
 
+    it("Renders a responsive modal", () => {
+      const component = getWrapper(props)
+      expect(component.find(ModalWrapper).props().isResponsive).toBe(true)
```
zephraph: `isResponsive` -> `isFullscreenResponsive`

zephraph:
I'm good with this. The test needs to be updated, but after that I think it's good to go. 

Merge on green.

PR #1355, opened 2018-10-03
[FIX BNMOrouting] Extracts redirects for order app into routes config

sweir27:
This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-453

Thanks to @damassi for figuring this out with me! 

This PR moves all routing/redirect logic (that is meant to happen _before_ render) into the root of the `OrderApp`. At first I tried to stick this logic into the sub-routes (i.e. `Payment/index`) but kept getting issues with trying to reference client-side methods on the server. (While doing this I noticed that the existing redirect implementation was also broken for the same reason).

One way to fix that ☝️ is to repeat logic in the `render()` function and in `componentDidMount` in order to redirect (and not render anything server-side), but only in the cases we specify.

This becomes unwieldy. Turns out `found` has redirect handling, and sticking it in the root route means we can be selective about only fetching the data we need to make this decision.

cc @zephraph

### DIFF
--- a/src/Apps/Order/OrderApp.tsx
+++ b/src/Apps/Order/OrderApp.tsx
@@ -81,18 +81,10 @@ export class OrderApp extends React.Component<OrderAppProps, OrderAppState> {
   }
 
   render() {
-    const { children, location, router, order, params } = this.props
+    const { children, order } = this.props
 
     if (!order) {
       return <ErrorPage code={404} />
-    }
-
-    if (
-      order &&
-      order.state !== "PENDING" &&
-      !location.pathname.includes("status")
-    ) {
-      router.replace(`/order2/${params.orderID}/status`)
     }
 
     return (
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -162,6 +162,7 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
   }
 
   onMutationError(errors, errorModalTitle?, errorModalMessage?) {
+    console.error("Shipping/index.tsx", errors)
     this.setState({
       isCommittingMutation: false,
       isErrorModalOpen: true,
--- a/src/Apps/Order/__tests__/OrderApp.test.tsx
+++ b/src/Apps/Order/__tests__/OrderApp.test.tsx
@@ -20,10 +20,9 @@ jest.mock("react-stripe-elements", () => ({
 }))
 
 describe("OrderApp routing redirects", () => {
+  // FIXME: move to DevTools folder
   async function render(url, mockResolvers) {
-    const network = createMockNetworkLayer({
-      ...mockResolvers,
-    })
+    const network = createMockNetworkLayer(mockResolvers)
     const source = new RecordSource()
     const store = new Store(source)
     const environment = new Environment({ network, store })
@@ -48,6 +47,7 @@ describe("OrderApp routing redirects", () => {
     })
     expect(redirect).toBe(undefined)
   })
+
   it("redirects to the status route if the order is not pending", async () => {
     const { redirect } = await render("/order2/1234/shipping", {
       Order: () => ({
--- a/src/Apps/Order/__tests__/OrderApp.test.tsx
+++ b/src/Apps/Order/__tests__/OrderApp.test.tsx
@@ -1,15 +1,144 @@
+import { routes } from "Apps/Order/routes"
 import { ContextProvider } from "Artsy/Router"
 import { mount } from "enzyme"
+import { Resolver } from "found-relay"
+import createRender from "found/lib/createRender"
+import getFarceResult from "found/lib/server/getFarceResult"
 import React from "react"
-import { HeadProvider } from "react-head"
-import { Meta } from "react-head"
+import { HeadProvider, Meta } from "react-head"
 import { ErrorPage } from "../../../Components/ErrorPage"
 import { OrderApp } from "../OrderApp"
+
+import { createMockNetworkLayer } from "Artsy/Relay/createMockNetworkLayer"
+import { Environment, RecordSource, Store } from "relay-runtime"
 
 jest.mock("react-stripe-elements", () => ({
   Elements: ({ children }) => children,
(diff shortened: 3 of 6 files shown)


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -162,6 +162,7 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
   }
 
   onMutationError(errors, errorModalTitle?, errorModalMessage?) {
+    console.error("Shipping/index.tsx", errors)
```
sweir27: I'd accidentally removed this in a previous PR, but these log messages are super helpful for debugging.

src/Apps/Order/__tests__/OrderApp.test.tsx
```diff
@@ -1,16 +1,145 @@
...
+
 jest.mock("react-stripe-elements", () => ({
   Elements: ({ children }) => children,
   StripeProvider: ({ children }) => children,
+  CardElement: () => jest.fn(),
+  injectStripe: () => jest.fn(),
 }))
 
+describe("OrderApp routing redirects", () => {
+  async function render(url, mockResolvers) {
```
sweir27: At some point we may want to extract this to be used in other apps, but this is the only place it's needed for now.

artsyit:

Built with commit 12459e71242def773078096bebf9afd8e21e5e54

https://deploy-preview-1355--artsy-reaction.netlify.com

src/Apps/Order/OrderApp.tsx
```diff
@@ -81,20 +81,12 @@ export class OrderApp extends React.Component<OrderAppProps, OrderAppState> {
...
     }
 
-    if (
-      order &&
-      order.state !== "PENDING" &&
-      !location.pathname.includes("status")
-    ) {
-      router.replace(`/order2/${params.orderID}/status`)
-    }
-
```
damassi: 👋

src/Apps/Order/__tests__/OrderApp.test.tsx
```diff
@@ -1,16 +1,145 @@
...
+    const network = createMockNetworkLayer({
+      ...mockResolvers,
+    })
```
damassi: `const network = createMockNetworkLayer(mockResolvers)`

src/Apps/Order/__tests__/OrderApp.test.tsx
```diff
@@ -1,16 +1,145 @@
...
+    const source = new RecordSource()
+    const store = new Store(source)
+    const environment = new Environment({ network, store })
+
+    return await getFarceResult({
+      url,
+      routeConfig: routes,
+      resolver: new Resolver(environment),
+      render: createRender({}),
```
damassi: out of curiosity can this go away?

src/Apps/Order/__tests__/OrderApp.test.tsx
damassi: This might go well in the new DevTools folder once the other PR is merged

src/Apps/Order/__tests__/OrderApp.test.tsx
```diff
@@ -1,16 +1,145 @@
...
+          __typename: "Ship",
+        },
+        creditCard: {
+          id: "12345",
+        },
+      }),
+    })
+    expect(redirect).toBe(undefined)
+  })
+})
```
damassi: Glad to see that after the `render` util all of these tests are straightforward and simple

src/Apps/Order/routes.tsx
```diff
@@ -65,9 +71,45 @@ export const routes: RouteConfig[] = [
...
+      if (Component && props) {
+        const { location, order, params } = props as any
+
+        if (
+          order &&
+          order.state !== "PENDING" &&
+          !location.pathname.includes("status")
+        ) {
+          throw new RedirectException(`/order2/${params.orderID}/status`)
+        } else if (
```
damassi: It would be helpful if there was a short comment between each conditional branch

src/Apps/Order/Routes/Shipping/index.tsx
damassi: Sooner rather than later we should drop in a logger similar to what we have in prediction, which hooks right into sentry on `logger.error` (something I'd like to look at next quarter)

src/Apps/Order/__tests__/OrderApp.test.tsx
sweir27: I'll add a FIXME

src/Apps/Order/__tests__/OrderApp.test.tsx
sweir27: 😅 evidence of "the faff"

src/Apps/Order/__tests__/OrderApp.test.tsx
sweir27: nope, `getFarceResult` requires it to exist

src/Apps/Order/routes.tsx
```diff
@@ -65,9 +103,22 @@ export const routes: RouteConfig[] = [
...
+          }
+          creditCard {
+            id
+          }
         }
       }
     `,
+    render: ({ Component, props }) => {
+      if (Component && props) {
+        if (!shouldRedirect(props)) {
```
sweir27: Extracted this out as a function-- I didn't move it to its own file since it didn't feel too unwieldy but that's an easy thing to follow up on.

src/Apps/Order/routes.tsx
```diff
@@ -52,6 +58,38 @@ const confirmRouteExit = (
   return LEAVE_MESSAGING
 }
 
+const shouldRedirect = props => {
```
zephraph: Oh, nice!! I like this much better. 😄

sweir27: Extracted them into their own file

damassi: Cool, this is a little clearer -- with vscode click to definition the distance isn't too great, and also makes things easier to unit test in isolation

damassi:
LGTM 👍

PR #1357, opened 2018-10-03
 use new FollowProfile button for partner details

oxaudo:
This puts the follow profile button on the ArtwokDetails component. Not sure what should we do in case partner does not have a profile attached to it. Though I don't think it should ever be the case.

Also removing renderAbout check since now we supposed to render about tab in all the cases as it includes partner info.


artsyit:

Built with commit b211ff0a697331538edfc9f2ab8249f97aa37b25

https://deploy-preview-1357--artsy-reaction.netlify.com

artsyit:

Built with commit 0c9de329592d8d75b4241f56ce14e49de63fe34b

https://deploy-preview-1357--artsy-reaction.netlify.com

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -5,7 +5,7 @@ import styled from "styled-components"
 import { FullArtworkGrid, Tab, Tabs } from "Styleguide/Components"
 import { Col, Row } from "Styleguide/Elements/Grid"
 import { ArtistInfoQueryRenderer as ArtistInfo } from "./Components/ArtistInfo"
-import { ArtworkDetailsFragmentContainer as ArtworkDetails } from "./Components/ArtworkDetails"
+import { ArtworkDetailsQueryRenderer as ArtworkDetails } from "./Components/ArtworkDetails"
```
alloy: Is this only a query renderer while building out the artwork app or is it intended to stay one?

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -15,44 +16,28 @@ import { ArtworkDetails_artwork } from "__generated__/ArtworkDetails_artwork.gra
...
-    artwork.signatureInfo
-  if (
-    !renderAbout &&
-    !artwork.articles &&
-    !artwork.exhibition_history &&
-    !artwork.literature
-  ) {
-    return null
-  }
+  const { artwork, user, mediator } = props
```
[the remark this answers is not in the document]
damassi: Yup, would grab mediator from ContextConsumer rather than passing down, even if the distance is only parent to child. The less props passing the more decoupled things can be.

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -131,6 +97,8 @@ export const ArtworkDetailsQueryRenderer = ({
                 return (
                   <ArtworkDetailsFragmentContainer
                     artwork={props.artwork as any}
+                    user={user}
+                    mediator={mediator}
```
damassi: Same here

src/Apps/Artwork/Components/__stories__/ArtworkDetails.story.tsx
```diff
@@ -12,7 +12,10 @@ storiesOf("Styleguide/Artwork", module)
     return (
       <React.Fragment>
         <Section title="Artwork with every additional details available">
-          <ArtworkDetails artwork={AllAdditionalDetailsPresent as any} />
+          <ArtworkDetails
+            artwork={AllAdditionalDetailsPresent as any}
+            user={null}
```
damassi: does this user have to explicitly be `null`?

src/Apps/Artwork/ArtworkApp.tsx
damassi: This is new, below the fold react stuff that we're stitching into the old artwork page. The new page will follow a top-down fragment approach.

oxaudo: I think it's still to be determined with our work on responsive artwork page @alloy. For now it's actually used in current non reaction Artwork page through sticking so dropped it here to be looking at what production is looking at and for parity with ArtistInfo on this page.

But eventually we might need only one query renderer for the whole artwork page / or not...

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -50,6 +58,43 @@ export class ArtworkDetailsAboutTheWorkFromPartner extends React.Component<
...
+                            afterSignUpAction: {
+                              kind: "profile",
+                              action: "follow",
+                              objectId: partner.profile && partner.profile.id,
+                            },
+                          })
+                      }}
+                      render={profile => {
+                        const is_followed =
+                          (profile && profile.is_followed) || false
```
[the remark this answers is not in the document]
oxaudo: you right. Updating:)

src/Apps/Artwork/ArtworkApp.tsx
alloy: Gotcha, thanks 👍

damassi:
Latest changes look great 👍 mergin'

PR #1359, opened 2018-10-03
[FIX BNMO] Specify money field precision to avoid rounding

javamonn:
Implements [PURCHASE-494](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&modal=detail&selectedIssue=PURCHASE-494).

Metaphysics currently defaults to using 0 decimals of precision in fields that use the `amount` money formatting helper: https://github.com/artsy/metaphysics/blob/master/src/schema/fields/money.js#L24-L27. This causes cents values to be rounded to the nearest whole dollar in the order flow. By specifying an argument for `precision`, we can format prices without rounding.

An order with a whole number cost:
screen shot 2018-10-03 at 2 31 27 pm

An order with a cost that includes cents:
screen shot 2018-10-03 at 2 38 23 pm

### DIFF
--- a/src/Apps/Order/Components/TransactionSummary.tsx
+++ b/src/Apps/Order/Components/TransactionSummary.tsx
@@ -120,10 +120,10 @@ export const TransactionSummaryFragmentContainer = createFragmentContainer(
   TransactionSummary,
   graphql`
     fragment TransactionSummary_order on Order {
-      shippingTotal
-      taxTotal
-      itemsTotal
-      buyerTotal
+      shippingTotal(precision: 2)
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit 969d55c40d61c0b3443225cacb5b432b2c50d963

https://deploy-preview-1359--artsy-reaction.netlify.com

sweir27:
Wow, glad this change ended up being so trivial!

zephraph:
Heck yeah! Easy peasy. Great job @javamonn.

PR #1363, opened 2018-10-04
[Responsive] Don’t try to actually use media queries in tests.

alloy:
Addresses https://github.com/artsy/reaction/pull/1345#discussion_r222047185

Will comment inline.

### DIFF
--- a/src/DevTools/__test__/renderUntil.test.tsx
+++ b/src/DevTools/__test__/renderUntil.test.tsx
@@ -1,4 +1,6 @@
 import * as React from "react"
+import { Responsive } from "Utils/Responsive"
+import { MockBoot } from "../MockBoot"
 import { renderUntil } from "../renderUntil"
 
 class Component extends React.Component {
@@ -7,13 +9,24 @@ class Component extends React.Component {
   }
 
   componentDidMount() {
-    setImmediate(() => {
-      this.setState({ data: "ohai" })
-    })
+    this.setState(
+      {
+        data: "Loading",
+      },
+      () =>
+        setImmediate(() => {
+          this.setState({ data: "ohai" })
+        })
+    )
   }
 
   render() {
-    return <div>{this.state.data}</div>
+    return (
+      <div>
+        {this.state.data}
+        {this.state.data !== "Loading" && this.props.children}
+      </div>
+    )
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/Utils/Responsive/Responsive.tsx
```diff
@@ -71,7 +71,8 @@ export function createResponsiveComponents<M extends string>() {
       isSupportedEnvironment = () => {
         return (
           typeof window !== "undefined" &&
-          typeof window.matchMedia !== "undefined"
+          typeof window.matchMedia !== "undefined" &&
+          window.name !== "nodejs"
```
alloy: Because we use jsdom, these tests were passing, which in turn would lead to the media queries being kicked off and thus ignoring the initial matched one passed in as a prop.

src/DevTools/__test__/renderUntil.test.tsx
```diff
@@ -7,13 +9,24 @@ class Component extends React.Component {
   }
 
   componentDidMount() {
-    setImmediate(() => {
-      this.setState({ data: "ohai" })
-    })
```
alloy: This was leading to hard to debug crashes in tests.

artsyit:

Built with commit 46807122cc2e2ffab607320f3c2be7cc56e50c4e

https://deploy-preview-1363--artsy-reaction.netlify.com

src/Utils/Responsive/Responsive.tsx
damassi: @alloy - this is how we deal with it in our tests, which I think is simple and no nonsense:  

```
https://github.com/damassi/reaction/blob/51146296989d3a13c9d9409506074f25b8810df7/src/Styleguide/Components/__tests__/ArtistCard.test.tsx#L27-L29
```

alloy: I don’t think that’s a good solution as it is very surprising having to do this. Is there any reason to consider the test env a support env?

damassi: Can you elaborate a bit? Functions on `window` points to `global.window` which can be stubbed to whatever, per common jest conventions. If `matchMedia` is null, then the responsive component dimension fires immediately, as it it were on the server, and thus we can test different dimensions. This seems simpler than adding environment specific config into the component itself, which is leaky.

alloy: Oh I’m fine with moving that example you pointed to to `setup_jest.ts` so it applies to all tests, I just don’t want anybody having to spend any more time on having to figure this issue out.

(What I was asking about is if there’s ever a reason to not shortcut, regardless of implementation.)

zephraph: So what's the consensus here? Emptying out `matchMedia` in `setup_jest.ts` or using a check in the `isSupportedEnvironment` method? Looking through the issues, it seems like the jsdom team recommends using `navigator.userAgent.includes("jsdom")` to test if it's in jsdom.

alloy: I’ll update it to empty it in `setup_jest.ts`

PR #1365, opened 2018-10-04
[FIX ArtworkDetailsChecklist] Properly wrap context inside the row

oxaudo:
adress DISCO-332

This mainly super minor fix for the row alignment.


src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -134,9 +134,8 @@ export const ArtworkDetailsAboutTheWorkFromPartnerFragmentContainer = createFrag
           city
         }
         profile {
-          __id
+          ...FollowProfileButton_profile
```
oxaudo: This change allowed me to remove casting profile to `as any` above. But that is also duplicated some data here. I think it's still the correct way to go about definitions here, @damassi ?

artsyit:

Built with commit df1437cd76392ca4aee0d7cbe3d1e39a893f1ac8

https://deploy-preview-1365--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -134,11 +134,10 @@ export const ArtworkDetailsAboutTheWorkFromPartnerFragmentContainer = createFrag
...
           id
-          is_followed
           icon {
-            url
+            url(version: "square140")
```
oxaudo: Also noticed that on the site this image looked off compared to what was on partner details before. Realizing we need a properly square version rather then resizing the large one.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
damassi: This is correct 👍

PR #1366, opened 2018-10-04
[FIX BNMO] set height of CreditCardInput to prevent size changes

javamonn:
Sets a hardcoded height on the container wrapping `react-stripe-elements/CardElement` so that the size of the input is consistent before and after Stripe is able to initialize. The result:

Hardcoding the height is non-ideal and could present an issue if height of the inner iframe that Stripe mounts changes, but I couldn't find an alternate way to accomplish keeping a consistent input height. Some other approaches we could consider:

- The [demo linked](https://jsfiddle.net/attystripe/xux7qzch/) the [react-stripe-elements](https://github.com/stripe/react-stripe-elements#demo) repo takes the approach of _completely_ hiding the input prior to Stripe initializing. I think this would require us still assuming some height in order to leave empty space for the input in the payment form, or hiding the form completely.
- We could call [`injectStripe`](https://github.com/stripe/react-stripe-elements#injectstripe-hoc) earlier (on `/shipping`) so that Stripe is pre-initialized once the user navigates to `/payment`. This would reduce the load time of the input but not eliminate it: `Stripe.js` would already be fetched or in the process of fetching, but the iframe responsible for rendering the credit card input and other Stripe elements resources would still need to be fetched. To work around this we could maybe pre-render the `CardElement` offscreen. We would want to combine this approach with the changes here to handle cases where `/payment` is the entry point for a session.

### DIFF
--- a/src/Apps/Order/Components/CreditCardInput.tsx
+++ b/src/Apps/Order/Components/CreditCardInput.tsx
@@ -15,6 +15,8 @@ export const StyledCardElement = styled(CardElement)`
 // Re-uses old input border behavior
 const StyledBorderBox = styled(BorderBox).attrs<InputBorderProps>({})`
   ${inputBorder};
+  padding: 9px 10px;
+  height: 40px;
 `
 
 interface CreditCardInputProps {


src/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -15,6 +15,8 @@ export const StyledCardElement = styled(CardElement)`
 // Re-uses old input border behavior
 const StyledBorderBox = styled(BorderBox).attrs<InputBorderProps>({})`
   ${inputBorder};
+  padding: 9px 10px;
```
javamonn: I adjusted this down from `padding: 10px` set by `inputBorder` so that this input's height is consistent with our other inputs on `/shipping` - we're using `40px` height right now it looks like.

artsyit:

Built with commit 3926eb373fc9dfa1be54c5afd55a3999abfadd3f

https://deploy-preview-1366--artsy-reaction.netlify.com

PR #1370, opened 2018-10-04
limit number of locations displayed in ArtworkDetails

oxaudo:
adress https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&projectKey=DISCO&modal=detail

Here is the look:
screen shot 2018-10-04 at 5 20 23 pm

### DIFF
--- a/src/Apps/Artwork/Utils/__tests__/limitArrayOfStringsWithCount.test.tsx
+++ b/src/Apps/Artwork/Utils/__tests__/limitArrayOfStringsWithCount.test.tsx
@@ -1,0 +1,32 @@
+import { limitArrayOfStringsWithCount } from "../limitArrayOfStringsWithCount"
+
+describe("limitArrayOfStringsWithCount", () => {
+  it("returns empty array if no array of strings is given", () => {
+    const limited = limitArrayOfStringsWithCount(null, 3)
+    expect(limited).toEqual([])
+  })
+
+  it("returns initial array if limit number is more then elements in initial array", () => {
+    const limited = limitArrayOfStringsWithCount(
+      ["let", "there", "be", "light"],
+      8
+    )
+    expect(limited).toEqual(["let", "there", "be", "light"])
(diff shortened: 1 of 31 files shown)


artsyit:

Built with commit 26eb97dd4e0b6f7997e4adf63b11cb1a1b718777

https://deploy-preview-1370--artsy-reaction.netlify.com

src/Apps/Artwork/Utils/__tests__/limitArrayOfStringsWithCount.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const limited = limitArrayOfStringsWithCount(
+      ["let", "there", "be", "light"],
+      4
+    )
+    expect(limited).toEqual(["let", "there", "be", "light"])
+  })
+
+  it("returns limited array according to limit", () => {
+    const limited = limitArrayOfStringsWithCount(
+      ["let", "there", "be", "light"],
```
damassi: 🔥

src/Apps/Artwork/Utils/limitArrayOfStringsWithCount.tsx
```diff
@@ -0,0 +1,20 @@
+/**
+ * Helper function to limit array of strings for shorter displays
+ * @param initial array of strings to chech for the limit
+ * @param limit number of elements to keep
+ */
+export const limitArrayOfStringsWithCount = (
```
damassi: Rather than encoding implementation details, could this be shortened to `limitWithCount`? 

Also, would you mind adding a small example in the docblock up above so that it its clear at a glance what this does? Something similar to what's seen here (though of course simpler): https://github.com/artsy/reaction/blob/master/src/DevTools/MockRelayRenderer.tsx#L38

(If it seems useful to others might want to place in the root `/Utils` folder)

PR #1373, opened 2018-10-05
[FIX BNMO] Address checkout flow tabs QA issues

javamonn:
Addresses issues noted in [PURCHASE-492](https://artsyproduct.atlassian.net/browse/PURCHASE-492).

- `BuyNowStepper` tabs are no longer clickable
-  `BuyNowStepper` now extends the full width of the screen `@XS` breakpoint. This was a good opportunity to clean up a [`FIXME`](https://github.com/artsy/reaction/compare/master...javamonn:purchase-492-checkout-flow-tabs?expand=1#diff-f2080351f37e0c5c0bf5c1020b503c08L58) and push padding application out of `Boot` and into the route components themselves, as there are various situations like this where a route will want different container padding. This allows us to apply different container padding to the stepper / header than the screen body as well. For other apps that previously received the padding from `Boot`, I've updated so that their route component is wrapped by a component that applies the same amount of padding. I searched force for usage of `buildClientApp` and updated `artist`, `collect2` and `order2` appropriately. `artwork2` is also affected and updated but I'm not sure if this is used currently - it looks broken right now.
- `BuyNowStepper` tabs now have correct spacing between items `@XS` breakpoint.

Checkout flow:

screen shot 2018-10-05 at 3 17 42 pm
screen shot 2018-10-05 at 3 18 11 pm
screen shot 2018-10-05 at 3 20 08 pm
screen shot 2018-10-05 at 3 20 32 pm

Other apps affected by the padding change, should show no changes:

screen shot 2018-10-05 at 2 49 45 pm
screen shot 2018-10-05 at 2 50 22 pm


src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
...
+import styled from "styled-components"
+import { media } from "../../Components/Helpers"
+
+interface PaddingProps {
+  sizeXS?: SpacingUnit | 0
+  size?: SpacingUnit | 0
+}
+
+const responsivePadding = media.sm<PaddingProps>`
+  padding-right: ${p => space(p.sizeXS as SpacingUnit) || 0}px;
```
javamonn: Should palette be updated to handle `0` correctly? Right now, `0` is not part of the `SpacingUnit` enum, and `space(0)` is `undefined`.

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const HorizontalPadding = styled.div.attrs<PaddingProps>({})`
+  padding-right: ${p => space(p.size as SpacingUnit) || 0}px;
+  padding-left: ${p => space(p.size as SpacingUnit) || 0}px;
+  margin-right: auto;
+  margin-left: auto;
+  ${responsivePadding};
+`
+
+HorizontalPadding.defaultProps = {
```
javamonn: These defaults are equivalent to the padding previously applied in `Boot`. We could have each component pass them in as props instead to avoid setting defaults.

src/Artsy/Router/Components/Boot.tsx
```diff
@@ -53,19 +48,10 @@ export class Boot extends React.Component<BootProps> {
...
-                      <Responsive>
-                        {({ xs }) => {
-                          // FIXME: Padding should be moved out of here
-                          return (
-                            <Padding padding={xs ? 20 : 40}>
-                              {children}
-                              {process.env.NODE_ENV === "development" && (
-                                <BreakpointVisualizer />
-                              )}
-                            </Padding>
```
damassi: Awesome 👍 thanks for cleaning this up

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
+import { space, SpacingUnit } from "@artsy/palette"
+import styled from "styled-components"
+import { media } from "../../Components/Helpers"
```
damassi: This refers to the old media helper (which we need to add a deprecation note about so that people don't use it in new framework code). The proper import is 
```
import { media } from 'Styleguide/Elements/Grid'
```
(and make sure the path is absolute)

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+interface PaddingProps {
+  sizeXS?: SpacingUnit | 0
```
damassi: Do you think encoding the environment into this API prop is necessary? It might be a little less overhead to think about this component as an extension of `<Box>` and then see if this is still needed to achieve the same end result. Then people only have to think about one system.

src/Styleguide/Utils/HorizontalPadding.tsx
damassi: If `space(0) == 0` then why not just use `0`?

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
...
+  size?: SpacingUnit | 0
+}
+
+const responsivePadding = media.sm<PaddingProps>`
+  padding-right: ${p => space(p.sizeXS as SpacingUnit) || 0}px;
+  padding-left: ${p => space(p.sizeXS as SpacingUnit) || 0}px;
+`
+
+export const HorizontalPadding = styled.div.attrs<PaddingProps>({})`
```
damassi: The latest styled-components allows one to type things like 

```js
export const HorizontalPadding = styled.div<PaddingProps>` ... `
```

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
...
+  padding-right: ${p => space(p.size as SpacingUnit) || 0}px;
+  padding-left: ${p => space(p.size as SpacingUnit) || 0}px;
```
damassi: Is `p.size` being cast here due to the possibility that it includes `0`? In the interface, could remove `| 0` and simplify fall back to `p.size && space(p.size) || 0`. 

Another approach would be 

```js
export const HorizontalPadding = styled.div<PaddingProps>`
  ${p => p.size && css`
      padding-right: ... 
      padding-left: ... 
   `
  }
`
```

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
+import { space, SpacingUnit } from "@artsy/palette"
+import styled from "styled-components"
+import { media } from "../../Components/Helpers"
+
+interface PaddingProps {
```
damassi: Not _strictly_ necessary but good to note that with TypeScript's `--emitDeclaration` prop we've begun establishing a convention around naming that follows 

```
// MyFileName.tsx

export interface MyFileNameProps {} 

export class / function / whatever MyFileName<MyFileNameProps>
```

Any given file can have multiple interfaces so it doesn't work 100%, but for simple combinations this pattern seems to work well. It also helps avoid naming clashes.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -181,31 +182,69 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
 
     return (
       <>
-        <Row>
-          <Col>
-            <BuyNowStepper currentStep={"review"} />
-          </Col>
-        </Row>
+        <HorizontalPadding sizeXS={0}>
```
damassi: Why not `<HorizontalPadding p={0}>`?

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,26 @@
...
+  sizeXS?: SpacingUnit | 0
+  size?: SpacingUnit | 0
+}
```
damassi: If `size` and `sizeXS` refer explicitly to horizontal left and right padding and additional api methods outside of styled-system are necessary, then it would be clearer to follow `styled-system`'s naming model as closely as possible: 

```js
pxXS: SpacingUnit
px: SpacingUnit

File: src/Apps/Order/Routes/Shipping/index.tsx [hunk 11]
```diff
@@ -231,85 +232,72 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
     return (
       <>
-        <Row>
-          <Col>
-            <BuyNowStepper currentStep={"shipping"} />
-          </Col>
-        </Row>
+        <HorizontalPadding size={4} sizeXS={0}>
```
damassi: Another thought would be to piggy back on styled-system's [responsive props](https://github.com/jxnblk/styled-system/blob/master/docs/responsive-styles.md). So like 

```
<HorizontalPadding p={[0, 4]}>

**Comment by javamonn** [comment]
Updated this in response to feedback from @damassi:
- Use [`styled-system`'s approach to responsive style values](https://github.com/jxnblk/styled-system/blob/master/docs/responsive-styles.md), consolidates size props
- Updated `media` helper import and added a deprecated comment to the old definition
- Removed `defaultProps` from `HorizontalPadding` in favor of passing in values explicitly

File: src/Apps/Artwork/ArtworkApp.tsx [hunk 12]
```diff
@@ -28,7 +29,7 @@ export const ArtworkApp: React.SFC<Props> = props => {
   ]
 
   return (
-    <React.Fragment>
+    <HorizontalPadding px={[2, 4]}>
```
damassi: Might be good to move this to `HorizontalPadding.defaultProps = { px: [2, 4] }` since most apps seem to follow these values and then later if things diverge too much we could pretty easily refactor.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -186,31 +187,69 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
...
-              Content={
-                <>
-                  <Join separator={<Spacer mb={3} />}>
-                    <ShippingAndPaymentReview
-                      order={order}
-                      onChangePayment={this.onChangePayment.bind(this)}
-                      onChangeShipping={this.onChangeShipping.bind(this)}
-                      mb={xs ? 2 : 3}
-                    />
+            <HorizontalPadding px={[2, 4]}>
```
damassi: I like this double use pattern. Once this is merged I can go back and address my wacky negative margin stuff in `<NavigationTabs>` 😄

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,24 @@
+import { space } from "@artsy/palette"
+import styled, { css } from "styled-components"
+import { ResponsiveSpaceValue } from "styled-system"
+import { media } from "Styleguide/Elements/Grid"
+
+export interface HorizontalPaddingProps {
+  px: ResponsiveSpaceValue
```
damassi: 👍

src/Styleguide/Utils/HorizontalPadding.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const HorizontalPadding = styled.div<HorizontalPaddingProps>`
+  ${p =>
+    p.px[1] &&
+    css`
+      padding-right: ${space(p.px[1])}px;
+      padding-left: ${space(p.px[1])}px;
+    `} ${p =>
```
damassi: If you add a `;` after the ` `}` then the second `${p =>` can sit onto its own line and prettier wont reformat.  

Also, if possible can this go in sequence, like `p.px[0]` and then `px[1]`? It reads a little bit better.

src/Apps/Artwork/ArtworkApp.tsx
javamonn: Updated this to go back to setting via `defaultProps`. 😅 Waffled a bit, I thought it was a little strange for a general padding component to set a default used by our route components when it could be used for horizontal padding in general, but I agree that it's widely used so makes senes to have in `defaultProps`, at least for now. Additionally, it saves us from having to do extra array checking before reaching into indices.

PR #1374, opened 2018-10-05
[SPIKE: Responsive Component] Try out new Responsive component 

damassi:
Spike to see what the new responsive component feels like in practice, and any issues we might encounter.


artsyit:

Built with commit 309d6633d7843f40799b049e50387de1b853645e

https://deploy-preview-1374--artsy-reaction.netlify.com

artsyit:

Built with commit 3cedf134e3c2e5ff0d103eca469a51ddc3c1af97

https://deploy-preview-1374--artsy-reaction.netlify.com

damassi:
One thing I'd like to note is that there are quite a few areas that rely on conditional rendering for configuring props that are passed into a component, or whether a given branch of code is executed, unrelated to the dom. These items I wasn't able to effectively unwind with this new implementation.

src/Styleguide/Components/Footer.tsx
```diff
@@ -124,19 +129,27 @@ const FooterContainer: React.SFC<FlexDirectionProps & Props> = props => {
...
           <Flex justifyContent="space-between">
             <Flex alignItems="center" mb={4}>
-              {!xs && <Mark width="30px" height="30px" mr={2} />}
-              {xs && <Mark width="20px" height="20px" mr={2} />}
-              <Spacer mr={1} />
-              {!xs && <PolicyLinks />}
+              {/* NOTE: Responsive2 Is it possible to render out multiple instances
+                of a given breakpoint component? Like, interweve stuff? Noticing because
+                <breakpoint.xs> is used above, this xs / else isn't rendering properly
+            */}
```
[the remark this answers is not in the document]
alloy: I guess this could be simplified by not having the smart `else` behavior, but rather something like `<breakpoints.not breakpoints={[breakpoints.xs, breakpoints.md]} />`.

This definitely has a lot of `breakpoints` in it, so maybe the components should be capitalized after all so that we don’t need to prefix them with `breakpoints.`

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -34,35 +34,51 @@ export class ArtistAuctionResultItem extends Component<Props> {
...
+                        />
+                      </breakpoints.sm>
+                      <breakpoints.md>
+                        <SmallAuctionItem
+                          {...this.props}
+                          mediator={mediator}
+                          user={user}
+                        />
+                      </breakpoints.md>
+                      <breakpoints.else>
```
l2succes: @damassi is `breakpoints.else` dynamically defined based on adjacent `breakpoints`?

alloy: Yeah, which gets confusing especially when not nested directly under the Responsive component.

PR #1375, opened 2018-10-05
[FIX BNMO] Card number text should be 16px and expand the clickable area for the credit card input

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-493

The reason for WIP is that the vertical text alignment looks a bit off from the center of the input. Some of you are probably already aware of the fact that this is because of the typeface, but we may be able to apply an easy fix.

## Screenshot

screen_shot_2018-10-05_at_4_53_51_pm


artsyit:

Built with commit 286db2124b740a9aa898611ff31bcd78f2f16911

https://deploy-preview-1375--artsy-reaction.netlify.com

Issue #1367, opened 2018-10-04
Effective responsiveness


alloy:
For reasons such as efficiency (4) and SEO, we want to do server-side rendering in such a way that the response sent to the client is immediately usable without the need to re-render.

Alas our current solution, which essentially boils down to if/else branching per breakpoint, doesn’t fulfil these needs as it is practically impossible (1) to know what the breakpoint is that the client will want to display at SSR time.

Additionally, even when getting the initial breakpoint right, the performance of changing breakpoints will be worse when a browser has to evaluate JS, recalculate a component tree, diff DOM elements, and update the DOM; as opposed to the browser only needing to re-render the styling without any DOM changes. (2)

## Example

Consider a desktop browser requesting a page that is rendered as follows:

```jsx
<Responsive>
  {({ xs }) => {
    if (xs) {
      return <Box width={1} />
    } else {
      return <Box width={1/2} />
    }
  }}
</Responsive>
```

On the server we’ll render a `xs` version (because that’s currently our default), then once the browser receives the response it may start to render that version.

However, once React is fully booted it starts to render a virtual DOM version based on an _actual_ media query that may result in a different breakpoint than `xs`. If that vDOM does not match what the server sent, the server version is thrown away and the new version is used instead.

## Options

Note that these are just some bare-bones examples. Where possible it would be great to come up with prettier declarative APIs that use these techniques under the hood.

### Pure CSS

In cases where all differences in layout can be expressed in styling, we should be able to do something like:

```jsx
const ResponsiveBox = styled.div`
  @media screen and (max-width: 767px) {
    width: 100%;
  }
  @media screen and (min-width: 768px) {
    width: 50%;
  }
`

<ResponsiveBox />
```

When some components should only render on certain breakpoints, these may simply be hidden on those breakpoints using CSS. (3)

### styled-system

We’re using styled-system, which out of the box comes with [responsive options](https://github.com/jxnblk/styled-system#responsive-style-props) that, I assume–and we should verify this, do exactly what the above option does. For instance, rather than the above example, we should write `<Box width={[1, 1/2]} />`.

### More…

Please suggest yours!

## Notes

1. That is, without resorting to added complexity such as sniffing user-agent and even then it won’t be foolproof.
2. While this performance penalty is even noticeable on desktop machines, it’s even more important for mobile devices.
[...]

PR #1379, opened 2018-10-08
[WIP] New pattern for responsive pages

alloy:
Closes https://github.com/artsy/reaction/issues/1367

- [x] API
- [x] Core implementation
- [ ] Deal with issue of invisible components being mounted (which may lead to accidental side-effects).
      _A quick and dirty implementation exists, but needs to be taken to completion either now or in a next PR._
- [ ] Put examples of API usage in this PR description.
- [ ] Make a final call on the deprecation path for `Responsive`.
- [ ] Write guide for README repo that outlines process of writing responsive components:
  1. Use styled-system’s responsive props when layout only slightly adapts (TODO come up with better terminology for this)
  2. When design changes in a way that’s not solvable by responsive props, use `Media` API, but on the component nearest to the leaf of the tree. 
- [ ] Add a link to the README guide in the `Responsive` deprecation warning.


artsyit:

Built with commit 8ab7a1de4fb6d7c3f506c32ec7e85aa945cb167d

https://deploy-preview-1379--artsy-reaction.netlify.com

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,162 @@
...
+  any?: Array<{ query: string }>
+  all?: Array<{ query: string }>
+  children: React.ReactNode | MatchRenderProp
+}
+
+export class Match extends React.Component<MatchProps> {
+  render() {
+    if (this.props.any) {
+      const queries = this.props.any.map(matcher => matcher.query)
+      const prefix = this.props.not ? "not all and " : ""
```
alloy: This actually doesn’t work, see comment above https://github.com/artsy/reaction/pull/1379/files#diff-2aace99a6c76f1455d5427a1708c309eR5

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,162 @@
...
+ *   breakpoints.md = Query.create({ minWidth: 900, maxWidth: 1023 })
+ *   <Match not from={breakpoints.xs} to={breakpoints.md} />
+ *
+ * Which would then generate:
+ *
+ *   @media not all (max-width: 1023px)
+ */
+
+interface RenderPropStyleGeneratorOptions {
+  display?: "block" | "inline" | "inline-block"
```
damassi: I think `display: inherit` would keep this more flexible

alloy: Ah yes, of course 👌

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,162 @@
...
+ * TODO:
+ *
+ * Something like this doesn’t work, because an attribute can only be used once
+ * in a media query, presumably the last one wins:
+ *
+ *   <Match not any={[breakpoints.xs, breakpoints.sm, breakpoints.md]} />
+ *
+ * Generates:
+ *
+ *   @media not all and (max-width: 767px), not all and (max-width: 899px) and (min-width: 768px), not all and (max-width: 1023px) and (min-width: 900px)
```
damassi: What if `<Match not any={[breakpoints.xs, breakpoints.sm, breakpoints.md]} />` produced a series of media queries each on their own lines? 

```css
@media not all and (max-width: 767px) { ... }
@media not all and (max-width: 899px) and (min-width: 768px) { ... }
@media not all and (max-width: 1023px) and (min-width: 900px) { ... } 
...
```
Keywords only work once, but with a little duplication i think it will work just fine

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,162 @@
...
+ * To overcome this we could define these attributes as actual props and then introduce a range matcher. E.g.
+ *
+ *   breakpoints.xs = Query.create({ maxWidth: 767 })
+ *   breakpoints.md = Query.create({ minWidth: 900, maxWidth: 1023 })
+ *   <Match not from={breakpoints.xs} to={breakpoints.md} />
+ *
+ * Which would then generate:
+ *
+ *   @media not all (max-width: 1023px)
+ */
```
damassi: Could also go with something like https://github.com/morajabi/styled-media-query (https://github.com/morajabi/styled-media-query/blob/master/src/index.js)

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -36,27 +36,14 @@ export class ArtistHeader extends Component<Props> {
...
-                }
-              }}
-            </Responsive>
+            <>
+              <breakpoints.xs>
+                <SmallArtistHeader mediator={mediator} user={user} {...props} />
+              </breakpoints.xs>
+              <breakpoints.xs not>
+                <LargeArtistHeader mediator={mediator} user={user} {...props} />
+              </breakpoints.xs>
```
damassi:
```jsx
<breakpoints.xs>small</breakpoints.xs>
<breakpoints.xs not>everything else</breakpoints.xs>
```

I think we should continue iterating on feel of this API as its a bit unintuitive and error prone and not to conducive to large-scale refactors. Why remove `<breakpoints.else>`? That seemed pretty clear and maps to a good mental model at a glance. A factory function like you alluded to

```js
const { xs, rest } = makeBreakpoints([breakpoints.xs]) 

<xs>
  ...
</xs>
<rest>
  ...
</rest>
```
seems flexible while also clear.

alloy: `breakpoints.else` was too smart. It required to be directly nested under the `Responsive` component, which seemed both hard to explain and a bit much. (Imagine needing to wrap multiple levels in a render function with the Responsive component.)

But I like your example of upfront specifying what breakpoints will be used and thus being able to create the `else` case upfront. Care to sketch that out further?

src/Utils/Responsive/Media.tsx
zephraph: Oh, that's pretty awesome.

src/Apps/Artist/Components/ArtistHeader.tsx
damassi: Ah, updated comment with a new renderProps example which i think would be the most flexible since those values could be passed at runtime and / or expanded to other kinds of media queries.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -34,36 +34,27 @@ export class ArtistAuctionResultItem extends Component<Props> {
...
+                  user={user}
+                />
+              </Match>
+              <Match any={[breakpoints.lg, breakpoints.xl]}>
+                <LargeAuctionItem
+                  {...this.props}
+                  mediator={mediator}
+                  user={user}
+                />
+              </Match>
```
damassi: I'm inclined to think that rather than exporting `breakpoints` to be used directly, we create intermediary compositional components at the app (as opposed to lib) level: 

```jsx
export const breakpoints = {
  xs: (children) => <Match any={[breakpoints.xs]}>{children}</Match>
}
```

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,162 @@
...
+    const prefix = this.props.not ? "not all and " : ""
+    const generatedStyle = (options?: RenderPropStyleGeneratorOptions) => css`
+      display: none;
+      @media ${prefix}${this.props.query} {
+        display: ${(options && options.display) || "block"};
+        ${(options && options.style) || ""};
+      }
+    `
+
+    if (typeof this.props.children === "undefined") {
```
damassi: `if (!this.props.children)`

src/Apps/Artist/Components/ArtistHeader.tsx
alloy: I think we should avoid render props where possible, as I find them to make readability of large pieces of markup harder.

src/Utils/Responsive/Media.tsx
alloy: Nice, I like this pattern. I’m going to make another example that uses this instead.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
alloy: Yep, good point. Addressed in Media2 👍

damassi:
Commented in https://github.com/artsy/reaction/issues/1367#issuecomment-428285766, but latest commit look A+

alloy:
Alright, did a migration pass so looking through the full patch should now better give an idea of what the API looks like and I also implemented the ability to provide a context and limit the breakpoints that will output any content (which is used in various tests to only render `xs`).

Tomorrow I’ll cleanup and write some docs/examples, but for now please do leave any feedback you may have.

src/Apps/Artist/Routes/Articles/ArtistArticle.tsx
```diff
@@ -23,12 +25,14 @@ interface ArticleItemProps {
...
-        else return <LargeArticleItem {...props} />
-      }}
-    </Responsive>
+    <>
+      <AllMedia.Media at="xs">
+        <SmallArticleItem {...props} />
+      </AllMedia.Media>
+      <Media greaterThan="xs">
+        <LargeArticleItem {...props} />
+      </Media>
```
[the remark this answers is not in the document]
alloy: No sorry, ignore that, this is stashed code of me moving between machines and it has this relic of me trying to figure out why vscode wasn’t showing me docs for props.

zephraph:
This is _amazing_ 👏

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -34,36 +34,27 @@ export class ArtistAuctionResultItem extends Component<Props> {
...
-              </Responsive>
-
+              <Media at="xs">
+                <ExtraSmallAuctionItem
+                  {...this.props}
+                  mediator={mediator}
+                  user={user}
+                />
+              </Media>
+              <Media between={["sm", "lg"]}>
```
damassi: This makes me think -- would an expanded `at` make sense here (can be string or array)? Something like: 

```js
<Media at={['sm', 'md']}>...
```

From a design handoff to dev layout perspective i feel like this is (slightly) easier to think about.

src/Styleguide/Components/Pagination.tsx
```diff
@@ -30,12 +30,12 @@ export class Pagination extends React.Component<Props> {
...
-            if (xs) return <SmallPagination {...this.props} />
-            return <LargePagination {...this.props} />
-          }}
-        </Responsive>
+        <Media at="xs">
+          <SmallPagination {...this.props} />
+        </Media>
+        <Media greaterThan="xs">
+          <LargePagination {...this.props} />
+        </Media>
```
damassi: Its really nice being able to skip the renderProps ceremony

src/Utils/Responsive/DeprecatedResponsive.tsx
```diff
@@ -32,12 +33,37 @@ const shallowEqual = (a: MediaQueryMatches, b: MediaQueryMatches) => {
...
+        }
+        console.warn(
+          `[Responsive] The Responsive component has been deprecated in ` +
+            `favour of the Media component.${
+              ownerName
+                ? ` It is being used in the ${ownerName} component.`
+                : ""
+            }`
+        )
+      }
```
damassi: 👍

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,268 @@
...
+                ) {
+                  return null
+                }
+              }
+              query = `(min-width:${fromWidth}px) and (max-width:${toWidth -
+                1}px)`
+            }
+          }
+
+          const generatedStyle: RenderPropStyleGenerator = matchingStyle => css`
```
damassi: It's pretty awesome that with a little work this can be platform agnostic.

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,268 @@
...
+  "lessThan",
+  "greaterThan",
+  "greaterThanOrEqual",
+  "between",
+  "interaction",
+]
+
+// TODO: All of these props should be mutually exclusive. Using a union should
+//       probably be made possible by https://github.com/Microsoft/TypeScript/pull/27408.
+export interface Props<B, I> {
```
zephraph: Instead of this, could we use conditional types? Here's an example that (I think) is doing what we want. http://realfiction.net/2018/06/10/ts-conditional-types/

zephraph: Eh, nope. Scratch that. Isn't what we want. >_>

src/Utils/Responsive/Media.tsx
```diff
@@ -0,0 +1,268 @@
...
+    breakpoints: { [key: string]: number }
+    interactions: { [key: string]: (negated: boolean) => string }
+  }
+>(config: C) {
+  type Breakpoint = keyof C["breakpoints"]
+  type Interaction = keyof C["interactions"]
+
+  const MediaContext = React.createContext<{ onlyRender?: Breakpoint[] }>({})
+  const MediaContextProvider: React.SFC<{ onlyRender?: Breakpoint[] }> = ({
+    onlyRender,
```
damassi: Can you add a comment that describes this prop a bit?

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
alloy: Oh yeah, great point, I like that! 👌

src/Utils/Responsive/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import { createMedia } from "./Media"
+
+// We need this to be 0-based, whereas currently in palette xs is defined as 767
+const newThemeBreakpoints = {
+  xs: 0,
+  sm: 768,
+  md: 900,
+  lg: 1024,
+  xl: 1192,
+}
```
damassi: We should update Palette with these new values an publish a new release, i think it will be fine.

src/Utils/Responsive/Media.tsx
alloy: Yup yup. I’m now also thinking to rename it to `onlyRenderAt`, we can iterate and a separate one for interactions later on when we actually start using this on the server.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
zephraph: That's in addition to between? It could be useful if we had non-contiguous designs. I'm not sure how likely that is though.

src/Apps/Artist/Routes/AuctionResults/TableColumns.tsx
```diff
@@ -1,17 +1,21 @@
...
-        if (xs) return null
-        else if (sm || md) return <SmallTableColumns />
-        else return <LargeTableColumns />
-      }}
-    </Responsive>
+    <>
+      <Media at="sm">
+        <SmallTableColumns />
+      </Media>
+      <Media at="md">
```
zephraph: Could this not just be `<Media between={["sm", "md"]}>`? Or, like @damassi recommended above `<Media at={["sm", "md"]}>`

src/Utils/Responsive/index.ts
alloy: I guess we can update the media query for the `xs` breakpoint so that it won’t break existing `Responsive` usage, but do you think we’re ok with styled-system using a different value for `xs`?

src/Apps/Artist/Routes/AuctionResults/TableSidebar.tsx
```diff
@@ -26,12 +26,14 @@ interface Props {
...
-      {({ xs }) => {
-        if (xs) return <SmallTableSidebar {...props} />
-        else return <LargeTableSidebar {...props} />
-      }}
-    </Responsive>
+    <>
+      <Media at="xs">
+        <SmallTableSidebar {...props} />
+      </Media>
+      <Media greaterThan="xs">
```
zephraph: I really, really like this api. It looks better, reads easier, and feels cleaner than the switch above.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
alloy: Yeah I was thinking of leaving `between` for now. It’s already written and seems likely to be useful.

src/Apps/Artist/Routes/AuctionResults/TableColumns.tsx
alloy: Oh yeah, absolutely! 👍  Hah, I missed that they were rendering the same component.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
damassi: I think that `at={['sm', 'md']}` is easier to reason about since it maps 1:1 with our design assets (and thus will be more friendly over time), but that `between` should remain in the underlying API since its a powerful building block.

src/Apps/Artist/Routes/CV/CVItem.tsx
```diff
@@ -89,6 +89,12 @@ class CVItem extends Component<CVItemProps, CVItemState> {
     )
 
     return (
+      /**
+       * TODO:  replace with new responsive component
+       *
+       * Sometimes conditional rendering utilities are helpful, though the below
+       * would most likely benefit from a refactor since the code is a bit crazy
+       */
```
zephraph: I think this isn't as bad as it seems. Can I have a go at updating it?

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
damassi: screen shot 2018-10-10 at 1 51 23 pm

damassi: It seems a bit weird to have to mention XL in our code when thinking specifically about L and M

src/Apps/Artist/Routes/CV/CVItem.tsx
alloy: Not sure if you’re asking @damassi or me, but I’m more than happy for you to have a go 👍

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
alloy: I imagine `between` being useful when programatically using the API.

src/Utils/Responsive/index.ts
damassi: Oh shoot, I read this wrong. 🤔 Yeah, might be safest to return to this in another PR. Maybe add a FIXME around uniting this with Palette? We def need to make sure that styled-system, palette and Reaction (and other consumers of each) are all on the same page.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
zephraph: I definitely think we should leave `between`. I already ran across a usecase for `at` though... (or `notAt`?) 

https://github.com/artsy/reaction/pull/1379/files#diff-17af9a6729cede275ff0d1e7714fdc5bR130

That would be `<Media at={["xs", "lg", "xl"]}/>`

src/Utils/Responsive/Media.tsx
alloy: Wait, why not? I haven’t read extensively, but to me it does appear to do the thing we want.

src/Apps/Artist/Routes/CV/CVItem.tsx
damassi: heh, I think its fine for now, was just a bit hard to reason about in terms of the new API 😄 We'll have to return to all of these conditional items in time.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
zephraph: Meh, I read it wrong. That's only `lg` and `xl`. Component is complicated 😓

src/Apps/Artist/Routes/CV/CVItem.tsx
```diff
@@ -3,7 +3,7 @@ import { groupBy } from "lodash"
 import React, { Component } from "react"
 import styled from "styled-components"
 import { Col, Row } from "Styleguide/Elements/Grid"
-import { Responsive } from "Utils/Responsive"
+import { Media, Responsive } from "Utils/Responsive"
```
zephraph: So I had a go at converting this. Couldn't get it 100%, but the only differences will be attributes.

src/Apps/Artist/Routes/CV/CVItem.tsx
```diff
@@ -90,56 +90,58 @@ class CVItem extends Component<CVItemProps, CVItemState> {
...
             <CVItems className="cvItems">
-              {(xs || sm || md) && (
+              <Media between={["xs", "md"]}>
                 <Row>
                   <Col>
                     <Category size={xs ? "2" : "3"} weight="medium">
                       {this.props.category}
                     </Category>
-                    <Spacer mb={xs ? 0.5 : 1} />
+                    <Spacer mb={[0.5, 1]} />
```
zephraph: We just need to do this as a larger scale refactor

src/Apps/Artist/Routes/CV/CVItem.tsx
```diff
@@ -151,16 +153,23 @@ class CVItem extends Component<CVItemProps, CVItemState> {
                       onClick={() => this.loadMore()}
                       loading={this.state.isLoading ? true : false}
                       width={xs ? "100%" : ""}
```
zephraph: @damassi think this could be `width={[ "100%", "" ]}`? 🤔 

It looks like width is a responsive styled systems value...

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
alloy: I’m going to defer this until a next iteration, as it will require some (required anyways) refactoring first.

alloy:
Ok, this should pretty much be it. I’ll do another review myself tomorrow and then add some examples of usage to the PR description.

alloy:
I’ve added some TODOs to this PR.

I need to switch gears for a bit to prepare for next week, if anybody else wants to pickup taking 
https://github.com/artsy/reaction/pull/1379/commits/52b66cf19f8cb6f0db5e5d854dd0faa939f9fea6 to completion that would be stellar 🙏 or we can defer that to a subsequent PR. Whatever y’all think is best.

Likewise with the guide.

damassi:
Per offline discussion i've moved code to https://github.com/artsy/react-responsive-media and set up all the needed bits for releases.

damassi:
Closing per https://github.com/artsy/reaction/pull/1466

PR #1382, opened 2018-10-08
{WIP} [Collection] finish UI styling for collection Header component

xtina-starr:
This adds the rest of the styling for the collection header UI.

screen shot 2018-10-08 at 5 35 04 pm

### DIFF
--- a/src/Apps/Collect/CollectApp.tsx
+++ b/src/Apps/Collect/CollectApp.tsx
@@ -18,11 +18,23 @@ export interface CollectAppProps {
 
 export class CollectApp extends Component<CollectAppProps> {
   render() {
+    const collection = {
+      id: "minimalist-prints",
+      title: "minimalist prints",
+      image:
+        "https://d7hftxdivxxvm.cloudfront.net?resize_to=width&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2Fbzx0Po9dHb_GhcvbD_c7jQ%2Funtouched-jpg.jpg&width=2000&quality=80",
+      description:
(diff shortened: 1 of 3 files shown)


src/Apps/Collect/CollectApp.tsx
```diff
@@ -18,11 +18,23 @@ export interface CollectAppProps {
...
+      major_periods: [],
+      gene_ids: [],
+      artist_ids: [],
+    }
     return (
       <Fragment>
         <Flex flexDirection="column">
           {true ? (
-            <CollectionHeader />
+            <CollectionHeader collection={collection} />
```
xtina-starr: It seems the work flow you were using was just to render the Collections Header inside of the CollectApp. I assume once everything is hooked up we will just move things into `/Collect/routes.tsx`?

artsyit:

Built with commit 341130af125a8a6824f71124ef21f02b9fa34954

https://deploy-preview-1382--artsy-reaction.netlify.com

src/Apps/Collect/CollectApp.tsx
l2succes: Right, that's the route I was thinking initially but on second though, I think we should create a new top level app component for Collections (`CollectionApp`) for simplicity sake since the root query will be different. So we'd end up with

```
Apps/Collect
|- CollectApp.tsx
|- CollectionApp.tsx
|- routes.tsx
```

src/Apps/Collect/Components/Header/index.tsx
```diff
@@ -42,13 +65,18 @@ export class CollectionHeader extends Component {
   }
 }
 
-const Background = styled(Box)`
+const Background = styled(Box).attrs<{ headerImageUrl: string }>({})`
```
damassi: Styled components recently updated to support  

```jsx
const Background = styled(Box)<{ headerImageUrl: string }>`
  ...
`
```

xtina-starr: Nice!

src/Apps/Collect/CollectApp.tsx
xtina-starr: Okay I will refactor to go with the new approach.

l2succes:
Closing in favor of https://github.com/artsy/reaction/pull/1408

PR #1383, opened 2018-10-09
[Fix EntityHeader] Follow button should sit inline with text

oxaudo:
Adress https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&projectKey=DISCO&modal=detail&selectedIssue=DISCO-353


oxaudo:
The change archives this result:

screen shot 2018-10-09 at 10 24 37 am

Instead of this:
screen shot 2018-10-09 at 10 28 02 am

damassi:
👍

PR #1388, opened 2018-10-10
 Update artists for the sidebar with the new FollowArtist button

oxaudo:
The first step to update current sidebar to be a real deal.

Also adding bunch of actual sidebar renderers in the new SidebarLive story so that we can keep looking at the real data on staging.


oxaudo:
Not sure the FollowIcon here is up with the latest standards but it does look right and seem to be working, so in a name of speed here just reusing it:)

artsyit:

Built with commit ad28a7c3751c3e8401f2bf56f6fce835a9b00086

https://deploy-preview-1388--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarArtists.tsx
```diff
@@ -24,11 +27,31 @@ export class ArtworkSidebarArtists extends React.Component<ArtistsProps> {
...
+              signupIntent: "follow artist",
+              afterSignUpAction: {
+                kind: "artist",
+                action: "follow",
+                objectId: artist.id,
+              },
+            })
+          }}
+          render={({ is_followed }) => {
+            return <FollowIcon is_followed={is_followed} />
```
damassi: 👍

oxaudo:
One artist:
screen shot 2018-10-10 at 1 10 39 pm

Multiple artists:
screen shot 2018-10-10 at 1 11 38 pm

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarArtists.tsx
damassi: Can this be updated to follow normal js camelCase in the prop (`isFollowed`)? Or perhaps something a bit simpler like `<FollowIcon following>`.

src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -1,7 +1,9 @@
-import { Separator, Serif } from "@artsy/palette"
+import { Separator, Serif, Spacer } from "@artsy/palette"
+import { Box } from "@artsy/palette"
+import { ContextConsumer } from "Artsy/Router"
+import Spinner from "Components/Spinner"
```
damassi: import { Spinner } from '@artsy/palette'. (Not sure this has been updated in ArtistDetails, but should def come from palette.)

src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -80,3 +85,48 @@ export const ArtworkSidebarFragmentContainer = createFragmentContainer(
...
+                    artwork={props.artwork as any}
+                  />
+                )
+              } else {
+                return (
+                  <SpinnerContainer
+                    width="100%"
+                    height="100px"
+                    position="relative"
+                  >
```
damassi: Can these props be passed directly to `<Spinner>`? And if not, it might be worthwhile to update spinner in pallete to have a container.

src/Apps/Artwork/Components/__stories__/ArtworkSidebar/ArtworkSidebarArtists.story.tsx
```diff
@@ -7,19 +7,22 @@ import { ArtworkSidebarArtists as Artists } from "Apps/Artwork/Components/Artwor
 import React from "react"
 import { storiesOf } from "storybook/storiesOf"
 import { Section } from "Styleguide/Utils/Section"
+import { RelayStubProvider } from "Utils/RelayStubProvider"
```
damassi: Can this be moved to `DevTools` folder?

src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
alloy: Please use this function instead, as it will make sure tooling like renderRelayTree works, we can add error/retry handling in one place: https://github.com/artsy/reaction/blob/master/src/Artsy/Relay/renderWithLoadProgress.tsx

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -148,23 +149,7 @@ export const ArtistInfoQueryRenderer = ({ artistID }: { artistID: string }) => {
...
-                )
-              } else {
-                return (
-                  <Box width="100%" height="100px" position="relative">
-                    <Spinner />
-                  </Box>
-                )
-              }
-            }}
+            render={renderWithLoadProgress(ArtistInfoFragmentContainer as any)}
```
oxaudo: @alloy - updating here per your pointer. This def looks way simpler in this class! But to get it to compile I had to cast ArtistInfoFragmentContainer as any. Is that acceptable or am I missing something here?

damassi:
👍

src/Apps/Artwork/Components/ArtistInfo.tsx
alloy: Huh, interesting, I’m pretty sure I tested this to work. I’ll pull this down tomorrow and have a look.

alloy: Seems to work fine for me. Maybe your vscode needs a ‘Reload Window’?

screen shot 2018-10-11 at 01 36 35

PR #1389, opened 2018-10-10
[FEATURE collection] adds collections header

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-924

This extracts the Collection Header Component from https://github.com/artsy/reaction/pull/1382 into its own PR. How the data hooks up to the `CollectionApp` can be addressed in a separate PR. I think the shape of the data may change slightly once we have everything hooked up but I think this gets us pretty close to how things might look.


artsyit:

Built with commit 6a24a65f4bf9e79f38c68677ce34874e5e119d0b

https://deploy-preview-1389--artsy-reaction.netlify.com

Issue #1386, opened 2018-10-10
New Tracking Patterns


damassi:
After being tasked with implementing tracking on a few areas of the app I got a chance to dive into our new tracking patterns established around the time the new /artist app was built and it got me thinking about the possibility that we're engineering a solution to something that might not be a problem and that there might be a complexity cost at scale. 

New pattern: 
https://github.com/artsy/reaction/blob/master/src/Artsy/Analytics/Schema/Values.ts

How its used: 

```js
@track({ 
  context_module: Schema.Context.MinimalCtaBanner 
})
export class MinimalCtaBanner extends React.Component {
  @track({
    action_type: Schema.ActionType.AuthImpression,
    intent: Schema.ActionName.ViewEditorial,
    trigger: "click",
  })
  componentDidMount() {
    // no op
  }
}
```

Old pattern: 
https://github.com/artsy/force/tree/master/src/desktop/analytics 
https://github.com/artsy/force/blob/master/src/desktop/analytics/artwork.js (example context)

How its used:

The Analytics team passes us a snippet: 

```js
analytics.track('Click', {
  flow: 'Artwork about the work',
  context_module: 'About the Work (Partner)',
  type: 'Button',
  label: 'Read more',
})
```

Which we would then paste in via `react-tracking`: 

```js
@track({
  flow: 'Artwork about the work',
  context_module: 'About the Work (Partner)',
  type: 'Button',
  label: 'Read more',
  trigger: "Click",
})
```

In the new system, we take the values that Analytics passes us and we encode them into enums under specific buckets grouped by type, and then we import those types into our code and pass into `@track()`. 

What are the benefits of type safety in regards to our tracking code?

Things like `ActionType` — repetitive bits that we see duplicated all throughout old analytics code — would benefit. That way we don’t have things like “Click” or “Clickk”, etc. Those properties can be encoded in an enum or variable and make sense. 

<img width="376" alt="screen shot 2018-10-09 at 5 03 06 pm" src="https://user-images.githubusercontent.com/236943/46707281-20ef2a80-cbee-11e8-8179-41981ea43280.png">

But when looking at things like `ActionName`, these are unique values:

<img width="354" alt="screen shot 2018-10-09 at 4 38 06 pm" src="https://user-images.githubusercontent.com/236943/46707297-35cbbe00-cbee-11e8-8377-c84a1f61e087.png">

[...]

PR #1391, opened 2018-10-11
Add Tracking

damassi:
Adds tracking for: 

- https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&modal=detail&selectedIssue=DISCO-361&sprint=143
- https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&modal=detail&selectedIssue=DISCO-362&sprint=143
- https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&modal=detail&selectedIssue=DISCO-363&sprint=143
- https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=57&modal=detail&selectedIssue=DISCO-324

Additionally, adds a few new tracking fields to the main Trackables type: 
- `Flow`
- `Label`
- `Type`

so that things match up to tracking code segment props: 

```ms
analytics.track('Click', {
  context_module: 'Biography',
  flow: 'Artwork about the artist',
  type: 'Button',
  label: 'Read more',
})
```
-> 
```js
props.track({
  flow: Schema.Flow.ArtworkAboutTheWork,
  type: Schema.Type.Button,
  label: Schema.Label.ReadMore,
})
```

Also refactors `Schema.Context` to `Schema.ContextModule` to map 1:1 to segment property. 

As a follow up to the discussion started in https://github.com/artsy/reaction/issues/1386 around these new patterns, after working through these tickets there was definitely more duplicate data than expected across all three tickets that lends itself to a schema.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkDetails/index.tsx
+++ b/src/Apps/Artwork/Components/ArtworkDetails/index.tsx
@@ -13,14 +13,13 @@ import { ArtworkDetailsChecklistFragmentContainer as Checklist } from "./Artwork
 import { ArtworkDetails_artwork } from "__generated__/ArtworkDetails_artwork.graphql"
 import { ArtworkDetailsQuery } from "__generated__/ArtworkDetailsQuery.graphql"
 
-import { track, Track } from "Artsy/Analytics"
+import { track } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
+import { TrackingProp } from "react-tracking"
 
 export interface ArtworkDetailsProps {
   artwork: ArtworkDetails_artwork
-  tracking?: {
-    trackEvent: Track
-  }
+  tracking?: TrackingProp
 }
 
 const ArtworkDetailsContainer = Box
--- a/src/Apps/Artwork/Components/ArtworkDetails/index.tsx
+++ b/src/Apps/Artwork/Components/ArtworkDetails/index.tsx
@@ -1,7 +1,7 @@
 import { Box } from "@artsy/palette"
 import { renderWithLoadProgress } from "Artsy/Relay/renderWithLoadProgress"
 import { ContextConsumer } from "Artsy/Router"
-import React from "react"
+import React, { Component } from "react"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 import { Tab, Tabs } from "Styleguide/Components"
 import { ArtworkDetailsAboutTheWorkFromArtsyFragmentContainer as AboutTheWorkFromArtsy } from "./ArtworkDetailsAboutTheWorkFromArtsy"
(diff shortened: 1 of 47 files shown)


src/Artsy/Analytics/Schema/index.ts
```diff
@@ -7,24 +7,32 @@
...
 export type Trackables =
   | AuthenticationInteraction
   | ContextModule
   | ContextPage
+  | Flow
   | Interaction
+  | Label
   | Success
   | Failure
+  | Type
```
damassi: I was a bit unsure about these additions but it _seemed_ correct. These values should map 1:1 to the segment schema, right? 

so 

```js
context_module: 'Biography',
flow: 'Artwork about the artist',
type: 'Button',
label: 'Read more',
```
-> 
```
```js
context_module: ContextModule.Biography
flow: Flow.ArtworkAboutTheArtist,
type: Type.Button,
label: Label.ReadMore
```

And as more segment fields are added we'll update this similar to how I have above?

src/Apps/Artwork/Components/ArtworkDetails/index.tsx
```diff
@@ -23,32 +28,49 @@ export const ArtworkDetails: React.SFC<ArtworkDetailsProps> = props => {
...
   return (
     <ArtworkDetailsContainer pb={4}>
-      <Tabs>
-        <Tab name="About the work">
+      <Tabs
+        onChange={({ data }) => {
+          props.tracking.trackEvent({
+            flow: Schema.Flow.ArtworkAboutTheArtist,
+            type: Schema.Type.Tab,
+            label: data.trackingLabel,
```
damassi: Is it possible to look up an enum value dynamically? Like: 

```js
export enum Label {
  Foo = 'foo',
  Bar = 'bar',
  Baz = 'baz'
}
...

const foo = 'Foo'
...
label: Schema.Labels[foo]
```

alloy:
After all of these comments, I realise we should probably just have a linter rule, for which I filed https://artsyproduct.atlassian.net/browse/PLATFORM-973

src/Artsy/Analytics/Schema/index.ts
damassi: Would it be possible to dumb this down a bit and create a `Fields` interface that lists out segment properties? Then we can associate those properties with their associated enums in Values. It seems like starting from there we could lock things down further in the future if need be and the system becomes a little less confusing while we're still getting started and defining it.

alloy: The reason these shapes have separate interfaces is so you only use the right set of properties together. See e.g. https://github.com/artsy/reaction/blob/master/src/Artsy/Analytics/Schema/Interaction.ts, these are the shapes that either a `Interaction` or a `AuthenticationInteraction` should have.

Your event looks to be an `Interaction` which doesn’t have `flow` yet, so if the analytics team is sure that they could potentially see `flow` props on any interaction event then it should be added there.

damassi: (We can follow up in another PR)

alloy:
Merge on green

PR #1394, opened 2018-10-11
Updates addressErrors when any component of address is changed

ashfurrow:
This PR fixes [PURCHASE-564](https://artsyproduct.atlassian.net/browse/PURCHASE-564), a bug reported from QA where a user was asked to enter a postal code despite having selected a non-US/Canada country. This was a frontend validation issue, Exchange was fine.

It looks like we were running the validations for everything, but only keeping the component of the address that had changed. This would work if a change in one part of the address form wouldn't affect the validation of another part, but changing your country changes your postal code validation, 

The simplest fix, to me, is to just grab all the validation errors. @erikdstock I'm keen to get your input if that will be okay. I also added test coverage for this front-end validation.

### DIFF
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -16,6 +16,7 @@ export interface Address {
 }
 
 export type AddressErrors = Partial<Address>
+export type AddressTouched = Partial<{ [T in keyof Address]: boolean }>
 export type AddressChangeHandler = (
   address: Address,
   key: keyof Address
@@ -36,7 +37,8 @@ export interface AddressFormProps {
   defaultValue?: Partial<Address>
   billing?: boolean
   continentalUsOnly?: boolean
-  errors?: AddressErrors
+  errors: AddressErrors
+  touched: AddressTouched
 }
 
 interface AddressFormState {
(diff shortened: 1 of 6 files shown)


artsyit:

Built with commit 8c3c6b1b8f57b69095e523816902d5ec9d070509

https://deploy-preview-1394--artsy-reaction.netlify.com

sweir27:
Oh whoa, did not realize this was the fix. Good sleuthing @ashfurrow! 

Will wait for the go-ahead from @erikdstock .

erikdstock:
Ah of course!

ashfurrow:
Noting here for visibility, Erik has pointed out that this would cause the whole form to be invalid right away (since they're all empty). Formik uses a value called `touched` to keep track of this – Erik's tackling this and will send a PR. Thanks Erik!

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -70,6 +72,16 @@ export class AddressForm extends React.Component<
     })
   }
 
+  getError = (key: keyof Address): string => {
+    return (
+      (this.props.touched &&
+        this.props.touched[key] &&
+        this.props.errors &&
+        this.props.errors[key]) ||
+      ""
```
ashfurrow: @erikdstock For this, it looks like we're returning an empty string instead of `null` or `undefined` here. Should we change that? Or is it fine as-is?

ashfurrow:
Okay this should be good. Merge on green.

PR #1401, opened 2018-10-12
[FIX BNMO] Refetch data on every path within the orders app

sweir27:
We actually need this on every route because they all depend on data potentially updated and that can be changed by navigating backwards and forwards.

Also, I'm pretty sure I updated yarn yesterday-- is this change to `yarn.lock` in the right direction or the wrong one?


damassi:
👍 

Re yarn, not too sure. (I suspect someone else hasn't updated.)

PR #1407, opened 2018-10-13
[WIP] Setup PR based release process

### LINKED ISSUE #1095: [RFC] Change rules around automatic deploys, and add a CHANGELOG
## Proposal:

This proposes 2 changes:

- Move Reaction to deploy on every PR
- Enforce adding a user-facing CHANGELOG entry for Major and Minor changes

What the RFC aims to do is:

### Deployment

1. Someone makes a PR to Reaction
2. They add one of the following labels: `"Deploy: Patch"`, `"Deploy: Feature"` or `"Deploy: Breaking"`.
3. If they don't add a label, Peril will add  `"Deploy: Patch"` for you (basically making it the default)
4. When the PR is merged, the CI on master uses the GitHub API to determine which label is added and to deploy a version with that semantic version bump

### CHANGELOG

A Danger rule can be added so that any PRs which are "Deploy: Feature" or "Deploy: Breaking" should include a user-facing changelog entry.

## Reasoning

There are now 3 clients for Reaction. We want to increase visibility into what is changing, how and why. A list of PR titles are not useful enough to provide that kind of context in comparison to seeing what how useful the [Emission changelog](https://github.com/artsy/emission/blob/master/CHANGELOG.md) is for consumers and answering questions from product folk. 

## Exceptions:

There's a possibility that you could merge a PR without one of those labels, ideally that wouldn't happen but in that case I guess it would just not do a deploy.

## Additional Context:

You can see our discussion [in slack here](https://artsy.slack.com/archives/C02BC3HEJ/p1533661877000050) and [here](https://artsy.slack.com/archives/C2TQ4PT8R/p1533657945000171)

zephraph:
Fixes #1095

This PR will fundamentally rework how our deployment process happens. 

The changes are as follows:
- Changelog entries will be required
- The release process will no longer use semantic-release
- Releases will effectively happen on every PR merge*
- The PR must include a label that describes the type of change
  - `release:patch`
  - `release:minor`
  - `release:major`
  - `release:none`

Remaining work:

- [ ] Ensure changelog entries are required before merging (peril may already handle this)
- [ ] Publish releases to github
- [ ] Replace semantic release command (semantic release will be removed in a follow up PR once we've verified all functionality)
- [ ] Setup CircleCI config to support a git push
- [ ] Cover corner case where `git push` succeeds but `npm publish` fails
   - In this case the changelog will be updated with a version that may never show up in the releases. We'll need to check the tags and if the latest tag is of a lower version than the changelog then we'll want to group those changelog changes in with the master section and continue with the changelog release.

Follow up work:

- Remove semantic-release
- Pull release process out into a separate package
- Allow changelog entries in the PR body

---

*_Note: Due to timing issues w/ pull requests it's possible that multiple PRs will be grouped together in one release with the one of the largest change determining the version_


damassi:
@zephraph - q: 

> The PR must include a label that describes the type of change

How should one think about PRs that are unrelated to `@artsy/reaction` stuff? (tooling, internals, etc)

artsyit:

Built with commit 07b2133768561f0f945a1047d1b77107571a85f6

https://deploy-preview-1407--artsy-reaction.netlify.com

zephraph:
@damassi I was talking to @orta today about potentially having a `release:none` tag or something. If it truly didn't change the source then we probably wouldn't want to trigger a release for it. 

With that being said a lot of times making a tooling change _does_ effect the source, in which case we should always use standard semantic versioning (IMO).

damassi:
That makes sense. 

Another q: 

Also, why migrate away from `semantic-release` to something that's custom? There are lots of contributors and we're also following this pattern on other repos. Why not leverage command-line usage and hook into labels? I'm a little wary of roll-your-own solutions when there are many solid options in the community.

artsyit:

Built with commit 5d339de2bb1d91b247b92124cf3e3a2d62f2359f

https://deploy-preview-1407--artsy-reaction.netlify.com

zephraph:
We don't have to roll our own solution for actually publishing the package. I'm fine with using a tool for that, but semantic-release covers and entirely different usecase. The point of this PR is to essentially trigger releases on every PR (or most every PR) _without_ having to write a custom commit message. Every time master CI runs, it'll check if it's the latest release (if it's tagged). If it's not, it'll look for all changed PRs since the last release, gather up their changelog entries, update the changelog, update the version, commit all that and tag the commit w/ the version. When CI runs on a commit w/ the latest tag it'll automatically deploy (or try to deploy) that version. 

It's not really too crazy of a process and @orta has largely implemented most of the individual parts w/ danger. It's all about stitching things together. When this is done though, there should be very little custom code.

zephraph:
Let me get this first pass done. If you still have concerns, I'm more than willing to try to do it differently. Just trying to make this process as easy as possible. Definitely don't want to roll too much of my own stuff here 😅 .

damassi:
Cool cool, lets be sure to keep semantic-release's plugin architecture in mind during this POC, which can interface with things like [`conventional-changelog`](https://github.com/conventional-changelog/conventional-changelog) so a lot of work can be automated away.

alloy:
Seems like I didn’t actually post my comment previously. I like it! The one question I had basically echoes @sweir27’s question for an example.

zephraph:
Good questions @sweir27! The changelog entry in the PR isn't a necessary feature for this, but I thought it could be nice. I was thinking it could look something like the following...

```markdown
## Changes
- Added a release process that triggers on PR labels
- Added the ability to specify changelog entries through PR bodies
```

Then after the PR is merged, the initial master build grabs the PR (already does this to get the label) and can commit and update to the changelog. Ultimately though, as I said, it isn't necessarily required. We could just use a danger rule to enforce changelog updates. 

As for the default release type I was planning on adding a danger rule that fails CI if there isn't a label added. We could specify a default and we'd talked about that in the RFC. I'm just worried people would forget to check the label.

scripts/release/changelog.ts
```diff
@@ -0,0 +1,61 @@
...
+  return formatChangelog(changelog)
+}
+
+export const readChangelog = (changelogPath = CHANGELOG_PATH) =>
+  parseChangelog(changelogPath)
+
+export const writeChangelog = (
+  changelogMarkdown: string,
+  changelogPath = CHANGELOG_PATH
+) => write(changelogPath, changelogMarkdown)
```
damassi: Nice and simple

damassi:
@zephraph - would you mind turning the description list items into a check-list for tracking what has been completed? Makes it slightly easier to review as parts come in.

PR #1408, opened 2018-10-14
Collection page

l2succes:
This refactors /collect to have better naming for modules and refactors filter components to be reused for the collection app.

TODO:
- [x] fix refetch container for collect
- [x] fix refetch container for collection
- [ ] add tests for remaining components


artsyit:

Built with commit 432e5ba8fde38a7a1f95bcae15684fa13b8c5044

https://deploy-preview-1408--artsy-reaction.netlify.com

src/Apps/Collect/Components/Base/CollectRefetch.tsx
```diff
@@ -1,18 +1,16 @@
...
 import { isEqual } from "lodash"
 import React, { Component } from "react"
 import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
 import { Responsive } from "Utils/Responsive"
 import { CollectArtworkGridRefreshContainer as ArtworkFilter } from "./CollectArtworkGrid"
-interface CollectArtworkFilterRefetchProps {
+interface CollectRefetchProps {
   filters: any
-  viewer: CollectArtworkFilterRefetch_viewer
+  viewer: any
```
alloy: Is there a reason to type this as any? (E.g. limitation in our type emitting code?)

l2succes: I had to set it to `any` while I was renaming modules but I'll add the type back now 👍

alloy:
From skimming it seems like users of the filter components each need to define their own fragments and repeat all arguments etc, is that correct? If so, why not move the fragments back into the components and have them be generic by defining a “filter interface” in MP?

src/Apps/Collect/Components/Collection/CollectionRefetch.tsx
```diff
@@ -0,0 +1,139 @@
+import { isEqual } from "lodash"
+import React, { Component } from "react"
+import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
+import { Responsive } from "Utils/Responsive"
+import { CollectArtworkGridRefreshContainer as ArtworkFilter } from "../Base/CollectArtworkGrid"
+
+interface CollectionRefetchProps {
+  filters: any
```
eessex: types need updated here too

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return (
+      <CollectFrame>
+        <CollectionHeader
+          description={
+            <div dangerouslySetInnerHTML={{ __html: collection.description }} />
+          }
+          // image={collection.headerImage.large }
+          image="https://artsy-vanity-files-production.s3.amazonaws.com/images/kaws2.png"
+          // image_caption={collection.image_caption}
```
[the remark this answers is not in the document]
l2succes: Will break that up in a separate ticket / PR as we discussed

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+          slug={collection.slug}
+          title={collection.title}
+        />
+        <Box>
+          <CollectionFilterFragmentContainer collection={collection} />
```
damassi: `import { CollectionFilterFragmentContainer as CollectionFilter } ...`

src/Apps/Collect/Components/Base/CollectFilterContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <ContextConsumer>
+        {({ user, mediator }) => {
+          return (
+            <Responsive>
+              {({ xs }) => {
+                return (
+                  <FilterContainer
+                    isMobile={xs}
+                    user={user}
+                    mediator={mediator}
```
damassi: Rather than passing mediator around the component should pull from context

src/Apps/Collect/Components/Base/CollectFilterContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+                    mediums={mediumAggregation.counts as any}
+                  >
+                    {({ filters }) => (
+                      <CollectRefetchContainer
```
damassi: The convention we've been following in other relay app code is to typically export the full name + relay wrapper type and then rename at the import so as to not encode relay behavior into runtime code.

src/Apps/Collect/Components/Base/CollectFilterContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+            </Responsive>
+          )
+        }}
+      </ContextConsumer>
+    )
+  }
+}
+
+export const CollectFilterFragmentContainer = createFragmentContainer(
+  CollectFilterContainer,
```
damassi: Not sure how necessary this is in this instance, but render props code can be made more useful throughout a full class if passed in and accessed as props (similar to whats [happening here](https://github.com/artsy/reaction/blob/912e401b1c148fdb7c8c596d9d2f863da6c73b18/src/Apps/Comparables/Filter.tsx#L68-L88). Since the component is already been wrapped it makes this pattern easy to implement.

src/Apps/Collect/Components/Filters/index.tsx
```diff
@@ -0,0 +1,7 @@
+export { FilterContainer } from "./FilterContainer"
+
+export { MediumFilter } from "./MediumFilter"
+export { PriceRangeFilter } from "./PriceRangeFilter"
+export { SortFilter } from "./SortFilter"
+export { TimePeriodFilter } from "./TimePeriodFilter"
+export { WaysToBuyFilter } from "./WaysToBuyFilter"
```
damassi: So that the context renderProps pattern doesn't feel too verbose, could create a factory function that wraps and injects items in as props:

```jsx
export const MediumFilter = makeFilter(...)
```

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { CollectionFilterFragmentContainer } from "./Components/Collection/CollectionFilterContainer"
+import { CollectionHeader } from "./Components/Collection/Header"
+
+interface CollectionAppProps {
+  collection: CollectionApp_collection
+}
+
+export class CollectionApp extends Component<CollectionAppProps> {
+  render() {
+    const collection = this.props.collection
```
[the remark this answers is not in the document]
l2succes: na there isn't 😔

l2succes:
@alloy I was actually going that route but there are some minor differences between the queries for various filter use-cases. The hardest one to deal with being the `RefetchContainer`, I couldn't figure out how to make the refetch query generic so I reverted to creating separate containers for collect and collection but with some metaphysics changes we should be able to make that work.

alloy:
> @alloy I was actually going that route but there are some minor differences between the queries for various filter use-cases. The hardest one to deal with being the `RefetchContainer`, I couldn't figure out how to make the refetch query generic so I reverted to creating separate containers for collect and collection but with some metaphysics changes we should be able to make that work.

Gotcha 👍

Let‘s follow up on that once this is shipped, although not sure where 🤔 Should it be a tech-debt ticket on Grow, DiscoValuate (cc @mzikherman), or seeing as it touches multiple teams maybe on client-infra?

src/Apps/Collect/CollectionApp.tsx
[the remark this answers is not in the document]
l2succes: Yeah, we'll go with that option
@xtina-starr will update as part of the responsive header work / passing in collection as a prop

src/Apps/Collect/Components/Base/CollectFilterContainer.tsx
l2succes: Yup, makes sense

l2succes:
Alright thanks for the feedback everyone, everything's been addressed. We'll go ahead and merge this and follow up with some other fixes in a separate PR.

l2succes:
@alloy I think client-infra could make sense for this. We can also have grow do it as part of iterating on collection.

alloy:
Cool, can you file one?

l2succes:
I created it on Platform (Client-Infra) https://artsyproduct.atlassian.net/browse/PLATFORM-1010 as you may get to it first but we can always move it.

PR #1417, opened 2018-10-15
beginning of test for artwork Details

oxaudo:
This adds a super simple test for the following not present for auction works.

Not sure how I feel about testing the actual html for strings but not clear how to test for an actual component render here. I think that'll do for now.

### DIFF
--- a/src/Apps/Artwork/Components/__tests__/ArtworkDetails.test.tsx
+++ b/src/Apps/Artwork/Components/__tests__/ArtworkDetails.test.tsx
@@ -1,0 +1,44 @@
+import { cloneDeep } from "lodash"
+import { graphql } from "react-relay"
+import { renderRelayTree } from "../../../../DevTools"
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit c5cd2438e91e2a17ec41115c95ac80f4254eb396

https://deploy-preview-1417--artsy-reaction.netlify.com

src/Apps/Artwork/Components/__tests__/ArtworkDetails.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      `,
+      mockResolvers: {
+        Artwork: () => response,
+      },
+    })
+  }
+
+  it("renders a correct component tree for artwork with all details", async () => {
+    const wrapper = await getWrapper()
+    expect(wrapper.html()).toContain("About the work")
```
damassi: `const html = wrapper.html()`
(then pass in all the way down)

PR #1422, opened 2018-10-16
Migrates to use NPM package for linter rule

ashfurrow:
(Not urgent, I know you're out of office so no need to review now!)

I finished publishing [tslint-plugin-relay](https://github.com/relay-tools/tslint-plugin-relay), my first NPM package! It only has one rule, which I took from this repo, so I figured it's time for this repo to point to the new package. I've tested that this still works locally.


alloy:
Sweeeeeet!

PR #1424, opened 2018-10-17
Adds mediator to OrderApp storybooks

sweir27:
We're currently unable to open the storybooks for the order app since they now require `mediator` to be present in the context. 

Note that the `ShippingAndPaymentSummary` storybook is still broken, but for a different reason that will take some more investigation.


artsyit:

Built with commit d40daecbf1a8ba342b097322795a120fb631e147

https://deploy-preview-1424--artsy-reaction.netlify.com

PR #1426, opened 2018-10-18
[Collection] add test to CollectionApp

l2succes:
Related: https://artsyproduct.atlassian.net/browse/GROW-949

### DIFF
--- a/src/Apps/Collect/__tests__/CollectionApp.test.tsx
+++ b/src/Apps/Collect/__tests__/CollectionApp.test.tsx
@@ -1,0 +1,211 @@
+import { mount } from "enzyme"
+import React from "react"
+import { graphql } from "react-relay"
+
+import { Provider } from "unstated"
+import { MockBoot, MockRelayRenderer, renderUntil } from "../../../DevTools"
+import { CollectionAppFragmentContainer as CollectionApp } from "../CollectionApp"
+import { FilterContainer } from "../Components/Filters"
+import { FilterState } from "../FilterState"
+
+jest.unmock("react-relay")
+
+describe("CollectionApp", () => {
+  let filterState: FilterState = null
+
+  beforeEach(() => {
+    filterState = new FilterState({
+      tracking: {
+        trackEvent: jest.fn(),
+      },
+    })
+  })
+
+  it("renders a relay tree correctly", async () => {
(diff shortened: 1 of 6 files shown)


artsyit:

Built with commit e495e82c0c3312518f732563f504308784e307db

https://deploy-preview-1426--artsy-reaction.netlify.com

src/Apps/Collect/__tests__/CollectionApp.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+  beforeEach(() => {
+    filterState = new FilterState({
+      tracking: {
+        trackEvent: jest.fn(),
+      },
+    })
+  })
+
+  it("renders a relay tree correctly", async () => {
+    const tree = await renderUntil(
```
alloy: Was there a reason you didn’t use `renderRelayTree`?

l2succes: It didn't work for me when I first tried but I may have done something wrong. so I ended up testing different conditionals to stop the render loop. 
Also I wanted to understand the underlying parts of it but we can always update it after

alloy: Great, understanding the parts is always a good thing 👍 Let’s connect on Monday about why it didn’t work for you.

l2succes: 👍sounds good

alloy: Is there by any chance a nested QueryRenderer that isn’t using renderWithLoadProgress?

l2succes: I don't think so, there's only one root QueryRenderer. I'll try switching it again to see what happens

alloy: Or if there’s any other special work that happens that may have been deferred until a later render pass? In which case you should use the `renderUntil` parameter for `renderRelayTree` (which appears to still be called `until` in the doc) https://github.com/artsy/reaction/blob/bdc26827d58b78b5c13a6603d6b8768269c67c4e/src/DevTools/renderRelayTree.tsx#L25-L27

PR #1429, opened 2018-10-19
[PATCH collection] fixes mobile image and other minor QA items

xtina-starr:
This mainly fixes the header image for mobile but also addresses some other minor QA items.

- Increases header title and subtitle font size in mobile.
- Replaces category placeholder.
- Removes the padding on the side of description.

Regarding the extra margin at the top, it looks like that is something that would need to be addressed in for as that where it is being set. I noticed that the same also happens on the artist page.


artsyit:

Built with commit d96950ccd8a8c17e68f266eac51059550cc92cc1

https://deploy-preview-1429--artsy-reaction.netlify.com

PR #1430, opened 2018-10-21
[typings] Remove unneccesary any casts.

alloy:
@oxaudo Not sure why you needed to add these, maybe there’s something wrong with your local env?


oxaudo:
Thanks!

I'll investigate what's up if you are saying it should not be the case and it's not giving you any warnings locally.

alloy:
Please do 👍

PR #1431, opened 2018-10-22
Use scroll management for found

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-572

This PR fixes a general problem with our `found` (router) setup. It was not preserving browser behaviour regarding scroll position when navigating between pages. Luckily someone made this `found-scroll` package which is a drop-in decorator providing the correct behaviour. I had a little play around in the order app and it seems to work perfectly.

I tried to calculate what impact it has on bundle size, but webpack keeps running out of memory when I try to build force right now. I'm guessing it's < 1kb judging from the look of the code but could be wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,7 @@
     "formik": "^0.11.11",
     "found": "^0.3.14",
     "found-relay": "^0.3.1",
+    "found-scroll": "^0.1.5",
     "history": "^4.6.1",
     "isomorphic-fetch": "^2.2.1",
     "jsonp": "^0.2.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -157,6 +157,7 @@
     "formik": "^0.11.11",
     "found": "^0.3.14",
     "found-relay": "^0.3.1",
+    "found-scroll": "^0.1.5",
```
damassi: Ah yeah, this is also the pattern in React Router (of which this `found` is basically a rewrite of v3).

damassi:
👍 👍

PR #1432, opened 2018-10-22
[type checking test files] ArtistBio, etc

ds300:
### See #1166 for background and approach

In this exciting second edition, we:

- Replaced ArtistBio tests that used the presentational component with ArtistBio tests that use the fragment container.
- Sprinkled some `any` dust around a few other files

🕺


damassi:
LGTM! I was going to pick up some of this work this week (it's "write test week" on discover team), so great timing

PR #1433, opened 2018-10-22
@l2succes => [Collections] use hardcoded medium list

eessex:
Addresses https://artsyproduct.atlassian.net/browse/GROW-947

- Use hardcoded list of mediums for filter if none are returned from query
- Adds `.cache-loader` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 .awcache
 .cache
+.cache-loader
 .env
 .storybook/.DS_Store
 .storybook/manager.js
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 .awcache
 .cache
+.cache-loader
```
alloy: This is an old directory that you can `rm -rf`, it doesn’t need to be ignored anymore.

eessex: Maybe worth leaving there for a minute since it was recently deprecated?  Was surprised to have 1500 changed files on a PR today, and had no reference for what `.cache-loader` referred to.

eessex: Or just shouting out to #dev slack that we should remove

alloy: In retrospect, I think leaving the directory in the ignore list for now is probably the most workable solution 👍

PR #1435, opened 2018-10-22
add specs for ArtworkDetailsFromPartner

oxaudo:
This mainly adds specs but in process uncovered what I think might have been a bug. Basically we would never have avatar with initials for partner if icon url was not present. Corrected that, but maybe side effect that now we will display icon with initials for auctions as well. Not sure if that is what we wanted.

screen shot 2018-10-22 at 1 39 21 pm

Previously we would have no icons neither for auctions not for partners with profile image.


src/Styleguide/Components/EntityHeader.tsx
```diff
@@ -28,7 +28,7 @@ export const EntityHeader: SFC<EntityHeaderProps> = ({
 
   return (
     <Container onClick={() => handleContainerClick()} hasLink={!!href}>
-      {imageUrl && (
+      {(imageUrl || initials) && (
```
oxaudo: That is the change in question @damassi

PR #1441, opened 2018-10-23
Starting to add ArtworkSidebar spec

oxaudo:
Created one spec for the whole Sidebar but think all the specifics should be tested in each components.

Also renamed fixtures to follow the structure of the component implementation.
In general - not sure if it would be more convenient if Fixtures lived inside the components. Found myself constantly jumping between fixtures and components.


artsyit:

Built with commit e5ddd884cdf0329a4b8e585dc0fc8644a51363b9

https://deploy-preview-1441--artsy-reaction.netlify.com

damassi:
LGTM 👍 - lets continue building this out as more real content starts coming online in the new page.

PR #1451, opened 2018-10-24
Feature/offer page

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-354

Desktop

https://zpl.io/2yy0ero

image

Mobile:

https://zpl.io/aw1DnGg

image

Notice that formatting the input field is not possible with standard type=number fields. We need to iterate on that, but there's plenty of prior art.


artsyit:

Built with commit 5ebba899d8483baed2bd0655d9ded0a85c387ea6

https://deploy-preview-1451--artsy-reaction.netlify.com

src/Apps/Order/Components/OrderStepper.tsx
```diff
@@ -0,0 +1,32 @@
...
+  | {
+      currentStep: "Offer" | "Shipping" | "Payment" | "Review"
+      makeOfferFlow: true
+    }
+  | {
+      currentStep: "Shipping" | "Payment" | "Review"
+      makeOfferFlow: false
+    }
+
+export const OrderStepper: React.SFC<OrderStepperProps> = ({
```
sweir27: Nice 🙏 . This will extend nicely when we add the counteroffer flow.

src/Apps/Order/Components/TransactionSummary.tsx
```diff
@@ -53,6 +56,10 @@ export class TransactionSummary extends React.Component<
       textOverflow: "ellipsis",
     } as any
 
+    const isMakeOfferFlow =
```
sweir27: General question throughout: Do we want to be referring to this as the `makeOfferFlow`? or just `offerFlow`? The "Make" seems a little redundant + tied to current business naming.

On the other hand it might be a nice distinction here to talk about the `makeOfferFlow` vs. the `counterofferFlow`. 🤷‍♀️ food for thought but interested in what others think.

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const dollarDigits = dollars.split("")
+
+  for (let i = dollarDigits.length - 3; i > 0; i -= 3) {
+    dollarDigits.splice(i, 0, ",")
+  }
+
+  const formattedDollars = dollarDigits.join("")
+
+  if (cents === "00") {
+    return `$${formattedDollars}`
```
sweir27: We should at least add a TODO here about supporting an actual currency if that's known.

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  state = {
+    offerValue: null,
+    isCommittingMutation: false,
+    isErrorModalOpen: false,
+    errorModalTitle: null,
+    errorModalMessage: null,
+  }
+
+  componentDidMount() {
+    this.props.mediator.trigger("order:offer")
```
sweir27: Heh, adding in analytics early! This will need a corresponding update to force here: https://github.com/artsy/force/blob/9648208dae29e94fcddd76cdfd71eb846f27e53e/src/desktop/apps/order2/client.js#L33 

You could also just leave out this + the `@track`s for now and we can add later when we have specs (although this one for pageview we'll certainly want).

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  onContinueButtonPressed: () => void = () => {
+    this.setState({ isCommittingMutation: true }, () => {
+      if (this.props.relay && this.props.relay.environment) {
+        // TODO: commit mutation
+        new Promise(resolve => setTimeout(resolve, 1000)).then(() => {
+          this.setState({ isCommittingMutation: false })
+          this.setState({
+            isErrorModalOpen: true,
+            errorModalTitle: "Congratulations",
```
sweir27: hehe nice. To make sure all the code here is used, maybe we can call `onMutationError` here

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+export interface OfferState {
+  offerValue: number | null
+  isCommittingMutation: boolean
+  isErrorModalOpen: boolean
+  errorModalTitle: string
+  errorModalMessage: string
+}
+
+function formatMoney(money: number | undefined): string | null {
```
sweir27: Would be good to have some tests for this behavior (even unit tests for the method would be 👍 since it has a lot of logic). Wonder if we also want to pull it into a separate `util` folder.

src/Apps/Order/Components/TransactionSummary.tsx
damassi: I think `offer` and `counteroffer` is how its used in everyday language, so that maps nicely.

src/Apps/Order/Routes/Offer/index.tsx
l2succes: You can actually use `toLocaleString` to format a number into it's currency format. See (https://github.com/artsy/reaction/blob/master/src/Styleguide/Components/PriceRange.tsx#L50)

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+                        Continue
+                      </Button>
+                    )}
+                  </Flex>
+                }
+                Sidebar={
+                  <Flex flexDirection="column">
+                    <TransactionSummary
+                      order={order}
+                      mb={xs ? 2 : 3}
```
damassi: `mb={[2, 3]}`

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+
+const OfferRouteWrapper = props => (
+  <ContextConsumer>
+    {({ mediator }) => {
+      return <OfferRoute {...props} mediator={mediator} />
+    }}
+  </ContextConsumer>
+)
```
damassi: 👍

src/Apps/Order/Routes/Offer/index.tsx
damassi: Aw nice, was not aware of this.

l2succes:
My only feedback is use the built in converter to convert a number to a currency string

```
(49999).toLocaleString("en-US", {
      style: "currency",
      currency: "USD",
      minimumFractionDigits: 0})
"$49,999"
```

ds300:
Great feedback everyone, thanks!! I think I've taken care of everything except for testing the redirect logic, which is temporary until we have some way to determine whether an order is make-offerable. Does that sound OK to you @sweir27 ?

@l2succes :scream: I had no idea `toLocaleString` can do currencies, that's amazing! Thanks for pointing it out :bow:

src/Apps/Order/Routes/__tests__/Offer.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+      `,
+      mockResolvers: {
+        Order: () => UntouchedOrder,
+      },
+      renderUntil: wrapper => wrapper.find(Input).length > 0,
+    })
+  }
+
+  it("renders", async () => {
+    const wrapper = await getWrapper()
```
damassi: If `getWrapper` doesn't accept params might be better to move to 

```js
beforeAll(async () => { 
  wrapper = await getWrapper()
})
``` 
so that it can be used without needing `async / await` on every function.

sweir27:
Looks good @ds300 !

I'm still wondering about `isMakeOfferFlow` vs. `isOfferFlow`-- we can always follow up but interested  in what you think.

ds300:
@sweir27 ah yes, I forgot that one!

src/Apps/Order/Routes/__tests__/Offer.test.tsx
ds300: good idea!

ds300:
This should be 🍏now. I'm away until Monday so if there's any more feedback, feel free to update this branch on my behalf 🙏

PR #1453, opened 2018-10-24
[PATCH collection] throw Not Found error when collection do not exist

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-962

screen shot 2018-10-24 at 2 43 38 pm

### DIFF
--- a/src/Apps/Collect/CollectionApp.tsx
+++ b/src/Apps/Collect/CollectionApp.tsx
@@ -1,5 +1,6 @@
 import { Box } from "@artsy/palette"
 import { CollectionApp_collection } from "__generated__/CollectionApp_collection.graphql"
+import { HttpError } from "found"
 import React, { Component } from "react"
 import { Meta, Title } from "react-head"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 81bc4a36820024354b448b81b1dfb9f8174b6ead

https://deploy-preview-1453--artsy-reaction.netlify.com

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -13,8 +14,15 @@ interface CollectionAppProps {
...
 export class CollectionApp extends Component<CollectionAppProps> {
+  collectionNotFound = collection => {
+    if (!collection) {
+      throw new HttpError(404)
+    }
+  }
+
   render() {
     const { collection } = this.props
+    this.collectionNotFound(collection)
```
l2succes: I would move that to `componentDidMount` or `componentWillReceiveProps`, basically move it out of render

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -13,8 +14,15 @@ interface CollectionAppProps {
 }
 
 export class CollectionApp extends Component<CollectionAppProps> {
+  collectionNotFound = collection => {
+    if (!collection) {
```
l2succes: Also we may want a better check then `collection` because `collection` can be `null` before the page actually loads. We should use the relay request status code

PR #1454, opened 2018-10-24
adding specs for sidebar metadata

oxaudo:
This brings the metadata to almost production state. We are still missing the explanation modal for Classification sting, but otherwise this should be in a ready to put on artwork page state(just for this component:).

I'll work on classification modal next.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
@@ -1,6 +1,7 @@
 import { Box, Serif } from "@artsy/palette"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { get } from "Utils/get"
 import { ArtworkSidebarClassificationFragmentContainer as Classification } from "./ArtworkSidebarClassification"
 import { ArtworkSidebarSizeInfoFragmentContainer as SizeInfo } from "./ArtworkSidebarSizeInfo"
 import { ArtworkSidebarTitleInfoFragmentContainer as TitleInfo } from "./ArtworkSidebarTitleInfo"
@@ -16,15 +17,17 @@ export class ArtworkSidebarMetadata extends React.Component<
 > {
   render() {
     const { artwork } = this.props
+    const lotLabel = get(
+      artwork,
+      a => a.is_biddable && a.sale_artwork.lot_label
+    )
     return (
       <Box>
-        {artwork.is_biddable &&
-          artwork.sale_artwork &&
-          artwork.sale_artwork.lot_label && (
-            <Serif size="2" weight="semibold" color="black100">
-              Lot {artwork.sale_artwork.lot_label}
-            </Serif>
-          )}
+        {lotLabel && (
+          <Serif size="2" weight="semibold" color="black100">
+            Lot {lotLabel}
+          </Serif>
+        )}
         <TitleInfo artwork={artwork} />
         {artwork.edition_sets.length < 2 && <SizeInfo piece={artwork} />}
         <Classification artwork={artwork} />
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
@@ -1,4 +1,4 @@
-import { Box } from "@artsy/palette"
+import { Box, Serif } from "@artsy/palette"
 import React from "react"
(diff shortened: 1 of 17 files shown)


artsyit:

Built with commit 367258152895d23c7022956b00a12cc47c86ebf0

https://deploy-preview-1454--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
```diff
@@ -18,6 +18,13 @@ export class ArtworkSidebarMetadata extends React.Component<
     const { artwork } = this.props
     return (
       <Box>
+        {artwork.is_biddable &&
+          artwork.sale_artwork &&
+          artwork.sale_artwork.lot_label && (
```
damassi: Maybe: 
```js
const lotLabel = get(artwork, p => p.is_biddable && p.artwork.sale_artwork.lot_label)

{lotLabel && 
 ...
}
```
? 

Basically the same, but i think using `get` is a little less error prone.

src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarArtists.test.tsx
```diff
@@ -62,5 +62,10 @@ describe("ArtworkSidebarArtists", () => {
       const wrapper = await getWrapper(MultipleArtists)
       expect(wrapper.html()).not.toContain("Follow")
     })
+    it("separates artist names by comma", async () => {
+      const wrapper = await getWrapper(MultipleArtists)
+      // 2 artists are separared by 1 ,
+      expect(wrapper.html().match(", ").length).toBe(1)
```
damassi: Might be a little better to assert against the actual artist names + the comma? Then can loose the comment and have the test itself speak for the behavior.

src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarMetadata.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        }
+      `,
+      mockResolvers: {
+        Artwork: () => response,
+      },
+    })
+  }
+  describe("for non editioned artwork", () => {
+    it("displays title and year", async () => {
+      const wrapper = await getWrapper()
```
alloy: Seeing as all the wrappers are the same inside these describe sections, you could define the wrapper once from a beforeAll function and then all the tests don’t even need to be async.

Probably reduces some noise and a slightly faster pattern, performance wise (negligible on its own, but noticeable when a test suite gets large).

src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarMetadata.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+      expect(wrapper.html()).toContain("<i>Easel (Vydock)</i>, 1995")
+    })
+    it("displays medium", async () => {
```
alloy: Can you add white-space between the tests? 🙏

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
oxaudo: I guess get is a bit confusing here because of how we need to check for `is_biddable` and not only to unpack `p.artwork.sale_artwork.lot_label`. But happy to change if you think it's less error prone.

damassi: It's debatable! no strong prefs at all.

src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarMetadata.test.tsx
damassi: Ohh great idea. Also eliminates the annoying missing `async / await` thing that constantly arises.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarMetadata.tsx
oxaudo: I guess setting it to const eliminates long repeat inside the clause - so you win:)

oxaudo:
@alloy, @damassi - I think I've addressed all the comments here. Merging.
Thanks for suggestions! They all make sense:)

alloy:
Nice 👌

PR #1438, opened 2018-10-22
Storybook fix

ashfurrow:
This PR fixes the `ShippingAndPaymentSummary` story and switches to new `<></>` fragment syntax.

The problem with the storybook was that the query we were feeding in was requesting (and getting back from the mock network layer) an `Order` labelled `ecommerceOrder`, not `order` [as we do in the actual query](https://github.com/artsy/reaction/blob/dc6c47f1927c2f961aa73b09dbff5c24c2ffc64d/src/Apps/Order/routes.tsx#L42).

PR #1460, opened 2018-10-25
MockQueryRenderer improvements

ashfurrow:
This PR is a followup from #1438. When we discussed my change in the Purchase team knowledge share the next day, we agreed that better surfacing of errors in storybooks would be helpful for debugging (as there are apparently other broken storybooks in Reaction).

So my solution was to turn `MockRelayRenderer` into an [error boundary](https://reactjs.org/docs/error-boundaries.html), put any caught errors into the component's `state`, and use that state to indicate something went wrong. It looks like `renderWithLoadProgress` _was_ catching errors (as marked in the comments) but was returning `null` instead, leading to blank screens:

screen shot 2018-10-25 at 4 08 52 pm

A blank screen is probably what we want, for users, so I didn't want to change the behaviour of `renderWithLoadProgress`.

If developers opened the console and scrolled through all the logs of a broken storybook, they might find this error:

```
Warning: createFragmentSpecResolver: Expected prop `order` to be supplied to `Relay(ShippingAndPaymentSummary)`, but got `undefined`. Pass an explicit `null` if this is intentional.
```

But that's not a great developer experience. This PR surfaces the error from the component in the browser:

screen shot 2018-10-25 at 3 54 30 pm

I'm open to feedback on wording, etc.

I also removed an errant `console.log` I accidentally left in #1438.

### DIFF
--- a/src/DevTools/MockRelayRenderer.tsx
+++ b/src/DevTools/MockRelayRenderer.tsx
@@ -132,7 +132,7 @@ export class MockRelayRenderer<
 
     if (this.state.caughtError) {
       const { error, errorInfo } = this.state.caughtError
-      console.log({ error, errorInfo })
+      console.error({ error, errorInfo })
       return `Error occurred while rendering Relay component: ${error}`
     }
 
--- a/src/DevTools/MockRelayRenderer.tsx
+++ b/src/DevTools/MockRelayRenderer.tsx
@@ -23,6 +23,13 @@ export interface MockRelayRendererProps<
   mockResolvers: IMocks
 }
 
+export interface MockRelayRendererState {
+  caughtError: {
+    error: any
+    errorInfo: any
+  }
+}
+
 /**
  * Renders a tree of Relay containers with a mocked local instance of the
  * metaphysics schema.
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit 462f26785b758d88bea50281f057eb5ff8cca09e

https://deploy-preview-1460--artsy-reaction.netlify.com

src/DevTools/__test__/MockRelayRenderer.test.tsx
```diff
@@ -34,4 +39,36 @@ describe("MockRelayRenderer", () => {
...
+      />
+    )
+    tree.setState({
+      caughtError: {
+        error: new Error("Hey it's an error!"),
+        errorInfo: {},
+      },
+    })
+    setTimeout(() => {
+      console.log(tree.text())
```
alloy: Stray loggy?

src/DevTools/__test__/MockRelayRenderer.test.tsx
```diff
@@ -34,4 +39,36 @@ describe("MockRelayRenderer", () => {
...
+      expect(tree.update().text()).toEqual(
+        "Error occurred while rendering Relay component: Error: Hey it's an error!"
+      )
+      done()
+    }, 10)
```
alloy: Instead of this timeout you could also use the renderUntil utility.

src/DevTools/MockRelayRenderer.tsx
```diff
@@ -115,6 +130,12 @@ export class MockRelayRenderer<
       )
     }
 
+    if (this.state.caughtError) {
+      const { error, errorInfo } = this.state.caughtError
+      console.log({ error, errorInfo })
```
alloy: This should be `console.error`

src/DevTools/__test__/MockRelayRenderer.test.tsx
ashfurrow: Ah, yes! Good point. 

I played around with a bit and was initially flummoxed by some React/Enzyme lifetime stuff around the initial state of the component I set, but then I realized that I don't need the timeout at all 😅 I must've copied it from the other test, sorry about that! I've updated the other test to use `renderUntil`.

PR #1461, opened 2018-10-25
add modal with details to ArtworkSidebarClassification

oxaudo:
This code also puts back the scrolling into modals because disabling the scroll actually introduced more severe bugs in other places than what was reported here: https://artsyproduct.atlassian.net/browse/GROW-798

This change sets the overflow to auto which should work just fine unless default OS settings are overwritten. @l2succes was mentioning that there is the idea of writing the custom scroll implementation to fix this issue but that seems like way to much trouble that could introduce other bugs than the issue that we are trying to fix here....


oxaudo:
The look:
screen shot 2018-10-25 at 5 12 54 pm

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarClassification.tsx
```diff
@@ -1,6 +1,8 @@
-import { Box, Serif } from "@artsy/palette"
+import { Box, Sans, Serif } from "@artsy/palette"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import styled from "styled-components"
+import Modal from "../../../../Components/Modal/Modal"
```
damassi: `from `Components/Modal/Modal'`

oxaudo: Oh.. Sorry.... Fixing

artsyit:

Built with commit 621340857706e00a81cc0a942453dcdcd073a455

https://deploy-preview-1461--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarClassification.tsx
```diff
@@ -36,3 +70,80 @@ export const ArtworkSidebarClassificationFragmentContainer = createFragmentConta
...
   `
 )
+
+const ClassificationLink = styled.span`
+  cursor: pointer;
+  text-decoration: underline;
+`
+
+const ClassificationDetails = () => {
+  const classificationOptions = [
```
oxaudo: Also this is not very kosher. I think we might want to get those from MP since there we already have a class that lists all this strings. But I was not sure if the best thing is to add it to artwork or do a separate call for AttributionClasses so just put it inline for now.

oxaudo:
@damassi - I've updated addressing your comments here. Still need to write the test for `ArtworkSidebarClassification` here, but I think this might be mergeable regardless since it fixes the modals view issue.

damassi:
@oxaudo - one last q -- in the screenshot it looks like the text is overflowing the container. Assuming this is an incorrect screenshot and that area gets clipped at the bottom?

oxaudo:
Yes, @damassi. Design seems to suggest this treatment: https://app.zeplin.io/project/5b113f413ede4bb10dbfdc42/screen/5b16c46a68e34a929863d4c3

Though we might want to put ok button fixed so that the clipping is more obvious maybe? Also I used the cta for the dialog for the ok button. Which makes the button black. If having it white is important - I think we could add it as a property for Modal. Need to check with Jeffrey.

damassi:
The modal content can be whatever, so the CTA should come from the person creating the modal content, e.g., 

https://github.com/artsy/reaction/blob/master/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionDetailsModal.tsx#L40-L54

Inside of the `LotDetails` component just below the button is defined: 

https://github.com/artsy/reaction/blob/master/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionDetailsModal.tsx#L107-L111

The `<Modal>` component itself is only supposed to control general behavior such as appearing on the screen / fading background out, closing, etc.

oxaudo:
@damassi - yes, I actually started following this example that you are pointing to above. But then realized that designs for ClassificationsDetail modal actually predate the Reaction Modal. Conceptually in this context we are the person creating this dialog and Ok is the call for action here. The action we want people to take on this dialog - read content and just click Ok.

Either way - It's probably more attention to that problem than it deserves :) I'll just change the CTA into button inside the modal content if Jeffrey confirms that this black style for the button is the correct one.

damassi:
Sounds good 👍 Might also be good also to ping Jeffrey about unifing the designs across contexts so that we don't have too many different modal / dialog implementions floating around.

oxaudo:
@damassi - confirmed with @Jeffrey that we want to standard new button for the dialog CTA. So updated that.

Ok is a special case that should be displayed as `OK` instead of camel case. So passing 'OK' as a string for the button with OK here.

Also added couple of basic specs for the dialog. Think this is ready to be merged unless you have other suggestions here.

src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarClassification.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  describe("for artwork without classification", () => {
+    beforeAll(async () => {
+      wrapper = await getWrapper({ attribution_class: null })
+    })
+
+    it("does not render anything", () => {
+      expect(wrapper.html()).toBeNull()
+    })
+  })
+})
```
damassi: Thanks for adding specs!

PR #1462, opened 2018-10-26
[Tests] Fix RelayStubProvider errors in Jest

damassi:
Some background context is in this [Slack thread](https://artsy.slack.com/archives/CAZG82X4G/p1540517734000100) but in short I found myself continuously fighting with the mock network layer and realized that if we could make it so that `<RelayStubProvider>` (which we use in Storybooks and elsewhere) doesn't error out in the test environment then that's likely all that we'll need for most general tests since it provides for a way of passing in data without needing mocks. It renders exactly the same as if we were going through the full  GraphQL resolution flow, but without the setup complexity.

@alloy, @mzikherman, @sweir27 - thoughts? 

I've posted some comments inline.


artsyit:

Built with commit b5f473b1c6e86b14d610fbabd58b3608df165914

https://deploy-preview-1462--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Shows/__test__/Shows.test.tsx
```diff
@@ -1,55 +1,43 @@
...
-      ),
-    })
+  function getWrapper(breakpoint: Breakpoint = "xl") {
+    return mount(
+      <RelayStubProvider>
+        <MockBoot breakpoint={breakpoint}>
+          <ShowsRoute viewer={ShowsFixture as any} />
+        </MockBoot>
+      </RelayStubProvider>
+    )
```
damassi: The diff here between the old `getWrapper` and the new is a good example of before / after

src/Apps/Artist/Routes/Shows/__test__/Shows.test.tsx
```diff
@@ -1,55 +1,75 @@
...
-        query Shows_Test_Query($artistID: String!) {
-          viewer {
-            ...Shows_viewer
-          }
-        }
-      `,
-      mockResolvers: {
-        Viewer: () => ({ viewer: { artist_currentShows: CVFixture } }),
-        Artist: () => ({ artist: CVFixture }),
-        ShowConnection: () => showsConnection,
```
damassi: As `viewer` is the top-level query here, if `Viewer` was mockable in the sense that the JSON data returned from it would result in the rest of the resolvers returning correct data this would be a bit more usable but in this case each little piece needs to be properly mocked out.

alloy:
I agree that this is proving tricky. However, I think we should fix the trickiness of assembling the payload instead and not bypass testing the containers.

Eg that Relay error about missing data should bubble up. I have an idea about how to get a type for the full expected query which should help in pointing out missing data in a fixture. When those things are in I don’t think this should be any different from bypassing containers.

src/Apps/Artist/Routes/AuctionResults/__tests__/AuctionResults.test.tsx
```diff
@@ -1,47 +1,33 @@
...
-jest.unmock("react-relay")
-
 describe("AuctionResults", () => {
   let wrapper: ReactWrapper
 
-  const getWrapper = async (breakpoint: Breakpoint = "xl") => {
-    return await renderRelayTree({
-      Component: AuctionResultsRoute,
-      query: graphql`
-        query AuctionResults_Test_Query($artistID: String!) {
```
damassi: Another thing about using `RelayStubProvider` is we don't pollute our `__generated__` folder with a bunch of test-related queries, and also don't need to run `yarn relay` while writing tests.

alloy:
Not really, we’re just passing data _through_ them, the containers are not being exercised. These tests would bypass testing if the fragments are right, and, probably more importantly, they would bypass testing refetch containers and pagination containers being configured correctly. For example, tests that check if any of the filtering works as expected can’t really be done by bypassing the container, unless you’re going to mock out the entire API, which I’m not a big fan of.

alloy:
(If there currently are no tests yet that truly exercise the filtering, then that example should probably be added to have something that can better be evaluated.)

damassi:
@alloy - by filtering do you mean asserting against, say, a `this.props.relay.refetch()` response?

alloy:
Yeah exactly, or paginate. These things require testing a properly configured container, a fragment that takes variables, etc.

alloy:
If we were to ever only do static fragments, then I agree with you that this approach would probably sufficient, as we have TS to point out problems and we would probably easily see if they work at all at runtime. But seeing as we do have more complex situations, I’d prefer to work towards a situation that is equal (or close enough) to this.

damassi:
That makes sense. Let me look at creating an example of that. It seems ok to stub out methods on the relay object, since that sort of testing has already taken place in metaphysics, and many errors are caught by the relay compiler. By the time we get to the UI testing phase we have a lot of other assurances about data behavior, but not many assurances about rendering or interaction. The latter two things can be asserted against without a fully mocked environment.

alloy:
Absolutely! I definitely agree with you that we should get it into that state 👍

alloy:
@damassi Could you do me a solid and leave a failing example in of the issues with connections you had?

alloy:
I also now realise that our devtools are probably insufficient for dealing with testing those complex scenarios, as `renderUntil` only works for the initial setup so we should refactor `renderUntil` into an actual part of the enzyme API so it can invoked on a wrapper at any time (and also be PR’ed upstream to enzyme) 🤔

alloy:
@damassi It looks like I should be able to generate types for the exact query that gets made, as it gets stored in each operation artifact, e.g. https://github.com/artsy/reaction/blob/173650de99de2cf4961b275ce11f8684b41da701/src/__generated__/Articles_Test_Query.graphql.ts#L21-L86

damassi:
> Could you do me a solid and leave a failing example in of the issues with connections you had?

Yup! I'll add a repo PR. (And looking over other stuff now.)

damassi:
Ok, i've swapped out the re-swapped things with the `renderRelayTree` (but left the fix in for `<RelayStubProvider>` since that's useful for simple components. Gonna merge these tests.

PR #1464, opened 2018-10-26
[Test] Remove unused var

damassi:
Noticed master was failing `tsc`


williardx:
Thanks @damassi for fixing this!

PR #1465, opened 2018-10-26
[Do Not Merge] Example test resolver issue

damassi:
When playing with some tests using the new `renderRelayTree` I noticed that its not exactly clear how to populate connection data correctly or to track down where problems occur:

To repo: 
- `yarn install` to apply patch package (which just drops a `console.log` in the right spot in `relay-mock-network-layer`)
- `yarn jest src/Apps/Artist/Routes/CV/__tests__/CV.test.tsx --watch --verbose false` (the verbose bit is to preserve console.logs) 

When first running the test you'll see a log of response data coming from `relay-mock-network-layer` that _seems_ correct. After that log is a log from `CVItem` logging out its props. Looking inside, the first and second items have an undefined `showsConnection`. This was eventually semi-fixed due to adjustments made to the mocking code (enough to assume the structure is valid to get the test fully passing), and then finally there's the fully-correct implementation (that's commented out). 

Another thing to note is that in `src/__generated__/CV_Test_Query.graphql.ts` it doesn't list out all of the required types needed to mock the resolvers as `Show` and `ShowConnection` is missing. A developer would have to infer it from the query key name though I imagine there might be some way for tooling to grab the correct info. 

I guess this again goes back to how to help surface issues for devs trying to mock out complex pages that cross over a few areas in MP, as it took a while to find the thread that led to the issue due to two silent errors.

Here's a screencap of CV.test: 

test

### DIFF
--- a/src/Apps/Artist/Routes/CV/CVItem.tsx
+++ b/src/Apps/Artist/Routes/CV/CVItem.tsx
@@ -77,6 +77,7 @@ class CVItem extends Component<CVItemProps, CVItemState> {
     ))
 
   render() {
+    console.log(this.props)
     if (
       !this.props.artist.showsConnection ||
       !this.props.artist.showsConnection.edges.length
--- a/src/Apps/Artist/Routes/CV/__tests__/CV.test.tsx
+++ b/src/Apps/Artist/Routes/CV/__tests__/CV.test.tsx
@@ -5,10 +5,7 @@ import React from "react"
 import { graphql } from "react-relay"
 import { Breakpoint } from "Utils/Responsive"
 
-import {
-  CVFixture,
-  showsConnection,
-} from "Apps/__test__/Fixtures/Artist/Routes/CV"
+import { CVFixture } from "Apps/__test__/Fixtures/Artist/Routes/CV"
 
 jest.unmock("react-relay")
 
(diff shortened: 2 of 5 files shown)


patches/relay-mock-network-layer+1.2.0.patch
```diff
@@ -0,0 +1,124 @@
...
+-            // Trigger Relay error in case of GraphQL errors (or errors in mutation response)
+-            // See https://github.com/facebook/relay/issues/1816
+-
+-            result => {
+-                if (result.errors && result.errors.length > 0) {
+-                    return Promise.reject(new RelayMockNetworkLayerError(result.errors));
+-                }
+-                return Promise.resolve(result);
+-            }
+-        );
```
alloy: This patch is changing this fix related to better reporting errors, so that’s probably not intentional?

damassi: Just prettier reformatting, its all here: https://github.com/artsy/reaction/pull/1465/files/df0cd142d766c3abd6b0fb717008b6be0e9df6af#diff-d03959419fb73774b0d069d3d973b986R94

damassi: ohh i see what you mean, at the top

damassi: Yeah not sure where that top portion was even coming from? I popped off that file but if you add a log here from node modules it should work the same: https://github.com/1stdibs/relay-mock-network-layer/blob/master/index.js#L29

damassi: Ahh, interesting: I know what this issue is: When `patch-package` ran it pulled in [your latest PR from upstream - 1.3.0](https://github.com/1stdibs/relay-mock-network-layer/pull/6) and performed a diff from there, vs diffing against the version currently installed in my local `node_modules` folder (1.2.1).

alloy:
Thanks!

damassi:
Closing this since @ds300 and @alloy are on it

PR #1467, opened 2018-10-29
give image default aspect ratio in case we don't get one from gravity

oxaudo:
@mzikherman - based on discussion here: https://artsy.slack.com/archives/CAZG82X4G/p1540825452041200

I think this might fix the immediate issue I was seeing when image is not displayed at all and there is no width for the item:
screen shot 2018-10-29 at 11 59 33 am

This fix might make those images scale strangely but I think it's still better than current situation.

### DIFF
--- a/src/Components/Artwork/FillwidthItem.tsx
+++ b/src/Components/Artwork/FillwidthItem.tsx
@@ -13,6 +13,7 @@ import { Mediator } from "Artsy/SystemContext"
 import styled, { StyledComponentClass } from "styled-components"
 
 const IMAGE_QUALITY = 80
+const FALLBACK_RATIO = 1
 
 const Image = styled.img`
   width: 100%;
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 7cdbcdaffd9cd3e62f9efc6f1adec25f09272570

https://deploy-preview-1467--artsy-reaction.netlify.com

alloy:
The same is done in Emission https://github.com/artsy/emission/blob/904e47e620acb1313dfe9ce383df2998f359871a/src/lib/Components/OpaqueImageView.tsx#L64-L68 👍

oxaudo:
Oh cool @ally :) That validates my thinking here!

orta:
Good thinking Matt - just cheeked, an image has `aspect_ratio` in metaphysics already, so it could be used from there instead

oxaudo:
Currently `aspect_ratio` could come in as null from MP. Are we all cool with updating it to a default of 1 in MP? @mzikherman, @alloy, @orta ?

orta:
Yep - [Eigen does this](https://github.com/artsy/eigen/blob/e1c1be0f1e5c9b80fa5082cd0f224d2bfd1fe0f4/Artsy/Models/API_Models/Partner_Metadata/Artwork.m#L207-L210), [Emission also](https://github.com/artsy/emission/blob/904e47e620acb1313dfe9ce383df2998f359871a/src/lib/Components/OpaqueImageView.tsx#L67)

oxaudo:
Closing this one then and will update MP. Thanks all for feedback!

alloy:
Good call!

PR #1472, opened 2018-10-31
refactoring some tests

oxaudo:
This removes some async tests where they are not necessary and moving the get a wrapper call to beforeAll function.

Also adding couple of minot checks. I think both ArtworkDetails and ArtworkArtist details are tested to bare minimum. Going forward we should try to add more comprehensive tests once we discover bugs and that I should be sufficient while moving forward.

Let me know what you think here.

### DIFF
--- a/src/Apps/Artwork/Components/__tests__/ArtistInfo.test.tsx
+++ b/src/Apps/Artwork/Components/__tests__/ArtistInfo.test.tsx
@@ -24,15 +24,17 @@ describe("ArtistInfo", () => {
   }
   let wrapper
 
-  beforeAll(async () => {
-    wrapper = await getWrapper()
-  })
+  describe("ArtistInfo for artwort with complete artist info", () => {
+    beforeAll(async () => {
+      wrapper = await getWrapper()
+    })
 
-  it("renders a correct component tree", () => {
-    expect(wrapper.find("EntityHeader").length).toBe(1)
-    expect(wrapper.find("ArtistBio").length).toBe(1)
-    expect(wrapper.find("MarketInsights").length).toBe(1)
-    expect(wrapper.find("SelectedExhibitions").length).toBe(1)
+    it("renders a correct component tree", () => {
+      expect(wrapper.find("EntityHeader").length).toBe(1)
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit 5efd7d23b831678713f6b9826e09f8f5585a1bbb

https://deploy-preview-1472--artsy-reaction.netlify.com

src/Apps/Artwork/Components/__tests__/ArtistInfo.test.tsx
```diff
@@ -22,9 +22,13 @@ describe("ArtistInfo", () => {
       },
     })
   }
+  let wrapper
 
-  it("renders a correct component tree", async () => {
-    const wrapper = await getWrapper()
+  beforeAll(async () => {
+    wrapper = await getWrapper()
+  })
```
damassi: Since the first test case below is an outlier, I don't think we should wrap this in a beforeAll -- if _most_ tests didn't need to be booted with data then it would make sense.

oxaudo:
@damassi - My thinking was that we might need to add more specs for default data state, so I wanted to do a generic setup that would allow that. But you are right - since it's not happening now kinda silly to do this extra work. 

Either case - I moved things around once more - to make the idea more clear (hopefully).

src/Apps/Artwork/Components/__tests__/ArtistInfo.test.tsx
```diff
@@ -22,19 +22,25 @@ describe("ArtistInfo", () => {
...
-  it("renders a correct component tree", async () => {
-    const wrapper = await getWrapper()
-    expect(wrapper.find("EntityHeader").length).toBe(1)
-    expect(wrapper.find("ArtistBio").length).toBe(1)
-    expect(wrapper.find("MarketInsights").length).toBe(1)
-    expect(wrapper.find("SelectedExhibitions").length).toBe(1)
+  describe("ArtistInfo for artwort with complete artist info", () => {
+    beforeAll(async () => {
+      wrapper = await getWrapper()
+    })
```
damassi: This is better 👍

damassi:
Changes look good! 👍

PR #1473, opened 2018-10-31
[PATCH display] fix title slide overflow

eessex:
Looks like the latest versions of `react-slick` insert inline style `display: block;` on all slides.  This is fine for images, but causes a bug on the title slide which must display as flex instead.

And ugly fix with a bang, but not sure how else to get around the inline styles. 

screen shot 2018-10-31 at 13 12 51


package.json
```diff
@@ -183,7 +183,7 @@
     "react-responsive-decorator": "^0.0.1",
     "react-router": "^4.2.0",
     "react-sizeme": "^2.4.4",
-    "react-slick": "^0.23.1",
+    "react-slick": "^0.23.2",
```
damassi: React-slick is pretty fragile -- can you be sure to check the new artist page and Recently Viewed slider in storybooks before merging?

eessex: Yep already checked out all the existing slideshows, looking good.

l2succes:
Merge on green

PR #1474, opened 2018-11-01
Fix a bug where the sort option is always set to "-decayed_merch" on both Collect and Collection

yuki24:
While I was working on https://artsyproduct.atlassian.net/browse/GROW-976 I noticed that the `sort` options is ignored entirely on Collect and Collection. This PR updates the `prepareVariables` function to respect it properly so the page the page results will be consistent after a page refresh.

### DIFF
--- a/src/Apps/Collect/__tests__/routes.test.tsx
+++ b/src/Apps/Collect/__tests__/routes.test.tsx
@@ -48,6 +48,20 @@ describe("Routes", () => {
       })
     })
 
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit c5e2ba03706ee012cdbe2a6dbbc31b203b29d06e

https://deploy-preview-1474--artsy-reaction.netlify.com

l2succes:
Nice catch @yuki24!

PR #1475, opened 2018-11-01
User should see see the same results after a page refresh on the collection page

yuki24:
finishes https://artsyproduct.atlassian.net/browse/GROW-976. Also this PR depends on https://github.com/artsy/reaction/pull/1474.

It seems a bit tedious but we need to pass in all the arguments down to the `CollectionApp_collection` fragment otherwise page results will be inconsistent after a page refresh. I wasn't sure if this is the right solution, so definitely open to suggestions and opinions for how this should be handled.


artsyit:

Built with commit 6b7a4c4ad97cd348e284c638704ef27cb1b2a459

https://deploy-preview-1475--artsy-reaction.netlify.com

yuki24:
Now that #1474 is merged this is no longer a WP and ready for reivew.

PR #1476, opened 2018-11-01
[WIP] First pass in adding initialOffer mutation

ashkan18:
depends on https://github.com/artsy/metaphysics/pull/1365


artsyit:

Built with commit 970913ed4aa571b9ac2cbee64572fd10d18c773a

https://deploy-preview-1476--artsy-reaction.netlify.com

ashkan18:
whatever the issue is, i can see App/Order stories are broken in storybook because of it too

src/Apps/Order/Components/TransactionSummary.tsx
```diff
@@ -94,13 +94,7 @@ export class TransactionSummary extends React.Component<
...
           {isOfferFlow ? (
             <>
-              <Entry
-                label="Your offer"
-                value={
-                  offerOverride ||
-                  /* TODO: || order.offer.value or whatever */ "—"
-                }
-              />
+              <Entry label="Your offer" value={offerTotal} />
```
ds300: this should be value={offerOverride || offerTotal} so that the value updates when you type on the initial offer screen. Although maybe we should rename offerOverride to offerTotalOverride now.

src/Apps/Order/Components/TransactionSummary.tsx
```diff
@@ -92,9 +94,13 @@ export class TransactionSummary extends React.Component<
...
           {isOfferFlow ? (
             <>
-              <Entry label="Your offer" value={offerTotal} />
+              <Entry label="Your offer" value={offerOverride} />
               {Boolean(itemsTotal) && (
-                <Entry label="List price" secondary value={itemsTotal} />
+                <Entry
+                  label="List price"
+                  secondary
+                  value={itemsTotal || offerTotal}
```
ds300: i think this should just be `value={itemsTotal}`

src/Apps/Order/Components/TransactionSummary.tsx
```diff
@@ -92,9 +94,13 @@ export class TransactionSummary extends React.Component<
         <StackableBorderBox flexDirection="column">
           {isOfferFlow ? (
             <>
-              <Entry label="Your offer" value={offerTotal} />
+              <Entry label="Your offer" value={offerOverride} />
```
ds300: I think this should be `<Entry label="Your offer" value={offerOverride || offerTotal} />`

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -192,7 +180,18 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
...
-                    <TransactionSummary order={order} mb={[2, 3]} />
+                    <TransactionSummary
+                      order={order}
+                      mb={[2, 3]}
+                      offerOverride={
+                        this.state.offerValue &&
+                        this.state.offerValue.toLocaleString("en-US", {
+                          style: "currency",
+                          currency: "USD",
+                          minimumFractionDigits: 0,
```
ds300: There's another ticket in Jira for adding '.00' after the price here, so you could change `minimumFractionDigits` to 2 here and take care of that already.

src/Apps/Order/Components/TransactionSummary.tsx
ds300: I think this should be `value={itemsTotal || offerTotal}`? The list price should only reflect the line items list prices, not the current offer, right?

src/Apps/Order/Components/TransactionSummary.tsx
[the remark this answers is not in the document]
ashkan18: oh put this `|| offerTotal` on the wrong place! good catch! added it to List Price!

src/Apps/__test__/Fixtures/Order.ts
```diff
@@ -86,8 +89,15 @@ export const UntouchedOrder = {
   },
 }
 
+export const UntouchedOfferOrder = {
+  ...UntouchedBuyOrder,
+  mode: "OFFER",
+  state: "PENDING",
+  taxTotal: null,
```
ds300: #superminor `taxTotal` and `state` can be omitted here? Seems they have the same values in `UntouchedBuyOrder`.

src/Apps/__stories__/OrderApp.story.tsx
```diff
@@ -24,7 +25,15 @@ const Router = props => (
...
 
 storiesOf("Apps/Order Page/Offer", module).add("Empty", () => (
-  <Router initialRoute="/orders/123/offer" />
+  <Router
+    initialRoute="/orders/123/offer"
+    mockResolvers={mockResolver({
+      ...UntouchedOfferOrder,
+      mode: "OFFER",
+      requestedFulfillment: null,
+      state: "PENDING",
```
ds300: #superminor I think you don't need to overwrite `mode` and `state` here, since they have the same values in `UntouchedOfferOrder`.

src/Apps/__stories__/OrderApp.story.tsx
```diff
@@ -33,16 +42,27 @@ storiesOf("Apps/Order Page", module)
...
       mockResolvers={mockResolver({
-        ...UntouchedOrder,
+        ...UntouchedBuyOrder,
         requestedFulfillment: null,
       })}
       initialRoute="/orders/123/shipping"
     />
   ))
   .add("Review", () => <Router initialRoute="/orders/123/review" />)
+  .add("Empty", () => (
```
ds300: This seems to be a duplicate of the story above on line 27: "Apps/Order Page/Offer"?

ashkan18:
thanks for help and review @ds300 going to merge

PR #1477, opened 2018-11-02
Get rid of schema.json

ds300:
cc @ashkan18 

We had this weird situation where we had two sources of truth for the graphql schema: `data/schema.json` and `data/schema.graphql`. Only the `schema.graphql` file was being updated by our scripts, so we had to update the `schema.json` file manually.

This PR makes it so we don't use the `schema.json` file at all.


artsyit:

Built with commit be6ad252ab7485e9b4f3ebb4072cecd2f31ac78f

https://deploy-preview-1477--artsy-reaction.netlify.com

alloy:
Ah I see, makes sense 👍

PR #1480, opened 2018-11-02
Use a more structured type for ModalOptions.afterSignUpAction

yuki24:
It seems like the `afterSignUpAction` field is never a string, but an object with a required `action` and optional `objectId` and `kind`. This PR makes it more explicit and be type-safe.

### DIFF
--- a/src/Components/Authentication/Types.ts
+++ b/src/Components/Authentication/Types.ts
@@ -34,6 +34,12 @@ export interface FormProps {
   onBackButtonClicked?: (e: Event) => void
 }
 
+interface AfterSignUpAction {
+  action: "save" | "follow" | "editorialSignup"
+  objectId?: string
+  kind?: "artist" | "artworks" | "gene" | "profile" | "show"
+}
+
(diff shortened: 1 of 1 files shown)


src/Components/Authentication/Types.ts
```diff
@@ -34,6 +34,12 @@ export interface FormProps {
   onBackButtonClicked?: (e: Event) => void
 }
 
+interface AfterSignUpAction {
+  action: "save" | "follow" | "editorialSignup"
```
yuki24: The `"editorialSignup"` type isn't used now but that's the keyword I'm going to use for https://artsyproduct.atlassian.net/browse/GROW-858.

artsyit:

Built with commit 4a9726ddef343736cd8fd3e55dc973a1a39a129a

https://deploy-preview-1480--artsy-reaction.netlify.com

src/Components/Authentication/Types.ts
l2succes: Makes sense 👍

PR #1481, opened 2018-11-02
Use orbs, queue releases

zephraph:
This PR does 2 things

1. Introduces CircleCI orbs to the mix to reduce duplicate configurations that are scattered across projects. Check out [the repo](https://github.com/artsy/orbs)!
2. Forces master builds to queue so that we don't have timing issues w/ PR merges

Essentially this means that only 1 master build will be allowed to proceed at a time. Secondary builds will hold a container for a maximum of 2 minutes before timing out and failing the build.


artsyit:

Built with commit 8bcaf1481c456dae9d911fb409e701ef97565cc6

https://deploy-preview-1481--artsy-reaction.netlify.com

.circleci/config.yml
```diff
@@ -1,117 +1,92 @@
...
       - checkout
-      - <<: *restore_cache
+      - yarn/load_dependencies
       # Setup the .npmrc with the proper registry and auth token to publish
       - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
       - run: yarn semantic-release
+  pre-deploy:
+    executor: node/build
+    steps:
+      - queue/until_front_of_line:
```
joeyAghion: Interesting. We could use something similar to prevent builds and pushes of new `staging` docker images while a `production` release (which does a promote of the `staging` image) is pending.

zephraph:
ughhhhhhh. Minor config breaking changes without having a validator for it make me sad. Workflows requires a version in < 2.0 but > 2.1 will _break_ if the workflow has a version. 

https://discuss.circleci.com/t/circleci-2-1-config-overview/26057

zephraph:
This PR is blocked by https://github.com/eddiewebb/circleci-queue/pull/9. We need to be able to enable queuing only on a certain branch and there isn't really a mechanism to do so within circle-ci currently. Job conditionals won't work because they can't use environment variables and filters won't work because there's not a way to require a workflow job to depend on another job that runs only on a particular branch.

zephraph:
Unblocked!! 🕺

zephraph:
It's green! @orta this is good to go/review. Please, if you merge this, do so with *squash and merge*. I've got a billion garbage commits in here.

zephraph:
I also want to give a huge shoutout to @eddiewebb for making this PR possible and working with me to get the functionality we needed. You and the whole CircleCI team do great work! <3

.circleci/config.yml
```diff
@@ -1,117 +1,89 @@
...
-
-filter_master: &filter_master
-  filters:
-    branches:
-      only:
-        - master
+orbs:
+  node: artsy/node@volatile
+  yarn: artsy/yarn@volatile
+  queue: eddiewebb/queue@volatile
```
orta: Any chance a comment could link to the two artsy references?

zephraph: Do you mean links to the repo?

eddiewebb: Although docs will use `volatile` as easy way to reference the latest, I would suggest pointing to specific versions for each orb for deterministic behavior over time.

orb cli or online [registry](https://circleci.com/orbs/registry/) are easy easy to check version and source

eddiewebb: Actually, easiest way is to pick a job that passed, and look at the top of the config file, https://circleci.com/gh/artsy/reaction/15151#config/containers/0 and you'll see

```
# Orb 'artsy/node@volatile' resolved to 'artsy/node@0.0.1'
# Orb 'artsy/yarn@volatile' resolved to 'artsy/yarn@0.0.1'
# Orb 'eddiewebb/queue@volatile' resolved to 'eddiewebb/queue@1.0.107'
```

zephraph: Cool, thanks. I'll update it.

orta:
Squashed & merged 👍  nice one @zephraph

PR #1485, opened 2018-11-02
Alternate payment form

sweir27:
This is a big PR, so I'll try to break it down. (Also it'll likely need a rebase/update when https://github.com/artsy/metaphysics/pull/1367 is merged)

This adds a payment form that will be used in the user/settings view:
payment-page

Roughly, it's a component that'll get stitched into Force with the following component structure:
- `UserSettingsPayments`: The outer-level component. It will embed the `SavedCreditCards` view if there are any saved credit cards. It always renders the `PaymentForm`.
- `SavedCreditCards` displays a list of credit cards that are connected to the current user. You can Remove them and remain on the page.
- `PaymentForm` is adapted from the payment form used in the checkout flow. Since there is slightly different validation logic and it's sending a different mutation, I feel OK about not trying to overload the other component to handle both of these cases. It's actually not so bad since we've extracted `AddressForm` and the `CreditCardInput`.
- `PaymentFormWrapper` just wraps the `PaymentForm` with the stripe context.


artsyit:

Built with commit 64fca28d0d4ba9419da1d2b9012a76467ce094e7

https://deploy-preview-1485--artsy-reaction.netlify.com

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+      phoneNumber: true,
+    }
+  }
+
+  onSubmit = () => {
+    const billingAddress = this.getSelectedBillingAddress()
+    const { me } = this.props
+    this.setState({ isCommittingMutation: true }, () => {
+      const errors = this.validateAddress(this.state.address)
+      if (Object.keys(errors).filter(key => errors[key]).length > 0) {
```
damassi: Would it make sense to maybe encompass this bit within `validateAddress` so that it can be checked like 

`if (errors.hasErrors)` 

(or something)

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+  render() {
+    const { error, isCommittingMutation } = this.state
+
+    return (
+      <Theme>
+        <>
+          <Responsive>
+            {({ xs }) => {
+              return (
+                <Flex flexDirection={xs ? "column" : "row"}>
```
damassi: Out of curiousity does this work? 

```
flexDirection={['column', 'row']}
```

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+      address_country: country,
+    }
+  }
+
+  onCreditCardAdded(me, store, data): void {
+    const {
+      createCreditCard: { creditCardOrError },
+    } = data
+
+    if (creditCardOrError.creditCardEdge) {
```
damassi: Can you add a small comment describing whats happening for those not familiar with relay internals?

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+        variables: {
+          input: { token },
+        },
+        updater: (store, data) => this.onCreditCardAdded(me, store, data),
+      }
+    )
+  }
+
+  private onMutationError(errors, errorModalMessage?) {
+    console.error("PaymentForm.tsx", errors)
```
damassi: (When will merges his PR will want to update this)

src/Components/Payment/PaymentFormWrapper.tsx
```diff
@@ -0,0 +1,58 @@
...
+import PaymentForm from "./PaymentForm"
+
+declare global {
+  interface Window {
+    Stripe?: (key: string) => stripe.Stripe
+    sd: {
+      STRIPE_PUBLISHABLE_KEY: string
+    }
+  }
+}
```
damassi: https://github.com/artsy/reaction/blob/master/typings/sharify.d.ts#L18

src/Components/Payment/PaymentFormWrapper.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export class PaymentFormWrapper extends React.Component<
+  PaymentFormWrapperProps,
+  PaymentFormWrapperState
+> {
+  state = { stripe: null }
+
+  constructor(props) {
+    super(props)
+  }
```
damassi: This can be removed

src/Components/Payment/SavedCreditCards.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { CreditCardDetails } from "Apps/Order/Components/CreditCardDetails"
+import { ErrorModal } from "Components/Modal/ErrorModal"
+import React from "react"
+import { commitMutation, graphql, RelayProp } from "react-relay"
+import { ConnectionHandler } from "relay-runtime"
+import styled from "styled-components"
+
+interface SavedCreditCardsProps {
+  creditCards: any
+  me: any
```
damassi: I wonder if there's a `me` interface anywhere in reaction?

src/Components/Payment/SavedCreditCards.tsx
```diff
@@ -0,0 +1,181 @@
...
+interface CreditCardsState {
+  isErrorModalOpen: boolean
+  isCommittingMutation: boolean
+}
+
+interface CreditCardProps {
+  creditCard?: any
+  me: any
+  relay?: RelayProp
+}
```
damassi: Can this and `SavedCreditCardsProps` be shared?

src/Components/Payment/SavedCreditCards.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+const SpinnerContainer = styled.div`
+  padding-right: 30px;
+  position: relative;
+`
+
+export class SavedCreditCards extends React.Component<SavedCreditCardsProps> {
+  constructor(props) {
+    super(props)
+  }
```
damassi: Can remove

src/Components/Payment/SavedCreditCards.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  render() {
+    return (
+      <Theme>
```
damassi: Will all of these components have a common root? If so maybe can create a Form component and then put `<Theme>` in there?

src/Components/Payment/UserSettingsPayments.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  render() {
+    const creditCardEdges = get(this.props, props => props.me.creditCards.edges)
+    const creditCards = creditCardEdges.map(({ node: creditCard }) => {
+      return creditCard
+    })
+
+    return (
+      <>
+        <Theme>
```
damassi: Can likely remove Theme and outer `<>`

src/Components/Payment/UserSettingsPayments.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  ...UserSettingsPayments_me
+                }
+              }
+            `}
+            render={({ props }) => {
+              if (props) {
+                return <UserSettingsPaymentsFragmentContainer me={props.me} />
+              } else {
+                return null
+              }
```
damassi: Would it make sense to use [`renderWithLoadProgress`](https://github.com/artsy/reaction/blob/master/src/Artsy/Relay/renderWithLoadProgress.tsx) here?

src/Components/Payment/__tests__/UserSettingsPayments.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+jest.mock("react-stripe-elements", () => ({
+  Elements: ({ children }) => children,
+  StripeProvider: () => <div />,
+  CardElement: () => jest.fn(),
+  injectStripe: () => jest.fn(),
+}))
+
+describe("UserSettingsPayments", () => {
+  beforeAll(() => {
+    // @ts-ignore
```
damassi: Is this still needed?

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+        }
+      })
+    })
+  }
+
+  render() {
+    const { error, isCommittingMutation } = this.state
+
+    return (
+      <Theme>
```
damassi: I mention something about this below, but this theme could likely be hoisted into the main parent `UserSettingsPayments`

damassi:
One meta question is: how sharable is a lot of this code across the other (new) payment form? I'm thinking things like the validators and tests and so on -- are they duplications or are there minor differences?

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+import React, { Component } from "react"
+import { commitMutation, graphql, RelayProp } from "react-relay"
+import { injectStripe, ReactStripeElements } from "react-stripe-elements"
+import { ConnectionHandler } from "relay-runtime"
+import { Responsive } from "Utils/Responsive"
+
+export interface PaymentFormProps
+  extends ReactStripeElements.InjectedStripeProps {
+  relay?: RelayProp
+  me: any
```
damassi: I wonder where a generic `Me` interface could go? Maybe [here](https://github.com/artsy/reaction/blob/master/src/Artsy/SystemContext.tsx#L19)?

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+    const usOrCanada = country === "US" || country === "CA"
+    return {
+      name: validatePresence(name),
+      addressLine1: validatePresence(addressLine1),
+      city: validatePresence(city),
+      region: usOrCanada && validatePresence(region),
+      country: validatePresence(country),
+      postalCode: usOrCanada && validatePresence(postalCode),
+    }
+  }
```
damassi: Part of my meta question about reusability -- if this code is being used in the other form maybe it makes sense to hoist it up into a `Order/Utils/formValidators.js` file?

src/Components/Payment/SavedCreditCards.tsx
```diff
@@ -0,0 +1,181 @@
...
+  relay?: RelayProp
+}
+
+interface CreditCardsState {
+  isErrorModalOpen: boolean
+  isCommittingMutation: boolean
+}
+
+interface CreditCardProps {
+  creditCard?: any
```
damassi: Is there an interface that can be pulled from Stripe?

src/Components/Payment/SavedCreditCards.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+export class CreditCard extends React.Component<
+  CreditCardProps,
+  CreditCardsState
+> {
+  state = { isErrorModalOpen: false, isCommittingMutation: false }
+
+  render() {
+    return (
+      <>
```
damassi: no need to wrap

src/Components/Payment/__tests__/PaymentForm.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+jest.mock("react-stripe-elements", () => ({
+  CardElement: props => <div {...props} />,
+  injectStripe: args => args,
+}))
+
+import { commitMutation, RelayProp } from "react-relay"
+import {
+  Address,
+  AddressForm,
+} from "../../../Apps/Order/Components/AddressForm"
```
damassi: Now that we have type-checking in on tests all of these relative paths throughout test can be absolute

src/Components/Payment/SavedCreditCards.tsx
sweir27: They're  slightly  different since `savedCreditCards` needs an array. I'm updating  the  types now though to be clearer!

src/Components/Payment/SavedCreditCards.tsx
sweir27: They're slightly different since `savedCreditCards` needs an array. I'm updating the types now though to be clearer!

src/Components/Payment/SavedCreditCards.tsx
sweir27: I think it does since  it's including two components at the  root level (`<BorderBox` and `<ErrorModal`)

src/Components/Payment/UserSettingsPayments.tsx
sweir27: This is the top-level component, so I think the `Theme` is needed  here?

src/Components/Payment/UserSettingsPayments.tsx
sweir27: I think so... need to dig in since updating  gives me a TS error:
```
Property 'me' is missing in type 'MappedFragmentProps<Pick<any, string | number | symbol>> & ComponentRef'.```

File: src/Components/Payment/__tests__/UserSettingsPayments.test.tsx (hunk 13 above)
Comment (sweir27) [reply src/Components/Payment/__tests__/UserSettingsPayments.test.tsx]: Seems to be

File: src/Apps/Order/Routes/Shipping/index.tsx [hunk 20]
```diff
@@ -236,7 +234,7 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
       phoneNumber,
     } = address
     const usOrCanada = country === "US" || country === "CA"
-    return {
+    const errors = {
```
sweir27: This validator  is  slightly different  from the one shared by the payment forms, so decided to keep it in this file.

src/Components/Payment/PaymentFormWrapper.tsx
sweir27: I updated  this to just  pull in `sd` directly. I'm actually not sure why it was extracted  separately before, but this appears to work?

damassi: Not sure? I think `STRIPE_PUBLISHABLE_KEY` needs to go in the `sharify.d.ts` as well as the `webpack.config` file -- @williardx also recently had to update this for his logging work.

sweir27: updated

sweir27:
@damassi assuming the build is 🍏, I think this is good-to-go! Mind taking a final pass?

damassi:
Yup! 👁

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -245,18 +243,23 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
       postalCode: usOrCanada && validatePresence(postalCode),
       phoneNumber: validatePresence(phoneNumber),
     }
+    return {
+      errors,
+      hasErrors: Object.keys(errors).filter(key => errors[key]).length > 0,
```
damassi: #minor, but would be nice to separate a complex operation from object assignment, i.e., 

```js
const hasErrors = Object.keys(errors).filter(key => errors[key]).length > 0,

return {
  errors,
  hasErrors
}
```

src/Apps/Order/Utils/formValidators.ts
```diff
@@ -0,0 +1,23 @@
...
+    name: validatePresence(name),
+    addressLine1: validatePresence(addressLine1),
+    city: validatePresence(city),
+    region: usOrCanada && validatePresence(region),
+    country: validatePresence(country),
+    postalCode: usOrCanada && validatePresence(postalCode),
+  }
+  return {
+    errors,
+    hasErrors: Object.keys(errors).filter(key => errors[key]).length > 0,
```
damassi: same here

src/Components/Payment/PaymentForm.tsx
```diff
@@ -0,0 +1,264 @@
...
+      phoneNumber: true,
+    }
+  }
+
+  onSubmit = () => {
+    const billingAddress = this.getSelectedBillingAddress()
+    const { me } = this.props
+    this.setState({ isCommittingMutation: true }, () => {
+      const { errors, hasErrors } = validateAddress(this.state.address)
+      if (hasErrors) {
```
damassi: 👍

PR #1490, opened 2018-11-05
Collections app scaffold

eessex:
Sets up basic scaffold for `/collections` app to display header, meta, and grid of collections.

Currently it shows all collections, and will require some updates once we are able to query by category - https://artsyproduct.atlassian.net/browse/GROW-989

screen shot 2018-11-05 at 11 04 13


artsyit:

Built with commit ab054ef51097417fec9c00363902b7eac1ae238d

https://deploy-preview-1490--artsy-reaction.netlify.com

src/Apps/Collections/CollectionsApp.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </FrameWithRecentlyViewed>
+      </>
+    )
+  }
+}
+
+const HeaderContainer = styled(Box)`
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-end;
```
yuki24: Any reason we can't use the styled system syntax?

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -0,0 +1,77 @@
...
+            </CollectionsContainer>
+          </Box>
+        )}
+      </Responsive>
+    )
+  }
+}
+
+const CollectionsContainer = styled(Row)`
+  justify-content: space-between;
```
yuki24: Same here

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -0,0 +1,77 @@
...
+`
+
+const CollectionContainer = styled(Col)<{
+  hasBorderTop: boolean
+  hasMarginRight: boolean
+  isFullWidth: boolean
+}>`
+  border-bottom: 1px solid ${color("black10")};
```
yuki24: Rather than manually adding a border to each component we need add a border to, we should probably re-use the [`<Join>`](https://palette.artsy.net/#/docs-join) or [`<Spacer>`](https://palette.artsy.net/#/docs-spacer) component and take out one of these depending on the screen size.

src/Apps/Collections/routes.tsx
```diff
@@ -0,0 +1,25 @@
...
+    Component: CollectionsApp,
+    query: graphql`
+      query routes_MarketingCollectionsAppQuery {
+        collections: marketingCollections {
+          ...CollectionsApp_collections
+        }
+      }
+    `,
+    render: ({ props, Component }) => {
+      if (!props) {
```
yuki24: I could be wrong, but I think `props` always evaluates to `{}` or an object that has properties at runtime. Is this check really necessary?

src/Apps/Collections/CollectionsApp.tsx
eessex: I'm not familiar, what is the system syntax?

src/Apps/Collections/Components/CollectionsGrid.tsx
eessex: Looks good, will check this out

src/Apps/Collections/routes.tsx
l2succes: Yeah it should be there, when `props` is null, it means relay hasn't gotten any data back yet so we shouldn't render the relay container. We could actually have a `<Loading />` instead of `null` here but since we're using SSR, it's not needed.

See https://github.com/relay-tools/found-relay#render

l2succes:
@eessex Code looks good but as @yuki24 mentioned let's use `<Flex>` and `<Box>` and other layout components from palette as needed.

Also we'll have to rework these queries with groups shortly after, it might've made more sense to start with the KAWS work first. But i'll leave it up to you to decide.

src/Apps/Collections/CollectionsApp.tsx
yuki24: You can specify CSS properties as React props on the `Box` component. Also, you may want to use the `<Flex>`component since the `<Box>` component does not accept flex properties:

```tsx
<Flex justifyContent="space-between" alignItems="flex-end">
  ...
</Flex>
```

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -0,0 +1,52 @@
...
+              {collections.map((collection, index) => {
+                const hasBorderTop = xs ? index === 0 : index < 3
+
+                return (
+                  <Flex
+                    width={xs ? "100%" : "30%"}
+                    flexDirection="column"
+                    key={index}
+                  >
+                    {hasBorderTop && <Separator />}
```
yuki24: Nice 👍

PR #1491, opened 2018-11-05
Set default image quality to 80 from 95

l2succes:
Makes our default image quality (size) smaller.

This PR will also cut the header image size by over 50% on collections with no noticeable difference in quality
screen shot 2018-11-05 at 12 27 59 pm


artsyit:

Built with commit 6bd0fe360ab45dc7f2c740addedefc7d2c052b29

https://deploy-preview-1491--artsy-reaction.netlify.com

zephraph:
🙌 So good!

src/Apps/Collect/Components/Collection/Header/index.tsx
```diff
@@ -71,8 +71,9 @@ export class CollectionHeader extends Component<Props> {
                     mt={xs ? 0 : 3}
                     mb={3}
                     headerImageUrl={resize(collection.headerImage, {
-                      width: imageWidth * 2,
-                      height: imageHeight * 2,
+                      width: imageWidth,
+                      height: imageHeight,
```
xtina-starr: Perhaps you already did, but if you haven't it might be worth checking the quality of the header image in mobile and making sure it isn't degraded. I think that's the initial reason why you suggested increasing the image size here.

l2succes: Yeah that's a good point. I just checked again on mobile and the quality does go down a lot on retina displays so what we can do is only add `* 2` only for mobile

eessex: Does this affect retina screens overall or just mobile?

PR #1494, opened 2018-11-05
Make artwork banner real

oxaudo:
Mainly want to get some feedback on approach and figure out one error that is puzzling me. I'll ask some questions inline.

Once that is done I'll add some specs and remove WIP.

The view with names included(not committed here because it would not compile properly:( )
screen shot 2018-11-05 at 4 38 53 pm

And here is the small version for now:
screen shot 2018-11-05 at 5 20 21 pm


artsyit:

Built with commit 8a844f481ec14cacac76bed2136bdd62e30841f0

https://deploy-preview-1494--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+            variables={{ artworkID }}
+            query={graphql`
+              query ArtworkBannerQuery($artworkID: String!) {
+                artwork(id: $artworkID) {
+                  ...ArtworkBanner_artwork
+                }
+              }
+            `}
+            render={renderWithLoadProgress(
+              ArtworkBannerFragmentContainer as any
```
oxaudo: Also again will not render for me locally without casting to `any` here. Might be able to remove that once I figure out this name thing. I home:)

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -41,10 +34,6 @@ export class ArtworkDetailsAboutTheWorkFromPartner extends React.Component<
     // noop
   }
 
-  renderProfileImage(imageUrl?: string, initials?: string) {
```
oxaudo: This is a leftover from some previous refactors. Not used any more.

oxaudo:
Current look with actual names stored in Fixtures by running the query against production for listed works.

screen shot 2018-11-05 at 4 38 53 pm

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -32,12 +32,7 @@ export const ArtworkApp: React.SFC<Props> = props => {
     <HorizontalPadding>
       <Row>
         <Col sm={8}>
-          <Banner
-            src="https://picsum.photos/110/110/?random"
-            badge="In show"
-            headline="Francesca DiMattio: Boucherouite"
-            subHeadline="Salon 94"
-          />
+          <ArtworkBanner artworkID={props.artwork.id} />
```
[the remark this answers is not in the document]
damassi: Yup, it's mostly in place. I was going to look at seeing what remains right now. In the meantime as we're building items out this is a good approach and easy to refactor out later.

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { Banner } from "./Banner"
+
+import { ArtworkBanner_artwork } from "__generated__/ArtworkBanner_artwork.graphql"
+import { ArtworkBannerQuery } from "__generated__/ArtworkBannerQuery.graphql"
+
+export interface ArtworkBannerProps {
+  artwork: ArtworkBanner_artwork
+}
+
+export class ArtworkBanner extends React.Component<ArtworkBannerProps> {
```
damassi: If not using `state` this should be an SFC

src/Apps/Artwork/Components/ArtworkBanner/Banner.tsx
```diff
@@ -3,7 +3,8 @@ import React from "react"
 import { Responsive } from "Utils/Responsive"
 
 export interface BannerProps {
-  src: string
+  imageUrl: string
```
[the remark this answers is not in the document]
damassi: We should use the EntityHeader component and not recreate: https://artsy-reaction.netlify.com/?selectedKind=Styleguide%2FComponents&selectedStory=EntityHeader&full=0&addons=1&stories=1&panelRight=0&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel

https://github.com/artsy/reaction/blob/master/src/Styleguide/Components/__stories__/EntityHeader.story.tsx

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+      case "ArtworkContextAuction":
+        const auctionImage = get(partner, p => p.profile.icon.url)
+        return (
+          <Banner
+            imageUrl={auctionImage}
+            initials={partner.initials}
+            badge="In auction"
+            headline={context.name}
+            subHeadline={partner.name}
+          />
```
damassi: Can be replaced with EntityHeader -- just need to account for the badge

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        )
+      case "ArtworkContextFair":
+        const fairImage = get(context, c => c.profile.icon.img.url)
+        const initials = get(context, c => c.profile.initials)
+        return (
+          <Banner
+            imageUrl={fairImage}
+            initials={initials}
+            badge="At fair"
+            // headline={context.name}
```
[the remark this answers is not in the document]
damassi: Not sure -- @mzikherman?

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+export class ArtworkBanner extends React.Component<ArtworkBannerProps> {
+  render() {
+    const { context, partner } = this.props.artwork
+    if (!context) return null
+    // imageUrl: image for avatar
+    // initials: fallback partner initials in case image is not there.
+    // badge: in auction / at fair / in show
+    // headline:  auction / fair / show name
+    // subHeadline: partner name
+    switch (context.__typename) {
```
damassi: Nice and clear 👍

src/Apps/Artwork/Components/__stories__/Banner.story.tsx
```diff
@@ -1,30 +1,31 @@
...
+          initials="OU"
           badge="In show"
           headline="Francesca DiMattio: Boucherouite"
           subHeadline="Salon 94"
         />
       </Section>
       <Section title="Small Banner">
         <SmallBanner
-          src="https://picsum.photos/110/110/?random"
+          imageUrl="https://picsum.photos/110/110/?random"
```
damassi: I think this file can go away once we implement `EntityHeader` and instead rely on `ArtworkBanner`

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
oxaudo: I was thinking about that too, but I believe entity header currently have different fonts/sizes for each row there and is not following the same responsive logic. Might be a good point to bring to @design team to see if there is a way we can consolidate those or they are different entities indeed.

This whole Banner component seems like it should not be specific to artwork. But don't see it used in any other designs currently.

src/Apps/Artwork/Components/ArtworkBanner/Banner.tsx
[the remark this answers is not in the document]
damassi: @briansw - Thoughts on this?

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
damassi: Yeah, looking at the designs its different enough to warrant a different component. We should make sure that the props are roughly the same just in case we might want to unify at a later point.

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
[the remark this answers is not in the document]
mzikherman: That is odd, I'm not sure! It should, `name` is a valid field on this types, so not being included in the generated file (and also no Relay compiler/TS complaints) _is_ really weird!

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+          name
+          href
+          is_auction
+          is_closed
+          is_open
+          live_start_at
+          live_url_if_open
+        }
+        ... on ArtworkContextFair {
+          name
```
oxaudo: @alloy - wondering if you might have an idea of what is going on here. We are getting context for the artwork in this query that can be one of 3 types: ArtworkContextAuction, ArtworkContextFair or ArtworkContextPartnerShow. Based on what is the type we are requesting proper object in MP (https://github.com/artsy/metaphysics/blob/master/src/schema/artwork/context.js#L44). All the underlying objects have name/href properties. And  running this query on MP directly returns proper data. However defining it here as part of ArtworkBannerFragmentContainer - the name/href is dropped for some reason for Fairs and Shows. It's not included in src/__generated__/ArtworkBanner_artwork.graphql.ts below. Any thoughts?

src/__generated__/ArtworkBanner_artwork.graphql.ts
```diff
@@ -0,0 +1,337 @@
...
+        readonly name: string | null;
+        readonly href: string | null;
+        readonly is_auction: boolean | null;
+        readonly is_closed: boolean | null;
+        readonly is_open: boolean | null;
+        readonly live_start_at: string | null;
+        readonly live_url_if_open: string | null;
+    } | {
+        readonly __typename: "ArtworkContextFair";
+        readonly is_active: boolean | null;
```
oxaudo: @alloy - Here. I was expecting to see name/href as well.

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+      )
+    case "ArtworkContextFair":
+      const fairImage = get(context, c => c.profile.icon.img.url)
+      const initials = get(context, c => c.profile.initials)
+      return (
+        <Banner
+          imageUrl={fairImage}
+          initials={initials}
+          meta="At fair"
+          name="Fair name goes here" // {context.name}
```
oxaudo: @alloy - so name would display properly here, but will give me an error and would not allow to push:)

src/Apps/Artwork/Components/__stories__/Banner.story.tsx
oxaudo: @damassi - I left it here as a test for responsive versions of the Banner. The actual data for the Artwork Banner is storied in ArtworkBanner file.

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
alloy: Yeah this is a thing on your env, because it works fine again for me when I remove it and also last time when I PR’ed removing them CI was fine too. Let’s do a quick pair on this.

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+          type
+          status
+          thumbnail: cover_image {
+            img: resized(width: 70, height: 70, version: "square") {
+              url
+            }
+          }
+        }
+      }
+      # FIXME: There is a bug in the Relay transformer used before generating Flow types, and thus also our TS type
```
oxaudo: @alloy - Thanks for pointing me to the example of the same issue and this workaround. It did work smoothly here.

src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+    artworkContextPartnerShow,
+    partner,
+  } = props.artwork
+
+  // imageUrl: image for avatar
+  // initials: fallback partner initials in case image is not there.
+  // meta: in auction / at fair / in show
+  // name:  auction / fair / show name
+  // subHeadline: partner name
+  if (artworkContextAuction) {
```
oxaudo: I think this series of `if`s here with the corresponding shortcut returns actually read clear here and make it more explicit that only one banner will be displayed at a time. 

But if we do decide to go with proper query without aliasing once it's a possibility - the change is easy here.

src/Apps/Artwork/Components/ArtworkBanner/Banner.tsx
```diff
@@ -3,33 +3,32 @@ import React from "react"
...
-          else return <LargeBanner {...this.props} />
-        }}
-      </Responsive>
-    )
-  }
+export const Banner: React.SFC<BannerProps> = props => {
+  return (
+    <Responsive>
+      {({ xs }) => {
+        if (xs) return <SmallBanner {...props} />
```
oxaudo: @damassi - is there a new patter to handle this responsiveness now?

src/Apps/Artwork/Components/__tests__/ArtworkBanner.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  describe("ArtworkBanner for artwork with auction banner", () => {
+    beforeAll(async () => {
+      wrapper = await getWrapper(ArtworkAuctionBannerFixture)
+    })
+
+    it("renders a correct data for the auction", () => {
+      const html = wrapper.html()
+      expect(html).toContain("In auction")
+      // expect(html).toContain("Doyle: Post-War & Contemporary Art")
```
oxaudo: @alloy - even though the alias seemed to be picked up properly now, the final data does not include object that is inside of the context data. I suspect it is because context itself in turn loads either Auction/Fair or Show and appends data there. Not sure how to fix that part.

src/Apps/Artwork/Components/__tests__/ArtworkBanner.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      `,
+      mockResolvers: {
+        Artwork: () => ({
+          ...response,
+          // FIXME: Need to figure out how to return proper values for aliesed objects
+          context: (_source, _args, _context, info: GraphQLResolveInfo) => {
+            const alias = info.fieldNodes[0].alias!.value
+            const x = response[alias]
+            return x
+          },
```
alloy: For context, I’m surprised that graphql-tools’ mocking doesn’t look for aliases when returning a static payload, so need to figure out if that’s supposed to work or not

oxaudo:
@damassi - ok - merging this while I have it without conflicts!

PR #1495, opened 2018-11-05
Use APP_URL to allow Force to override the APP_URL

yuki24:
In staging and development I've observed weird behavior where newly signed-up users are redirected to production. We should be able to switch to `APP_URL` environment variable safely and fall back to `www.artsy.net` so the existing behaviour is preserved.


eessex:
Nice change 👍

PR #1500, opened 2018-11-06
[PATCH makeOffer] Transaction Summary UI tweaks

pepopowitz:
This PR makes a few modifications to the `<TransactionSummary>` component, mostly visual. It includes:

1. A refactor of the `<Entry>` helper component into two helper components - `<Entry>` and `<SecondaryEntry>`. `<Entry>` previously took a prop named `secondary`, which toggled the look of the entry item slightly. It was two very separate code paths, though, and refactoring to two separate components made making changes to the entry items more intuitive.
2. Larger font sizes for primary entry items at breakpoints larger than "xs". This was accomplished with the `<Responsive>` component, because I didn't see a way in [palette.artsy.net](https://palette.artsy.net/#/docs-typography) to send an array of font sizes to typographical components. I did raise that question in the #design-system Slack channel though, and it sounds like we might be able to add this ability to palette. That shouldn't hold up this PR, though, in my opinion.
3. Modification to the secondary entry item to use black60 instead of black30, to match the Zeplin prototype.

### DIFF
--- a/src/Apps/Order/Components/TransactionSummary.tsx
+++ b/src/Apps/Order/Components/TransactionSummary.tsx
@@ -2,6 +2,7 @@ import { TransactionSummary_order } from "__generated__/TransactionSummary_order
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { get } from "Utils/get"
+import { Responsive } from "Utils/Responsive"
 
 import {
   Box,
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit aedb133fad79adf024679560e5e649cd5eeeef40

https://deploy-preview-1500--artsy-reaction.netlify.com

PR #1503, opened 2018-11-07
[FIX OfferAmount] Switch to offerAmount and pass amount and currencyCode

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-639

# Solution
Follow up with https://github.com/artsy/metaphysics/pull/1371


artsyit:

Built with commit 5d976c2c96a785a0fd169455ddcebcd84c857996

https://deploy-preview-1503--artsy-reaction.netlify.com

damassi:
@ashkan18 - note that the formatting in the description won't trigger a release -- and its also a formatted a bit incorrectly. Needs to be in a commit message and need to follow all caps and so on from the README.

ashkan18:
oh thanks @damassi ! I thought i can just say `[FIX something]` in the title and that'll do it. Although this title of PR when merged gets into a merge commit message, would that be good enough?

PR #1504, opened 2018-11-07
small visual fix for works for you render

oxaudo:
@damassi - I think the latest reshuffle brought this issue: https://artsyproduct.atlassian.net/browse/DISCO-489

Seems like the simple fix here and it makes sense. Cause once the data is loaded the content is re rendered.

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -52,20 +52,22 @@ export const ArtworkApp: React.SFC<Props> = props => {
           <ArtworkDetails artwork={props.artwork} />
         </Col>
       </Row>
-      <Row>
-        <Col sm={8}>
-          <ArtistInfo artistID={props.artwork.artist.id} />
-        </Col>
-      </Row>
-
-      {typeof window !== "undefined" && (
+      {props.artwork.artist && (
         <Row>
-          <Col>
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -52,22 +52,24 @@ export const ArtworkApp: React.SFC<Props> = props => {
...
         </Col>
       </Row>
-      <Row>
-        <Col sm={8}>
-          <ArtistInfo artistID={props.artwork.artist.id} />
-        </Col>
-      </Row>
-
-      {typeof window !== "undefined" && (
+      {props.artwork.artist && (
```
oxaudo: Also looked at some Cultural maker artworks - those don't have artists and will crash here.

Technically I think there should be a test for this stuff but not sure if we have any of those yet for the top level page.

artsyit:

Built with commit 8dac383bfaa2b84cd5759b26b7556e86f7a6a7c4

https://deploy-preview-1504--artsy-reaction.netlify.com

src/Apps/Artwork/ArtworkApp.tsx
damassi: We're just in the cleanup phase now, all of this stuff is gonna be moved around. Not ready to be tested (but will be tested soon).

damassi:
👍

Issue #1506, opened 2018-11-07
Input fields lose focus after each keypress


sweir27:
Able to reproduce this in storybooks: https://artsy-reaction.netlify.com/?selectedKind=Apps%2FOrder%20Page&selectedStory=Shipping%20-%20Untouched%20Order&full=0&addons=1&stories=1&panelRight=0&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel

After each keypress, the field loses focus (I also see a warning in the browser about switching to the new Responsive component).

PR #1511, opened 2018-11-08
[PATCH Order Forms] Fix bug that made it impossible to enter more than 1 character.

damassi:
Fixes https://github.com/artsy/reaction/issues/1506

@alloy - so this gets weird: 

- I've added an `AddressForm` story with two `<Section>`s
- One section contains two media elements and wraps an example `GoodAddressForm` containing no media elements and everything works (though the form state is lost when matched against different breakpoints -- fine and expected since things are being mounted / unmounted)
- The other section contains `BadAddressForm` and inside of that there are two media elements + the form. This one reproduces the error. 

form


artsyit:

Built with commit 3b85cf984626a10df1a0f8d4c4c7924a8a586635

https://deploy-preview-1511--artsy-reaction.netlify.com

damassi:
As an aside, is it possible to delete 

screen shot 2018-11-07 at 9 30 13 pm

from reaction and instead rely exclusively on `DynamicResponsive` from `@artsy/react-responsive-media`? It's a bit confusing and i'm not sure what's being called from where / whats needed! It _seems_ like it would be fine but would like to get your take.

alloy:
Alright, going for it so this is fixed for people waking up now.

src/Apps/Order/Routes/__tests__/Offer.test.tsx
```diff
@@ -54,13 +54,13 @@ describe("Offer InitialMutation", () => {
...
     input.props().onChange({ currentTarget: { value: "1" } } as any)
-    expect(transactionSummary.update().text()).toContain("Your offer$1")
+    expect(transactionSummary.text()).toContain("Your offer$1")
 
     input.props().onChange({ currentTarget: { value: "1.23" } } as any)
-    expect(transactionSummary.update().text()).toContain("Your offer$1")
+    expect(transactionSummary.text()).toContain("Your offer$1")
 
     input.props().onChange({ currentTarget: { value: "1023.23" } } as any)
-    expect(transactionSummary.update().text()).toContain("Your offer$1,023")
```
damassi: oooh so this was a bug represented by the other bug -- i had no idea how this worked before and was so confused.

alloy: Somewhat, yeah, but there’s tons of tests where (apparently) we already needed to update the tree after state changes, so these felt natural. But yep, the need went away after the fix 👍

PR #1513, opened 2018-11-08
[FEATURE SEO] Add a bread crumb list to <head> search results

yuki24:
partially addresses https://artsyproduct.atlassian.net/browse/GROW-952

This adds a `<script>` tag that hosts the information for the website structure (a.k.a structured data):

screen shot 2018-11-08 at 10 22 46 am


artsyit:

Built with commit 6e07b06cf072769114d49d385145f541663e3d7c

https://deploy-preview-1513--artsy-reaction.netlify.com

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
```diff
@@ -0,0 +1,148 @@
...
+            is_acquireable
+            is_price_range
+            price
+            price_currency
+            title
+            artists {
+              name
+            }
+            image {
+              url(version: "large")
```
yuki24: We may not have to use the `large` image. let me know if this needs to be updated.

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    // here the filtering is necessary so we can re-use the artwork list shown in the page (could include
+    // non-acquireable artworks) without making an extra request.
+    const artworksForSeoProduct = artworks_connection.edges.filter(
+      edge => edge.node.is_acquireable && edge.node.partner
+    )
+
+    return artworksForSeoProduct.map(({ node }, index) => {
+      const { artists, is_price_range, partner, price } = node
+      const location = partner.locations && partner.locations[0]
```
yuki24: Here the location is picked up based on the array index, which may not be the location where the piece is stored, but I wasn't sure how I could pick the right location.

yuki24:
Alright, getting a green build was a bit tough but this PR is good to review.

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
eessex: Does the seller need to be limited to one?

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
```diff
@@ -0,0 +1,148 @@
...
+        : null
+      const isInstitution = partner.type === "Institution"
+
+      return (
+        <Product
+          key={node.__id}
+          data={{
+            name: node.title,
+            image: node.image.url,
+            description: node.meta.description,
```
eessex: Probably worth checking if `node.image` and `node.meta` exist, we came upon some bad errors when deep objects came up undefined when implementing this for artist pages.

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
```diff
@@ -0,0 +1,148 @@
...
+                      : {
+                          minPrice: formatCurrency(price.split("-")[0]),
+                          maxPrice: formatCurrency(price.split("-")[1]),
+                        },
+                    priceCurrency: node.price_currency,
+                    availability: AVAILABILITY[node.availability],
+                    seller: {
+                      "@type": "ArtGallery",
+                      name: partner.name,
+                      image: partner.profile.icon.url,
```
eessex: Same here -- not all profiles have icons

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
eessex: this can probably use the `thumbnail` size, I believe that is what we use elsewhere in JSONld

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
yuki24: Could you send me some artwork examples that don't have `image` or `meta` so I can test this?

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
yuki24: Good question. From our perspective, no, but structured data may not allow the data to have multiple sellers. I'll double-check once I have access to Google (I'm in Beijing and have no access to Google)

src/Apps/Collect/CollectApp.tsx
```diff
@@ -36,6 +41,8 @@ export class CollectApp extends Component<CollectAppProps> {
         <Meta property="og:description" content={description} />
         <Meta property="twitter:description" content={description} />
         <Link rel="canonical" href={canonicalHref} />
+        <BreadCrumbList items={[{ path: "/collect", name: "Collect" }]} />
```
xtina-starr: I'm wondering if we should expand this to some how handle the case for when there is a medium for the collect page? I think so because there is mention of it in the ticket under "For Collect/Medium, the breadcrumb structure can be:". 

This would handle url such as `/collect/painting`. I think you could pass in `canonicalHref` which takes this into account.

src/Apps/Collect/Components/Seo/BreadCrumbList.tsx
```diff
@@ -0,0 +1,46 @@
...
+              "@type": "ListItem",
+              position: 1,
+              item: {
+                "@id": APP_URL,
+                name: "Artsy",
+              },
+            },
+            ...items.map(({ path, name }, index) => ({
+              "@type": "ListItem",
+              position: index + 2,
```
xtina-starr: I'm just curious but why do we add 2 instead of 1 to the index to calculate the position.

src/Apps/Collect/CollectApp.tsx
yuki24: Thanks for catching this! Yeah there should probably another path for mediums. I'll update this component so all mediums will be crawled.

src/Apps/Collect/Components/Seo/BreadCrumbList.tsx
yuki24: Yeah this looks tricky. The example in the official doc has a list where the index starts with `1` while in programminng it starts with `0`. There's a top-level breadcrumb link for www.artsy.net as the first element, so this `+2` is important to make the `itemListElement` valid. I'll add a comment about this so the context won't be lost.

src/Apps/Collect/CollectApp.tsx
yuki24: Now this is tested with `CollectApp.test.tsx` and covers tow cases:

 * `Artsy > Collect` on https://www.artsy.net/collect
 * `Artsy > Collect > Painting` on https://www.artsy.net/collect/painting

src/Apps/Collect/Components/Seo/SeoProductsForArtworks.tsx
```diff
@@ -0,0 +1,154 @@
...
+    // non-acquireable artworks) without making an extra request. Also, seller image is a required field
+    // so excluding those that don't have `partner.profile.icon.url`.
+    const artworksForSeoProduct = artworks_connection.edges.filter(
+      edge =>
+        edge.node.is_acquireable &&
+        edge.node.partner &&
+        edge.node.partner.profile &&
+        edge.node.partner.profile.icon &&
+        edge.node.partner.profile.icon.url
+    )
```
damassi: Does't look as cool as the above but could also use `get`! 
```js
import { get } from `Utils/get`

const artworksForSeoProduct = artworks_connection.edges.filter(edge => {
  const found = get(edge, e => e.node.is_acquireable && e.node.partner.profile.icon.url)
  return found
}
```

yuki24:
Marking this PR WIP again as all the artworks on the `/collect` page are now assigned to partner `"Mocktion Demo Partner Causality Bug Test"` which has no icon url and there's no way to self-QA. I'll double-check the data tomorrow and see if this is really merge-able.

yuki24:
I've manually added an icon to the demo partner for QA and now this is good for review.

src/Apps/Collect/CollectApp.tsx
```diff
@@ -36,6 +49,16 @@ export class CollectApp extends Component<CollectAppProps> {
         <Meta property="og:description" content={description} />
         <Meta property="twitter:description" content={description} />
         <Link rel="canonical" href={canonicalHref} />
+        <BreadCrumbList
+          items={[
+            { path: "/collect", name: "Collect" },
+            mediumMeta && {
+              path: `/collect/${mediumMeta.id}`,
+              name: mediumMeta.name,
+            },
```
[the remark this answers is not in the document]
yuki24: It is common to include meta tags for SEO in the components that represent a page, and having to import `sharify` for every `< BreadCrumbList>` seems like an opportunity for consolidation. This would keep high-level components cleaner because environment variables won't be spread out. Though I also see why the direct import from `react-head` with a string interpolation with `APP_URL` makes sense. The two items are required to handle two different routes:

 * When it routes to `/collect`:
    * The `items` props will be `[{ path: "/collect", name: "Collect" }]`
    * So Google will show `Artsy > Collect`
 * When it routes to e.g. `/collect/painting`:
    * The `items` props will be `[{ path: "/collect", name: "Collect" }, { path: "/collect/panting", name: "Painting" }]`
    * So Google will show `Artsy > Collect > Painting`

I think it makes more sense that mediums belong to `Collect` rather than the top-level `Artsy` since they are options on the Collect page. The top-level `Artsy` link is added automatically by the `<BreadCrumbList>` component.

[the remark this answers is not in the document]
xtina-starr: Yeah I understand the need for meta tag data but I guess I don't fully understand why you have `mediumMeta` and check for its presence versus checking for the presence of `medium` and do something like `/collect/${medium}`. We could probably even refactor `canonicalHref` to `canonicalPath`and just have the path (i.e. `/collect/${medium}` or `/collect`) then on line 50 do something like `sd.APP_URL + canonicalPath`. I was just thinking that lines 31-38 seem redundant and wasn't sure if I was missing something or the reason why they are needed. 

Regarding the routing, `canonicalHref` currently handles resolving to whatever path was routed to based on whether the medium parameter exists, but again maybe there is something I'm not considering or don't understand. Perhaps we can chat about it later.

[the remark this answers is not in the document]
yuki24: I used the `mediumMeta.name` that comes from the GraphQL query (the `aggregations` field holds mediums) so we don't have to worry about updating the breadcrumb list when we add/update/remove mediums on the backend side. This also aligns with how the medium checkboxes are rendered and reduces the cost of manual work that needs to be done on the front end side. The downside is, as you pointed out, there would be a bit more, possibly duplicate, logic that finds and checks the existence of the medium specified.

I've added a new field `breadcrumbTitle` to `CollectMediumMetadata.ts` to remove the logic for now. These titles aren't the exact same but should be good enough to represent breadcrumb list data.

src/Apps/Collect/Components/Seo/BreadCrumbList.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from "react"
+import { Meta } from "react-head"
+
+const { APP_URL } = process.env
```
l2succes: it's a bit unsafe to rely on `process.env` here, I would use sharify

PR #1515, opened 2018-11-08
Updating responsive for details for work from partner

oxaudo:
Not sure how it was overlooked before.


artsyit:

Built with commit ee568fd2871cd7911a67bd2e8d6263aa6d8a3489

https://deploy-preview-1515--artsy-reaction.netlify.com

oxaudo:
@alloy, @damassi - Btw... doing global search in reaction for `import { Responsive } from "Utils/Responsive"` - I still see a good amount of those. I guess we can tackle those one by one later, right?

We do `export * from "./DeprecatedResponsive"` in the responsive file so none of that should technically break.

alloy:
Yup, the intent was to do _a_ sweep, but otherwise mostly allow for new work to use the new Media component and others can iterate and fix deprecation warnings.

PR #1523, opened 2018-11-09
🔥blazing fast builds 🔥

zephraph:
_Let's get these builds running faster, faster_.

Let's see how much time we can trim... ⏳

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,7 @@ jobs:
       # Runs jest tests with 4 concurrent workers. Without limiting it to 4
       # workers Jest will spawn many memory hungry workers and ultimately
       # starve the job for memory.
-      - run: yarn test -w 4
+      - run: yarn test -w 5
   update-cache:
     <<: *setup_workspace
     steps:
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 03722ac5d618e79a20c8d2c5aae0b23292ea7182

https://deploy-preview-1523--artsy-reaction.netlify.com

artsyit:

Built with commit 9571fee8ede1c20200cfc93def66e9cc5c72fd93

https://deploy-preview-1523--artsy-reaction.netlify.com

zephraph:
image
image
image
image
image

😱

zephraph:
So I've got builds to the sub-50 second range. 

I did two things: 

1. Update the yarn caching policy to have a faster general case. Essentially there's no more cache fallback and I removed `yarn --check-files`. That means when a dependency changes it'll take longer than it did previously, but the general case is faster
2. Updated the test executors by 2. This will increase the memory consumption but hopefully not to a level that'll start starving other jobs. We'll have to kind of keep an eye on it. Tests are our slowest thing though and that trend will only continue as we write more of them.

orta:
Impressive that it's just moving some strings around 👍

zephraph:
Isn't that most coding at the end of the day? 😄

PR #1524, opened 2018-11-09
[WIP] Import apps from force

alloy:
* Remove freeform SystemContext props, because it’s making it real hard to refactor when arbitrary data can be passed into the context by the host (Force) and used anywhere in Reaction’s tree. Also improves Found typings.
* Defines a ‘app’ API that will allow a host (Force) to specify context props and render its own layout with the Reaction app blocks.
* Adds a dev-server that hosts a bare-bones version of the Artist app.


package.json
```diff
@@ -48,7 +48,8 @@
     "test:watch": "jest --watch --runInBand",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
     "type-check": "tsc --pretty --noEmit",
-    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'"
+    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'",
+    "server:start": "babel-node --extensions .js,.jsx,.ts,.tsx -r node_modules/dotenv/config /Users/eloy/Code/Artsy/reaction/src/Apps/Artist/server.tsx"
```
alloy: Start server

src/Apps/Artist/client.ts
```diff
@@ -0,0 +1,64 @@
+// import { buildClientApp } from "reaction/Artsy/Router/client"
```
alloy: This is a straight import of Force’s client file, so this needs to be made to work, which requires a working webpack setup. Where possible, it would be nice to share config with the storybook setup.

src/Apps/Artist/server.tsx
```diff
@@ -0,0 +1,105 @@
...
+  layoutBlocks
+) => {
+  return await stitch({
+    layout: "data/dev-server.handlebars",
+    config: {
+      styledComponents: true,
+    },
+    blocks: layoutBlocks as any,
+  })
+}
```
alloy: This should probably move out of this file and be part of Reaction’s dev-server setup, instead of being exported.

artsyit:

Built with commit abc6f9945ed3404b03f977b6bd56a9be279e50f8

https://deploy-preview-1524--artsy-reaction.netlify.com

src/Apps/Artist/server.tsx
```diff
@@ -0,0 +1,105 @@
...
+app.get(
+  "/artist/:artistID*",
+  requestHandler((req, res) => ({
+    // initialMatchingMediaQueries: res.locals.sd.IS_MOBILE ? ["xs"] : undefined,
+    // user: req.user && req.user.toJSON(),
+    // isEigen: res.locals.sd.EIGEN,
+    // mediator,
+  }))
+)
+app.listen(3000, () => console.log(`Example app listening on port 300!`))
```
alloy: Quick ‘n dirty

package.json
orta: `/Users/eloy/Code/Artsy/reaction` ^

src/Apps/Artist/server.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * is fine.
+ */
+type RenderLayoutHandler = (
+  req: Request,
+  res: Response,
+  layoutBlocks: LayoutBlocks
+) => Promise<any>
+
+type ReactionApp = (
+  contextBuilder: (req: Request, res: Response) => ContextProps,
```
alloy: Allow’s the host (Reaction dev-server/Force) to specify `SystemContext` props.

src/Apps/Artist/server.tsx
```diff
@@ -0,0 +1,105 @@
...
+  renderLayout?: RenderLayoutHandler
```
alloy: Asks the host (Reaction dev-server/Force) to render its layout and include the head/body/footer blocks that were renderer by the Reaction app.

package.json
alloy: lol

alloy:
Note that this first pass is still very raw and pushed only to handoff to @zephraph to get started on the webpack setup.

data/dev-server.handlebars
```diff
@@ -0,0 +1,14 @@
...
+  {{{css}}}
+</head>
+
+<body>
+  <div id="react-root">
+    {{{body}}}
+  </div>
+</body>
+
+</html>
```
alloy: Not sure which templating language is the simplest, just went with handlebars for now, but absolutely not married to that.

alloy: I actually wondered if it were possible to use React components for this, but currently Stitch doesn’t accept that.

zephraph: Being that stitch already uses handlebars for stuff, I think this is good. 👍

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -44,15 +53,11 @@ export function buildServerApp(config: RouterConfig): Promise<Resolve> {
...
-            // TODO: The docs seem to indicate that if there’s a redirect there
-            //       will not be a status.
-            //       https://github.com/4Catalyzer/found#server-side-rendering
-            status,
-          })
+        if (isRedirect(farceResult)) {
+          resolve(farceResult)
           return
+        } else {
+          farceResult.element
```
damassi: `return farceResult.element`?

alloy: Actually I think this was me just testing the typing, namely that using the isRedirect type guard would narrow the union down correctly. So this can be removed.

src/Apps/Artist/client.ts
alloy: To be able to test and track the performance of these apps on CI

alloy:
I want to integrate reaction apps into force in a way that allows iterative migration and still let force do all the things it still does, eg let force still render its header.

damassi:
@alloy - mind if we close this?

alloy:
Yep, I think that makes sense, just keep in mind it exists for when we talk about this topic again in the future.

PR #1528, opened 2018-11-09
[PATCH Responsive] Fix issue with SSR rendered content not showing after rehydration.

alloy:
https://github.com/artsy/react-responsive-media/pull/18


zephraph:
Purchase flow seems to be rendering good now. 👍

PR #1529, opened 2018-11-09
Offer status page

ashfurrow:
This PR addresses [PURCHASE-366](https://artsyproduct.atlassian.net/browse/PURCHASE-366), paired on with @pepopowitz. It includes an update to styled-system which might be removed from another PR from @pepopowitz. 

This actually fixes an issue with margins in the `<Message>` component, which were previously only ever 20px on desktop. You can see what that looks like here: https://artsy-reaction.netlify.com/?selectedKind=Apps%2FOrder%20Page%2FBuy%20Now%2FStatus&selectedStory=fulfilled%20%28ship%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel Just flagging this as something that was fixed /cc @briansw 

We also tried replacing `<br />` tags with the more semantic meaning like  `<p>` tags (however, `<p>` tags can't contain `<div>`s like `<Serif>`). A discussion for another time 😄

### DIFF
--- a/src/Apps/Order/Routes/Status/index.tsx
+++ b/src/Apps/Order/Routes/Status/index.tsx
@@ -33,72 +33,6 @@ export class StatusRoute extends Component<StatusProps> {
     }
   }
 
-  messageCopy = () => {
-    const { order } = this.props
-    const { state, requestedFulfillment } = order
-    switch (state) {
-      case "SUBMITTED":
-        return (
-          <>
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit 42075c3d930840063e30653649cb00aab92de89a

https://deploy-preview-1529--artsy-reaction.netlify.com

PR #1534, opened 2018-11-12
[BREAKING Responsive] Update react-responsive-media and refactor to use it.

alloy:
Review app is here https://a507075eae78711e893cb1211419c762-1673852081.us-east-1.elb.amazonaws.com


artsyit:

Built with commit fb2cd6a681057021aed4e5cd1e819d06c203236f

https://deploy-preview-1534--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -305,39 +305,36 @@ class Filter extends Component<Props> {
                         )}
                       </Media>
                       <Media greaterThan="xs">
-                        <Sidebar width="25%" mr={2}>
-                          <Filters />
-                        </Sidebar>
+                        <Filters />
```
alloy: One to keep in mind, dimensions relative to parent size should go _outside_ the `Media` component.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -26,31 +26,58 @@ export interface Props extends ContextProps {
   user: User
 }
 
+// TODO: This whole component should be refactored to use less `Media` decisions
```
alloy: These components are slightly too complicated to do immediately, I wanted to get the layout fixed first.

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -305,39 +305,36 @@ class Filter extends Component<Props> {
...
+                      {this.showZeroState ? (
+                        this.renderZeroState({
+                          user,
+                          mediator,
+                        })
+                      ) : (
+                        <ArtworkFilter
+                          artist={this.props.artist}
+                          artistID={this.props.artist.id}
+                          columnCount={xs || sm || md ? 2 : 3}
```
damassi: I wonder if there's a responsive helper, like 
```
columnCount={responsive([2, 3, 4])}
```

alloy: I’m actually wondering if we can’t just use `Media` for this too. Why did we think we couldn’t do that?

The only possible issue I see is having to on-the-fly calculate all the columns, but for that we could indeed make the count a responsive array, and then internally we just keep calculated layouts that we amend on each pagination. That way it’s probably even a bit faster than the current version (as currently it does the calculation on size change. The the problem is also abstracted away and we can work on it in the future in isolation.

What do you think?

alloy: Well, we don’t even need to do the cache ahead optimization for the first step, just abstracting the issue by making the API responsive is enough so users can start doing the better thing.

damassi: It seems like it would work, but I think that pretty soon we'll be using simple horizontal rows. At that point I'm not sure we'll be worrying about column counts and instead deferring to an artwork item width to define the layouts that we need -- 25% = 4, 33% = 3, 50% = 2, etc.

alloy: Makes sense 👍

src/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -60,6 +65,20 @@ export class ArtworkGridContainer extends React.Component<
     }
   }
 
+  columnBreakpoints = memoize(createColumnBreakpoints, isEqual)
```
damassi: The intent might be slightly clearer in the future if `memoize` was renamed to `memoizeOnce`

src/Apps/WorksForYou/WorksForYouFeed.tsx
```diff
@@ -100,7 +100,7 @@ export class WorksForYouFeed extends Component<Props, State> {
 
                 <ArtworkGrid
                   artworks={node.artworksConnection}
-                  columnCount={3}
+                  columnCount={[3]}
```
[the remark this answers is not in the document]
alloy: Agreed, will take a look tomorrow.

src/Components/ArtworkGrid/__tests__/ArtworkGrid.test.tsx
```diff
@@ -28,88 +30,126 @@ const TestContainer = createFragmentContainer(
...
+      ])
+      expect(createColumnBreakpoints([1, 2, 2, 3])).toEqual([
+        [1, ["xs"]],
+        [2, ["sm", "md"]],
+        [3, ["lg", "xl"]],
+      ])
+      expect(createColumnBreakpoints([2, 2, 2, 3])).toEqual([
+        [2, ["xs", "sm", "md"]],
+        [3, ["lg", "xl"]],
+      ])
```
damassi: 👍

alloy:
Ok, ready for final review.

src/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -1,20 +1,25 @@
+import { Flex } from "@artsy/palette"
+import { MediaBreakpointProps } from "@artsy/react-responsive-media/dist/Media"
```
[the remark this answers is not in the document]
alloy: Instead, for now, I’m going to try to keep it as unnecessary as possible to need to import this at all, it also requires a bunch of generics gymnastics.

src/Apps/Artist/Routes/AuctionResults/TableSidebar.tsx
```diff
@@ -25,70 +25,31 @@ interface Props {
...
+              options={SORTS}
+              selected={filters.state.sort}
+              onSelect={filters.setSort}
+            />
 
-              <Spacer mb={2} />
-            </Flex>
-          </Col>
+            <Spacer mb={2} />
+          </Flex>
```
damassi: Nice clean up 👍

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/ArtworkFilterArtworkGrid.tsx
```diff
@@ -15,7 +15,7 @@ import {
 
 interface Props {
   artistID: string
-  columnCount: number
+  columnCount: number | number[]
```
damassi: 👍

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -31,7 +35,7 @@ export function buildServerApp(config: RouterConfig): Promise<Resolve> {
         const historyMiddlewares = [queryMiddleware]
         const resolver = new Resolver(relayEnvironment)
         const render = createRender({})
-        const headTags = []
+        const headTags = [<style type="text/css">{MediaStyle}</style>]
```
damassi: This makes sense, but I wonder if there might be somewhere less hidden?

src/Utils/Responsive/index.tsx
```diff
@@ -17,13 +16,16 @@ const newThemeBreakpoints = {
...
+      "not all and (pointer: coarse), not all and (-moz-touch-enabled: 1)",
   },
 })
 
 export const Media = ReactionMedia.Media
 export const MediaContextProvider = ReactionMedia.MediaContextProvider
+export const createMediaStyle = ReactionMedia.createMediaStyle
+export const SortedBreakpoints = ReactionMedia.SortedBreakpoints
+export const findBreakpointsForWidth = ReactionMedia.findBreakpointsForWidth
+export const valuesWithBreakpointProps = ReactionMedia.valuesWithBreakpointProps
```
damassi: 👍

alloy:
Thanks, buddy! 🙏

src/Artsy/Router/buildServerApp.tsx
alloy: Not sure where. Also keep in mind that it’s easy to find this place if you go to the symbol in the main setup file and use TS to find references to it.

src/Apps/Collect/Components/Filters/FilterContainer.tsx
```diff
@@ -99,7 +99,11 @@ export class FilterContainer extends React.Component<
                 <Mobile filters={filters} />
               </Media>
               <Media greaterThan="xs">
-                <Desktop filters={filters} />
+                {(className, renderChildren) =>
+                  renderChildren && (
+                    <Desktop filters={filters} className={className} />
+                  )
+                }
```
damassi: @alloy - can you explaiun why you would do this over just 

```js
<Media greaterThan="xs">
  <Desktop filters={filters} className={className} />
</Media>
```
?

alloy: Values that are essential to determining the overal layout should either go _outside_ of `Media` usage or the wrapper div should be skipped entirely so that the 25% becomes relative again to the parent of `Media`. (I tried to explain in the commit message too https://github.com/artsy/reaction/pull/1534/commits/cde859bfa0389898118fd9b83552888b6b9713c8.)

There’s probably a better way to refactor this, but I didn’t want to get distracted and so went with the easy route that I knew would work.

alloy: (Skipping the wrapper div is done by using a render prop)

alloy: Here’s a better refactor example https://github.com/artsy/reaction/pull/1534/files#diff-e3efa0e5ac989dfe1a9073a3fc6c8384

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
zephraph: That's going to cause a lot of bugs 😅

src/Apps/Collect/Components/Filters/FilterContainer.tsx
alloy: Put differently, just as with outer margins, a child component should not make overall layout decisions, instead the parent that composes the various components should be in charge.

zephraph:
wheeeeewww... LET'S DO THIS!

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
damassi: I presume these overhead issues can be smoothed out?

Issue #1535, opened 2018-11-13
Line-height is extremely large on a few elements after responsive typography update


pepopowitz:
When we merged [this PR](https://github.com/artsy/reaction/pull/1531), we ended up with a couple of elements that suddenly had very large line-heights. See [here](https://github.com/artsy/reaction/pull/1531#issuecomment-437709780) and [here](https://github.com/artsy/reaction/pull/1531#issuecomment-437709971) for screenshots.

In looking at these examples, it looks like the "extra tall" components are all specifying a `lineHeight={18}` prop, like this:

```
  <Serif mb={1} size="3t" color="black100" lineHeight={18}>
    Country
  </Serif>
```

I suspect that this broke when we updated the styled-system dependency - we noticed similar issues in the development of #1531, where it seemed to switch from a default of pixels to a default of ems. 

We should either add back support for `lineHeight={18}`, or update these elements to specify `lineHeight="18px"`.

pepopowitz:
@alloy @zephraph Do either of you have thoughts on whether `lineHeight={18}` should be supported as it was previously, or these lines should be changed to `lineHeight="18px"`?

zephraph:
We shouldn't be explicitly setting the lineHeight like this. If it _has_ to be set, we should do what you said in the latter part of your suggesting... just set it to `18px`. This is a code smell to me though.

alloy:
I’m no CSS expert, but isn’t `line-height` also used to vertically center text in a box? Anyways, in terms of the actual fix I’ll defer to you two 👍

pepopowitz:
In these examples, `line-height` is being used to adjust the height of the label, to match the `line-height` set by the other input fields. They are different, because....

Most input fields on the forms use the `<Input>` component - which uses the [Garamond s17 font specified in src/assets/fonts.tsx](https://github.com/artsy/reaction/blob/master/src/Assets/Fonts.tsx#L20) for the label. That gives them a line height of 1.1em. 

These one-off fields render a `<Serif>` component for the label. They specify `size="3t"`, which gives them a matching font-size, but a non-matching line-height (20px instead of 18px). 

I'm going to change these three examples to specify line-height of "1.1em", to match the `<Input>` component labels more precisely.

zephraph:
This was fixed a while back. Gonna close the issue.

PR #1538, opened 2018-11-13
Fix elements that were specifying lineHeight={18}; specify lineHeight as 1.1em instead.

pepopowitz:
This PR addresses #1535. 

I chose 1.1em instead of 18px, because that is the line-height of the other input fields that these are intending to match.


artsyit:

Built with commit 216bb00bfeee92f3066c1f044e195fa05ac14dff

https://deploy-preview-1538--artsy-reaction.netlify.com

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -102,7 +102,7 @@ export class AddressForm extends React.Component<
 
         <TwoColumnSplit>
           <Flex flexDirection="column" pb={1}>
-            <Serif mb={1} size="3t" color="black100" lineHeight={18}>
+            <Serif mb={1} size="3t" color="black100" lineHeight="1.1em">
```
damassi: Since line height is baked [directly into the typography component size](https://github.com/artsy/palette/blob/master/src/Theme.tsx#L241-L244) it should rarely be overwritten.

zephraph: We generally don't use em for the units. If this was supposed to be `18px`, just hard code it to that.

pepopowitz: @damassi The line height is only being specified here because the majority of the form doesn't use the Typography components. Most of the form uses [Input components](https://github.com/artsy/reaction/blob/master/src/Components/Input.tsx#L58), which [render the input labels](https://github.com/artsy/reaction/blob/master/src/Components/Input.tsx#L253) using font sizes that were [specified here in reaction](https://github.com/artsy/reaction/blob/master/src/Assets/Fonts.tsx#L20. 

The line heights defined here in reaction are different than the ones baked into Typography; hence these three overrides.

zephraph: Oh, I see. Sorry about that! Yeah, looks good 👍

pepopowitz: @zephraph I can change it to 18px, but all of the fields in the affected forms are using em for units. See the trail of links in my comment above.

That's why I opted for 1.1em instead - the other inputs are all using 1.1em....these fields specifying 18px are outliers.

damassi:
(Merging per slack convo; cc @sweir27)

PR #1537, opened 2018-11-13
Fix instagram embeds

eessex:
Addresses https://artsyproduct.atlassian.net/browse/GROW-979

Looks like instagram has changed their api's cors handling, so `jsonp` was no longer working and returned CORB errors.

Instagram links have been updated to use `isomorphic-fetch`, twitter still needs to use `jsonp`.

PR #1546, opened 2018-11-14
Collections groupings

eessex:
Addresses https://artsyproduct.atlassian.net/browse/GROW-990,
https://artsyproduct.atlassian.net/browse/GROW-986,
and https://artsyproduct.atlassian.net/browse/GROW-999

- Updates CollectionApp query to hit `marketingCategories` (grouped) endpoint)
- Updates `CollectionGrid` to use new `Media` component
- Crops collection headerImage to display as square
- Links `/collection/:id` header category back to respective category


src/Apps/Collections/CollectionsApp.tsx
```diff
@@ -28,13 +28,22 @@ export class CollectionsApp extends Component<CollectionsAppProps> {
             justifyContent="space-between"
             alignItems="flex-end"
           >
-            <Serif size="8">Collections</Serif>
+            <Serif size="8">
+              <h1 className="title">Collections</h1>
```
l2succes: is the className used to set some styles here? if that's the case I would rather use styled-components.

```javascript
const Title = styled.h1`
  ...
`
```

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -1,52 +1,63 @@
...
+            const imageUrl =
+              collection.headerImage &&
+              crop(collection.headerImage, {
+                width: 50,
+                height: 50,
+              })
+
+            return (
+              <Flex width={["100%", "30%"]} flexDirection="column" key={index}>
+                <Media at="xs">{index === 0 && <Separator />}</Media>
```
l2succes: 👍

src/Apps/Collections/CollectionsApp.tsx
eessex: added for SEO based on Nicolette's recommendation here, not used for actual syles - https://artsyproduct.atlassian.net/browse/GROW-1004

eessex: maybe it doesn't actually need a class though, now that i look closer

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -1,52 +1,63 @@
...
+      <Box pb={80} id={slugify(name)}>
+        <Sans size="3" weight="medium" pb={15}>
+          {name}
+        </Sans>
+
+        <Flex flexWrap="wrap" justifyContent="space-between">
+          {collections.map((collection, index) => {
+            const imageUrl =
+              collection.headerImage &&
+              crop(collection.headerImage, {
```
l2succes: I would not set a width here to preserve the correct aspectRatio

eessex: The Entity component sizes the image to 50x50, and expects a square file -- otherwise they are squeezed to a square: https://artsy.slack.com/archives/CD0B3F620/p1542210908007300?thread_ts=1542146808.001600&cid=CD0B3F620

yuki24:
This also seems to have the same build issue as https://github.com/artsy/reaction/pull/1537#issuecomment-438889005. Do we need Elan in this repo?

artsyit:

Built with commit 640b8cd66d03b64035e34f0a16f51b8ca8d6e410

https://deploy-preview-1546--artsy-reaction.netlify.com

src/Apps/Collect/Components/Collection/Header/index.tsx
```diff
@@ -4,6 +4,7 @@ import React, { Component } from "react"
 import styled from "styled-components"
 import { ReadMore } from "Styleguide/Components/ReadMore"
 import { Col, Grid, Row } from "Styleguide/Elements/Grid"
+import { slugify } from "underscore.string"
```
alloy: Not sure how this passed the `type-check` CI phase, locally I’m getting a failure because this package was not added as a dependency to Reaction.

alloy: Added in https://github.com/artsy/reaction/pull/1597

orta: @eessex want to pair on removing this dependency? The slug for a collection should come from the API response, trying to re-create it from the client will end up inconsistent

eessex: This is an id for a collection category, which doesn't actually have a slug -- I have the dependency in my local reaction right now?

eessex: I'm open to adding unique ids to the category, but will require some additional work in Kaws cc @l2succes

alloy: > I have the dependency in my local reaction right now?

Might have been a transitive dependency of another dependency that got updated/removed? Force has a hard dependency on it, so the code does work when used inside Force, but Reaction itself did not have one so not sure how else that could have worked… Unless maybe you ran `npm install underscore.string`?

PR #1548, opened 2018-11-15
Add analytics tracking for "read more" in Collection pages

yuki24:
finishes https://artsyproduct.atlassian.net/browse/GROW-983

This PR adds `{ "context_module": "CollectionDescription", "type": "button" }` to the existing analytics payload. Here is the actual payload sent to Segment:

analytics

### DIFF
--- a/src/Apps/Collect/CollectionApp.tsx
+++ b/src/Apps/Collect/CollectionApp.tsx
@@ -1,5 +1,7 @@
 import { Box } from "@artsy/palette"
 import { CollectionApp_collection } from "__generated__/CollectionApp_collection.graphql"
+import { track } from "Artsy/Analytics"
+import * as Schema from "Artsy/Analytics/Schema"
 import { FrameWithRecentlyViewed } from "Components/FrameWithRecentlyViewed"
 import { HttpError } from "found"
 import React, { Component } from "react"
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 4ee198cea0c1c8c3c4acb417771ee817ae7dcd6a

https://deploy-preview-1548--artsy-reaction.netlify.com

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -14,6 +16,9 @@ interface CollectionAppProps {
   collection: CollectionApp_collection
 }
 
+@track({
```
xtina-starr: I'm going to go ahead and merge this but I was curious as to why you didn't add the decorator to the [`<CollectionHeader>`](https://github.com/artsy/reaction/blob/master/src/Apps/Collect/Components/Collection/Header/index.tsx) component?

yuki24: Given that the `<ReadMore>` is a generic component, I was thinking what actually adds context is a page.

 * "Where did this event for `<ReadMore>` come from?"
 * "It's on the page X"

So it seemed most natural to me to specify the context in the component that represents the entire page (which in this case is `<CollectionApp>`). However, there would be naming conflict if the child components of it have a decorator that overrides it, so we may have to be careful for how we use the `track` decorator.

xtina-starr: I'm revisiting this since it was mentioned recently in [slack](https://artsy.slack.com/archives/C9SATFLUU/p1544469914156700). This definitely feels like something we may want to sync up with the analytics team about because I could also see a case where there may be more than one read more component on a page and it would be difficult to identify which one was clicked.

PR #1549, opened 2018-11-15
[FIX display] Fixes an issue where display logo is not centered properly

yuki24:
fixes https://artsyproduct.atlassian.net/browse/GROW-998

It seems that the logo image is already centered and the `width: 100%` could be removed. There may be some historical reasons why this is there, but this seems to be working without it on all the responsive sizes we have support for, with logo images in a different aspect ratio.

## Before

screen shot 2018-11-15 at 17 44 02

## After

screen shot 2018-11-15 at 17 42 56


eessex:
This looks good besides the ❌

PR #1553, opened 2018-11-15
Implement action icons

iskounen:
JIRA Ticket:
- https://artsyproduct.atlassian.net/browse/DISCO-459

Changes:
- Decoupled the save and share icons from the image carousel.
- Modified the share widget to match the designs in Zeplin.
- Created a separate story for the image carousel.

Screenshot:
disco-459


artsyit:

Built with commit a15f550853d739d56889bf8efa88302e416c9899

https://deploy-preview-1553--artsy-reaction.netlify.com

.env.oss
```diff
@@ -9,3 +9,5 @@ WEBPACK_DEVTOOL=cheap-module-eval-source-map
 # 3. and get the values `sd.CURRENT_USER.id` and `sd.CURRENT_USER.accessToken`
 USER_ID=
 USER_ACCESS_TOKEN=
+
+APP_URL=
```
damassi: 👍

src/Apps/Artwork/Components/ArtworkImages.tsx
```diff
@@ -5,13 +5,20 @@ import { renderWithLoadProgress } from "Artsy/Relay/renderWithLoadProgress"
...
 interface ArtworkImagesProps {
   artwork: ArtworkImages_artwork
 }
 
 const ArtworkImages: React.SFC<ArtworkImagesProps> = props => {
-  return <ImageCarousel images={props.artwork.images} />
+  return (
+    <ImageCarousel
+      images={props.artwork.images}
+      actions={[<SaveButton />, <ShareButton />]}
```
damassi: This is totally fine but note that another (slightly more flexible) way of writing this would be to use a fragment:
```jsx
actions={
  <>
     <SaveButton />
     <ShareButton />
  </>
}
```
That way we're staying inside of more common JSX idioms.

src/Apps/Artwork/Components/ImageCarousel.tsx
```diff
@@ -130,6 +130,7 @@ const ActionButtonsContainer = styled.div`
 
 interface ImageCarouselProps {
   images: any // Array<{ uri: string; aspectRatio: number }>
+  actions?: ReactChild[]
```
damassi: `actions?: JSX.Element`

src/Apps/Artwork/Components/SaveButton.tsx
```diff
@@ -0,0 +1,24 @@
+import { color, Flex } from "@artsy/palette"
+import Icon from "Components/Icon"
+import React from "react"
+import styled from "styled-components"
+
+const SaveIconWrapper = styled(Flex)`
+  position: relative;
+  justify-content: center;
+`
```
damassi: While there is value in idiomatic styled components naming, it's a bit simpler to just use flex directly and inline the props: 

```jsx
<Flex justifyContent='center' position='relative'>
  <SaveIcon name="heart" />
</Flex>
```

src/Apps/Artwork/Components/SaveButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const SaveIcon = styled(Icon)`
+  color: ${color("black100")};
+  cursor: pointer;
+`
+
+export class SaveButton extends React.Component {
```
damassi: When not using state best to just use an SFC:

```jsx
export const SaveButton: React.SFC = (props) => { ... }
```

src/Apps/Artwork/Components/ArtworkImages.tsx
iskounen: I wasn't aware of fragments, so I initially tried to do this by wrapping the buttons in a div tag. However, I decided to pass these buttons as a prop so that we could pass more than action buttons as children of an image carousel if it becomes necessary in the future.

src/Apps/Artwork/Components/Share.tsx
```diff
@@ -1,22 +1,34 @@
-import { Flex, Sans, Separator } from "@artsy/palette"
+import { color, Flex, Sans, Separator } from "@artsy/palette"
 import { IconName } from "Assets/Icons"
 import Icon from "Components/Icon"
 import React from "react"
 import styled from "styled-components"
 
 interface ShareProps {
   url: string
+  handleClose?: () => void
```
damassi: Conventionally we use `handleClose` inside of a class for attaching to an action, and then the external "props" API is `onClose`.

src/Apps/Artwork/Components/ArtworkImages.tsx
iskounen: 👍

src/Apps/Artwork/Components/ImageCarousel.tsx
iskounen: 👍

src/Apps/Artwork/Components/Share.tsx
```diff
@@ -54,33 +73,38 @@ export class Share extends React.Component<ShareProps> {
...
-              type="text"
-              readOnly
-              value={this.props.url}
-              innerRef={input => (this.input = input)}
-              onClick={this.selectURL}
-            />
-          </SansGrow>
-          <a onClick={this.selectURL}>
+      <Container>
+        <CloseIconWrapper>
```
damassi: At this point we should avoid manually defining pixel values and instead defer to our [palette / styled-system theme](https://github.com/artsy/palette/blob/master/src/Theme.tsx#L122). This can be simplified to 

```jsx
<Box top={1} right={1}>
  <CloseIcon name="close" onClick={this.props.onClose} />
</Box>
```

src/Apps/Artwork/Components/SaveButton.tsx
iskounen: Good call.

src/Apps/Artwork/Components/ShareButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+import Icon from "Components/Icon"
+import React from "react"
+import { data as sd } from "sharify"
+import styled from "styled-components"
+import { Share } from "./Share"
+
+const ShareIconWrapper = styled(Flex)`
+  position: relative;
+  justify-content: center;
+`
```
damassi:
```jsx
<Flex position='relative' justifyContent='center'> ... </Flex>
```

src/Apps/Artwork/Components/ShareButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const ShareIcon = styled(Icon)`
+  color: ${color("black100")};
+  cursor: pointer;
+`
+
+const ShareWidgetWrapper = styled.div`
+  position: absolute;
+  bottom: 40px;
+`
```
damassi:
```js
<Box bottom={4} position='absolute'> ... </Box>
```

src/Apps/Artwork/Components/ShareButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface State {
+  isOpen: boolean
+}
+
+export class ShareButton extends React.Component<any, State> {
```
damassi: `any` -> `null`

src/Apps/Artwork/Components/ShareButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+  private close: () => void
```
damassi: This can be removed and refactored as below

src/Apps/Artwork/Components/ShareButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(props: any) {
+    super(props)
+
+    this.close = () => {
+      this.setState({ isOpen: false })
+    }
```
damassi: Rather than creating functions in the constructor its better to define them directly on the class.

src/Apps/Artwork/Components/ShareButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+    this.state = {
+      isOpen: false,
+    }
+
+    this.open = this.open.bind(this)
+  }
+
+  open() {
+    this.setState({ isOpen: true })
+  }
```
damassi: Since we use ES6 class properties we can take advantage of arrow functions and how they bind to the outer scope and remove the constructor completely (as well as `bind`): 

```jsx
export class ShareButton extends React.Component<null, State> {
  state = {
    isOpen: false
  }

  open = () => {
     this.setState({ isOpen: true })
  }

  close = () => {
     this.setState({ isOpen: false })
  }

  render() { ... } 
}
```

src/Apps/Artwork/Components/__stories__/ArtworkImages.story.tsx
```diff
@@ -3,76 +3,20 @@ import { storiesOf } from "storybook/storiesOf"
...
+  return (
+    <>
+      <Section title="Multiple images">
+        <Col sm="8">
+          <ArtworkImagesQueryRenderer artworkID="andy-warhol-lenin-fs-ii-dot-402-1" />
+        </Col>
+      </Section>
+      <Section title="Single image">
+        <Col sm="8">
+          <ArtworkImagesQueryRenderer artworkID="pablo-picasso-david-et-bethsabee" />
```
damassi: 👍

src/Apps/Artwork/Components/ShareButton.tsx
iskounen: @damassi, the refactored code seems to have been cut out.

iskounen:
@damassi thank you so much for your comments!! I'm sorry it took longer than I had estimated, but this is great feedback.

I agree that the save and share buttons should be a single component, so I'll try to make that happen once I've made the corrections you recommended above.

iskounen:
I made the changes recommended above. I also merged the Save and Share buttons (and stories) into a shared .tsx file, but stopped short of moving shared code into a "parent" component due to time constraints.

alloy:
Aw yeah, nice one @iskounen 👏 🍾

PR #1564, opened 2018-11-20
[PATCH collections] collections UI qa

xtina-starr:
Address QA items for collections page

screen shot 2018-11-19 at 5 00 47 pm

### DIFF
--- a/src/Styleguide/Components/EntityHeader.tsx
+++ b/src/Styleguide/Components/EntityHeader.tsx
@@ -1,4 +1,12 @@
-import { Avatar, Box, Flex, Sans, Serif } from "@artsy/palette"
+import {
+  Avatar,
+  Box,
+  Flex,
+  Sans,
+  Serif,
+  SerifProps,
+  SerifSize,
+} from "@artsy/palette"
 import React, { SFC } from "react"
 import styled, { css } from "styled-components"
 
@@ -8,6 +16,8 @@ interface EntityHeaderProps {
   initials?: string
   meta?: string
   name: string
+  nameFontSize?: SerifSize
+  nameFontWeight?: SerifProps["weight"]
   showFollow?: boolean
   FollowButton?: JSX.Element
 }
@@ -17,6 +27,8 @@ export const EntityHeader: SFC<EntityHeaderProps> = ({
   imageUrl,
   initials,
   name,
+  nameFontSize,
+  nameFontWeight,
   meta,
   FollowButton,
 }) => {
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit eafe724b2d9c19fa2eff0828c59d0c7c1b5ce40d

https://deploy-preview-1564--artsy-reaction.netlify.com

src/Styleguide/Components/EntityHeader.tsx
```diff
@@ -35,7 +47,11 @@ export const EntityHeader: SFC<EntityHeaderProps> = ({
       )}
 
       <Flex flexDirection="column" justifyContent="center" width="100%">
-        <Serif size="3" weight="semibold" color="black100">
+        <Serif
+          size={nameFontSize ? nameFontSize : "3"}
+          weight={nameFontSize ? nameFontWeight : "semibold"}
```
zephraph: Are these configurations for repsonsiveness or does it just need be changed based on the context in which the component renders?

xtina-starr: @zephraph it's definitely based on context ([the QA ticket for context](https://artsyproduct.atlassian.net/browse/GROW-1014)). We were a little iffy about changing the component. I went ahead and opened the PR to get a sense of what we should be doing in these cases. 

We also thought that Owen might not be aware that is a component used globally which might have impacted he's decision to make these minor UI changes. Since Owen is out, @nicoleyeo do you have any input?

cc: @eessex

nicoleyeo: Hi — hmmm I do not recommend altering "entity-header" here. It's visually similar but is a component with a very different use.

If I understand the context of this component correctly (a small representation of a collection), this should be called something like "entity-collection" and is actually more similar to "entity-artist" or "entity-gallery" which we have not formalized yet in Palette (or Reaction, really).

For simplicity, I'd recommend creating a new "CollectionEntity" (in Reaction) to be safe and avoid breaking other areas. @xtina-starr lmk if this needs further detail, happy to discuss.

(@zephraph note for us to formalize "Entity" components in Palette soon).

Question: Is there a specific reason why we are building this in Reaction and not in Palette?

owendodd:
Just wanted to weigh in quickly! Didn’t realize this was a global component, so def disregard my minor UI tweaks. I think it makes sense to just keep existing components as much as possible, as we will be iterating on this page in bigger ways in the future.

PR #1570, opened 2018-11-20
[PATCH logging] Add `ErrorWithMetadata` class and use when logging errors with metadata

williardx:
This is a somewhat in-the-weeds PR, so please let me know if something doesn't make sense!

### Problem
In https://github.com/artsy/reaction/pull/1479, I created a [method to send errors to Sentry](https://github.com/artsy/reaction/blob/30b0755443b48f9c64b8000ce2312d5de54548d0/src/Utils/errors.tsx#L18) and added that method to the newly-created [`logger.error` method](https://github.com/artsy/reaction/blob/30b0755443b48f9c64b8000ce2312d5de54548d0/src/Utils/logger.ts#L16). The idea was that if you're logging an error you likely want to send that error to an error-capturing service anyway.

The issue is that the code doesn't capture mutation errors from Metaphysics, and we need it to do that. On mutation error, we receive an object that represents an error and contains multiple relevant fields for the error. Currently we just pass in that object to the logger, but since [it's not actually an `Error` object we don't end up sending it to Sentry](https://github.com/artsy/reaction/blob/30b0755443b48f9c64b8000ce2312d5de54548d0/src/Utils/logger.ts#L17). We want to convert this object into an error (which Sentry expects), log it and send the error and extra fields to Sentry so that we can properly track mutation errors.

### Solution

Turns out this isn't straightforward! Here are some possible approaches and their pitfalls:

1. Create a new `Error` and pass in the stringified error object from Metaphysics as the message.

This will work, but we won't be taking advantage of Sentry's ability to associate the metadata with the error since all the metadata will be in the message. We'll have to parse JSON from the error message in Sentry in order to understand what the error is, which makes the data more difficult to deal with.

2. Create a new `Error` from what we get back from Metaphysics and pass the error along with the metadata as an `ErrorInfo` object to the logger.

`ErrorInfo` is an object that we pass along to Sentry to associate metadata with that error. This sounds good, but is made difficult by the fact that there's no good way to pick out the `ErrorInfo` in the logger input (which doesn't have any limitations on what it can accept). While `ErrorInfo` _does_ have an interface that we currently use to [figure out which of the inputs implements `ErrorInfo`](https://github.com/artsy/reaction/blob/30b0755443b48f9c64b8000ce2312d5de54548d0/src/Utils/logger.ts#L18), we don't actually need it like I originally thought. Sentry doesn't have any limitations on what we pass in as extra fields, so there's no need to check. `ErrorInfo` can  be any object, and so with that change there's no way to distinguish between `ErrorInfo` and another object in the logger input.

3. Create a custom `ReactionError` class and add metadata as a property of that object. Get rid of the `ErrorInfo` object, and use the metadata on the `ReactionError` object when setting the metadata for Sentry. Use this error for all errors we want to send to Sentry, including our mutation errors.

This could work, but it feels a little strange. It's something new that other developers will have to think about when doing error handling. [`componentDidCatch` in `ErrorBoundary` returns an error and metadata object separately](https://github.com/artsy/reaction/blob/30b0755443b48f9c64b8000ce2312d5de54548d0/src/Components/ErrorBoundary.tsx#L11), so we'd have to convert that error into a new `ReactionError` with the associated metadata to properly capture everything, which is awkward.

4. Don't send an error to Sentry when we call `logger.error`, and instead send errors to Sentry separately with `sendErrorToService`.

This works, too, but is also awkward because we probably want to call `sendErrorToService` whenever we call `logger.error`. Still, this is nice because we don't have to worry about what's happening in `logger.error` in order to send an error to Sentry.

**After all that, option (4) seems like the most reasonable option to me**, despite the fact that I don't like having to use `sendErrorToService` all over the place. We could create a new wrapper method `logAndSendErrorToService` to get around that. It does make the logger a somewhat useless wrapper for `console`, but I don't see an immediate reason to get rid of it.

Thoughts / other ideas?

### DIFF
--- a/src/Apps/Order/Routes/Offer/index.tsx
+++ b/src/Apps/Order/Routes/Offer/index.tsx
@@ -17,6 +17,7 @@ import {
 } from "react-relay"
 import { Col, Row } from "Styleguide/Elements/Grid"
 import { HorizontalPadding } from "Styleguide/Utils/HorizontalPadding"
+import { ErrorWithMetadata } from "Utils/errors"
 import { get } from "Utils/get"
 import createLogger from "Utils/logger"
 import { Media } from "Utils/Responsive"
@@ -109,7 +110,7 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
   }
 
   onMutationError(errors, errorModalTitle?, errorModalMessage?) {
-    logger.error(errors)
+    logger.error(new ErrorWithMetadata(errors.message || errors.data, errors))
     this.setState({
       isCommittingMutation: false,
       isErrorModalOpen: true,
--- a/src/Apps/Order/Routes/Offer/index.tsx
+++ b/src/Apps/Order/Routes/Offer/index.tsx
@@ -98,7 +98,12 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
             } = data
 
             if (orderOrError.error) {
-              this.onMutationError(orderOrError.error)
+              this.onMutationError(
+                new ErrorWithMetadata(
+                  orderOrError.error.code,
+                  orderOrError.error
+                )
+              )
             } else {
               this.props.router.push(`/orders/${this.props.order.id}/shipping`)
             }
@@ -109,8 +114,8 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
     })
   }
 
-  onMutationError(errors, errorModalTitle?, errorModalMessage?) {
-    logger.error(new ErrorWithMetadata(errors.message || errors.data, errors))
+  onMutationError(error, errorModalTitle?, errorModalMessage?) {
+    logger.error(error)
     this.setState({
       isCommittingMutation: false,
       isErrorModalOpen: true,
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -40,6 +40,7 @@ import {
 import { injectStripe, ReactStripeElements } from "react-stripe-elements"
 import { Col, Row } from "Styleguide/Elements/Grid"
 import { HorizontalPadding } from "Styleguide/Utils/HorizontalPadding"
+import { ErrorWithMetadata } from "Utils/errors"
 import createLogger from "Utils/logger"
 import { Media } from "Utils/Responsive"
 
@@ -414,7 +415,7 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
   }
 
   private onMutationError(errors, errorModalMessage?) {
-    logger.error(errors)
+    logger.error(new ErrorWithMetadata(errors.message || errors.data, errors))
     this.setState({
       isCommittingMutation: false,
       isErrorModalOpen: true,
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -319,7 +319,7 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
             })
           } else {
             if (errors) {
-              errors.map(err => this.onMutationError(err))
+              errors.forEach(this.onMutationError.bind(this))
             } else {
               const mutationError = creditCardOrError.mutationError
               this.onMutationError(
@@ -374,7 +374,7 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
             this.props.router.push(`/orders/${this.props.order.id}/review`)
           } else {
             if (errors) {
-              errors.map(err => this.onMutationError(err))
+              errors.forEach(this.onMutationError.bind(this))
             } else {
               const orderError = orderOrError.error
               this.onMutationError(
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -318,11 +318,15 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
               creditCardId: creditCardOrError.creditCard.id,
             })
           } else {
-            this.onMutationError(
-              errors || creditCardOrError.mutationError,
-              creditCardOrError.mutationError &&
-                creditCardOrError.mutationError.detail
-            )
+            if (errors) {
+              errors.map(err => this.onMutationError(err))
+            } else {
+              const mutationError = creditCardOrError.mutationError
+              this.onMutationError(
+                new ErrorWithMetadata(mutationError.message, mutationError),
+                mutationError.detail
+              )
+            }
           }
         },
         onError: this.onMutationError.bind(this),
@@ -369,7 +373,14 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
           if (orderOrError.order) {
             this.props.router.push(`/orders/${this.props.order.id}/review`)
           } else {
-            this.onMutationError(errors || orderOrError)
+            if (errors) {
+              errors.map(err => this.onMutationError(err))
+            } else {
+              const orderError = orderOrError.error
+              this.onMutationError(
+                new ErrorWithMetadata(orderError.code, orderError)
+              )
+            }
           }
         },
         onError: this.onMutationError.bind(this),
@@ -414,8 +425,8 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
     )
   }
 
-  private onMutationError(errors, errorModalMessage?) {
-    logger.error(new ErrorWithMetadata(errors.message || errors.data, errors))
+  private onMutationError(error, errorModalMessage?) {
+    logger.error(error)
     this.setState({
       isCommittingMutation: false,
       isErrorModalOpen: true,
--- a/src/Components/Payment/PaymentForm.tsx
+++ b/src/Components/Payment/PaymentForm.tsx
@@ -222,7 +222,7 @@ class PaymentForm extends Component<PaymentFormProps, PaymentFormState> {
             window.scrollTo(0, 0)
           } else {
             if (errors) {
-              errors.map(err => this.onMutationError(err))
+              errors.forEach(this.onMutationError.bind(this))
             } else {
               const mutationError = creditCardOrError.mutationError
               this.onMutationError(
--- a/src/Components/Payment/PaymentForm.tsx
+++ b/src/Components/Payment/PaymentForm.tsx
@@ -221,11 +221,15 @@ class PaymentForm extends Component<PaymentFormProps, PaymentFormState> {
             this.cardElement && this.cardElement.cardInputElement.clear()
             window.scrollTo(0, 0)
           } else {
-            this.onMutationError(
-              errors || creditCardOrError.mutationError,
-              creditCardOrError.mutationError &&
-                creditCardOrError.mutationError.detail
-            )
+            if (errors) {
+              errors.map(err => this.onMutationError(err))
+            } else {
+              const mutationError = creditCardOrError.mutationError
+              this.onMutationError(
+                new ErrorWithMetadata(mutationError.message, mutationError),
+                mutationError.detail
+              )
+            }
           }
         },
         onError: this.onMutationError.bind(this),
@@ -261,8 +265,8 @@ class PaymentForm extends Component<PaymentFormProps, PaymentFormState> {
     )
   }
 
-  private onMutationError(errors, errorModalMessage?) {
-    logger.error(new ErrorWithMetadata(errors.message || errors.data, errors))
+  private onMutationError(error, errorModalMessage?) {
+    logger.error(error)
     this.setState({
       isCommittingMutation: false,
       isErrorModalOpen: true,
--- a/src/Components/Payment/PaymentForm.tsx
+++ b/src/Components/Payment/PaymentForm.tsx
@@ -15,6 +15,7 @@ import React, { Component } from "react"
 import { commitMutation, graphql, RelayProp } from "react-relay"
 import { injectStripe, ReactStripeElements } from "react-stripe-elements"
 import { ConnectionHandler } from "relay-runtime"
+import { ErrorWithMetadata } from "Utils/errors"
 import createLogger from "Utils/logger"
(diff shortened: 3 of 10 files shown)


artsyit:

Built with commit ee1705a6f2a4d4d239b2e9c6c7a6d53be9d06030

https://deploy-preview-1570--artsy-reaction.netlify.com

src/Utils/__tests__/errors.test.ts
```diff
@@ -1,19 +1,9 @@
...
-    it("returns true if an object implements the ErrorInfo interface", () => {
-      const errorInfo = { componentStack: "someString", anotherKey: 0 }
-      expect(isErrorInfo(errorInfo)).toBe(true)
-    })
-    it("returns false if an object does not implement the ErrorInfo interface", () => {
-      const notErrorInfo = { anotherKey: 0 }
-      expect(isErrorInfo(notErrorInfo)).toBe(false)
-    })
-  })
   describe("#sendErrorToService", () => {
```
dleve123: thoughts on adding a separate test case that tests that `setExtra` gets called as many times as there are keys in `errorInfo` object passed in, and restructuring this test case to be focused on expecting `captureException` to be called?

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -110,6 +111,7 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
 
   onMutationError(errors, errorModalTitle?, errorModalMessage?) {
     logger.error(errors)
+    sendErrorToService(new Error(errors.code), errors)
```
damassi: Why not add a config argument to `logger.error` and send there? By circumventing the logger to deal with localized issue it creates drift for other devs leading to less error coverage. Could do something like 

```
logger.error(errors, { sendErrorToService: true /* the default /* })
```
which then wraps the above call-set. (Not suggesting the above exactly, but you get what I mean :) With the way its been refactored there are now three `errors` objects being passed about which increases complexity and mental overhead, essentially crippling the function of the logger. If a user is already importing the logger it should be assumed to be the canonical  location for extending functionality, regardless of local conditions, and extended via configuration.

src/Utils/__tests__/errors.test.ts
dleve123: Also, I think we should test the `if` on L11 – maybe we can stub `process.env.NODE_ENV` somehow?

williardx: Yep, that all makes sense to me -- I just deleted these so that the code would compile. I'll add unit tests once there's consensus this is the right direction.

src/Apps/Order/Routes/Offer/index.tsx
williardx: Hmm, I don't totally follow you — could give a more detailed example of how a config would help here? The problem from my perspective is that we want to send everything through the logger but the logger can’t always determine what it should use to call `sendErrorToService`, and it doesn’t seem like a config would help that at first glance.

williardx: I'm still missing something, here. How does this help with the problem that we can't figure out with certainty which input contains error metadata?

sweir27: I think @damassi 's approach makes sense too-- but @williardx what do you mean by your last point?

williardx:
@alloy I haven't looked at that particular case, so I'm glad you brought that up. `creditCardOrError.mutationError` should be an object (i.e. formatted Gravity error from Metaphysics), but I'm not 100% certain what `errors` will be. I was previously assuming I'd generally get back an object from a mutation error, but maybe that's not the case?

alloy:
I’d like to avoid adding any further configuration, the API is already setup in such a way that you can tell it to report the error to Sentry by wrapping non-error objects in an `Error`. So my suggestion is to try to work out the details of _why_ an ‘error’ is not being reported that you do would want to be reported and let that guide the design.

src/Apps/Order/Routes/Offer/index.tsx
williardx: Here's a better description of the input disambiguation problem I've been mentioning: 

A mutation fails, and we get back a response that looks like the following: `errors = { code: 'some_code', type: 'some_type', data: { ... } }`. In order for the logger to recognize this as an error, we need to wrap this in an `Error` object. However, `Error` doesn't automatically support adding the extra metadata that we have for this error -- its arguments are a message, filename and line number. 

So, we can create a new error, say `Error(errors.code)`, and the logger will find the error. We pass in the rest of metadata as a separate "error info" object (`errorInfo`), giving us: `logger.error(new Error(errors.code), errorInfo)`.

However, the logger doesn't have any restrictions as to what input it can take. We can find an error by checking for objects that are an instance of `Error`, but we can't do the same for our error info object since it's just a plain object. Someone could end up passing in multiple objects to the logger: `logger.error(new Error(errors.code), errorInfo, "some message", someOtherObj)`. There's no way to know that `errorInfo` contains the metadata we want and that `someOtherObj` doesn't because we don't know ahead of time what `errorInfo` will contain. I don't know why we might have multiple objects there, but I don't want to discount the possibility or restrict the logger. (Previously we could tell because we were looking for objects that implemented a particular `ErrorInfo` interface, but it turns out we don't need that and that we should just let error info objects be generic.)

From what I see, adding a config doesn't actually solve ☝️because knowing whether or not we need to send an error isn't the problem -- it's figuring out if there's metadata to associate with the error.

alloy: It sounds like we need to define a `Error` subclass for these mutation errors that override the `message` property to return the right information. The `logger.error` implementation shouldn’t need to care about anything else than the `message` property.

Here’s an example I recently wrote https://github.com/1stdibs/relay-mock-network-layer/blob/master/RelayMockNetworkLayerError.js

williardx: Do you think this `Error` subclass would just be for mutation errors, or should it be used more broadly? If `sendErrorToService` receives some errors that contain metadata and some that have it as a separate object it gets confusing, so I would think we'd want some new general error class that can be used anytime we want to attach metadata to an error. 

This is scenario (3) in the PR description. The `ReactionError` class could be simple -- I don't _think_ we need to override the message:

```
class ReactionError extends Error { 
  metadata: object
  constructor(message, metadata = {}) {
    super(message)
    this.metadata = metadata
  }
}
```

We could then nix the `errorInfo` object in `sendErrorToService` and just pull from `error.metadata` if that property exists:

```
export const sendErrorToService = (error: Error | ReactionError) => {
  Sentry.withScope(scope => {
    if (error instanceof ReactionError) {    
      Object.keys(error.metadata).forEach(key => {
        scope.setExtra(key, error.metadata[key])
      })
    }
    Sentry.captureException(error)
  })
}
```

There's a little weirdness in `componentDidCatch` in `ErrorBoundary` since we get an error and immediately convert it into `ReactionError`, but thinking again that seems all right.

williardx:
Just paired with @alloy and decided to go with scenario (3) instead (i.e., creating a generic `ErrorWithMetadata` class) -- will update the PR shortly.

src/Utils/errors.tsx
```diff
@@ -4,22 +4,26 @@ export class NetworkError extends Error {
   response: any
 }
 
-interface ErrorInfo {
-  componentStack: string
+export class ErrorWithMetadata extends Error {
+  metadata: object
+  decoratedMessage: string
+  constructor(message, metadata = {}) {
+    super(message)
```
williardx: @alloy I decided not to make the `decoratedMessage` the actual error message since the metadata can be very long and obscure what's going on.

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -109,7 +110,7 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
   }
 
   onMutationError(errors, errorModalTitle?, errorModalMessage?) {
-    logger.error(errors)
+    logger.error(new ErrorWithMetadata(errors.message || errors.data, errors))
```
williardx: One snag here is that it's difficult to tell what the "message" is when we get back an error payload from Metaphysics. Sometimes we get back a specific message, and other times we get back stringified JSON data that might contain an error message. This can make for some noisy error messages in the latter case since that object might contain a lot of key/value pairs.

alloy: Isn’t that dependent on the call-site? If so, I think you should create the error there instead.

williardx: Yes! You're right, and since we know the type of the error this problem takes care of itself. I'll update that.

src/Utils/errors.tsx
```diff
@@ -4,22 +4,26 @@ export class NetworkError extends Error {
...
+    this.decoratedMessage = message + " | " + JSON.stringify(metadata)
```
damassi: Is `this.decoratedMessage` being used anywhere?

damassi: Ah, I see, its being passed along within `this` down below.

williardx: No, not currently -- it'd be reasonable just to get rid of it until we need it.

src/Utils/errors.tsx
```diff
@@ -4,22 +4,26 @@ export class NetworkError extends Error {
...
-
-export const sendErrorToService = (error: Error, errorInfo?: ErrorInfo) => {
+export const sendErrorToService = (error: Error | ErrorWithMetadata) => {
   Sentry.withScope(scope => {
-    Object.keys(errorInfo).forEach(key => {
-      scope.setExtra(key, errorInfo[key])
-    })
+    if (error instanceof ErrorWithMetadata) {
+      Object.keys(error.metadata).forEach(key => {
+        scope.setExtra(key, error.metadata[key])
```
damassi: Could also use `Object.entries` here: 

```js
Object.entries(error.metadata).forEach(([key, value]) => {
  scope.setExtra(key, value)
  ... 
})

File: src/Apps/Order/Routes/Payment/index.tsx [hunk 7]
```diff
@@ -317,11 +318,15 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
               creditCardId: creditCardOrError.creditCard.id,
             })
           } else {
-            this.onMutationError(
-              errors || creditCardOrError.mutationError,
-              creditCardOrError.mutationError &&
-                creditCardOrError.mutationError.detail
-            )
+            if (errors) {
+              errors.map(err => this.onMutationError(err))
```
alloy:
```suggestion
              errors.forEach(this.onMutationError.bind(this))
```

alloy: (Or bind to `this` with class property syntax and an arrow function.)

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -368,7 +373,14 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
           if (orderOrError.order) {
             this.props.router.push(`/orders/${this.props.order.id}/review`)
           } else {
-            this.onMutationError(errors || orderOrError)
+            if (errors) {
+              errors.map(err => this.onMutationError(err))
```
alloy: Ditto

src/Components/Payment/PaymentForm.tsx
```diff
@@ -220,11 +221,15 @@ class PaymentForm extends Component<PaymentFormProps, PaymentFormState> {
             this.cardElement && this.cardElement.cardInputElement.clear()
             window.scrollTo(0, 0)
           } else {
-            this.onMutationError(
-              errors || creditCardOrError.mutationError,
-              creditCardOrError.mutationError &&
-                creditCardOrError.mutationError.detail
-            )
+            if (errors) {
+              errors.map(err => this.onMutationError(err))
```
alloy: Ditto

src/Utils/__tests__/errors.test.ts
```diff
@@ -1,24 +1,27 @@
...
+    const scope = { setExtra: jest.fn() }
+    const err = new Error("some error")
+    const errWithMetadata = new ErrorWithMetadata("some message", {
+      foo: "bar",
     })
-    it("returns false if an object does not implement the ErrorInfo interface", () => {
-      const notErrorInfo = { anotherKey: 0 }
-      expect(isErrorInfo(notErrorInfo)).toBe(false)
+    it("does not call setExtra on scope if the error has no metadata", () => {
+      scope.setExtra.mockReset()
```
alloy: Creating the scope from a beforeEach removes the need to do this.

src/Utils/__tests__/errors.test.ts
```diff
@@ -1,24 +1,27 @@
...
+  describe("#reportError", () => {
+    const scope = { setExtra: jest.fn() }
+    const err = new Error("some error")
+    const errWithMetadata = new ErrorWithMetadata("some message", {
+      foo: "bar",
     })
-    it("returns false if an object does not implement the ErrorInfo interface", () => {
-      const notErrorInfo = { anotherKey: 0 }
-      expect(isErrorInfo(notErrorInfo)).toBe(false)
+    it("does not call setExtra on scope if the error has no metadata", () => {
```
alloy: Can you add newlines between blocks?

src/Utils/errors.tsx
```diff
@@ -4,22 +4,27 @@ export class NetworkError extends Error {
   response: any
 }
 
-interface ErrorInfo {
-  componentStack: string
+export class ErrorWithMetadata extends Error {
+  metadata: object
+  decoratedMessage: string
+  constructor(message, metadata = {}) {
```
alloy: Can you add a newborn before the method definition?

src/Utils/__tests__/errors.test.ts
damassi: I'm not sure if this is related, but lately I've noticed sporadic automatic white-space removal. Not sure if its related to prettier precommit hook or tslint autofix (or something else entirely) but whatever it is is def new. Not sure how to reproduce.

williardx: This time it was me 😄

PR #1584, opened 2018-11-23
[Artist/Overview] Remove stray curly brace.

alloy:
Phew, thought there was yet another bug related to the responsive work that we only found out after deploying, but appears to be just a stray curly brace 😅 

screen shot 2018-11-23 at 16 14 42


src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -116,7 +116,7 @@ class OverviewRoute extends React.Component<OverviewRouteProps, State> {
                 {bioLen < MAX_CHARS.xs ? (
                   <>
                     <Genes artist={artist} />
-                    <Spacer mb={1} />}
```
damassi: Software is so stressful :/

PR #1586, opened 2018-11-26
Respond page Pt 1

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-665

Hello friends!

This ticket turned out to be a much bigger than expected! And this is only part 1!

The bulk of the problem was that we had coupled some things when working on the buy/initial-offer flow that needed to be decoupled here for use in other combinations. Those were `ShippingAndPayment(Summary|Review)` and `TransactionSummary`. I broke these into four individual components:

`ShippingAndPayment(Summary|Review)` became
- `ShippingSummaryItem`
- `CreditCardSummaryItem`

I didn't need to make separate review versions, because the only difference is the 'change' link, which is now a prop.

`TransactionSummary` became
- `ArtworkSummaryItem`
- `TransactionDetailsSummaryItem`

Also in this PR I needed to refactor the `OrderStepper` again since we have another set of steps to deal with.

Sorry about the size of this PR. I originally implemented it as one big commit 🙈 but went through and re-wrote it in four increments to make it a bit easier to digest hopefully.

Still to come in Part 2:

- Appropriate copy for TransactionDetailsSummaryItem (`Your offer` vs `Seller's offer`)
- 'Seller's offer' section with 'offer history' button.
- Add enzyme tests for the respond page

**EDIT:** One more thing in this PR: Needed to add a new web-specific `Collapse` component because the cross-platform one in palette is buggy when expanding to show text. I couldn't figure out how to fix that one so I think it would be best to try out a new web-only one here and then if it works out we can try to find a way to move it into palette.

Here's a demo of the problem (also available in storybook)

bad-collapse

Also I improved the timing function for the transition. The curve in palette's `Collapse` gets too shallow for the end and you get these noticeable 1-pixel jumps that make it look slow and drunk.


artsyit:

Built with commit 86cee60a2a85baa5711a963ad7ad749db4d76284

https://deploy-preview-1586--artsy-reaction.netlify.com

damassi:
Cool that all of the responsiveness in the PR came directly from `styled-system` -- no need for `<Media>` or `<Responsive>` 🎉

ds300:
@sweir27 
> I know we had some tests already for these routes which included checking for the existence of various content. Do you feel like they are sufficient to catch any potential bugs with the refactor? I poked in storybooks and things look 👍 but would be good to do a pass if you haven't already.

Yeah I'm quite confident in the robustness of these changes. Most of it is statically checked by typescript and relay anyhow. Had a good click through the BN flow in force and everything looked the same and worked the same.

src/Components/WebCollapse.tsx
```diff
@@ -0,0 +1,54 @@
...
+    if (!this.wrapperRef) {
+      return
+    }
+    if (this.props.open && this.wrapperRef.style.height === "0px") {
+      // open
+      this.wrapperRef.style.height = "auto"
+      const actualHeight = this.wrapperRef.offsetHeight
+      this.wrapperRef.style.height = "0px"
+      setTimeout(() => {
+        this.wrapperRef.style.height = actualHeight + "px"
```
sweir27: Will this give us an issue if we add height to the expanded component (i.e. an error message under the input?). If so we may need to re-think.

I'm also wondering if the name of this component can be a bit more descriptive-- maybe `StaticCollapse`? Or is the reason it's called `WebCollapse` because it uses styles/offsets that are not available in RN? Since reaction is only used on the web it sounds a little confusing in this context.

sweir27:
@ds300 this looks 👍!! Thanks for answering all of my q's. 

Feel free to merge it when you're happy with the naming of that component. 😄

PR #1589, opened 2018-11-26
Update with latest MP changes

ashkan18:
# Problem
Tests fail after updating to latest MP schema.

# Cause
We introduced `OrderInterface` in https://github.com/artsy/metaphysics/pull/1386, after updating to schema with that change, we need to update our mocked resolvers to also properly resolve and Order to `BuyOrder`.


artsyit:

Built with commit df2aaaea21a8d66f13ef58c513f524625526a8ed

https://deploy-preview-1589--artsy-reaction.netlify.com

PR #1594, opened 2018-11-26
Stylint fixes for dev-time

orta:
Stylelint is **not** on CI, which I think is OK, but this fixes a lot of the common issues that came up when I ran `yarn stylelint-project`. If I can get emission in a good shape with stylelint, then I'll add style lint's extension to the omakase extension set.


src/Components/Publishing/Sections/ImageSetPreview.tsx
```diff
@@ -192,12 +195,14 @@ const SubTitle = styled.div`
...
-  ${unica("s14", "medium")} ${pMedia.xs`
+  ${unica("s14", "medium")};
+  ${pMedia.xs`
     ${unica("s12", "medium")}
   `};
 `
 const SubTitleCount = styled.div`
-  ${unica("s14")} margin-left: 20px;
+  ${unica("s14")};
+  margin-left: 20px;
```
damassi: @orta - did stylelint catch these things?

orta: Yes

orta: The error is more like "there is two rules on the same line" kinda thing

PR #1597, opened 2018-11-27
[PATCH package] Bump relay network lib to fix silent errors issue

alloy:
Fixes https://artsyproduct.atlassian.net/browse/DISCO-528
For artsy/post_mortems#41

Also adds a dependency missing from #1546


ashkan18:
THAAANKS! 🎉 Nice find!! is this something we can verify easily?

alloy:
You can verify it if you setup Force with local Reaction and MP at the right commit. I’ve done it and can confirm this fixes it, I wouldn’t say it’s entirely trivial to do so.

damassi:
👍

PR #1600, opened 2018-11-27
My last offer transaction summary

ashkan18:
# Change
https://artsyproduct.atlassian.net/browse/PURCHASE-654
For OfferOrders we get the offer amount and list price from `order.myLastOffer` instead of getting the data from `order.itemsTotalCents`

This PR updates with latest MP where we now have access to `OfferOrder` and can access offer related fields like `myLastOffer` and from there we use `amount`, `shipping` and `tax` total cents to populate transaction summary.

Had to do some refactoring for our mocked data to get tests and storybooks working.

# Followup
We need to show proper `buyerTotal` for Offers. We may want to handle that from Exchange->MP end.


artsyit:

Built with commit 530dde37a0761d4e820a29bb1a031c189e224dc5

https://deploy-preview-1600--artsy-reaction.netlify.com

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -20,52 +20,53 @@ export class TransactionDetailsSummaryItem extends React.Component<
...
-        />
+        <Entry label="Tax" value={this.taxDisplayAmount(order)} />
         <Spacer mb={2} />
-        <Entry label="Total" value={buyerTotal} final />
+        <Entry label="Total" value={order.buyerTotal} final />
       </StackableBorderBox>
     )
   }
 
+  private shippingDisplayAmount = order => {
```
ds300: #minor 

public/private doesn't matter at all in react components since everything is essentially private by default (you never really interact with component instances outside of the components themselves). So leaving everything public is fine and the normal way of doing things.

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -20,52 +20,53 @@ export class TransactionDetailsSummaryItem extends React.Component<
...
+  private taxDisplayAmount = order => {
+    return order.mode === "BUY"
+      ? this.formattedAmount(order.taxTotal, order.taxTotalCents) || "—"
+      : this.formattedAmount(
+          order.myLastOffer.taxTotal,
+          order.myLastOffer.taxTotalCents
+        ) || "—"
+  }
+
+  private priceDisplayComponent = (order, offerOverride) => {
```
ds300: #superminor 

A convention that I've seen in react codebases is that component methods that return UI are prefixed with `render`, so this would be `renderPriceEntry` or similar

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -391,7 +394,7 @@ describe("Shipping", () => {
 
   describe("Offer-mode orders", () => {
     it("shows an active offer stepper if the order is an Offer Order", () => {
-      const offerOrder = { ...UntouchedBuyOrder, mode: "OFFER" }
+      const offerOrder = { ...UntouchedBuyOrder, ...UntouchedOfferOrder }
```
ds300: Seems like this could be just `const offerOrder = UntouchedOfferOrder`?

ashkan18:
Addressed all comments, merging to unblock https://github.com/artsy/reaction/pull/1603/files

PR #1603, opened 2018-11-27
Updates offer flow to use correct mutation while submitting an offer

sweir27:
Addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-646

We need to update the Make Offer flow to send the correct mutation when a user submits their offer-order.

Since the responses for the two mutations is currently the same, I updated it to re-use the `onCompleted` handler, but have separate wrappers for sending the actual mutation (this gives us nice TS support for the mutation/inputs/etc.)

There are a ton of changes because I had to update the schema so I had access to the new mutation!

I'll likely need to rebase after  https://github.com/artsy/reaction/pull/1600 goes in.


artsyit:

Built with commit 0c5c8dea3c1be3ccc07a872d0e517e2eca69cba7

https://deploy-preview-1603--artsy-reaction.netlify.com

sweir27:
Rebased!

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -113,48 +120,53 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
...
+            variables: {
+              input: {
+                offerId: this.props.order.myLastOffer.id,
+              },
+            },
+            onCompleted: result => {
+              const {
+                ecommerceSubmitOrderWithOffer: { orderOrError },
+              } = result
+              this.onSubmitCompleted(orderOrError)
```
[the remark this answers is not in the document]
sweir27: Totally possible-- some of them are the same, and since we haven't yet fleshed out which ones we want to handle specially, I think we can wait to implement something until we do.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -163,6 +175,48 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
     }
   }
 
+  private onSubmitCompleted = orderOrError => {
```
ds300: #superminor #nonblocking `public` and `private` don't really matter with react component classes, since actual instances are never really exposed to anybody. There's a lot of other fields marked `private` in this class, so really super nonblocking.

src/Apps/Order/Routes/__tests__/Review.test.tsx
```diff
@@ -51,124 +56,133 @@ describe("Review", () => {
...
-  it("shows an error modal when there is an error in submitOrderPayload", () => {
-    const component = getWrapper(defaultProps)
+    it("takes the user back to the /payment view", () => {
+      const component = getWrapper(defaultProps)
+      component
+        .find(StepSummaryItem)
+        .last()
+        .find("a")
+        .simulate("click")
+      expect(pushMock).toBeCalledWith("/orders/1234/shipping")
```
ds300: should this be /payment?

sweir27: Yes! Turned out we weren't resetting the mock. Good catch 🎉

sweir27:
I think this is good for a final review!

ashkan18:
lets do this! looks like all unrelated file changes are basically MP related where `partner` has become `externalPartner` for some cases

PR #1607, opened 2018-11-28
Fix null myLastOffer

ashkan18:
# Problem
Clicking make offer, creates the offer order but then goes to a page with error saying we missing amount.

# Cause
When we create an Offer Order we don't have `myLastOffer` yet since collector hasn't set their offer yet and from the `TransactionDetailItem` component we try to access `order.myLastOffer.amount` to populate transaction summary section.

# Solution
Add conditional code to check if we have `myLastOffer` and if we don't just show `-`.


artsyit:

Built with commit 59eda9a921bd5ad279bd3cca7c70b32734479b7a

https://deploy-preview-1607--artsy-reaction.netlify.com

sweir27:
Nice!! Thanks for the quick fix here 👏 👏

PR #1615, opened 2018-11-29
Queue CI master builds for more stable deployments

zephraph:
It's taken a while to get back to this point. This feature was initially introduced #1481. It unfortunately had to be reverted because it would fail on forks. I PR'd a fix (https://github.com/eddiewebb/circleci-queue/pull/13) so that issue should be resolved. 

Unfortunately there isn't an incredibly easy way to test this. Ultimately if the build works on master, we're in the green. I'll merge this and test it out after office hours so I can revert if there's a failure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   node: artsy/node@0.1.0
   yarn: artsy/yarn@0.0.2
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 9e564c03e16725e3ae28777d9d839755c887a8d9

https://deploy-preview-1615--artsy-reaction.netlify.com

PR #1617, opened 2018-11-30
Add offer buyer total

ashkan18:
# Problem
Show correct total for offer order.

# Cause
When we switched to show offer totals in https://github.com/artsy/reaction/pull/1600, we didn't set correct total for offer orders since we didn't have the amount coming from Exchange/MP yet.

# Solution
`buyerTotal` was added to `Offer` in https://github.com/artsy/metaphysics/pull/1404, here we get latest MP schema (which causes all sort of other unrelated changes) and based on `order.mode` we either show `order.buyerTotal` or `order.myLastOffer.buyerTotal` for total field.


artsyit:

Built with commit 685799081762c2f1a3d8ce2c5b96abb31d0a7fe9

https://deploy-preview-1617--artsy-reaction.netlify.com

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -163,6 +172,12 @@ export const TransactionDetailsSummaryItemFragmentContainer = createFragmentCont
           shippingTotalCents
           taxTotal(precision: 2)
           taxTotalCents
+          buyerTotal(precision: 2)
+          buyerTotalCents
+        }
+        lastOffer {
```
ds300: Seems like this doesn't actually use `lastOffer` right now. It probably should on the `/status` page though, but that can be another PR. For now let's just remove it?

ashkan18: makes sense, i wasn't sure where it's used.

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -66,6 +70,15 @@ export class TransactionDetailsSummaryItem extends React.Component<
     }
   }
 
+  buyerTotalDisplayAmount = order => {
+    switch (order.mode) {
+      case "BUY":
```
williardx: I wonder if it's worth having these as string constants somewhere else?

ashkan18: good point! @ds300 where would be a good place to have these type of consts?

ds300: I was going to say "it doesn't matter, typescript will tell you if you're putting the wrong string in there so having a const somewhere is not valuable" But then I noticed that in this case `order` has type `any`, which is bad and unnecessary.

I'll make a suggested change thing for this if I can figure out how to do that.

We should really turn on `noImplicitAny` at some point...

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -66,6 +70,15 @@ export class TransactionDetailsSummaryItem extends React.Component<
     }
   }
 
+  buyerTotalDisplayAmount = order => {
```
ds300:
```suggestion
  buyerTotalDisplayAmount = () => {
    const { order } = this.props
```

ds300: should also do the same kind of thing for the other methods here. There's no reason to be using `any` all over the place by accident.

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
orta: I wonder, if there's a way to incrementally add `noImplicitAny`

ashkan18: 👍 with new approach (not passing order to the methods) i think we are good here,

damassi: This is a sort of devils advocate perspective, but from a team-wide engineering perspective is `noImplicitAny` really necessary? We already get pretty good typing and security and its been nice seeing people "come online" with TypeScript from all different backgrounds. Turning that setting on seems somewhat pedantic (not to use the term in in a derogatory sense, just that it might not necessarily solve any real problems we've faced and seems to appeal to the more advanced TypeScript users in the group). Either way, I think that might be RFC territory.

ds300: :+1: to an RFC. I totally feel you not wanting to increase friction for not much benefit. IME it's not much friction at all to be forced to add types to all function parameters, and it can be really helpful to prevent errors early and give intellisense more coverage. Obviously I'm biased though 😅

PR #1619, opened 2018-11-30
Adds a warning when there are more than 15 modfied gnerated files

orta:
We want to try grab a copy of re-ordered generated files when it happens, so this adds a danger rule to highlight it in the PR, then we can look at the computer closer to the moment


javamonn:
Would this be port-able to Emission as well?

orta:
It can, yes

PR #1559, opened 2018-11-16
Adds a link to the partner in an artwork page under 'about this work'

orta:
Fixes DISCO-511 

For me: https://artsyproduct.atlassian.net/browse/DISCO-511

PR #1621, opened 2018-11-30
Only support linking to a partner when the partner is publicly available

orta:
Extends our PR https://github.com/artsy/reaction/pull/1559 to only allow linking when the partner is classed as a public one. There are fake partners (in a DB sense) which we create for things like Auctions, which we shouldn't link to.


artsyit:

Built with commit 9cc8f2de70fe521be0617428e7e269f374bf9bc7

https://deploy-preview-1621--artsy-reaction.netlify.com

PR #1624, opened 2018-12-02
[BREAKING SSR] Make use of @artsy/detect-responsive-traits

alloy:
As described on the tin, this renders less markup for devices that we can detect.

### DIFF
--- a/src/Artsy/Router/buildServerApp.tsx
+++ b/src/Artsy/Router/buildServerApp.tsx
@@ -12,6 +12,7 @@ import { getUser } from "Utils/getUser"
 import { createMediaStyle } from "Utils/Responsive"
 import { trace } from "Utils/trace"
 import { RouterConfig } from "./"
+import { matchingMediaQueriesForUserAgent } from "./matchingMediaQueriesForUserAgent"
 
 interface Resolve {
   ServerApp?: ComponentType<any>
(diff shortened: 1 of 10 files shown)


artsyit:

Built with commit c914cca7a99a1d1ee29995cac012fc374b2330d8

https://deploy-preview-1624--artsy-reaction.netlify.com

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -26,19 +27,19 @@ interface Resolve {
 // No need to invoke this for each request.
 const MediaStyle = createMediaStyle()
 
-export function buildServerApp(config: RouterConfig): Promise<Resolve> {
+export function buildServerApp(
+  config: RouterConfig & { userAgent: string | undefined }
```
damassi: Why not add `userAgent` to `RouterConfig`?

src/Artsy/Router/matchingMediaQueriesForUserAgent.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Find the breakpoints and interactions that the server should render
+ */
+export function matchingMediaQueriesForUserAgent(
+  userAgent: string
+): MatchingMediaQueries {
+  const device = findDevice(userAgent)
+  if (!device) {
+    return undefined
+  } else {
+    const supportsHover = device.touch ? "notHover" : "hover"
```
damassi: Not a blocker, but I wonder if there's a better way to encapsulate `hover` (and other media queries of that kind). Might be a bit premature tho.

src/Artsy/SystemContext.tsx
```diff
@@ -17,12 +16,6 @@ export interface SystemProps {
    */
   user?: User
 
-  /**
-   * Media queries that are passed to Responsive when an app first mounts. Useful
-   * for SSR rendering views for mobile devices.
-   */
-  initialMatchingMediaQueries?: MatchingMediaQueries
```
damassi: This probably relates to my question above -- whats the thinking in removing from here?

src/Artsy/Router/buildServerApp.tsx
alloy: I figure that would make it required by the client app too, which isn’t what we’d want.

src/Artsy/SystemContext.tsx
alloy: I wanted to capture detection of devices as part of `buildServerApp`, rather than each callsite having to determine the device and pass in this list.

PR #1630, opened 2018-12-04
Parameterize copy in TransactionDetailsSummaryItem

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-678

On the `/status` page, we want the `TransactionDetailsSummaryItem` to represent the latest *submitted* offer, which is present on the `lastOffer` field. Meanwhile, on all previous steps of the initial offer flow, we want it to represent the buyer's last *pending* offer, which is available on the `myLastOffer` field.

During the counteroffer flow, we again want it to represent the latest *submitted* offer (https://zpl.io/a8N4opQ), except on the `/review` screen if the user makes a counteroffer (https://zpl.io/bPm8JX6).

So here I added a prop to the `TransactionDetailsSummaryItem` component called `useLastSubmittedOffer` which causes it to use `lastOffer`. By default it uses `myLastOffer`.

The main effect of this is that it can show either 'Your offer' or 'Seller's offer' now: 

image

image

I also consolidated the stories for this component.

### DIFF
--- a/src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
+++ b/src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
@@ -14,6 +14,7 @@ import {
 export interface TransactionDetailsSummaryItemProps extends FlexProps {
   order: TransactionDetailsSummaryItem_order
   offerOverride?: string | null
+  useLastSubmittedOffer?: boolean
 }
 
 export class TransactionDetailsSummaryItem extends React.Component<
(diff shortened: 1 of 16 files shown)


artsyit:

Built with commit 597db8c243bb6ea74f78c72ce60eefda9436a2cc

https://deploy-preview-1630--artsy-reaction.netlify.com

ds300:
> This only impacts status page, right? for all previous pages we still use myLastOffer.

Exactly, yes!

PR #1631, opened 2018-12-04
Update npmignore to include only necessary client artifacts

zephraph:
I noticed that we were shipping a lot of unnecessary stuff with reaction. 

New package size: 1.9MB, 2322 files
Old package size: 2.0MB, 2717 files

Maybe not a _huge_ improvement, but it's better.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,13 +1,6 @@
-.awcache
-.vscode
-.env*
-.storybook
-.fusebox
-.babelrc
-
-externals
-src
-__mocks__
-reaction-force.js
-reaction-force.js.map
-gulpfile.js
+# Ignore everything except what's explicitly whitelisted
+*
+!dist/**
+dist/**/__mocks__/*
+dist/**/__tests__/*
+dist/**/__stories__/*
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 347df2ad6dae68c5d1a7e982fb480cf20dd6d351

https://deploy-preview-1631--artsy-reaction.netlify.com

.npmignore
```diff
@@ -1,13 +1,6 @@
...
-.babelrc
-
-externals
-src
-__mocks__
-reaction-force.js
-reaction-force.js.map
-gulpfile.js
\ No newline at end of file
+# Ignore everything except what's explicitly whitelisted
```
alloy: So much negation in this file... isn’t this including all files and then ignore some?

zephraph: Nope. This is an ignore file. So the `*` is ignore everything. The `!` just says don't ignore this thing.

zephraph: So it's like... ignore everything. Don't ignore dist stuff. But _do_ ignore __mocks__, __tests__, and __stories__ in dist.

alloy: Yeah sorry, I meant to comment on the `!dist/**` part. It basically includes everything first and then ignores some again.

PR #1633, opened 2018-12-04
Move offers and lastOffer to use OfferOrder fields

ashkan18:
# Problem
We've moved offer-related fields under order under `OfferOrder`. All offer-related fields under `Order` are now deprecated and should be accessed with
```graphql
... on OfferOrder {
...
}
```

# Solution
Move `lastOffer` and `offers` under `OfferOrder` fragment.

# Followup
remove these fields from MP and then Exchange.


artsyit:

Built with commit 401895e87a5a58f3684f79a3ac2257eda4d3359c

https://deploy-preview-1633--artsy-reaction.netlify.com

PR #1638, opened 2018-12-05
 Add auction partner name, estimate and premium

oxaudo:
This still needs a mediator event in Force for auction Buyer's premium.

Or it might be worth to just rebuild this modal here in reaction. But seems like eigen is also still using Force html page for that, so I figured - I'll go with a mediator call for now.

@cavvia - also adding analytics call for the 'buyer's premium' click. Figured it's not going to hurt to track it.

### DIFF
--- a/src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.test.tsx
+++ b/src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.test.tsx
@@ -1,0 +1,65 @@
+import { ArtworkWithEstimateAndPremium } from "Apps/__tests__/Fixtures/Artwork/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo"
+import { ArtworkSidebarAuctionPartnerInfoFragmentContainer } from "Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo"
+import { renderRelayTree } from "DevTools"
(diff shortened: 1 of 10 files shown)


src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  describe("ArtworkSidebarAuctionPartnerInfo", () => {
+    beforeEach(() => {
+      artwork = Object.assign({}, ArtworkWithEstimateAndPremium)
+    })
+
+    it("displays partner name, estimate and premium", async () => {
+      const wrapper = await getWrapper(artwork)
+
+      expect(wrapper.text()).toContain(artwork.partner.name)
```
damassi: I would avoid writing tests that are bound to fixtures -- its slightly safer (and easier to grok) working with plain strings.

artsyit:

Built with commit 1674b666896410a00b371dc14bd35c8211221f55

https://deploy-preview-1638--artsy-reaction.netlify.com

src/Apps/Artwork/Components/__tests__/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.test.tsx
oxaudo: The string is coming from a fixture:) But will happily update!

oxaudo:
Corrected the test and merging.

PR #1639, opened 2018-12-05
Test the respond page

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-679

Also fixed a couple of things

- Make sure to use the `lastOffer` field for the transaction summary
- Added the response choice parameters to the window.alert for the continue button so we can test that the data is being set properly already.

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Respond.test.tsx
+++ b/src/Apps/Order/Routes/__tests__/Respond.test.tsx
@@ -1,9 +1,16 @@
+import { Button } from "@artsy/palette"
 import {
   Buyer,
   OfferOrderWithShippingDetails,
   Offers,
   OfferWithTotals,
 } from "Apps/__tests__/Fixtures/Order"
+import { ArtworkSummaryItemFragmentContainer } from "Apps/Order/Components/ArtworkSummaryItem"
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit 1ff9cb20e59c2c38ab13dd8f91a9f8ad3765f017

https://deploy-preview-1639--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Respond.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+  it("renders", () => {
+    const component = getWrapper()
+    const input = component.find(Input)
+    expect(input.text()).toContain("Your offer")
+  })
+
+  it("Shows the stepper", () => {
+    const component = getWrapper()
+    const stepper = component.find(OrderStepper)
+    expect(stepper.text()).toMatch("RespondReview")
```
sweir27: Can we add an assertion here to make sure that `Respond` is the active step? We do a similar thing [here](https://github.com/artsy/reaction/blob/master/src/Apps/Order/Routes/__tests__/Shipping.test.tsx#L396).

PR #1640, opened 2018-12-05
WIP: Fixes an issue where Collection index page links are not registered by Google

yuki24:
fixes https://artsyproduct.atlassian.net/browse/GROW-1043

This replaces the manual `location.href = href` so Google will be able to recognize that they are actually links.

### DIFF
--- a/src/Apps/Collections/Components/CollectionsGrid.tsx
+++ b/src/Apps/Collections/Components/CollectionsGrid.tsx
@@ -1,6 +1,15 @@
-import { Box, Flex, Sans, Separator, Spacer } from "@artsy/palette"
+import {
+  Avatar,
+  Box,
+  Flex,
+  Link,
+  Sans,
+  Separator,
+  Serif,
+  Spacer,
+} from "@artsy/palette"
 import React, { Component } from "react"
-import { EntityHeader } from "Styleguide/Components/EntityHeader"
+import styled from "styled-components"
 import { slugify } from "underscore.string"
 import { crop } from "Utils/resizer"
 import { Media } from "Utils/Responsive"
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit f35a0efba8221d3b9d1a148ec40d7f465515f411

https://deploy-preview-1640--artsy-reaction.netlify.com

eessex:
I'm not sure why this was implemented with `onClick` rather than an actual link -- @damassi or @oxaudo do you have context, are there non-link applications this would break?

damassi:
I remember it being specific, like I think it was to capture the whole container (but not too sure). @oxuado would know more, but feel free to update / iterate. I’m traveling today so can’t test.

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -65,3 +80,9 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
     )
   }
 }
+
+const LinkWithoutBorder = styled(Link)`
+  &:focus {
+    border: none;
```
yuki24: This extra border seems unneeded but I'm just patching this for now.

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -41,13 +50,19 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
...
+                  href={`/collection/${collection.slug}`}
+                >
+                  <Flex py={2} alignItems="center">
+                    <Avatar size="xs" src={imageUrl} />
+
+                    <Serif ml={1} size="4t">
+                      {collection.title}
+                    </Serif>
+                  </Flex>
+                </LinkWithoutBorder>
```
yuki24: I ended up replacing the entire `<EntityHeader>` with palette components. The `<Flex>` component makes things simpler here so it doesn't seem like re-using the `<EntityHeader>` component is necessary.

Also, it would be interesting to make a `<FlexLink>` component if it's a common pattern to wrap elements with an `a` tag:

```jsx
<FlexLink py={2} alignItems="center" href={`/collection/${collection.slug}`}>
  <Avatar size="xs" src={imageUrl} />
  <Serif ml={1} size="4t">
    {collection.title}
  </Serif>
</FlexLink>
```

yuki24:
@eessex @oxaudo @damassi This is good for review.

With regards to the semantics here, [wrapping `div` elements with an `a` tag is no longer invalid since HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements#Block-level_vs._inline) so we should be good to go:

> The distinction of block-level vs. inline elements was used in HTML specifications up to 4.01. In HTML5, this binary distinction is replaced with a more complex set of content categories.

src/Apps/Collections/Components/CollectionsGrid.tsx
damassi: @yuki24 - We're already using Entity header in a number of places; by replacing it outright we're introducing drift. If anything, this should be moved into palette and the previous EntityHeader callsites should be updated.

damassi: List of locations: 

screen shot 2018-12-10 at 1 39 42 pm

yuki24:
I guess now I understand why an `a` tag was not used in the first place - it would add extra styles that `<EntityHeader>` doesn't need, and this PR would break it when `text-decoration: none` is not specified. I'm closing this PR and will come back with a more reliable solution. Thanks! @eessex @damassi @oxaudo!

PR #1643, opened 2018-12-05
[EOY] Culture scaffold

eessex:
Sets up layout and desktop styles for EOY culture feature.
Some responsive styles are included, but will wrap those up in a separate ticket.


artsyit:

Built with commit b9db683cf3745355122b7ea957cbaf10b3ca6ad6

https://deploy-preview-1643--artsy-reaction.netlify.com

src/Components/Publishing/EditorialFeature/Components/Eoy2018Culture.tsx
```diff
@@ -1,21 +1,306 @@
...
+                <div dangerouslySetInnerHTML={{ __html: introText }} />
+              )}
+            </Box>
+          </Box>
+        </TextContainer>
+
+        {chapters.map((chapter, i) => {
+          const isDark = i % 2 === 0
+          return (
+            <ChapterWrapper isDark={isDark} key={i}>
```
yuki24: How about using the `<Box>` component instead of defining a `<ChapterWrapper>`?

```tsx
<Box {...(isDark ? { color: "white", backgroundColor: "black100" } : {})}>
</Box>
```

src/Components/Publishing/EditorialFeature/Components/Eoy2018Culture.tsx
```diff
@@ -1,21 +1,306 @@
...
 
-export const Eoy2018Culture: React.SFC<ArticleProps> = props => {
-  return (
-    <ArticleWrapper>
-      <Nav canFix={false} transparent />
-      <FeatureLayout {...props} />
-    </ArticleWrapper>
-  )
+interface State {
+  chapters: any[]
```
yuki24: Does this need to be a state or a private member of the class is good enough? It doesn't seem like we update the `chapters` in this component.

src/Components/Publishing/EditorialFeature/Components/Eoy2018Culture.tsx
```diff
@@ -1,21 +1,306 @@
...
+  sectionIsHeader = section => {
+    const isText = section.type === "text"
+    const isHeader = isText && section.body.includes("<h1>")
+    return isHeader
+  }
+
+  makeSectionArrays = () => {
+    const {
+      article: { sections },
+    } = this.props
```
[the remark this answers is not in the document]
eessex: I'm not sure what you mean, I think this will still be fine to test -- but yeah, haven't gotten around to that yet ... have a ticket to handle tests separately, the implementation is still a bit in flux until we confirm all the content with editorial.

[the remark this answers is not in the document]
yuki24: I think it's more straightforward to just pass in the `props.article.sections` to `makeSectionArrays` so we don't have to look around to understnad where the data comes from. At a glance it looks mysterious when `makeSectionArrays` is called with no arguments in a `constructor`. What it does is data massagiing, so it would be good to extract it as a plain function.

Also, it has a good amount of complexity that deserves to have its own set of tests, rather than being tested _through_ React.

src/Components/Publishing/EditorialFeature/Components/Eoy2018Culture.tsx
eessex: Was trying to make it a static variable on the class, but got some type errors. I think maybe your suggestion of passing sections from props makes sense to avoid this.

src/Components/Publishing/EditorialFeature/Components/Eoy2018Culture.tsx
eessex: A good suggestion but ended up needing another style declaration, so keeping for now.

eessex:
@yuki24 @l2succes ok to merge?

PR #1645, opened 2018-12-06
[wip] fix mock fixtures

ds300:
For context: https://artsy.slack.com/archives/CC56D129X/p1543370134025100


alloy:
This is looking great!

I’ve been thinking about how to do the same guidance for fields other than these types that we wanted to override from graphql-tools and I think we could do it nicely by providing our own ‘default resolver’, which the `graphql-js` API accepts. That way we can do two things:

* Use the same `complain` implementation when the requested key does not exist in the stubbed data.
* Support aliased fields.

src/DevTools/createMockNetworkLayer/index.ts
```diff
@@ -1,14 +1,37 @@
+import { GraphQLFieldResolver, responsePathAsArray } from "graphql"
 import { IMocks } from "graphql-tools/dist/Interfaces"
 import getNetworkLayer from "relay-mock-network-layer"
 import { Network } from "relay-runtime"
 import schema from "../../../data/schema.graphql"
 import FormattedNumber from "./CustomScalars/formatted_number"
 
+const complain = (info, type) => {
```
sweir27: 🙏

alloy:
I don’t know why, but from my MacBook I wasn’t able to push any new commits to your fork, so here’s a patch that you should apply which finishes the alias support https://gist.github.com/alloy/59b502ebdb77b907c8fe8bc2a2c4f12b

src/DevTools/createMockNetworkLayer/index.ts
```diff
@@ -1,14 +1,37 @@
...
 
-export const createMockNetworkLayer = (mockResolvers: IMocks) => {
+const complain = (info, type) => {
+  const path = responsePathAsArray(info.path).join("/")
+  const message = `A mock for field at path '${path}' of type '${type}' was expected but not found.`
+  console.error(message)
+  throw new Error(message)
+}
+
+export const createMockNetworkLayer = (mockResolvers: IResolvers) => {
```
alloy: I wonder if we should even allow specifiying resolvers rather than just an object of data, because the resolvers are hard to understanding for people and if we emir types for the expected data the resolvers functions wouldn’t be part of that either.

I think we should rename this new implementation and leave the previous one in place and deprecate it. The new version can then only accept pure data objects. If it all works out, then we can probably even drop the graphql-tools dependency in the future and use just the graphql-JS API

ds300: Would it never be desirable to have a resolver return different items during the same query?

alloy: That’s what I’m starting to question and so think this transition could be a good time to tests that out. I think pure data objects are best understood by all for fixture usage.

ds300:
> What happens when source is frozen/sealed?

It falls back to complaining that it can't find the '__id' field.

> Do we even really need to mutate the source instead of just returning a value here?

I was worried that returning different IDs for the same object could be more likely to result in unexpected behaviour than mutating an object once. FWIW there is precedent for this kind of thing. ImmutableJS would mutate objects to store a unique hash on them, and that never seemed to be a cause of errors.

Another alternative would be to use a global es6 Map with the fixtures as keys and their ids as values. That would be fine perf-wise I think, and would avoid both problems.

alloy:
> I was worried that returning different IDs for the same object could be more likely to result in unexpected behaviour

Good point.

> Another alternative would be to use a global es6 Map with the fixtures as keys and their ids as values. That would be fine perf-wise I think (WeakMap could prevent memory leaks in case we do long-running tests some day), and would avoid both the mutation and inconsistent ID problems.

I like the `WeakMap` approach 👍

ds300:
closing in favour of #1854

PR #1647, opened 2018-12-06
Add view for accepting an offer

williardx:
Creates a view for accepting an offer at `/order/:id/review/accept`. Resolves [PURCHASE-365](https://artsyproduct.atlassian.net/browse/PURCHASE-365). Done in collaboration with @pepopowitz!

This PR is _just_ for the view. There are no corresponding mutations or tests just yet.

### What changed
- On selecting "Accept offer" and clicking "Continue" on `/order/:id/respond`, navigate to `/order/:id/review/accept`. 
- Added a new `renderHeaderEntry?` prop to `TransactionDetailSummaryItem`. This is a function that returns a stateless React component representing a header for the `TransactionDetailSummaryItem` component, e.g., "Accept seller's offer" with a "change" link. Most of the new offer review pages have a header like this.
- Modified the price entry component in `TransactionDetailSummaryItem` to show "Seller's offer" or "Your offer" depending on the last offer. Updated GraphQL query to fetch `fromParticipant`, which in turn updated many other autogenerated GraphQL files.
- Added route for `/review/accept` page. The route is _not_ nested under `review` because we ran into problems with Relay running unrelated / duplicate queries related to the `review` page. This route is also conceptually distinct from the original `review` page, which is really a page for reviewing a buy now submission.
- Added view for `/review/accept` 

screen shot 2018-12-06 at 1 24 06 pm

screen shot 2018-12-06 at 1 24 30 pm


artsyit:

Built with commit d4ddaa03e712d46169c32796e1d76d8b66cf29b8

https://deploy-preview-1647--artsy-reaction.netlify.com

src/Apps/Order/redirects.tsx
```diff
@@ -42,11 +42,6 @@ export const shouldRedirect = props => {
     throw new RedirectException(artworkID ? `/artwork/${artworkID}` : "/")
   }
 
-  if (order.state !== "PENDING" && !location.pathname.includes("status")) {
-    // Redirect to status page if the order is no longer PENDING (means it can't be edited anymore)
-    throw new RedirectException(`/orders/${params.orderID}/status`)
-  }
-
```
ds300: This file changed a lot in #1629

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -62,7 +62,9 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
 
   onContinueButtonPressed: () => void = () => {
     this.setState({ isCommittingMutation: true }, () => {
-      window.alert("You did a click!")
+      if (this.state.responseOption === "ACCEPT") {
+        this.props.router.push(`/orders/${this.props.order.id}/review/accept`)
+      }
```
ds300: :+1:

Tests for this page were added in #1639 so you'll need to pull those changes in and then, for the 'ACCEPT' option only, replace the window.alert stuff with the above route change, and then a test assertion like this https://github.com/artsy/reaction/blob/c93f27bfbdf92ab0e8851e0ca4d4e89fe6745716/src/Apps/Order/Routes/__tests__/Review.test.tsx#L72

pepopowitz:
@ds300 I think I've got most of the merge conflicts resolved locally. Most things fell into place pretty smoothly. 

I still want to prove to myself that I merged correctly, before committing. I could use some help figuring out why I can't click on the "make offer" button locally - I think I don't have something set up properly. I'll reach out via Slack.

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -23,6 +24,7 @@ export class TransactionDetailsSummaryItem extends React.Component<
     const { offerOverride, order, ...others } = this.props
     return (
       <StackableBorderBox flexDirection="column" {...others}>
+        {this.props.renderHeaderEntry ? this.props.renderHeaderEntry() : null}
```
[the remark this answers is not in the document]
ds300: I've got a fix for this coming in #1650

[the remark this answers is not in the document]
pepopowitz: This is great, thank you! We had a hunch there was something already existing that we could use, but weren't sure where to find it. We'll dig into this.

src/Apps/Order/Routes/Respond/index.tsx
pepopowitz: Thanks for the tip on the test. I think I got the merge of index.tsx correct, but I'll definitely have to go add the test.

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+    isCommittingMutation: false,
+  }
+
+  onSubmit: () => void = () => {
+    console.log("SUBMITTED!")
+  }
+
+  render() {
+    const { order } = this.props
+    console.log("ORDER", order)
```
sweir27: Extra!

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <TwoColumnLayout
+            Content={
+              <Flex
+                flexDirection="column"
+                style={isCommittingMutation ? { pointerEvents: "none" } : {}}
+              >
+                <Media at="xs">
+                  <ArtworkSummaryItem order={order} />
+                </Media>
+                <Media at="xs">
```
sweir27: Can we put this `Spacer` into the above `<Media at "xs"` component?

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+                  <Button
+                    onClick={this.onSubmit}
+                    loading={isCommittingMutation}
+                    size="large"
+                    width="100%"
+                  >
+                    Submit
+                  </Button>
+                  <Spacer mb={2} />
+                  <Sans textAlign="center" size="2" color="black60">
```
sweir27: I put a note on @ds300 's PR as well, but at some point we should probably extract this little component since we keep repeating it

src/Apps/Order/Routes/Respond/index.tsx
pepopowitz: We updated the corresponding test.

src/Apps/Order/Routes/Accept/index.tsx
pepopowitz: Removed!

src/Apps/Order/Routes/Accept/index.tsx
williardx: Moving this spacer into the same query causes the bottom border of `ArtworkSummaryItem` to disappear at xs viewports:

screen shot 2018-12-07 at 12 52 04 pm

Any ideas for how to get around that?

williardx: Thanks! That worked.

src/Apps/Order/Routes/Accept/index.tsx
ds300: That was done and is now merged in. It's called `ConditionsOfSaleDisclaimer`

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -63,10 +63,16 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
...
-      } else {
-        window.alert(`You decided to ${responseOption}.`)
+      switch (responseOption) {
+        case "COUNTER":
+          window.alert(`You decided to COUNTER with ${offerValue}.`)
+          break
+        case "ACCEPT":
+          this.props.router.push(`/orders/${this.props.order.id}/accept`)
+          break
+        default:
```
pepopowitz: We should probably log an error in the default case, instead of window alerting.

src/Apps/Order/routes.tsx
```diff
@@ -171,6 +172,17 @@ export const routes: RouteConfig[] = [
...
+        path: "review/accept",
+        Component: AcceptRoute,
+        query: graphql`
+          query routes_AcceptQuery($orderID: String!) {
+            order: ecommerceOrder(id: $orderID) {
+              ...Accept_order
+            }
+          }
+        `,
+      },
```
pepopowitz: We should add cacheConfig { force: true }

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -63,10 +63,16 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
...
-        window.alert(`You decided to COUNTER with ${offerValue}.`)
-      } else {
-        window.alert(`You decided to ${responseOption}.`)
+      switch (responseOption) {
+        case "COUNTER":
+          window.alert(`You decided to COUNTER with ${offerValue}.`)
+          break
+        case "ACCEPT":
+          this.props.router.push(`/orders/${this.props.order.id}/accept`)
+          break
```
pepopowitz: This URL is incorrect: should be `orders/{id}/review/accept`

pepopowitz:
We received some feedback while looking over this PR during Purchase Knowledge Share. I'm aggregating it in this comment, for a potential follow-up PR.

* The setup code in the tests & story seems like it will be repeated for other similar flows. Can we find a way to extract the common setup, so (a) it can be reused, and (b) the tests/story is more clear about what makes it unique?
* We should log an error instead of using `window.alert` in the default case of /src/Apps/Order/Routes/Respond/index.tsx (https://github.com/artsy/reaction/pull/1647#discussion_r240667013)
* The "accept" URL in src/Apps/Order/Routes/Respond/index.tsx is incorrect (https://github.com/artsy/reaction/pull/1647#discussion_r240670068)
* We should be using `cacheConfig: { force: true }` in src/Apps/Order/routes.tsx (https://github.com/artsy/reaction/pull/1647#discussion_r240669367)

PR #1648, opened 2018-12-06
[FIX analytics] fixes an issue where Artworks impressions not tracked on Collect and Collections pages

yuki24:
fixes https://artsyproduct.atlassian.net/browse/GROW-1037

This adds an `artwork-item` CSS class and `data-id` to the `<GridItem>` component that represents an artwork, so [Force will automatically pick up the artworks in the viewport](https://github.com/artsy/force/blob/17937ad75ec4dd87538ffdfa690e1ac27b2cdd2a/src/desktop/analytics/impressions.js#L6-L24) and send them to Segment. There is a lot of places where this component is used and a careful QA is required (and thus WIP), but implementation-wise this should be good for review.

artwork-impressions


artsyit:

Built with commit 1b95f2ee1d4c8dfdc8885eb2b110c1475a43e3d3

https://deploy-preview-1648--artsy-reaction.netlify.com

yuki24:
I should also add a few comments otherwise these `artwork-item` and `data-id` look very wasteful.

xtina-starr:
@yuki24 so the `Artwork impressions` is handled by force analytics which looks for this classname?

yuki24:
@xtina-starr Yes, that's exactly what Force does here: https://github.com/artsy/force/blob/17937ad75ec4dd87538ffdfa690e1ac27b2cdd2a/src/desktop/analytics/impressions.js#L6

xtina-starr:
@yuki24 I think this is good to merge. Should we remove the WIP label and monitor/QA in staging?

yuki24:
When 34 components depend on one component it's very hard to QA... but let's just merge and see how it goes on staging.

yuki24:
I've added a comment so this is good to go.

xtina-starr:
Agreed and I do see that this classname is used for some styling, but lets take a look in staging.

yuki24:
So I've looked to see where the `.artwork-item` is used in Force and found the following usages:

 * https://github.com/artsy/force/blob/d22009ec6604a6fc538238a41eaa9b025179a9ff/src/desktop/components/artwork_columns/index.styl#L12
 * https://github.com/artsy/force/blob/d22009ec6604a6fc538238a41eaa9b025179a9ff/src/desktop/apps/show/components/artwork_columns_metaphysics/index.styl#L12
 * https://github.com/artsy/force/blob/d22009ec6604a6fc538238a41eaa9b025179a9ff/src/desktop/components/fillwidth_row/index.styl#L8

These are all scoped to a narrow CSS class to be very specific. Given that we don't actively use CSS classes in Reaction (most of those are styled-components or palette components), it should be safe to add a `.artwork-item` to the `<GridItem>`component.

The other thing I found is that this PR actually added (back?) `Artwork impressions` to the Artist page, the Gene page, and the Works for You page as well:

 * https://staging.artsy.net/artist/andre-breton
 * https://staging.artsy.net/gene/engaged-with-traditional-african-art
 * https://staging.artsy.net/works-for-you (if you don't see any artworks try searching with a popular artist name like "Picasso")

If you open the developer console you should see `TRACKED:  Artwork impressions ...`, but not in production. I believe this should be okay, as we want to track as many artwork impressions as possible. But do let me know if this is over-tracking.

And another thing I noticed, is that we aren't tracking artwork impressions for related artworks on the Artwork page (e.g. https://staging.artsy.net/artwork/leonardo-da-vinci-the-last-supper). I see that something's embedded into the tag, but not exactly the same as the `artwork-item` and `data-id`. Perhaps it's used for something else, but at the same time it leaves a question of "Should we track them?".

PR #1650, opened 2018-12-07
/review/counter page UI

ds300:
Hello friends!

This implements the initial UI for the counter offer review page. It also does a couple of other things:

- Expands on the the redirect logic refactor from a few days ago to allow decoupling the route hierarchy from the redirect rules hierarchy. e.g. this route and a few others fall under `/review/*` but should not share the same logic as the main `/review` page.
- Builds the `TransactionDetailsSummary` on top of `StepSummaryItem` rather than `StackableBorderBox`, to allow us to specify the `title` and `onChange` props that make this happen:  
   image
- Also updates the `TransactionDetailsSummaryItem` to allow us to specify which 'context' price appears below the main price. e.g. 
   image
   vs
   image

 
Desktop

image

Mobile

image

We are still working on the mutation side of things so that will come in a follow-up PR.

### DIFF
--- a/src/Apps/Order/Components/ConditionsOfSaleDisclaimer.tsx
+++ b/src/Apps/Order/Components/ConditionsOfSaleDisclaimer.tsx
@@ -1,0 +1,13 @@
+import { Link, Sans, SansProps } from "@artsy/palette"
+import React from "react"
+
+export const ConditionsOfSaleDisclaimer: React.SFC<
+  Partial<SansProps>
+> = props => (
+  <Sans size="2" color="black60" {...props}>
+    By clicking Submit, I agree to Artsy’s{" "}
+    <Link href="/conditions-of-sale" target="_blank">
+      Conditions of Sale.
+    </Link>
+  </Sans>
+)
--- a/src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
+++ b/src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
@@ -2,28 +2,27 @@ import { TransactionDetailsSummaryItem_order } from "__generated__/TransactionDe
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 
-import {
-  Flex,
-  FlexProps,
-  Sans,
-  Serif,
-  Spacer,
-  StackableBorderBox,
-} from "@artsy/palette"
+import { Flex, Sans, Serif, Spacer } from "@artsy/palette"
+import { StepSummaryItem, StepSummaryItemProps } from "Styleguide/Components"
 
-export interface TransactionDetailsSummaryItemProps extends FlexProps {
+export interface TransactionDetailsSummaryItemProps
+  extends StepSummaryItemProps {
   order: TransactionDetailsSummaryItem_order
   offerOverride?: string | null
   useLastSubmittedOffer?: boolean
+  offerContextPrice?: "LIST_PRICE" | "LAST_OFFER"
 }
 
 export class TransactionDetailsSummaryItem extends React.Component<
   TransactionDetailsSummaryItemProps
 > {
+  static defaultProps: Partial<TransactionDetailsSummaryItemProps> = {
+    offerContextPrice: "LIST_PRICE",
+  }
   render() {
     const { offerOverride, order, ...others } = this.props
     return (
-      <StackableBorderBox flexDirection="column" {...others}>
+      <StepSummaryItem {...others}>
         {this.renderPriceEntry()}
         <Spacer mb={2} />
         <Entry label="Shipping" value={this.shippingDisplayAmount()} />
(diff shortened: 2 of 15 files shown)


artsyit:

Built with commit 154458fc2b9613b1210ee651665377687e66421a

https://deploy-preview-1650--artsy-reaction.netlify.com

src/Apps/Order/Routes/Counter/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+          contactEmail="orders@artsy.net"
+          detailText={this.state.errorModalMessage}
+          headerText={this.state.errorModalTitle}
+        />
+      </>
+    )
+  }
+}
+
+const ConditionsOfSaleDisclaimer = () => (
```
sweir27: Do you think it's worth extracting this into a separate component? Looks like we'll have to use it in each of the flows.

src/Apps/Order/redirects.tsx
```diff
@@ -101,6 +102,19 @@ const goToReviewIfOrderIsPending: RedirectPredicate = ({ order }) => {
   }
 }
 
+const goToRespondIfMyLastOfferIsNotMostRecentOffer: RedirectPredicate = ({
+  order,
+}) => {
+  if (
+    order.myLastOffer &&
+    order.lastOffer &&
+    moment(order.myLastOffer.createdAt).isAfter(order.lastOffer.createdAt)
```
sweir27: Instead of comparing timestamps, should we just verify that the `id`s on these offers are the same?

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -2,36 +2,35 @@ import { TransactionDetailsSummaryItem_order } from "__generated__/TransactionDe
...
-export interface TransactionDetailsSummaryItemProps extends FlexProps {
+import { Flex, Sans, Serif, Spacer } from "@artsy/palette"
+import { StepSummaryItem, StepSummaryItemProps } from "Styleguide/Components"
+
+export interface TransactionDetailsSummaryItemProps
+  extends StepSummaryItemProps {
   order: TransactionDetailsSummaryItem_order
   offerOverride?: string | null
   useLastSubmittedOffer?: boolean
+  offerContextPrice?: "LIST_PRICE" | "LAST_OFFER"
```
sweir27: #minor but since we're adding a new param to this component, it would be nice to add a story to `TransactionDetailsSummaryItem` so we can visualize.

src/Apps/Order/redirects.tsx
```diff
@@ -161,25 +184,34 @@ export const shouldRedirect = ({
     return false
   }
 
-  const locationParts = location.pathname.split("/").slice(3)
-
-  function traverse(node: RedirectRecord, path: string[]): void {
+  function traverse(node: RedirectRecord, path: string): void {
```
sweir27: Do you think the current test coverage in `redirects` is enough to handle the new logic that we're adding here?

It seems to cover our current cases but I'm wondering if we should add some unit-level tests for this method.

src/Apps/Order/redirects.tsx
```diff
@@ -139,6 +153,15 @@ const redirects: RedirectRecord = {
         goToPaymentIfPaymentIsNotCompleted,
       ],
     },
+    {
+      path: "review/counter",
```
sweir27: Can we add some tests to our current `redirects` test file to capture this new case? 😄 

And also, thanks for adding it!! Nice to see this logic is easy to extend. 🎉

src/Apps/Order/redirects.tsx
ds300: 🤔 I think they shouldn't be the same. If the user makes a counteroffer they will be different. Also they will be different before the buyer makes a counteroffer because the buyer made an initial offer at some point. So I think we need to check that the current `myLastOffer` came after the current `lastOffer`

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
ds300: I'll add a test as well.

src/Apps/Order/redirects.tsx
ds300: I think that the test coverage would be good enough if we add a test for the "review/counter" route, but also that unit tests which explicitly talk about this method would be really helpful for people who need to modify it in the future.

ds300: How do you feel about us doing that in a separate PR?

src/Apps/Order/redirects.tsx
sweir27: Ohhh duh you're totally right. Don't know where I was this morning 😅

src/Apps/Order/redirects.tsx
sweir27: Seems 👍 !

ds300: Added a ticket: https://artsyproduct.atlassian.net/browse/PURCHASE-717

src/Apps/Order/Components/ConditionsOfSaleDisclaimer.tsx
```diff
@@ -0,0 +1,13 @@
+import { Link, Sans, SansProps } from "@artsy/palette"
+import React from "react"
+
+export const ConditionsOfSaleDisclaimer: React.SFC<
```
sweir27: Thank you!

PR #1651, opened 2018-12-07
Track edition set option selection in RAP sidebar for BNMO

dblandin:
Render radio buttons for each edition set option within the Responsive Artwork Page sidebar for works that are enrolled in either of both of the Buy Now / Make Offer e-commerce programs.

2018-12-07-123512_871x673_scrot

PR #1652, opened 2018-12-07
Don't show radio buttons if artwork is not e-commerce enabled

dblandin:
Follow-up to #1651 

Don't show the radio buttons at all if the artwork is not enrolled in either Buy Now or Make Offer programs.

2018-12-07-134105_759x654_scrot

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -53,7 +53,11 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
   }
 
   renderEdition(edition) {
-    const isEcommerceEnrolled = edition.is_acquireable || edition.is_inquireable
+    const { artwork } = this.props
+    const isEcommerceEnrolled =
+      (artwork.is_acquireable || artwork.is_offerable) &&
+      (edition.is_acquireable || edition.is_offerable)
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 3d10aaab71f52a66580e52f02a8deadf99eaf4d4

https://deploy-preview-1652--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -62,27 +64,30 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
+      return <Row>{editionFragment}</Row>
+    }
   }
   renderEditions() {
     const editions = this.props.artwork.edition_sets
     const editionsFragment = editions.map((edition, index) => {
       return (
         <React.Fragment key={edition.__id}>
-          <Box p={2}>{this.renderEdition(edition)}</Box>
+          <Box p={2}>{this.renderEdition(this.props.artwork, edition)}</Box>
```
damassi: #minor Since this is a `class` maybe pull `artwork` and `addition` from `this.props` from within `renderEdition`?

dblandin: I think I can do that for `artwork` but I think the edition might need to come through the map iterator in `renderEditions`.

PR #1660, opened 2018-12-10
First pass in fixing createToken stripe not showing errors

ashkan18:
https://artsyproduct.atlassian.net/browse/PURCHASE-719

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Payment.test.tsx
+++ b/src/Apps/Order/Routes/__tests__/Payment.test.tsx
@@ -297,6 +297,19 @@ describe("Payment", () => {
     ).toContain("Your card number is invalid.")
   })
 
+  it("shows an error modal when CreateToken raises an error", () => {
+    stripeMock.createToken.mockImplementation(() => {
+      throw new Error("something failed")
+    })
+
+    const paymentRoute = getWrapper(testProps)
+    fillAddressForm(paymentRoute, validAddress)
+
+    paymentRoute.find(ContinueButton).simulate("click")
+
+    expect(paymentRoute.find(ErrorModal).props().show).toBe(true)
+  })
+
   it("commits setOrderPayment mutation with Gravity credit card id", () => {
     stripeMock.createToken.mockReturnValue({
       then: func => func({ token: { id: "tokenId" } }),
--- a/src/Apps/Order/Routes/__tests__/Payment.test.tsx
+++ b/src/Apps/Order/Routes/__tests__/Payment.test.tsx
@@ -1,4 +1,4 @@
-import { Checkbox, CheckboxProps, Sans } from "@artsy/palette"
+import { Button, Checkbox, CheckboxProps } from "@artsy/palette"
 import { mockTracking } from "Artsy/Analytics"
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 5b63f73cf4149a1667d642d6432d557c207d96b8

https://deploy-preview-1660--artsy-reaction.netlify.com

ashkan18:
@ds300, me and @sweir27 started looking at this yesterday, we kind of got stuck on getting the test pass. The problem seems to be, current test mock the `createToken` method to return results but in this case we want to mock it as promise. Something like
```js
stripeMock.createToken.mockImplementationOnce(() => Promise.resolve(jest.fn()))
```
but still tests are failing with some weird errors. When had a chance can you take a look and see if there is something we can do to solve this. thanks

ds300:
@ashkan18 haven't tested but it should probably be

```ts
stripeMock.createToken.mockImplementationOnce(() => Promise.reject(new Error("couldn't create token :(")))
```

ds300:
@ashkan18 still working on this. So so close to finishing. This close 👌 but need to step out for a few hours. Will push what I have anyway.

ashkan18:
THAAAAAAANKS for looking into this!

ds300:
err accidentally made a new branch, will fix later and push it here https://github.com/ds300/reaction/tree/fix/payemnt-issue

ds300:
@ashkan18 all done. The main problem was that promises were being mocked in too minimal a way in this test file, so I rewrote it to use actual promises that can be chained and support `catch` etc.

@damassi does the change to the sentry init logic make sense to you? It was being setup during tests because `sd.NODE_ENV` was `undefined`, which was overwriting jest's timers mocks somehow.

yuki24:
Nice! Thanks for catching this!

ashkan18:
this is awesome! thanks for dealing with this headache friendly issue :)

src/Apps/Order/Routes/__tests__/Payment.test.tsx
```diff
@@ -282,25 +278,45 @@ describe("Payment", () => {
...
+    stripeMock.createToken.mockReturnValue(
+      Promise.reject(new Error("something failed"))
+    )
+    const paymentRoute = getWrapper(testProps)
+    fillAddressForm(paymentRoute, validAddress)
     paymentRoute.find(ContinueButton).simulate("click")
 
+    // flush promise queue
+    await Promise.resolve()
+    await Promise.resolve()
```
sweir27: Why do we have this twice?

ds300: jest has an internal promise queue. when you do 

    await Promise.resolve()

it pushes that new promise to the end of the queue and execution continues when the queue gets emptied up to and including that new promise.

However, if during that time more promises get added to the queue, then those will still be unresolved. If your test depends on those extra promises being resolved and some effects happening, you will need to flush the queue again. And maybe again. And so on. Depends how long your promise chains get. 

Something like this https://github.com/facebook/jest/issues/2157#issuecomment-279171856 might mitigate that. Worth trying!

PR #1662, opened 2018-12-10
adding style to FollowArtistPopover

oxaudo:
For now I'm removing the react-popover addition and just putting this pop up statically below the follow element.

We can figure out later - what artsy popover elements should look like.
For now also using the existing size for Avatar. May need to introduce a smaller `xxs` size to palette if @design decides that it is crutial.

For now the popover looks like that:
screen shot 2018-12-10 at 3 28 14 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,6 @@
     "react-markdown": "^2.5.0",
     "react-oembed-container": "^0.3.0",
     "react-overlays": "^0.8.3",
-    "react-popover": "^0.5.10",
     "react-relay": "https://github.com/alloy/relay/releases/download/v1.5.0-artsy.3/react-relay-1.5.0-artsy.3.tgz",
     "react-relay-network-modern": "^2.4.0",
     "react-relay-network-modern-ssr": "^1.2.2",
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit 3985b119ef4019dd96fe6c73b58ffe7669cadf80

https://deploy-preview-1662--artsy-reaction.netlify.com

package.json
```diff
@@ -175,7 +175,6 @@
     "react-markdown": "^2.5.0",
     "react-oembed-container": "^0.3.0",
     "react-overlays": "^0.8.3",
-    "react-popover": "^0.5.10",
```
damassi: 👍

src/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -52,16 +49,19 @@ interface State {
   openSuggestions: boolean
 }
 
+const SuggestionsPopoverContainer = styled(Box)`
+  position: absolute;
+  z-index: 1000;
```
damassi: Mind experimenting with the lowest possible index?

damassi:
Looks great 👍

src/Components/FollowButton/FollowArtistButton.tsx
oxaudo: Hah... Sure. I'm not even sure why the contact button was above this pop up. So I went with the random really high number. Will definitely try to make it a more reasonable number.

PR #1673, opened 2018-12-11
[FIX SEO] Fixes an issue where Collection index page links are not registered by Google

yuki24:
I ended up basically just patching the `<CollectionsGrid>`  component to make the links traceable. There's definitely some work that needs to be done on the palette side and I'll follow up once this is fixed.

fixes https://artsyproduct.atlassian.net/browse/GROW-1043

### DIFF
--- a/src/Apps/Collections/Components/CollectionsGrid.tsx
+++ b/src/Apps/Collections/Components/CollectionsGrid.tsx
@@ -1,5 +1,6 @@
-import { Box, Flex, Sans, Separator, Spacer } from "@artsy/palette"
+import { Box, Flex, Link, Sans, Separator, Spacer } from "@artsy/palette"
 import React, { Component } from "react"
+import styled from "styled-components"
 import { EntityHeader } from "Styleguide/Components/EntityHeader"
 import { slugify } from "underscore.string"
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 2936eda9163ea339822530f82b347c672f154566

https://deploy-preview-1673--artsy-reaction.netlify.com

src/Apps/Collections/Components/CollectionsGrid.tsx
```diff
@@ -65,3 +67,9 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
     )
   }
 }
+
+const LinkWithoutBorder = styled(Link)`
```
oxaudo: Wondering if this will be a common issue in other places. Since this focus is not even currently displayed in most places - maybe it's ok to just kill it in Palette for now?

Or maybe add an option - similar to how `noUnderline` is added. Maybe have something like `noFocusOutline`?

oxaudo:
Also looks like snapshot for test need to be updated.

yuki24:
closing in favor of https://github.com/artsy/reaction/pull/1679

PR #1675, opened 2018-12-11
adding key to the avatar used in follow artist popover after following an artist

oxaudo:
#minor.

Also not sure it looks much better without having a loader in there or something. This blank space looks like jittering as well...

follow


artsyit:

Built with commit d1a793b1e4c4962d527029a1d52cf71eeb0d8179

https://deploy-preview-1675--artsy-reaction.netlify.com

damassi:
LGTM 👍  - we def need to start thinking of a proper image / loader for things like this but totally fine for now.

PR #1676, opened 2018-12-11
WIP => [Artwork] Add new slider

damassi:
#### WIP

Updates the Artwork slider to use React-slick as it abstracts away all of the complexity and we get mobile stuff for free. 

Still need to clean up mobile styles and add share (though those will be changing anyway). 

slider


damassi:
@alloy - check it out and let me know what you think. I know that the other image slider was a lot of work but we get everything for free with this one and its quite a bit simpler.

alloy:
Works for me :+1:

src/Apps/Artwork/Components/ImageBrowser/ImageBrowser.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const hasMultipleImages = this.props.images.length > 1
+
+    return (
+      <Container>
+        <Row>
+          {hasMultipleImages && (
+            <Col sm={1}>
+              <ArrowButton
+                direction="left"
+                onClick={() => this.slider.slickPrev()}
```
oxaudo: is the function name clickPrev and clickNext ?

damassi: We're using react-slick

PR #1677, opened 2018-12-11
Add accept offer mutation and corresponding status screen

williardx:
Done in collaboration with @pepopowitz !

On clicking submit on `/orders/$orderId/review/accept`, we send a mutation that accepts the offer and navigates to the correct status screen.

Resolves [PURCHASE-680](https://artsyproduct.atlassian.net/browse/PURCHASE-680) and [PURCHASE-710](https://artsyproduct.atlassian.net/browse/PURCHASE-710)

### What Changed
- Adds mutation to accept an offer and error handling for failed mutations
- On successful mutation, navigates to  `/orders/$orderId/status` with the correct status header and message for an accepted offer
- Cleans up code according to comments in https://github.com/artsy/reaction/pull/1647:
  - Corrects URL from `/orders/$orderId/accept` to `/orders/$orderId/review/accept`
  - Adds `cacheConfig: { force: true }` to `review/accept` route in `routes.tsx`
  - Logs an error instead of running `window.alert` if no action is recognized on pressing "continue" on the offer respond view


artsyit:

Built with commit d3123e3d4b5c10a5c31ae221f093625db2d8de89

https://deploy-preview-1677--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Respond.test.tsx
```diff
@@ -189,24 +190,27 @@ describe("Offer InitialMutation", () => {
         .onClick({})
 
       expect(mockPushRoute).toHaveBeenCalledWith(
-        `/orders/${testOrder.id}/accept`
+        `/orders/${testOrder.id}/review/accept`
       )
     })
 
-    it("Declining the seller's offer works", () => {
```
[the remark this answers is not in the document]
pepopowitz: Reverting things to make it easier for @ashfurrow and @iskounen makes sense. I also just merged their stubbed changes, so we can deal with the merge conflicts ourselves, too.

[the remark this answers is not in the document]
ashfurrow: Thanks!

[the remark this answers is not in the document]
pepopowitz: Resolved!

src/Apps/Order/Routes/__tests__/Accept.test.tsx
```diff
@@ -122,4 +130,89 @@ describe("Offer InitialMutation", () => {
...
+          .find("Button")
+          .props() as any
+        expect(buttonProps.loading).toBeTruthy()
+      })
+
+      const submitButton = component.find(Button).last()
+      submitButton.simulate("click")
+    })
+
+    it("hides the button spinner mutation completes", () => {
```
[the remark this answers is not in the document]
pepopowitz: Corrected! Thanks for catching that.

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -33,22 +42,93 @@ interface AcceptProps {
...
+    errorModalMessage: null,
   }
 
   onSubmit: () => void = () => {
-    console.log("SUBMITTED!")
+    this.setState({ isCommittingMutation: true }, () => {
+      if (this.props.relay && this.props.relay.environment) {
+        commitMutation<AcceptMutation>(this.props.relay.environment, {
+          mutation: graphql`
+            mutation AcceptMutation($input: buyerAcceptOfferInput!) {
```
mzikherman: #naming nit, this feels like a slightly generic-y name, and all of these queries/mutations in Reaction are global.

We've been namespacing (but not with a particularly strict rule or format AFAIK) in other areas, maybe this should be `AcceptOfferMutation`?

Here's an example of a more verbose such fragment: https://github.com/artsy/reaction/blob/988ac9d4f5cc363db4e686f8bfa83235dbae4e81/src/Apps/Collect/Components/Collection/CollectionFilterContainer.tsx#L55

In some cases, like the 'Save' button for an artwork, something genericy- like `Save_artwork`, and `SaveArtworkMutation` make sense. But, I'm also not great at naming, so just wanted to throw it out there!

williardx: That sounds good to me! I'll make that change.

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -145,6 +225,13 @@ export class Accept extends Component<AcceptProps, AcceptState> {
             }
           />
         </HorizontalPadding>
+        <ErrorModal
+          onClose={this.onCloseModal}
+          show={this.state.isErrorModalOpen}
+          contactEmail="orders@artsy.net"
```
iskounen: Since Reaction is open source, should this come from an environment variable instead?

williardx: That makes sense to me, though I'd prefer to knock that out in a separate PR since it affects a bunch of other files.

src/Apps/Order/Routes/__tests__/Respond.test.tsx
```diff
@@ -210,6 +211,25 @@ describe("Offer InitialMutation", () => {
...
+        .props()
+        .onSelect({ selected: true, value: "UNRECOGNIZED VALUE" })
+
+      component
+        .find(Button)
+        .last()
+        .props()
+        .onClick({})
+
+      expect(logger.error).toHaveBeenCalledWith(
```
iskounen: I think that this test should verify that an unrecognized action doesn't modify the order, instead of verifying that a message is logged.

williardx: That's a good point, but we won't have that test available, really, until the counter offer flow is finished since none of the other actions actually modify the order.

williardx: Though part of my response there is assuming that there isn't a straightforward way to test if the order is modified or not at this point, and instead would have to ensure that `commitMutation` isn't called

williardx: Curious for your thoughts here @pepopowitz

iskounen: I was thinking about the tests for reject-offer that Ash and I wrote in Exchange when I made this comment. In that case we made sure that the status of the mocked order was still "submitted". After reading your comment, I realize that Reaction is a purely front-end application so my suggestion doesn't make sense.

However, I still think that there is a better way to test this than checking the logger. Is there an error message that would be displayed to the user when an unrecognized action is attempted?

pepopowitz: I think the assertion that we log an error has value. It says "we will find out when users find an unexpected path through our code."

I also think it makes sense to assert that nothing was changed on the offer, but as you say @williardx, I think the best way to do that in this type of test is to assert that `commitMutation` wasn't called. 

My vote would be "let's have this one test assert that `commitMutation` wasn't called but a logger was". Verification that nothing was changed on the order itself would be great as an end-to-end test, but I don't think that fits here. (Which raises the question - do we have a place for end-to-end tests in this app?)

sweir27: @pepopowitz we do not have "end-to-end" tests in the app! 😞 The closest we currently have is the router tests, which visit a route and _can_ assert what's shown on the page.

This case seems rare (is it even possible?) so it seems like we have a few options:
- Show a generic error modal (we can test that this appears, as we do elsewhere)
- Default to an action like "accept" (since it just kicks them to the review screen, there's not much danger here)

williardx: Hm yeah, I'm not sure how this case would happen in practice at the moment. Given that, just going to the accept screen seems like a very simple fallback. If we need better error handling around this case we can always add it later when we know we need it.

iskounen: Thanks for explaining that to me. What do you guys think about including the action in the logged message and asserting that `UNRECOGNIZED VALUE` is included?

src/Apps/Order/Routes/Accept/index.tsx
pepopowitz: This might fit well into our upcoming "extract common things" knowledge share/mob programming session.

src/Apps/Order/Routes/__tests__/Respond.test.tsx
pepopowitz: We absolutely should do what you're describing @williardx but this conversation made me think more about why this impossible "default" logic branch exists. I think it's because we're using an open-ended string instead of something like a string enum.

This seems like the kind of thing that would be solved by "making impossible states impossible" with a string enum (or something...I'm no TypeScript expert). Then we wouldn't need a default case, because all actually-valid cases would have been covered. 

No action required by this comment....I'm just thinking out loud.

pepopowitz: Er, disregard my comment above. I missed this line, where we ARE limiting the valid values - 

 `responseOption: "ACCEPT" | "COUNTER" | "DECLINE" | null`

Maybe we just want to remove the `null` option & `default` case, since we think that's impossible?

williardx: That makes a lot of sense to me! Feels like a very clean and simple approach 💯

PR #1678, opened 2018-12-11
Fix load progress rehydration

zephraph:
## So, this is a thing.

---

This issue is caused by a rehydration error where React couldn't differentiate between the `SpinnerContainer` and the `RelayContainer`. Ultimately, React doesn't fully reconcile the DOM and only checks the elements in the tree on rehydration. This creates and issue where the content loaded from the `RelayContainer` is injected _inside_ the spinner and ends up... well... spinning. 

The fix is to ensure the dom elements _aren't_ the same. It's hacky, but I honestly don't know how else best to solve this.


artsyit:

Built with commit 997c9a30464515cb517885e261d57a112af9bdaa

https://deploy-preview-1678--artsy-reaction.netlify.com

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -3,7 +3,13 @@ import React from "react"
 import { ReadyState, RelayContainer } from "react-relay"
 import styled from "styled-components"
 
-const SpinnerContainer = styled.div`
+/**
+ * WARNING: Do _not_ change this element to something common like a div. If the
+ * element of this container is the same as the element used in the RelayContainer
+ * then then rehydration can fail and cause the RelayContainer to recieve styles
```
alloy:
```suggestion
 * then rehydration can fail and cause the RelayContainer to receive styles
```

alloy:
Small comment, but otherwise 👍 Merge at will.

oxaudo:
🥇 to @zephraph for taking on this strangeness -:)

PR #1680, opened 2018-12-12
Patch graphql .mjs file to fix storybooks no-load error

ds300:
context: https://artsy.slack.com/archives/C2TQ4PT8R/p1544560372121700

This bug first appeared in https://github.com/artsy/reaction/commit/2334f3006a414b1bab8ee7bb17af03d55172e887

That commit introduced quite a few dependency updates. One of them (I’m guessing the webpack update) caused storybook’s webpack to resolve `.mjs` files in preference of `.js` files if both are present for the same filename, which is fine.

The problem was that we are using a patched version of `graphql` which specifically disabled the ‘__id’ error message shown above. This patch was applied to a `.js` file but I discovered that there is also a `.mjs` version of that file which was now the version being loaded. So I patched that file too and the problem went away.

edit: also fixed the ArtworkSummaryItem story, which had an unrelated data mocking issue.

### DIFF
--- a/src/Apps/Order/Components/__stories__/ArtworkSummaryItem.story.tsx
+++ b/src/Apps/Order/Components/__stories__/ArtworkSummaryItem.story.tsx
@@ -1,11 +1,11 @@
 import { Flex } from "@artsy/palette"
 import { ArtworkSummaryItem_order } from "__generated__/ArtworkSummaryItem_order.graphql"
+import { mockResolver, UntouchedBuyOrder } from "Apps/__tests__/Fixtures/Order"
 import { MockRelayRenderer } from "DevTools"
 import React from "react"
 import { graphql } from "react-relay"
 import { storiesOf } from "storybook/storiesOf"
 import { Section } from "Styleguide/Utils/Section"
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 89296b6de5b58e64f8670a3afb4e241acd15b242

https://deploy-preview-1680--artsy-reaction.netlify.com

src/Apps/Order/Components/__stories__/ArtworkSummaryItem.story.tsx
```diff
@@ -52,17 +41,10 @@ const render = (extraOrderProps?: Partial<ArtworkSummaryItem_order>) => {
...
-        }),
-        OrderParty: () => ({
-          ...order.seller,
-          __typename: "User",
-        }),
-      }}
+      mockResolvers={mockResolver({
+        ...UntouchedBuyOrder,
+        ...extraOrderProps,
+      })}
```
alloy: 👍

alloy:
LGTM 👌

damassi:
Awesome, thanks @ds300 👍 👍

PR #1684, opened 2018-12-12
Make palette a peer dependency of reaction

zephraph:
All consumers of reaction already have a direct dependency on palette. Having palette as a direct dependency increases the risk of having conflicts between versions in reaction and in the consumer of reaction. This removes that risk by making palette a peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "react": "^16.5.0",
     "react-dom": "^16.5.0",
     "react-router": "^4.2.0",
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 9b5b208265bf26a4bb96d3e8bac647779e42cb10

https://deploy-preview-1684--artsy-reaction.netlify.com

ds300:
Thanks @zephraph 🙇

PR #1694, opened 2018-12-13
Hook up counter offer mutation on respond page

ds300:
<s>Marked as [wip] because I haven't had a chance to test this e2e in force yet but figured I'd put it up for review in case anybody else wants to give that a go :heart: If you do and it works, please feel free to remove the [wip] tag and/or merge!</s>

https://artsyproduct.atlassian.net/browse/PURCHASE-693


artsyit:

Built with commit 2b80a5bba9dee9193868b5e1a666f6101a4c3fc6

https://deploy-preview-1694--artsy-reaction.netlify.com

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -61,11 +68,17 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
   } as RespondState
 
   onContinueButtonPressed: () => void = () => {
-    const { offerValue, responseOption } = this.state
+    const { responseOption } = this.state
     this.setState({ isCommittingMutation: true }, () => {
       switch (responseOption) {
         case "COUNTER":
-          window.alert(`You decided to COUNTER with ${offerValue}.`)
+          this.createCounterOffer(Number(this.state.offerValue || "0"))
```
sweir27: We'll have to update this when we add validation 😅

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -80,7 +93,62 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
...
+                      data
+                    }
+                  }
+                }
+              }
+            }
+          `,
+          variables: {
+            input: {
+              offerId: this.props.order.lastOffer.id,
```
sweir27: In retrospect maybe we should have named this input param `respondsToOffer`.

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -80,7 +93,62 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
...
+                new ErrorWithMetadata(
+                  orderOrError.error.code,
+                  orderOrError.error
+                )
+              )
+            } else {
+              resolve(orderOrError.order)
+            }
+          },
+          onError: reject,
```
sweir27: We should make a note to adopt this pattern elsewhere, the `resolve`/`reject` is 👍.

src/Apps/Order/Routes/__tests__/Respond.test.tsx
```diff
@@ -222,16 +233,103 @@ describe("Offer InitialMutation", () => {
...
         .onClick({})
 
-      // TODO: get rid of window.alert
-      expect(window.alert).toHaveBeenCalledWith(
-        `You decided to COUNTER with 84838.`
+      expect(commitMutationMock).toHaveBeenCalledTimes(1)
+
+      expect(commitMutationMock.mock.calls[0][1].variables)
+        .toMatchInlineSnapshot(`
+Object {
```
sweir27: This formatting is weird 😬.

But `inlineSnapshot` is so cool!

src/Apps/Order/Routes/__tests__/Respond.test.tsx
```diff
@@ -222,16 +233,103 @@ describe("Offer InitialMutation", () => {
...
+
+    component
+      .find(Button)
+      .last()
+      .props()
+      .onClick({})
+
+    expect(commitMutationMock).toHaveBeenCalledTimes(1)
+
+    await flushPromiseQueue()
```
sweir27: I so wish there was a way around having to add this to each test.

src/Utils/flushPromiseQueue.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * test utility to wait for all promises to be resolved
+ */
+export function flushPromiseQueue() {
+  return new Promise(r => setTimeout(r, 0))
```
sweir27: Can you explain what this is doing/why you changed from the `Promise.resolve()`? 😄

src/Apps/Order/Routes/Respond/index.tsx
ds300: I decided to try promisifying this as a test ahead of our 'refactor to remove duplication' work, and yeah it feels good 💆

src/Apps/Order/Routes/__tests__/Respond.test.tsx
ds300: Yup gotta put up with the weird formatting, but otherwise the DX of `inlineSnapshot` is :100:

src/Utils/flushPromiseQueue.ts
ds300: Remember I mentioned the internal promise queue? It might look like this

    things come in here -> [ E, D, C, B, A ] -> and go out here

The things in that queue aren't actually promises. They're actually *worker callbacks*. They get pushed onto the queue whenever one of four things happens:

- An unresolved promise with registered `.then` listeners is resolved
- An unresolved promise with registered `.catch` listeners is rejected
- Someone calls `.then` on an already-resolved promise
- Someone calls `.catch` on an already-rejected promise

That worker's job is to execute the `.then` or `.catch` callbacks that were registered with the promise. It doesn't happen synchronously for... reasons. Probably to avoid unbounded call stacks by default.

```ts
console.log("creating the promise")
const p = new Promise(resolve => {
  console.log("resolving the promise")
  resolve()
})
console.log("registering .then callback")
p.then(() => {
  console.log("promise was resolved")
})
console.log("finished setup")
```

outputs

```
creating the promise
resolving the promise
registering .then callback
finished setup
promise was resolved
```

Let's see how the promise queue changes over that time

```
// Starts out empty
// -> [] ->
creating the promise
resolving the promise
registering .then callback
// Adds a worker callback
// -> [ F ] ->
finished setup
// call stack drains and worker callback is dequeued and invoked
// -> [ ] ->
promise was resolved
```

So before, when we were doing `await Promise.resolve()` something like this was happening:

We pushed a new callback worker P onto the queue

    -> [ P, Other, Workers, Created, During, The, Test ] ->

Then the call stack drains and those other workers are executed, which enqueued more workers

    -> [ Yet, More, Workers, P ] ->

Then P is dequeued and executed, which means our test continues executing from the `await` point.

But those workers that are still left in the queue might have effects which are necessary for our test assertions to pass. So we can add more `await Promise.resolve()`s until the test passes, or we can replace the original `await` with this:

    await new Promise(resolve => setTimeout(resolve, 0))

*aside: I think the number there can be anything sufficiently low. As long as it's below ~10 it shouldn't make much any difference.*

So what happens if we do that instead? A worker is not immediately pushed onto the promise queue. Even though we're calling `.then` immediately (via `await`), the promise is not yet resolved. Refer back to the list of four conditions above if this is unclear.

We scheduled the `resolve` function with the regular JS event loop, which has a lower precedence than the promise queue. So this time `[ Other, Workers, Created, During, The, Test ]` get executed and then so do `[ Yet, More, Workers ]` before our `P` is ever queued up.

Once P is executed, all callbacks registered for resolved promises should have been taken care of, which means our assertions should be able to pass 🤞

src/Apps/Order/Routes/Respond/index.tsx
ds300: I don't think we'll forget, but will add a TODO for visibility :+1:

src/Utils/flushPromiseQueue.ts
sweir27: @ds300 Thank you for taking the time to write this!!

PR #1699, opened 2018-12-13
Show error modal on accept offer review page if charge capture fails

williardx:
Displays an error modal with the correct title and message if charge capture fails when accepting an offer on `/orders/$orderId/review/accept`.

Addresses [PURCHASE-715](https://artsyproduct.atlassian.net/browse/PURCHASE-715).

screen shot 2018-12-13 at 5 59 17 pm

screen shot 2018-12-13 at 5 59 26 pm

PR #1700, opened 2018-12-14
Buyer can reject an offer and see a status page

iskounen:
This PR addresses [PURCHASE-586](https://artsyproduct.atlassian.net/browse/PURCHASE-586) and [PURCHASE-709](https://artsyproduct.atlassian.net/browse/PURCHASE-709).

After merging this PR, a buyer will be able to:
- Click the submit button on the review screen to reject the offer and have its order canceled.
- See a status screen that confirms the rejection.
- Click a button on the status screen to be navigated back to the Artsy homepage.


src/Apps/Order/Routes/Reject/index.tsx
```diff
@@ -1,44 +1,126 @@
...
+              }
+            }
+          `,
+          variables: {
+            input: {
+              offerId: this.props.order.lastOffer.id,
+            },
+          },
+          onCompleted: data => {
+            console.log(data)
```
sweir27: 🙈

data/schema.graphql
```diff
@@ -776,7 +776,7 @@ type Artwork implements Node & Sellable {
   # Returns an HTML string representing the embedded content (video)
   embed(width: Int = 853, height: Int = 450, autoplay: Boolean = false): String
   edition_of: String
-  edition_sets: [EditionSet]
+  edition_sets(sort: EditionSetSorts): [EditionSet]
```
iskounen: **Note to reviewers:** I haven't made any changes that should have caused this change to the schema, but I committed it because GraphQL works in mysterious ways. Please let me know if this seems strange.

artsyit:

Built with commit 3e76910d909df40ac8c6cfb518e339189b8c3f42

https://deploy-preview-1700--artsy-reaction.netlify.com

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -146,6 +176,19 @@ const offerMessages = {
       <a href="mailto:orders@artsy.net">orders@artsy.net</a>.
     </>
   ),
+  CANCELED: () => (
+    <>
+      <p>
+        Thank you for your repsonse. The seller will be informed of your
```
sweir27: `response`

sweir27: We actually do have a generic "canceled" status page for offers that's slightly different from this, and actually much more similar to the generic canceled page for orders. (See zpl.io/VqlAOY7)

Since `buyerRejectedOffer` is the only _special_ canceled case we care about, I'm not sure where this logic should live but I think we should make sure to only show this message in that one case, not for all canceled offers.

src/Apps/Order/Routes/__tests__/Reject.test.tsx
```diff
@@ -82,4 +94,82 @@ describe("Buyer rejects seller offer", () => {
...
+
+      const buttonProps = component
+        .update()
+        .find("Button")
+        .props() as any
+      expect(buttonProps.loading).toBeFalsy()
+    })
+  })
+
+  it("shows an error modal when there is an error from the server", () => {
```
sweir27: I think I missed this in other PR reviews, but let's add another test case for something like a network error that would go through the `onError` handler. See [here](https://github.com/artsy/reaction/blob/40f0fc58ee373765ae1248906e4abf63ef0458b5/src/Apps/Order/Routes/__tests__/Review.test.tsx#L112-L121) for an example.

src/Apps/Order/Routes/Reject/index.tsx
iskounen: Fixed.

src/Apps/Order/Routes/Status/index.tsx
iskounen: Fixed.

src/Apps/Order/Routes/__tests__/Reject.test.tsx
iskounen: Fixed. We can share this test case with _accept_ and _counter_ when we consolidate the code for thos features.

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -92,24 +92,40 @@ export class Accept extends Component<AcceptProps, AcceptState> {
...
-            }
+            this.onSubmitCompleted(orderOrError)
           },
           onError: this.onMutationError.bind(this),
         })
       }
     })
   }
 
+  onSubmitCompleted(orderOrError) {
```
williardx: Huh, confusing to see these changes in here since they're from https://github.com/artsy/reaction/pull/1699.

williardx:
This looks good to me, great work!

src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -146,6 +176,22 @@ const offerMessages = {
       <a href="mailto:orders@artsy.net">orders@artsy.net</a>.
     </>
   ),
+  CANCELED: (props: StatusProps) =>
+    props.order.stateReason === "buyer_rejected" ? (
+      <>
+        <p>
+          Thank you for your repsonse. The seller will be informed of your
```
sweir27: Looks like this typo made it in 😅

iskounen: Oh, how embrassing... Fixed in #1703

PR #1711, opened 2018-12-17
Refactor shouldRedirect logic for Make Offer & add unit test coverage

pepopowitz:
This PR refactors the `shouldRedirect` method that we were using to identify when a user should be redirected in the Make Offer workflow. It also adds unit-test coverage to the refactored logic.

We saw two big opportunities for improvement when refactoring:
1. The logic was tied to the concept of an order, but didn't need to be. We made it more generic & reusable.
2. Implementation details of throwing exceptions for the `found` router were baked into the logic. We extracted it.

Test coverage of the new `getRedirect` function is at 100%, but there might still be cases we want to test. We're open to suggestions for tests that might be missing!


artsyit:

Built with commit fbf12d17ad921ce358bb3491feadbf542e847a3f

https://deploy-preview-1711--artsy-reaction.netlify.com

src/Apps/Order/__tests__/getRedirect.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      const rule = aRuleWithChildren([
+        {
+          path: "this-section/this-subsection",
+          rules: [aMatchingPredicateWithResult("this-section-route")],
+        },
+      ])
+
+      // REVIEW: this test fails if we change the path to
+      //  "/this-section////this-subsection/this-page"
+      // Is the intention to ignore extra slashes at ALL levels?
```
ds300: Just checked, `found` does not normalize paths, so we probably shouldn't either. I can't imagine when it'd be necessary anyway.

ds300:
Alright I'm happy with this if you are @pepopowitz 🍏

alloy:
I'm noticing that the rules in the `redirects.tsx` module are not type-checked and the data query still exists in a different module (`routes.tsx`), making it easy to break things. I'd like to move the data query part into the `redirects.tsx` module and fix type-checking, let me know if either/both of you want to take that on instead or pair.

ds300:
Good catch! 🔍 Happy for you to take care of this if you especially want to. Otherwise I'm sure Steve and I could slot it in today. Shouldn't take long.

The default type parameters in the `RedirectPredicate` and `RedirectRecord` types were supposed to be temporary while we refactored. Oops 🤦‍♂️

PR #1716, opened 2018-12-17
[FIX bnmo] Round offer values to two decimal places

williardx:
### Problem
We currently round offers to the nearest whole dollar. This is incorrect since we want to allow users to enter in cent values in offers.

Addresses [PURCHASE-747](https://artsyproduct.atlassian.net/browse/PURCHASE-747)

### Solution
Round offer values to two decimal places.


artsyit:

Built with commit 28c89a493299f8f25f7e619b0a34c52ac3b7d77f

https://deploy-preview-1716--artsy-reaction.netlify.com

artsyit:

Built with commit 95b53c07915f8ee74e61f43e817904d43d0f0faf

https://deploy-preview-1716--artsy-reaction.netlify.com

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -164,9 +164,9 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
                     defaultValue={null}
                     onChange={ev =>
                       this.setState({
-                        offerValue: Math.floor(
-                          Number(ev.currentTarget.value || "0")
-                        ),
+                        offerValue: +Number(
```
damassi: Is `+` needed since things are already being coerced with `Number()`?

williardx: Yes, because `toFixed` converts numbers to strings. It is a little weird, though, to convert to a string to a number to a string to a number... 🤔

williardx: Just updated this to use `Math.round` instead, which I think is clearer.

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -164,9 +164,10 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
                     defaultValue={null}
                     onChange={ev =>
                       this.setState({
-                        offerValue: Math.floor(
-                          Number(ev.currentTarget.value || "0")
-                        ),
+                        offerValue:
+                          Math.round(
+                            Number(ev.currentTarget.value || "0") * 100
```
sweir27: This feels like sort of an insane way to achieve what we want 😆but I trust you!

PR #1721, opened 2018-12-18
make getting url from images defensive

oxaudo:
Was getting some js errors for getting `.cropped` of null so searched for all .cropped and replaced those with `get` functions (just to be safe :). Thanks @damassi for adding this function in a first place!

### DIFF
--- a/src/Components/Onboarding/Steps/Artists/PopularArtists.tsx
+++ b/src/Components/Onboarding/Steps/Artists/PopularArtists.tsx
@@ -164,24 +164,27 @@ class PopularArtistsContent extends React.Component<Props, null> {
   render() {
     const artistItems = this.props.popular_artists.artists
       .filter(Boolean)
-      .map((artist, index) => (
-        <LinkContainer key={`popular-artists-${index}`}>
-          <ReplaceTransition
-            transitionEnterTimeout={1000}
-            transitionLeaveTimeout={400}
-          >
-            <ItemLink
-              href="#"
-              item={artist}
-              key={artist.id}
-              id={artist.id}
-              name={artist.name}
-              image_url={get(artist, a => a.image.cropped.url)}
-              onClick={() => this.onFollowedArtist(artist)}
-            />
-          </ReplaceTransition>
-        </LinkContainer>
-      ))
+      .map((artist, index) => {
+        const imageUrl = get(artist, a => a.image.cropped.url)
+        return (
+          <LinkContainer key={`popular-artists-${index}`}>
+            <ReplaceTransition
+              transitionEnterTimeout={1000}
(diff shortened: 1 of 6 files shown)


artsyit:

Built with commit aeb510cf363c5b47dab0e5a5cf749ff84e2bf5d1

https://deploy-preview-1721--artsy-reaction.netlify.com

src/Components/Onboarding/Steps/Artists/PopularArtists.tsx
```diff
@@ -175,7 +176,7 @@ class PopularArtistsContent extends React.Component<Props, null> {
               key={artist.id}
               id={artist.id}
               name={artist.name}
-              image_url={artist.image && artist.image.cropped.url}
+              image_url={artist.image && get(artist.image, i => i.cropped.url)}
```
damassi: `image_url={get(artist.image, i => i.cropped.url)}`

src/Components/Onboarding/Steps/Genes/GeneSearchResults.tsx
```diff
@@ -148,7 +149,7 @@ class GeneSearchResultsContent extends React.Component<Props, null> {
             key={item.id}
             id={item.id}
             name={item.name}
-            image_url={item.image.cropped.url}
+            image_url={get(item, i => i.image.cropped.url)}
```
damassi: into a variable (and all further cases)

src/Components/Onboarding/Steps/Artists/PopularArtists.tsx
oxaudo: It's inside `map` function so was not very obvious without changing the flow. Not even sure we need to update those. I'm not sure where this is used but figured it could never hurt to be defensive. But will update for your `artist` suggestion.

src/Styleguide/Components/ArtistCard.tsx
```diff
@@ -49,7 +50,7 @@ export const LargeArtistCard: SFC<Props> = props => (
   <BorderBox hover flexDirection="column" width="100%" height="254px">
     <Flex flexDirection="column" flexGrow="0" alignItems="center" pt={1}>
       {props.artist.image && (
-        <Avatar src={props.artist.image.cropped.url} mb={1} />
+        <Avatar src={get(props.artist.image, i => i.cropped.url)} mb={1} />
```
oxaudo: And here the whole Avatar is rendered conditionally on image object presence, but strangely seems like image might be missing `cropped` sometimes, though not sure if it's just on staging...

oxaudo:
@damassi - ok, replaced with variables where applicable. The change-log is def bigger:)

damassi:
👍

PR #1722, opened 2018-12-18
[PATCH publishing] remove launch message from tooltips

lilyfromseattle:
removed “This is a new feature. Tell us what you think.“ message from gene and artist tooltips


artsyit:

Built with commit efc14e9d01b8ec18c7123496ebf36a457ecc91e3

https://deploy-preview-1722--artsy-reaction.netlify.com

eessex:
I'm curious if the `NewFeature` message is used anywhere else?  Heard that we were planning to deprecate this, if it isn't in use outside this component may be worth removing the file.

PR #1724, opened 2018-12-18
[PATCH bnmo] Redirect to respond from status page if awaiting buyer response

williardx:
Addresses [PURCHASE-752](https://artsyproduct.atlassian.net/browse/PURCHASE-752).

Redirects to `orders/$orderId/respond` from the status page if we're awaiting for buyer response. 

We generally want to redirect to `/respond` from any route except the `/review/*` routes. If we're awaiting buyer response, though, then the order is guaranteed to be in a submitted state, which means that a visit to one of those URLs will redirect to status since they contain the `goToStatusIfOrderIsNotPending` redirect rule. Once we redirect to status we again redirect to `/respond`, giving us the effective redirect to `/respond` we want.


artsyit:

Built with commit 4f867078c645a8b34025de8c577fb6653aca30b5

https://deploy-preview-1724--artsy-reaction.netlify.com

src/Apps/Order/redirects.tsx
```diff
@@ -175,6 +181,7 @@ export const redirects: RedirectRecord<{
         goToReviewIfOrderIsPending,
         goToShippingIfShippingIsNotCompleted,
         goToPaymentIfPaymentIsNotCompleted,
+        goToRespondIfAwaitingBuyerResponse,
```
sweir27: Wooo nice!!

Curious why the route below got updated? 

Also, can we add a test for this one? 😄

williardx: I'm not sure why the route below got updated 🤔 pretty much every time I touch my local environment it changes.

I'm actually not sure where tests for these live -- it doesn't seem like any of the other redirects are tested?

sweir27: All of the tests live in here! https://github.com/artsy/reaction/blob/master/src/Apps/Order/__tests__/OrderApp.test.tsx

williardx: Ah, thanks!

PR #1725, opened 2018-12-18
Added checks for version mismatch errors when making an initial offer.

iskounen:
This PR addresses [PURCHASE-723](https://artsyproduct.atlassian.net/browse/PURCHASE-723) in which a user is shown an informative error message if they make an initial offer on an artwork whose underlying data has changed since the page loaded.

This situation would arise if an administrator is updating an artwork in the CMS while a visitor is viewing that artwork's page. If the administrator saves that artwork before the visitor has a chance to click the Make Offer button, then this error message should be triggered to prevent them from making an offer on the altered artwork.

### DIFF
--- a/src/Apps/Order/Routes/Offer/index.tsx
+++ b/src/Apps/Order/Routes/Offer/index.tsx
@@ -8,7 +8,7 @@ import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
 import { ContextConsumer, Mediator } from "Artsy/SystemContext"
 import { Input } from "Components/Input"
 import { ErrorModal } from "Components/Modal/ErrorModal"
-import { Router } from "found"
+import { RouteConfig, Router } from "found"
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 7b98cc763eb03ae19aee8b8780512b8f5d798dd2

https://deploy-preview-1725--artsy-reaction.netlify.com

sweir27:
This looks like the way you would do it, but we actually will only receive this error when we _submit_ the order (which happens on the `Review` step of the checkout flow).

#minor but the description of this is slightly misleading as well. If the artwork is changed while the user is on the artwork page, they would still be able to enter the checkout flow. This error is for when the artwork changes while they are in the middle of checking out.

I think you can just move this logic over to the `Review` page, but it should look the same!

Sorry if this wasn't clear from the initial ticket. 😞

iskounen:
Closing PR due to a misunderstanding of the user story.

PR #1732, opened 2018-12-19
[PATCH CurrentBidInfo] Use most recent bid for win info

oxaudo:
This bug was reported. https://www.notion.so/artsy/Missing-outbid-x-symbol-365c59caeaa843c0afd0923f8cf5366e

Looks like once you've been outbid you don't have an active bid any more. We should use your latest bid for the win/lose info.

I have some questions about intuitiveness of this display, but that is a different story. Will bring it up with Melanie, Jeffrey and Steven.


artsyit:

Built with commit fa367052079b1fea2b4c3d625b9853fa005368b0

https://deploy-preview-1732--artsy-reaction.netlify.com

damassi:
👍

PR #1733, opened 2018-12-19
[wip] Factor out error modal

ds300:
Hi friends! Here's the work we did to factor the error modal out of the offer page. :heart:

things to do:

- [x] factor error modal out into injected prop `showErrorModal`
- [x] refactor pages to use new stuff
- [x] fix existing tests
- [ ] new tests for `showErrorModal`


artsyit:

Built with commit bed3bc4f681e9942cf4aeee840410a5e09f85505

https://deploy-preview-1733--artsy-reaction.netlify.com

src/Apps/Order/ErrorModalContext.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+
+export interface ErrorModalProps {
+  showErrorModal: (
+    props: { title?: string; message?: string; ctaAction?: string }
+  ) => void
+}
+
+export const ErrorModalContext = React.createContext<ErrorModalProps>({
+  showErrorModal: null,
```
ds300: Whoops, this should be that default implementation with the console.error

src/Apps/Order/Routes/__tests__/Offer.test.tsx
```diff
@@ -111,7 +111,7 @@ describe("Offer InitialMutation", () => {
       component.find(Button).simulate("click")
     })
 
-    it("shows an error modal when there is an error from the server", () => {
+    it.only("shows an error modal when there is an error from the server", () => {
```
ds300: Whoops

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -24,7 +27,7 @@ import createLogger from "Utils/logger"
 import { Media } from "Utils/Responsive"
 import { offerFlowSteps, OrderStepper } from "../../Components/OrderStepper"
 
-export interface OfferProps {
+export interface OfferProps extends ErrorModalProps {
```
iskounen: This line caught my attention and made me wonder whether Context is the correct thing to use for building an error modal component that can be used by any step in the Make Offer flow.

The Offer component should only care about being able to show an error modal and tell it what to display. However, this component is now setup as a consumer of the error modal's Context to populate its props that inherit from the error modal's props, which seems more complex than it needs to be.

ds300: Good point! I'll change it back to only caring about the one `showErrorModal` prop.

src/Apps/Order/OrderApp.tsx
```diff
@@ -99,16 +127,63 @@ export class OrderApp extends React.Component<OrderAppProps, OrderAppState> {
...
+                showErrorModal={showErrorModal}
+                mediator={mediator}
+                {...props}
+              />
+            )
+          }}
+        </ContextConsumer>
+      )}
+    </ErrorModalContext.Consumer>
+  )) as any
```
damassi: Could this be wrapped at the [top-most route](https://github.com/artsy/reaction/blob/master/src/Apps/Order/routes.tsx#L64) (which then composes other child routes) skipping the need to wrap each individual route?

ds300:
closing in favour of the approach in #1772

PR #1735, opened 2018-12-19
[PATCH RelatedArtists] - minor link update

oxaudo:
For some reason linking was not working properly on staging. My suspicion is that this joining in onClick is iffy. So added a separate variable to use in onclick.


artsyit:

Built with commit 1730900aa049f3bb5bd536831a582c4b1a6bba25

https://deploy-preview-1735--artsy-reaction.netlify.com

PR #1739, opened 2018-12-20
Validate offer amount

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-721

@briansw we managed to get the input working like you wanted. i.e. preventing them from inputting anything except digits.

@pepopowitz I continued your work to factor out the input. Ended up keeping the state management in the index pages though, since I couldn't think of a way to push the state management out to a shared place without either duplicating it in the index pages anyway, or doing some over-engineering with render props and context and so forth. Let me know if you have ideas in that area! Again, this is something that I think Hooks would let us solve really nicely 😅 

bad-0

bad-no-input


artsyit:

Built with commit 13fe280426115a80600f5e95cf1b8759ad518dcc

https://deploy-preview-1739--artsy-reaction.netlify.com

ds300:
@sweir27 this should be 🍏

src/Apps/Order/Components/OfferInput.tsx
```diff
@@ -0,0 +1,42 @@
+import Input from "Components/Input"
+import React from "react"
+
+export interface OfferInputProps {
+  id: string
+  showError?: boolean
+  onChange: (value: number) => void
+}
+
+export class OfferInput extends React.Component<OfferInputProps> {
```
sweir27: I could see this (in the future) becoming a _type_ of input that we could reuse and inject with a unique error message. But this seems like a fine abstraction for now 👍

src/Apps/Order/Routes/__tests__/Offer.test.tsx
```diff
@@ -81,6 +79,35 @@ describe("Offer InitialMutation", () => {
...
+      component
+        .find(OfferInput)
+        .props()
+        .onChange(0)
+
+      expect(component.find(OfferInput).props().showError).toBe(false)
+
+      component.find(Button).simulate("click")
+
+      expect(component.find(OfferInput).props().showError).toBe(true)
```
sweir27: Instead of inspecting the props here, can we assert that an error message shows up with the text we expect?

pepopowitz:
@sweir27 "What does the interaction look like when you try to input a non-digit (i.e. a 1.50). Does it feel natural?" - 

This gif shows me entering the value 1a2b3cd4,.56 - you can see the cursor act as if I was typing something, but nothing actually shows up.

non-numeric

pepopowitz:
@sweir27 "In your second screenshot, it looks like it's showing an error when you send 12 as a value. Is this just because of odd data/state?"

This is definitely inconsistent, I'm not exactly sure what it's about. My initial assumption was that it was because it was in a storybook, but that doesn't appear to be the case - sometimes it works, sometimes it doesn't. 

When it doesn't, I get this error in my console - does it ring a bell?

```
Order/Routes/Offer/index.tsx | ErrorWithMetadata: Hello World
    at onCompleted (index.tsx?df81:109)
    at Object.onNext [as next] (commitRelayModernMutation.js?993d:71)
    at Object.next (RelayObservable.js?8fd7:561)
    at Object.eval [as next] (RelayObservable.js?8fd7:227)
    at Object.next (RelayObservable.js?8fd7:561)
    at Object.next (RelayObservable.js?8fd7:561)
    at eval (RelayObservable.js?8fd7:492)
    at _subscribe (RelayObservable.js?8fd7:609)
    at RelayObservable.subscribe (RelayObservable.js?8fd7:315)
    at Object.next (RelayObservable.js?8fd7:376) 
```

ds300:
> In your second screenshot, it looks like it's showing an error when you send 12 as a value. Is this just because of odd data/state?

Like steve said, that's because the gif was taken in storybook. I should have added a comment about that.

Still curious that it occasionally works if you keep trying. 🤔 Will investigate. (wouldn't happen in force though, because of our redirect rules)

src/Apps/Order/Routes/__tests__/Offer.test.tsx
ds300: done :+1:

ds300:
(This should be 🍏 again)

package.json
```diff
@@ -122,6 +122,7 @@
     "react-addons-test-utils": "^16.0.0-alpha.3",
     "react-docgen-typescript-loader": "^2.1.0",
     "react-dom": "^16.5.0",
+    "react-powerplug": "^1.0.0",
```
damassi: @ds300 - `react-powerplug` is indeed useful, but why not go with react upgrade for `<Toggle>` instead?

ds300: Yeah I wanted to. We need to upgrade force and volt first since they consume reaction, right?

ds300: Played around with this today. Turns out that storybooks story functions are not rendered as components but rather called as ordinary functions, so they don't support hooks anyway :cry: – and also after reading this https://github.com/reactjs/rfcs/pull/68#issuecomment-439314884 I'm not feeling too confident about using hooks in production code until the final release. So my motivation to upgrade react is flagging a bit.

damassi: no prob! thanks for the RFC link 👍

PR #1744, opened 2018-12-20
PURCHASE-613 - Adds make offer filter inputs

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-613
- Adds Make Offer filter inputs to the artist and collection pages

- Dependent on: 
https://github.com/artsy/metaphysics/pull/1445
https://github.com/artsy/gravity/pull/12058
https://github.com/artsy/metaphysics/pull/1447 
https://github.com/artsy/gravity/pull/12063

screen shot 2018-12-20 at 3 39 57 pm


artsyit:

Built with commit 30ba9715561bbe50a36a656c98466528ec508937

https://deploy-preview-1744--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -42,6 +42,7 @@ class Filter extends Component<Props> {
     return {
       hasForSaleArtworks: artist.counts.for_sale_artworks > 0,
       hasBuyNowArtworks: artist.counts.ecommerce_artworks > 0,
+      hasMakeOfferArtworks: artist.counts.has_make_offer_artworks,
```
jonallured: I'm unclear about this line. All the other `hasBLAHArtworks` keys in this object are being computed by comparing a count against zero - how is it that we have `artist.counts.has_make_offer_artworks` here as a boolean?

kierangillen: I chose to use a boolean because the count in all the other validations don't accurately compare to the actual scopes for that purchase type (e.g. the ecommerce count is checking only if the artworks have ecommerce, not inventory, for sale, etc). These count fields are only validating whether to gray out the input so it's not a big deal either way. But in the future we wouldn't want anyone to use them as actual counts for the artworks, so it's better to use a boolean.

jonallured:
Just paired with @kierangillen on this and was able to verify that we're sending (what we think!) is the right params to MP - here's some visual proof!
screen shot 2018-12-21 at 2 01 14 pm

jonallured:
Worked with @mzikherman to figure out what was missing on the MP side of things so I've opened https://github.com/artsy/metaphysics/pull/1452 and once that's merged I can finish this off.

jonallured:
@mzikherman I think this is GTG since https://github.com/artsy/metaphysics/pull/1452 is merged, but would love another look when you have a moment!! ❤️

mzikherman:
LGTM! We still need to add a `{enableMakeOffer && <CheckBox .... />}` kinda thing for the 'Make Offer' filter. The GraphQL fragments and queries, as well as the filter/state and component updates are fine to keep as-is.

jonallured:
@mzikherman ok, with 63e4f7c1cdeeb01133f7aab163f3b8c19f2d8c60 I've got a commit that takes a look at the env and if the `ENABLE_MAKE_OFFER` var is not set to `true` then it'll splice out the Make offer checkbox. I did a little refactoring here to have objects to construct this array of Checkboxes because I thought that was cuter.

Once we're ready to launch, then the idea would be to revert that single commit and we'd be good to go.

However, I had a hard time getting some typings to work the way I expected. In the one file [`filterState.setFilter`](https://github.com/artsy/reaction/pull/1744/commits/cad0ad25b7d40d17bcae853fea2ed8d079dc4b83#diff-e3efa0e5ac989dfe1a9073a3fc6c8384R178) doesn't seem to care about the string argument, but in the other file [`filters.setFilter`](https://github.com/artsy/reaction/pull/1744/commits/cad0ad25b7d40d17bcae853fea2ed8d079dc4b83#diff-fd83632a2a3cdecc3008126e997a8d63R40) throws a `Argument of type 'string' is not assignable to parameter of type '"major_periods"` type error - what gives??!!

jonallured:
Well looking at this with fresh eyes this morning seems to have cleared up some things for me! 😝 

I figured out that I could just add an interface to make TypeScript happy and then a couple test failures lead me to fix a few mistakes that I had made. This is green and good for another look @mzikherman!! ❤️

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -148,38 +152,50 @@ class Filter extends Component<Props> {
...
+        state: "at_auction",
+      },
+      {
+        hasWorks: this.existy.hasForSaleArtworks,
+        name: "Inquire",
+        state: "inquireable_only",
+      },
+    ]
+
+    if (ENABLE_MAKE_OFFER !== "true") {
```
mzikherman: I would avoid this check and just do a truthy check `if (ENABLE_MAKE_OFFER)`. That way, it can either be set to some value in Force env, or just be unset (`hokusai production unset ENABLE_MAKE_OFFER`), rather than needing to futz with a specific value. Just the presence should be sufficient.

This is also to avoid an annoyance between ENV vars (which all start out as strings), and the actual sharify values which already have a `JSON.parse` applied to them: https://github.com/artsy/force/blob/ab85589c1fd5c906b8561aadafacbc95250c90d2/src/desktop/config.coffee#L139

Also note that along with updating Reaction in Force with these changes, you'll need to add the `ENABLE_MAKE_OFFER` env var paperwork there (doesn't look like there is any such one yet, which makes me wonder how people are testing/QA'ing. Is there a lab feature or is it just assumed there will be no offerable works in production until its ready?).

jonallured: Gotcha - how does f5d30a5 look, is that what you were thinking?

jonallured: Oh and yes, I think you're right that there wasn't already a flag because production just doesn't have works in MO. So I guess this was the first change that required the flag!

jonallured:
Ok going to merge this, thanks everyone!! 🤗

PR #1746, opened 2018-12-21
[PATCH CurrentBidInfo] proper winning info

oxaudo:
I believe this should address https://artsyproduct.atlassian.net/browse/DISCO-598

This is a bit confusing, but I will get it right eventually:) So I believe that for winning info we need to look at `active_bid` and for your current max bid amount we should look at `most_recent_bid`. Those match in most winning cases except for the scenario when you go and increase max bid while your current bid is winning. In this case your most_recent_bid is not winning because it has not played yet (one would not bid against oneself) and the current active bid from you is still winning.


artsyit:

Built with commit 2ace05dbe0f3e627566559cbb29456d418141adf

https://deploy-preview-1746--artsy-reaction.netlify.com

PR #1736, opened 2018-12-19
Fetch awaitingResponseFrom in response from counter offer submit mutation

williardx:
Addresses [PURCHASE-764](https://artsyproduct.atlassian.net/browse/PURCHASE-764).

Fetches `awaitingResponseFrom` in the response for `CounterSubmitMutation` to refresh the cache. Without this, on successfully countering an offer the buyer is redirected from /status back to /respond since `awaitingResponseFrom` is stale.

PR #1748, opened 2018-12-21
Refetch data after accepting/declining an offer to avoid redirect loop

sweir27:
Follow up from: https://github.com/artsy/reaction/pull/1736 

We need to fetch this data here so when we advance to the next page it can correctly _not_ redirect back to the `/respond` page.

Addresses https://artsyproduct.atlassian.net/browse/PURCHASE-769

cc @alloy @mzikherman or others... I don't believe I should have so many changes to the `__generated__` files. I re-installed my node modules, but doesn't appear to have helped. Any ideas? I'm guessing I have an outdated version of... _something_ but I can't find the slack convo where we discussed this before 😞.

### DIFF
--- a/src/Apps/Order/Routes/Accept/index.tsx
+++ b/src/Apps/Order/Routes/Accept/index.tsx
@@ -69,6 +69,9 @@ export class Accept extends Component<AcceptProps, AcceptState> {
                     __typename
                     order {
                       id
+                      ... on OfferOrder {
+                        awaitingResponseFrom
+                      }
                     }
                   }
                   ... on OrderWithMutationFailure {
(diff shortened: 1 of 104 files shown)


artsyit:

Built with commit da04d43cb06eaf9a2a85b6a29f801e4d1b1e289b

https://deploy-preview-1748--artsy-reaction.netlify.com

alloy:
You should absolutely not have so many changes 😞

I don’t believe an actual understanding came out of previous discussions. Paging @orta, as I believe he was involved when it got resolved on @kierangillen’s machine.

Do you get any changes when you run `yarn relay`?

sweir27:
No changes on `yarn relay` but I'll see if either of them has any thoughts 😬.

ashkan18:
closing in favor of https://github.com/artsy/reaction/pull/1749 which is same PR without extra generated files.

kierangillen:
@sweir27 what happens if you run `yarn install` and then run `yarn relay`, also which node version are you using?

PR #1749, opened 2018-12-21
Follow up on #1748 to add awaitingResponseFrom

ashkan18:
Followup from #1748 where original PR had unrelated generated file changes.

> We need to fetch this data here so when we advance to the next page it can correctly not redirect back to the /respond page.

Addresses https://artsyproduct.atlassian.net/browse/PURCHASE-769


artsyit:

Built with commit 3c743093aea73c38dbc498ed1076bbc0a2475515

https://deploy-preview-1749--artsy-reaction.netlify.com

PR #1750, opened 2018-12-21
[FIX Router] Reject from build*App promises when GraphQL errors occur.

alloy:
This should lead to Force being able to handle these errors, which it already does well when it would occur on the server-side, but may need some more work when it were to occur on the client-side, e.g. https://github.com/artsy/force/blob/f4fd2da5eddf6821ecc3b23c174c000b745711e6/src/desktop/apps/artist/client.tsx#L28-L30


alloy:
These tests were a pain to figure out 😅

artsyit:

Built with commit a06ca0eb8007337c84b4dbef7103fb718f879f0e

https://deploy-preview-1750--artsy-reaction.netlify.com

src/Apps/Artist/routes.tsx
```diff
@@ -36,7 +37,7 @@ import { ShowProps } from "./Routes/Shows"
 // * `render` functions requires casting
 // * `Redirect` needs to be casted, as it’s not compatible with `RouteConfig`
 export const routes: RouteConfig[] = [
-  {
+  new Route({
```
alloy: Trying to think of how to _require_ the usage of this class 🤔

artsyit:

Built with commit 6ffe0b954a2fda93c29e6644a9bc07ec675ed4cc

https://deploy-preview-1750--artsy-reaction.netlify.com

src/Apps/Artist/routes.tsx
damassi: Is this totally required now? I know Taion mentioned in Slack that they wrapped their routes with a custom route to handle cases like this but I feel like we could find a way to hoist this up into some kind of factory function at the `buildClientApp` or `buildServerApp` level so that we're not introducing scope complexity and deviating from Found docs. 

Ideally we can keep this as close to Found as possible, or PR upstream, or find a way to hide this complexity at the framework level vs the `routes` definition.

src/Artsy/Router/Route.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { Spinner } from "@artsy/palette"
+import { LoadingClassName } from "Artsy/Relay/renderWithLoadProgress"
+import BaseRoute from "found/lib/Route"
+import React from "react"
+import styled from "styled-components"
+
+const SpinnerContainer = styled.figure`
+  width: 100%;
+  height: 100px;
+  position: relative;
```
damassi: `import { Spinner } from '@artsy/palette'`

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -27,16 +30,19 @@ interface Resolve {
 // No need to invoke this for each request.
 const MediaStyle = createMediaStyle()
 
-export function buildServerApp(
-  config: RouterConfig & { userAgent: string | undefined }
-): Promise<Resolve> {
+export interface ServerRouterConfig extends RouterConfig<RelayEnvironment> {
+  userAgent?: string
+}
```
damassi: 👍

src/Artsy/Router/Route.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    // This should only ever show when doing client-side routing.
+    if (!props) {
+      return (
+        <SpinnerContainer className={LoadingClassName}>
+          <Spinner />
+        </SpinnerContainer>
+      )
+    }
```
damassi: Does this work well with the `<PreloadLink>` router as seen on `/artist/:id`?

src/Artsy/Router/Route.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export class Route extends BaseRoute {
+  constructor(props) {
+    if (!(props.query || props.getQuery)) {
+      super(props)
+      return
+    }
+
+    super({
+      // getDataFrom: defaultGetDataFrom,
```
damassi: Should this have a `TODO`?

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -27,16 +30,19 @@ interface Resolve {
...
     new Promise(async (resolve, reject) => {
       try {
         const { context = {}, routes = [], url, userAgent } = config
         const user = getUser(context.user)
-        const relayEnvironment = createEnvironment({ user })
+        const relayEnvironment =
+          context.relayEnvironment || createEnvironment({ user })
         const historyMiddlewares = [queryMiddleware]
         const resolver = new Resolver(relayEnvironment)
         const render = createRender({})
```
damassi: Would it be possible (for consistency) to use our new `createRender` here?

src/Artsy/Router/index.ts
```diff
@@ -11,11 +12,11 @@ export { ContextProvider, ContextConsumer } from "Artsy/SystemContext"
 /**
  * Configuration used when creating a new Router app
  */
-export interface RouterConfig {
+export interface RouterConfig<E extends Environment = Environment> {
```
damassi: Why extend `Environment` here?

src/Apps/Artist/routes.tsx
alloy: What they [Taion] do is use JSX to define the routes, which also alleviates this issue and doesn't deviate from the Found docs. (It also replaces the word `children` for JSX nesting, which I personally find a little more intuitive, but not sure if that applies to others as well.)

I'm on the fence about either, if you feel strongly about staying with object literals and doing recursive work on each request, then I'll change it to do that.

src/Artsy/Router/Route.tsx
alloy: That'a above. This is just a container.

src/Artsy/Router/Route.tsx
alloy: There's one for it at the top of the file.

src/Artsy/Router/Route.tsx
alloy: From my understanding of what it should do, it works well.

src/Apps/Artist/routes.tsx
damassi: I do feel (slightly) strongly about it just based on previous experience working on big react router apps where JSX feels slighly more natural until the route tree gets to a certain size  and enough JS is interpolated inside of JSX props that an array of objects feels nicer, and in any case it's native JS so devs are already familiar. When `makeRouteConfig` takes in a JSX tree it's parsed and turned back into this anyway, and by us doing this behind the scenes it requires one less import on the devs side.

I guess what I'm saying is that JSX can quickly get unweildy for little gain and unless we're plan on nesting children 1000 levels deep I don't see any real perf implications.

src/Artsy/Router/index.ts
damassi: Not sure too sure! Was just curious. Having `E extends Environment` seems to imply that there's something more complex going on somewhere when really its just a setup issue for the type-checker.

src/Apps/Artist/routes.tsx
alloy: Got it, will update accordingly then 👍

src/Artsy/Router/Route.tsx
alloy: I did click around.

src/Artsy/Router/index.ts
alloy: Cleaned this up.

alloy:
Addressed all the feedback and will merge on green as this otherwise already received a 👍

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -52,7 +53,7 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
           getFarceResult({
             url,
             historyMiddlewares,
-            routeConfig: routes,
+            routeConfig: createRouteConfig(routes),
```
damassi: 👍

src/Artsy/Router/__tests__/buildClientApp.test.tsx
```diff
@@ -1,5 +1,5 @@
 import { ContextConsumer } from "Artsy"
-import { createEnvironment } from "Artsy/Relay/createEnvironment"
+import { createRelaySSREnvironment } from "Artsy/Relay/createRelaySSREnvironment"
```
damassi: 👍

PR #1756, opened 2018-12-27
[PATCH artwork] Fix bell icon

damassi:
Fix regression introduced here https://github.com/artsy/reaction/pull/1752 and addresses follow-up comments about checking if a work is an auction before evaluating.


artsyit:

Built with commit 44b1554b49df012ca2f7800ca80a3d1ea59eff40

https://deploy-preview-1756--artsy-reaction.netlify.com

PR #1767, opened 2018-12-28
Misc UI fixes for counteroffer pages

ds300:
fixes https://artsyproduct.atlassian.net/browse/PURCHASE-741

And a few other things:

- restructures the stories to be easier to browse
- fix a bunch of other spacing issues
- fix the helper placements
- add our responsive media framework's global css to the storybook setup

### DIFF
--- a/src/Apps/Order/Routes/Accept/index.tsx
+++ b/src/Apps/Order/Routes/Accept/index.tsx
@@ -214,13 +214,15 @@ export class Accept extends Component<AcceptProps, AcceptState> {
               <Flex flexDirection="column">
                 <Flex flexDirection="column">
                   <Media greaterThan="xs">
-                    <ArtworkSummaryItem order={order} />
+                    {className => (
+                      <ArtworkSummaryItem className={className} order={order} />
+                    )}
                   </Media>
                   <ShippingSummaryItem order={order} locked />
                   <CreditCardSummaryItem order={order} locked />
                 </Flex>
                 <Media greaterThan="xs">
-                  <Spacer mb={3} />
+                  <Spacer mb={2} />
                   <Helper artworkId={artwork.id} />
                 </Media>
                 <Media at="xs">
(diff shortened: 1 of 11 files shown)


artsyit:

Built with commit e7f912820e113dc5d63f7769c590f00ddb7e06a7

https://deploy-preview-1767--artsy-reaction.netlify.com

src/__stories__/config.js
```diff
@@ -1,13 +1,19 @@
...
 
 const req = require.context("../", true, /\.story\.tsx$/)
 
 function loadStories() {
   req.keys().forEach(filename => req(filename))
 }
 
+injectGlobal`
+  ${createMediaStyle()}
+`
```
damassi: ty!

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -214,18 +214,20 @@ export class Accept extends Component<AcceptProps, AcceptState> {
               <Flex flexDirection="column">
                 <Flex flexDirection="column">
                   <Media greaterThan="xs">
-                    <ArtworkSummaryItem order={order} />
+                    {className => (
+                      <ArtworkSummaryItem className={className} order={order} />
+                    )}
```
damassi: Out of curiosity, why was `className` needed here?

ds300: The `ArtworkSummaryItem` is based on the `StackableBorderBox` which needs to have other `StackableBorderBox`es as siblings in order for the interstitial borders to work correctly. `Media` adds an extra div around the artwork summary by default, which breaks the borders. So we needed to use the render props `Media` pattern to avoid that extra div layer and pass the className directly to the `StackableBorderBox`

damassi: word 👍 ty

PR #1769, opened 2018-12-28
[PATCH BidActions] - User should be able to watch live bidding when registration is closed and live bidding is in progress

oxaudo:
Keep slight difference in the language there but otherwise it's the same button for live auction regardless of one is registered or not.

Address https://artsyproduct.atlassian.net/browse/DISCO-617

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx
@@ -103,20 +103,16 @@ export class ArtworkSidebarBidAction extends React.Component<
           {({ user }) => {
             return (
               <Box>
-                {artwork.sale.is_registration_closed && !registeredToBid ? (
-                  <Button width="100%" size="large" mt={1} disabled>
-                    Registration closed
-                  </Button>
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit 1435564adc6d6ea22fb74219d1da41d6cdd5bc7c

https://deploy-preview-1769--artsy-reaction.netlify.com

damassi:
👍

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx
```diff
@@ -103,20 +103,16 @@ export class ArtworkSidebarBidAction extends React.Component<
...
-                )}
+                <Button
+                  width="100%"
+                  size="large"
+                  mt={1}
+                  onClick={() => this.redirectToLiveBidding(user)}
+                >
+                  {artwork.sale.is_registration_closed && !registeredToBid
+                    ? "Watch live bidding"
+                    : "Enter live bidding"}
```
damassi: @oxaudo - now that we're near completion we should make sure there are tests for these little bits. Can merge as is but lets follow up in a cleanup pass.

PR #1772, opened 2019-01-02
Make offer speed bumps

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-763

We needed something that looks just like the `ErrorModal`, but with an extra button. I could have added an extra button to the `ErrorModal` component, but that would have made for a confusing API. Additionally the name `ErrorModal` would no longer be appropriate. So I decided to make a more generic component called `ModalDialog`.

- `ErrorModal` was re-written in terms of `ModalDialog`.
- Created the `Dialogs` service to replace the work done in #1733 – this uses `unstated` to manage the modal state, and provides helper functions which abstract over the props and state management of the `ModalDialog` to provide different behaviours (currently only two behaviours, error messages (replacing `ErrorModal`) and an 'accept' dialog)
- Added the speed bumps for both the Offer page and the Respond page. Have simply duplicated the code for now because I couldn't think of an obvious way to share it between the pages. Open to ideas!

To reiterate: the approach to managing the modal dialogs in here replaces the one adopted in #1733. I'll close that PR and refactor the other page components to use `Dialogs` in a follow up PR.

offer-bumps

(ignore the errors, that's just because I was recording in storybook)


artsyit:

Built with commit f57ce547b9c530c705c204fc23ed525240aa930c

https://deploy-preview-1772--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Offer.test.tsx
```diff
@@ -171,19 +172,152 @@ describe("Offer InitialMutation", () => {
...
+          .last()
+          .simulate("click")
+
+        await flushPromiseQueue()
+        component.update()
+
+        const dialog = component.find(ModalDialog)
+        expect(dialog.props().show).toBe(false)
+        expect(commitMutation).toHaveBeenCalled()
+      })
```
damassi: 👍

src/Components/Modal/ModalDialog.tsx
```diff
@@ -0,0 +1,89 @@
+import { color, Flex, Sans } from "@artsy/palette"
```
damassi: All new components (for the time being) should go in `Styleguide/Components`. That's the current boundary for before / after our design system implementation. At a certain point we'll find a proper way to organize but that should happen in a refactor pass rather than within a feature.

damassi: Ah, I see - this has been grouped with other existing modal stuff.

src/Components/Modal/ModalDialog.tsx
```diff
@@ -0,0 +1,89 @@
...
+    secondary &&
+    css`
+      color: ${color("black30")};
+      &:hover {
+        color: ${color("black60")};
+      }
+    `};
+`
+
+// TODO: Generalize this button and move it to @artsy/palette
```
damassi: 👍 😉

src/Apps/Order/Dialogs.tsx
```diff
@@ -0,0 +1,157 @@
...
+// make it safe to inject dialogs on components nested within a page by
+// only injecting the actual modal component if it hasn't been injected already
+// otherwise you might end up with two or more instances of the modal appearing.
+const DialogContext = React.createContext({ isModalInTree: false })
+
+/**
+ * Injects the `dialogs` prop into the given page component
+ * @param Component
+ */
+export function injectDialogs<R extends { dialogs: DialogHelpers }>(
```
damassi: While I think this is a nice abstraction in terms of safety, I wonder if we might want to approach this from two (possibly) simpler angles: 
1) Move `<Subscribe to={[DialogContainer]}>` and `ModalDialog` into the [root `OrderApp`](https://github.com/artsy/reaction/blob/master/src/Apps/Order/OrderApp.tsx#L91) render method. Then, from some area of the subtree, we can show / hide by passing the dialog definition to unstated which will update and then render the modal directly from `OrderApp`. Then we never have to check for dups, or maintain the overhead of additional contexts. 
2) Pass a state change to unstated which then attaches a modal via a [portal](https://reactjs.org/docs/portals.html)

The above examples are just suggestions and I may be missing something. In any case the below is clear.

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -44,16 +44,14 @@ export interface RespondProps {
   mediator: Mediator
   relay?: RelayProp
   router: Router
+  dialogs: DialogHelpers
```
damassi: Maybe just `Dialogs`?

src/Apps/Order/Dialogs.tsx
```diff
@@ -0,0 +1,157 @@
...
+export class DialogContainer extends Container<DialogState> {
+  state: DialogState = {
+    props: {
+      show: false,
+      heading: null,
+      detail: null,
+      primaryCta: {
+        text: "",
+        action: () => void 0,
+      },
```
damassi: Any particular reason not to initialize `secondaryCta` and `onClose` here?

src/Apps/Order/Dialogs.tsx
```diff
@@ -0,0 +1,157 @@
...
+          primaryCta: {
+            text: continueButtonText,
+            action: accept,
+          },
+          secondaryCta: {
+            text: cancelButtonText,
+            action: reject,
+          },
+          onClose: reject,
+        },
```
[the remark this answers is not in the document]
ds300: Nope. It doesn't even shallow diff.

src/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarCommercial.test.tsx
```diff
@@ -18,7 +18,8 @@ import React from "react"
 import { commitMutation as _commitMutation, RelayProp } from "react-relay"
 
 import { ArtworkSidebarCommercial } from "Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial"
-import { ErrorModal, ModalButton } from "Components/Modal/ErrorModal"
+import { ErrorModal } from "Components/Modal/ErrorModal"
+import { ModalButton } from "Components/Modal/ModalDialog"
```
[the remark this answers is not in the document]
damassi: Yah, all the parts over the Palette.

src/Apps/Order/Dialogs.tsx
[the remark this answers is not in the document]
damassi: If not, I think this could lead to some confusion later. Might be better to just flatten things out or use something like [updeep](https://github.com/substantial/updeep) (though flattening would be better).

src/Apps/Order/Dialogs.tsx
```diff
@@ -0,0 +1,157 @@
...
+        text: "",
+        action: () => void 0,
+      },
+    },
+    forceCloseDialog: () => Promise.resolve(),
+  }
+
+  // closes the dialog without resolving consumers' promises
+  private closeDialog() {
+    return new Promise(resolve => {
```
damassi: I wonder if promises are necessary throughout this file and if the same behavior could be achieved in a synchronous way?

src/Apps/Order/Dialogs.tsx
damassi: Or better, rather than `showAcceptDialog` and `showErrorDialog` we can keep things generic and have the call-sites manage this info, instead passing down `showDialog`. If `error` and `accept` are used in numerous places then those values can be shared across files via a `constants` file (or something).

ds300: I had done 1 originally but couldn't get access to the modals in enzyme without wrapping the routes in an `<OrderApp>` which seemed to cause some unsavoury issues that I thought seemed more complex to tackle than doing it like this.

Now that you bring it up I do see a flaw in this pattern: if you have two siblings wrapped with injectDialogs but not the parent, then you still end up with duplicate ModalDialogs in the dom. Unlikely to happen and easy to spot during development, but still not ideal.

No idea about using a portal. I suspect that might cause enzyme troubles too?

I could try going back to something like 1 but abstracting the Subscribe+ModalDialog bit into a <ProvideDialogs> component that can be used easily in both OrderApp.tsx and test files. What do you think?

src/Apps/Order/Dialogs.tsx
ds300: `secondaryCta` is optional, and this initial value is not really used anyway* so there's no need to specify it. 

*It's only used so that we don't need to have logic for handling `null` or `undefined` when rendering the  actual `ModalDialog`

src/Apps/Order/Dialogs.tsx
ds300: `setState` is asynchronous and so is showing a user a dialog box and waiting for them to click somewhere, which are the two things I've used promises to model in this file. We'd probably be safe not waiting for setState to complete in the places it's used here, but better safe than sorry IMO. Other than that we could convert to callbacks which would mean we can test synchronously, but I really prefer promises as an API.

src/Apps/Order/Dialogs.tsx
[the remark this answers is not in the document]
ds300: Ha now I'm confused 🙃— what could be confusing later?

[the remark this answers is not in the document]
damassi: Meaning, if someone later called (hypothetically) `this.setState({ props: { heading: "fool" }})` and no change occurred. But if it were flattened and `props` was removed then it would trigger an update as expected.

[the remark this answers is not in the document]
ds300: I'm still not following. Calling `this.setState({ props: { heading: "foo" } })` would not work because `props` is of type `ModalDialogProps` and has a bunch of other required properties so TS would complain at you to add them. React doesn't merge state deeply, only at the top level.

src/Apps/Order/Dialogs.tsx
ds300: > Or better, rather than showAcceptDialog and showErrorDialog being managed at the state level, we can keep things generic and have the call-sites manage this info, instead passing down showDialog. If error and accept are used in numerous places then those values can be shared across files via a constants file (or something) from within the order app.

Oops sorry i missed this message earlier.

Yeah both are used in a bunch of places, we defo don't wanna be specifying the whole `ModalDialogProps` every time we show a dialog. So moving that formatting/interaction logic for the two types of dialog somewhere else would be good for tree shaking if the dialogs are ever used in other apps, but would also be a more cumbersome API as far as I can tell. Not worth it just yet IMO.

damassi: I would encourage keeping it simple and finding a way to stick with one main modal at the root and then use unstated to render with updated props / hide but not a blocker if you feel strongly about current approach 👍

src/Apps/Order/Dialogs.tsx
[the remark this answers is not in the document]
damassi: > React doesn't merge state deeply, only at the top level

That's what I'm saying -- namespacing the modal props under `props` breaks the default React state model since a change to a nested prop won't trigger an update (irrespective of TS validation). We should just remove the nested `props` object field and keep everything flat:

```js
state = {
  show: false,
  heading: null,
  detail: null,
  ... 
}
```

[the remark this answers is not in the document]
ds300: > a change to a nested prop won't trigger an update

Seems like an update is always triggered when you call `setState` unless you try to replace the `state` object with `null` or `undefined` —  👉 https://codesandbox.io/s/62098oo6mk

src/Apps/Order/Dialogs.tsx
ds300: Managed to do what you had in mind here.

- Got rid of the context business and figured out a way to insert the actual `ModalDialog` instance into OrderApp and also use it in tests. 
- Now the injected dialog helpers are just `show` and `hide` and all the logic about accept vs error is separate.

ds300:
Thanks for the awesome feedback on this @damassi ! :tada: :heart::heart:

src/Apps/Order/Dialogs.tsx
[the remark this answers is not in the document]
damassi: Ahh nice, thank you for this example 👍

src/Apps/Order/Dialogs.tsx
```diff
@@ -19,25 +19,76 @@ export class DialogContainer extends Container<DialogState> {
...
+      )}
+    </Subscribe>
+  )
+}
+
+export const ConnectedModalDialog = () => (
+  <Subscribe to={[DialogContainer]}>
+    {(dialogs: DialogContainer) => <ModalDialog {...dialogs.state.props} />}
+  </Subscribe>
+)
```
damassi: Perfect 👍

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -137,20 +142,20 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
...
+    return showAcceptDialog(this.props.dialog, {
       title: "Offer may be too low",
       message:
         "Offers within 20% of the list price are most likely to receive a response.",
     })
   }
 
   confirmOfferTooHigh() {
-    return this.props.dialogs.showAcceptDialog({
+    return showAcceptDialog(this.props.dialog, {
```
damassi: @ds300 - Following convention it seems like `injectDialog` would make these functions accessible via props (as they were before). I can see a dev in the future using `injectDialog` but then being confused about how to trigger behavior.

damassi: Also, it seems like a bit of indirection to pass an injected `dialog` prop into an imported function to display. Since we're using `injectDialog` it would be best to keep everything things constrained around a single point: 
```js
this.props.showAcceptDialog({
  title: 'Offer higher'
  ...
})
```
Otherwise there's a bit too much juggling.

src/Apps/Order/Routes/Respond/index.tsx
```diff
@@ -112,15 +117,15 @@ export class RespondRoute extends Component<RespondProps, RespondState> {
   }
 
   confirmOfferTooLow() {
-    return this.props.dialogs.showAcceptDialog({
+    return showAcceptDialog(this.props.dialog, {
```
damassi: Same here (and all further points)

PR #1775, opened 2019-01-02
Render SVG icons in Storybook

dblandin:
We have two methods of adding icons to Reaction:

- Via the `artsy-icons` font-family
- SVG-backed React components

Icons backed by `artsy-icons` are already rendered within Storybook under `Components -> Icons -> All Icons`.

This commit adds the SVG-backed React component icons to that Storybook page as well.

2019-01-02-161645_1189x274_scrot


artsyit:

Built with commit a1fc78efadf512e221571d06713aa61b203e7ae7

https://deploy-preview-1775--artsy-reaction.netlify.com

src/Components/__stories__/Icons.story.tsx
```diff
@@ -12,6 +13,54 @@ storiesOf("Components/Icons", module).add("All Icons", () => {
 
   return (
     <div style={{ margin: "20px" }}>
```
damassi: Might be better to use palette here and down. 

```jsx
import { Row, Col } from 'Styleguide/Elements/Grid'`

<Box m={2}>
 <Row>
    <Col p={1}>...</Col>
 </Row>
</Box>
```

dblandin: @damassi Let me know if I'm setting this up in a weird way. Had some trouble getting the column elements to wrap but seems to look all right in Storybook now.

damassi: One suggestion would be to just use flexbox (+ `flexWrap` property) and ditch the columns completely:

```js
const Icon = ({ children }) => <Box width='auto' ...other props...>{children}</Box> 
const icons = [Foo, Bar, Baz]

<Flex flexWrap="wrap">
  {icons.map(icon => <Icon>{icon}</Icon>}
</Flex>
```

dblandin: I'm encountering some difficultly aligning the icons to display in a neat grid with that approach. I'll start with this for now.

PR #1776, opened 2019-01-02
"Make Offer" analytics

pepopowitz:
A collaboration with @ds300. Addresses [PURCHASE-736](https://artsyproduct.atlassian.net/browse/PURCHASE-736).

This PR adds analytics coverage for Make Offer.

* Page views for all offer pages are tracked and tested.
* Clicks on the Buy Now/Make Offer buttons are tracked.
* Successfully submitted offers & counter-offers are tracked.
* Use of the `ContextConsumer` and `Mediator` objects was removed. Instead, we're calling `window.analytics` directly, via a HOC. We felt this was more appropriate for tracking these page-views. As a result, a follow-up once this is merged will be to remove [the code in Force that was listening to these mediator events](https://github.com/artsy/force/blob/master/src/desktop/apps/order2/client.js#L31).


artsyit:

Built with commit 0977c4575afed7d6ecc7507e3880e23e8d2ccb57

https://deploy-preview-1776--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -147,7 +150,12 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
       })
   }
 
-  handleCreateOrder = () => {
+  @track((props, state, args) => ({
+    action_type: Schema.ActionType.ClickedBuyNow,
+    flow: "buy now",
+    type: "button",
```
damassi: These strings should be moved into the Values file under `Schema`. So like:

```js
flow:  Schema.Flow.BuyNow,
type: Schema.Type.Button
```

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -209,7 +217,12 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
     })
   }
 
-  handleCreateOfferOrder = () => {
+  @track((props, state, args) => ({
+    action_type: Schema.ActionType.ClickedMakeOffer,
+    flow: "make offer",
+    type: "button",
```
damassi: Same here (and all further call-sites)

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -57,6 +60,10 @@ export class Accept extends Component<AcceptProps, AcceptState> {
     errorModalMessage: null,
   }
 
+  componentDidMount() {
+    trackPageView()
```
damassi: This could also be a decorator; e.g., 

```js
@track()
@trackPageView()
export class Accept extends Component {
  ...
}
```

pepopowitz:
@ds300 I committed changes to track the last two events (submitted_offer & submitted_counter_offer). 

There is an issue with the pageviews, though. With `trackPageView` moved to `OrderApp#componentDidMount`, the page view will only fire for the *first* page you go to in an order. The `OrderApp` component isn't unmounted/remounted as you step through the order/offer flow. I verified this behavior with analytics turned on locally.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -332,7 +345,7 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
             size="large"
             mt={1}
             loading={isCommittingCreateOfferOrderMutation}
-            onClick={this.handleCreateOfferOrder}
+            onClick={this.handleCreateOfferOrder.bind(this)}
```
damassi: `.bind` is a somewhat annoying detail regarding decorators

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -258,16 +258,8 @@ export class Accept extends Component<AcceptProps, AcceptState> {
...
-  <ContextConsumer>
-    {({ mediator }) => {
-      return <Accept {...props} mediator={mediator} />
-    }}
-  </ContextConsumer>
-)
-
 export const AcceptFragmentContainer = createFragmentContainer(
-  AcceptRouteWrapper,
+  trackPageViewWrapper(Accept),
```
damassi: Rather than wrapping here (and all below), can we just move this to the top as a decorator to keep things consistent since we're already using decorators? 

```js
@track()
@trackPageViewWrapper()
export class Accept extends React.Component { ... }
```

src/Apps/Order/Routes/Counter/index.tsx
```diff
@@ -230,16 +245,8 @@ export class CounterRoute extends Component<CounterProps, CounterState> {
...
-    {({ mediator }) => {
-      return <CounterRoute {...props} mediator={mediator} />
-    }}
-  </ContextConsumer>
-)
-
-export const CounterFragmentContainer = createFragmentContainer(
-  CounterRouteWrapper,
+export const CounterFragmentContainer = createFragmentContainer<CounterProps>(
+  trackPageViewWrapper(CounterRoute),
```
damassi: Use as `@trackPageViewWrapper()` at the top of the class

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -259,16 +258,8 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
...
-  <ContextConsumer>
-    {({ mediator }) => {
-      return <OfferRoute {...props} mediator={mediator} />
-    }}
-  </ContextConsumer>
-)
-
 export const OfferFragmentContainer = createFragmentContainer(
-  injectDialog(OfferRouteWrapper),
+  injectDialog(trackPageViewWrapper(OfferRoute)),
```
damassi: Speaking of which, `injectDialog` can also go up there:

```js
@track() 
@trackPageView()
@injectDialog()
export class OfferRoute extends React.Component { ... }
```

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -457,16 +452,8 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
...
-  <ContextConsumer>
-    {({ mediator }) => {
-      return <PaymentRoute {...props} mediator={mediator} />
-    }}
-  </ContextConsumer>
-)
-
 export const PaymentFragmentContainer = createFragmentContainer(
-  injectStripe(PaymentRouteWrapper),
+  injectStripe(trackPageViewWrapper(PaymentRoute)),
```
damassi: So can `injectStripe` 😄

src/Apps/Order/Utils/trackPageViewWrapper.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react"
+import { trackPageView } from "./trackPageView"
+
+export function trackPageViewWrapper<T>(
```
damassi: HOC's like this can be used as top-level decorators; since we're already using decorators best to keep things consistent.

src/Apps/Order/Routes/Offer/index.tsx
damassi: Note that the `trackPageView` and `injectDialog` decorators aren't being called (`()`) since they're not designed to accept configurtation. If they were, its just another level of currying:

```js
export function trackPageView(config) {
  return (ComponentToWrap) => {
     export class TrackPageView extends React.Component { ... }
  }
}
```

williardx:
Agree with @damassi that using `trackPageView` as a decorator would be cleaner, but otherwise looks good! Nice work @ds300 and @pepopowitz!

damassi:
@pepopowitz and I were pairing on this but were running into some type-checking issues around the decorator -- everything worked as expected! but type errors. Was hoping @ds300 might be able to help out on Monday, but for the time being this is totally sufficient to merge and iterate on!

Once the conflict is fixed feel free to merge 👍 

(A follow up PR with consistent decorators would be def appreciated -- might be cool do do a pair sesh so that we can better understand the TypeScript errors we were running into.)

pepopowitz:
Cool. The conflict has been resolved. It is late Friday, it's a non-trivial PR, and I am not in the office next week...so I'm going to avoid merging it myself. If someone wants to merge Monday morning, that'd be 🎉 !

PR #1784, opened 2019-01-03
Add Selected Career Achievements component (Artist Insights 2.0)

dblandin:
This commit adds the new Selected Career Achievements component for the Artists Insights 2.0 experiment.

ticket: https://artsyproduct.atlassian.net/browse/DISCO-535
spec: https://zpl.io/VDq4enq

2019-01-03-182615_1479x605_scrot
2019-01-03-182627_1470x668_scrot
2019-01-03-182635_1041x1083_scrot


dblandin:
A couple thoughts I have so far:

- It was a bit tricky to get the responsive column wrapping and the entity expansion to work together. I don't think this implementation perfectly adheres to the ordering found in the spec.

- Do we have a link or email address for the "let us know" CTA in the learn more modal?

Some TODOs:

- Still need to add the remaining icons and swap them in.

- Still need to enable/disable the different market insight / selected career achievement sections based on our A/B test selection.

cc/ @mzikherman @stephen-i-chan

src/Styleguide/Components/SelectedCareerAchievements.tsx
```diff
@@ -0,0 +1,308 @@
...
+        </Flex>
+        <Flex flexDirection="column">
+          <Box>
+            <Sans size="2">{label}</Sans>
+            {value && (
+              <Sans size="2" color="black60">
+                {value}
+              </Sans>
+            )}
+            {entities && entities.length && this.renderEntities(entities)}
```
damassi: Rather than plucking and passing `entities` into the `renderEntities` method, can it be plucked from within the method itself and thus keep everything co-located?

src/Styleguide/Components/SelectedCareerAchievements.tsx
```diff
@@ -0,0 +1,308 @@
...
+}
+
+export interface ArtistInsightProps {
+  type: string
+  label: string
+  value?: string
+  entities?: ReadonlyArray<string>
+}
+
+class ArtistInsight extends React.Component<ArtistInsightProps> {
```
damassi: This would probably be better as its own `ArtistInsight.tsx` file

src/Styleguide/Components/SelectedCareerAchievements.tsx
dblandin: I like it :+1:

src/Styleguide/Components/SelectedCareerAchievements.tsx
dblandin: Agreed! Moved that and the learn more modal into their own files.

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -78,6 +79,12 @@ class OverviewRoute extends React.Component<OverviewRouteProps, State> {
       <>
         <Row>
           <Col sm={colNum}>
+            {showMarketInsights && (
```
mzikherman: For now, we should make this admin only. Once we're satisfied that it looks good on the artist page and are ready to launch the AB test we can prepare a PR that swaps out the admin only guard w/ the correct AB stuff.

Also, we probably want to have an accompanying storybook for this component. So it might be easier to leave out any artist page integration in this PR, and just have this PR be the new component and storybook/tests. Then we can get easy visual QA in storybooks.

Then, we can do an admin-only guard and add it to the artist page in a later PR, and then after that we would need that last PR to remove the admin-only guard in favor of the AB test.

It may sound a bit tedious but I think it's better to get into a rhythm of small/iterative PR's, vs one big feature launch.

That might be overkill however, so I'll leave it up to you!

src/Styleguide/Components/ArtistInsight.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  render() {
+    const { label, value, entities } = this.props
+
+    return (
+      <Flex mt={1} alignItems="top" width={260}>
+        <Flex mr={1} flexGrow={0} alignItems="top">
+          <BlueChip />
```
mzikherman: cool, this can be passed in by the client or inferred by the 'type' of insight or whatever

src/Apps/Artist/Routes/Overview/index.tsx
dblandin: > Then, we can do an admin-only guard and add it to the artist page in a later PR, and then after that we would need that last PR to remove the admin-only guard in favor of the AB test.

That sounds great to me :+1:

src/Styleguide/Components/ArtistInsight.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </Flex>
+        <Flex flexDirection="column">
+          <Box>
+            <Sans size="2">{label}</Sans>
+            {value && (
```
mzikherman: If somehow `value` or `entities` was null/undefined/empty, you'd still render an icon (possibly, you may be refactoring that anyway). might be better to just be defensive with a `return null` and not rendering anything.

src/Styleguide/Components/ArtistInsightsModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+          </Serif>
+          <Serif
+            size="2"
+            mt={1}
+            textAlign="center"
+            color="black60"
+            lineHeight="24px"
+          >
+            Artist Insights exists to help art buyers understand the cultural
+            significance of artists on Artsy. We aim to contextualize artists
```
mzikherman: We should have someone (maybe from Editorial?) review the copy and make sure the grammar/language is correct.

artsyit:

Built with commit 99eec1f59b0dde3cc6e2cdca56d79d721f8a3b09

https://deploy-preview-1784--artsy-reaction.netlify.com

dblandin:
2019-01-08-175018_943x450_scrot
2019-01-08-175031_937x538_scrot

dblandin:
@mzikherman This should be ready for another :mag_right: !

src/Styleguide/Components/SelectedCareerAchievements.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <ArtistInsight
+        type={insight.type}
+        label={insight.label}
+        entities={insight.entities}
+      />
+    )
+  }
+
+  render() {
+    console.log(this.props)
```
mzikherman: stray log

src/Utils/getUser.ts
```diff
@@ -3,13 +3,15 @@ export function getUser(user: User | null | undefined): User | null {
 
   if (_user === undefined) {
     const id = process.env.USER_ID
+    const type = process.env.USER_TYPE || "User"
```
mzikherman: 👍

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -78,13 +80,28 @@ class OverviewRoute extends React.Component<OverviewRouteProps, State> {
...
-              <>
-                <MarketInsights artist={artist} />
-                <Spacer mb={1} />
-              </>
-            )}
-
+            <ContextConsumer>
+              {({ user }) => {
+                if (
+                  user.type === "Admin" &&
```
mzikherman: If the user is logged out, then `user` will be undefined so this should be `if(user && user.type === "Admin")`.

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -78,13 +80,28 @@ class OverviewRoute extends React.Component<OverviewRouteProps, State> {
...
         <Row>
           <Col sm={colNum}>
-            {showMarketInsights && (
-              <>
-                <MarketInsights artist={artist} />
-                <Spacer mb={1} />
-              </>
-            )}
-
+            <ContextConsumer>
```
mzikherman: Up to you, but to avoid this nesting and `ContextConsumer` component, can use a wrapper down below where the fragment container is being declared:

```
export const OverviewRouteFragmentContainer = createFragmentContainer(
  withContext(OverviewRoute),
...)
```

Then, you can just add a `user: User` prop to the props for this component, and access it via `this.props.user`.

mzikherman: (`withContext` is https://github.com/artsy/reaction/blob/a7b04e2adb6fad4f4c441e7d2e0e489694dcf849/src/Artsy/SystemContext.tsx#L74 )

dblandin:
@mzikherman Looks good via Force locally for the different permutations:

**Logged In Admin**

2019-01-08-184831_1295x694_scrot

2019-01-08-185014_476x705_scrot

---

**Logged In Non-Admin or Logged Out**

2019-01-08-184901_1300x478_scrot
2019-01-08-184934_534x700_scrot

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -255,6 +266,9 @@ export const OverviewRouteFragmentContainer = createFragmentContainer(
           }
         }
       }
+      insights {
```
mzikherman: Ah got it, so this is just used for an existy check (if the artist _just_ has these insights).

mzikherman:
Alright let's do it!

damassi:
@dblandin - Code looks 💯 Great work on this.

PR #1785, opened 2019-01-04
Visual pass at non auction sidebar

oxaudo:
This should address points that apply to non auction artworks from this issue: https://artsyproduct.atlassian.net/browse/DISCO-613

Tried to follow spec here: https://app.zeplin.io/project/5b113f413ede4bb10dbfdc42/screen/5c118d5babc861afbb29ce2d

The tricky part here is that bunch of those lines/spaces are conditional here. Spend a non trivial amount of time fiddling with spaces on containers how they were defined before. At the end I think the more sane strategy here is to let each content box to have no x spaces in themselves and do all the lines / horizontal spaces with Spaces and Separators. Ended up a bit more readable in my opinion.


artsyit:

Built with commit 24778f5a2e50d40961c20298f6d68234eb1eead1

https://deploy-preview-1785--artsy-reaction.netlify.com

damassi:
Looks great 👍

PR #1787, opened 2019-01-04
Update auto, should fix our custom tags

orta:
So we get https://github.com/intuit/auto-release/pull/128

PR #1789, opened 2019-01-04
Remove release label requirement

zephraph:
We interpreted the _onlyPublishWithReleaseLabel_ wrong. With this option set you have to include the actual release label for a release to happen. Not just the major, minor, patch label. 

Every since auto-release was updated in #1787 our deployments have stopped. Before that the patch versions were being released because the config wasn't reading properly and it was falling back to the default behavior of releasing everything as a patch. 

This should fix the releases not happening.

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -1,6 +1,5 @@
 {
   "no-version-prefix": true,
-  "onlyPublishWithReleaseLabel": true,
   "name": "Artsy",
   "email": "it@artsymail.com",
   "labels": {


zephraph:
merge on green

artsyit:

Built with commit 797de410af3aada9e40bd3fbd8e1f0f18f927ee2

https://deploy-preview-1789--artsy-reaction.netlify.com

orta:
I see, so we would have had to say "release" and "patch" to make it work with that option set, cool - thanks

PR #1790, opened 2019-01-04
Custom colors

eessex:
Sets up cleaner path for passing color and background color from the top level to all article components.  This PR covers custom colors in layout `feature`.

Also updates some components to use palette components/colors.

screen shot 2019-01-04 at 10 48 36

### DIFF
--- a/src/Components/Publishing/Sections/Author.tsx
+++ b/src/Components/Publishing/Sections/Author.tsx
@@ -1,91 +1,81 @@
-import { unica } from "Assets/Fonts"
+import { Box, Flex, Sans, space } from "@artsy/palette"
+import { pMedia } from "Components/Helpers"
+import Icon from "Components/Icon"
 import React from "react"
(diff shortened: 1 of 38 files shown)


artsyit:

Built with commit a80068967350e14aa433db8cc29cf904abbadea7

https://deploy-preview-1790--artsy-reaction.netlify.com

src/Components/Publishing/Sections/Author.tsx
```diff
@@ -94,3 +84,7 @@ const TwitterHandle = styled.a`
     margin: 0;
   }
 `
+
+Author.defaultProps = {
+  color: "black",
+}
```
yuki24: Could this be part of the function signature so we don't miss this declaration?

```tsx
export const Author: React.SFC<AuthorProps> = ({ color = "black",  author, ...props }) => {
  ...
}
```

PR #1807, opened 2019-01-07
Fix offer cancelled messages

ds300:
Hi! :wave:

https://artsyproduct.atlassian.net/browse/PURCHASE-784

This ticket was filed because the copy on the status page was wrong for cancelled offer orders. We were incorrectly handing the 'buyer_rejected' cancellation reason and not even handling any of the others. (see here for the full list https://github.com/artsy/exchange/blob/6fb90698b1c0c33c3a50341f460e6a90647f48d4/app/models/order.rb#L39)

Brian and Joanna had a sync and we decided to handle the following four reasons with specific copy: `buyer_rejected`, `seller_rejected`, `seller_lapsed`, and `buyer_lapsed`. The remaining reasons fall under `seller_rejected` for now.

So that's what I've added here.

In addition I decided to refactor the logic for deciding which copy to show on the status page. It was getting messy (which I believe was the reason we ended up handling `buyer_rejected` incorrectly in the first place) and these new requirements would have made it even messier. So apologies for the big diff. Hopefully it will be a little easier to see what's going on now.

Finally I noticed that we weren't handling the `REFUNDED` state 🙈 so I did that too (it's the same as the BN `CANCELED` status page).

### DIFF
--- a/src/Apps/Order/Routes/Status/index.tsx
+++ b/src/Apps/Order/Routes/Status/index.tsx
@@ -26,9 +26,11 @@ import { ShippingSummaryItemFragmentContainer as ShippingSummaryItem } from "../
 
 const logger = createLogger("Order/Routes/Status/index.tsx")
 
-interface StatusCopy {
+interface StatusPageConfig {
   title: React.ReactNode
   description: React.ReactNode
+  // default showTransactionSummary is true
+  showTransactionSummary?: boolean
 }
 
 export interface StatusProps {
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit bec6ede43e5dfc303e3cefd0323c2e3d5dd82f2c

https://deploy-preview-1807--artsy-reaction.netlify.com

ashkan18:
this looks great! thanks for refactoring, really easy to read now. Should we merge this before the release today?

ashkan18:
merging to get it ready for the release

PR #1809, opened 2019-01-07
[WIP] Visual QA fixes

williardx:
[PURCHASE-766](https://artsyproduct.atlassian.net/browse/PURCHASE-766)

This PR fixes a few small visual QA items. Screenshots here are after correction, while before correction can be found in the ticket.

WIP while waiting on answers to some questions.

1. Too much space between input field and summary on payment step @ XS.

screen shot 2019-01-07 at 12 27 21 pm

2. Dialog scrim should be `rgba(0,0,0, .6)`.

screen shot 2019-01-07 at 3 53 45 pm

3. Add space between two paragraphs on offer declined status page.

screen shot 2019-01-07 at 3 55 31 pm


src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -142,6 +142,7 @@ const offerMessages = {
           Thank you for your response. The seller will be informed of your
           decision to end the negotiation process.
         </p>
+        <br />
```
williardx: For some reason there's a space in Storybook but this `br` is needed here to create a space when running the full app. In Storybook, this will appear larger than it will appear in the app.

artsyit:

Built with commit 1ec6e19681a06f1432fa8caae78fc960fe297ccc

https://deploy-preview-1809--artsy-reaction.netlify.com

src/Apps/Order/Routes/Status/index.tsx
ds300: I looked into this. Indeed force is injecting extra css. It's from a file called `main_layout.styl`. Not sure what to do about it, but whatever we do it should result in having things look identical in both storybook and force.

@damassi @zephraph could use your input on this. Does it make more sense to remove the global styles from the order app config in force, or copy them into reaction's storybook setup?

zephraph: Oof, that's a challenge. Removing it all together might not be good because we're not sure what it could effect. Copying the styles over to storybooks doesn't feel great, but it might be the best solution.

PR #1815, opened 2019-01-07
Grab circle jobs from yarn orbs

zephraph:
I recently learned that jobs can actually be shared via orbs. This will vastly simplify our configs and the duplication we have across them. Now if only they shared workflows....


artsyit:

Built with commit 82ca971887d28216efebdc3da3b50347571c8a51

https://deploy-preview-1815--artsy-reaction.netlify.com

PR #1845, opened 2019-01-10
Add MO speedbump analytics

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-791

Only issue is that I couldn't figure out how to use `mockTracking` and `MockBoot` together so had to figure out a new way of mocking the event tracker.

### DIFF
--- a/src/Apps/Order/Routes/Offer/index.tsx
+++ b/src/Apps/Order/Routes/Offer/index.tsx
@@ -8,7 +8,8 @@ import { TransactionDetailsSummaryItemFragmentContainer as TransactionDetailsSum
 import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
 import { Dialog, injectDialog } from "Apps/Order/Dialogs"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
-import { ActionType, Flow, track } from "Artsy/Analytics"
+import { track } from "Artsy/Analytics"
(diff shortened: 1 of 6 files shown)


artsyit:

Built with commit b99d124a16c763356d6a0f8c78e7966d053ebd28

https://deploy-preview-1845--artsy-reaction.netlify.com

src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -8,6 +8,7 @@ import { TransactionDetailsSummaryItemFragmentContainer as TransactionDetailsSum
 import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
 import { Dialog, injectDialog } from "Apps/Order/Dialogs"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
+import { ActionType, Flow, track } from "Artsy/Analytics"
```
damassi: Conventionally this is imported as `import * as Schema`

williardx:
As an FYI I'm no longer the best person to review code, so I'll defer to @damassi and @ashfurrow's judgment here 😄

src/Apps/Order/Routes/Offer/index.tsx
ds300: :+1: I changed this

Although I'm no longer a fan of `import * as X` since vscode got auto-import but it only works with named imports :cry:

PR #1854, opened 2019-01-11
Incremental relay mocking fixes

ds300:
This replaces #1645 with an incremental approach to adopting a simpler relay data mocking method in tests and stories.

# How is it simpler?

## Disables some nasty mocking behaviour by relay-mock-network-layer

Remember seeing 'Hello world' all over the place when things go wrong? Remember unspecified union types being nodeterminstically resolved with random values? Well now you can forget about that mess!

## Easier mock data injection API

With the current way of mocking data, you need to mock the resolver functions for *types*.

e.g.

```ts
renderRelayTree({
  Component: Genes,
  query: graphql`
    query Genes_Test_Query {
      artist(id: "pablo-picasso") {
        ...Genes_artist
      }
    }
  `,
  mockResolvers: {
    Artist: () => GenesFixture,
  },
})
```

With this new way, you can mock *query results*, which is a lot easier to reason about.

```ts
renderRelayTree({
  Component: Genes,
  query: graphql`
    query Genes_Test_Query {
      artist(id: "pablo-picasso") {
        ...Genes_artist
      }
    }
  `,
  mockData: {
    artist: GenesFixture,
  },
})
```

## Error messages for missing data

We want to make sure that data injected in tests represents the data requested in production.

So now you might see a message like this when running a test or a story: 

```
RelayMockNetworkLayerError: A mock for field at path 'order/myLastOffer/createdAt' of type 'String' was expected but not found.
```

# Why incremental?

I want to merge this now to make it available for new tests, and to avoid a long-lived branch that touches the entire code-base :scream:


artsyit:

Built with commit 22f142991891970fb1f47eb5e89c390d6a0150c0

https://deploy-preview-1854--artsy-reaction.netlify.com

patches/relay-mock-network-layer+2.0.0.patch
```diff
@@ -0,0 +1,45 @@
...
+ 
+     /** https://www.apollographql.com/docs/graphql-tools/mocking.html#Mocking-interfaces */
+     preserveResolvers = false,
+@@ -51,7 +52,9 @@ module.exports = function getNetworkLayer({
+         });
+ 
+         // Add mocks, modifies schema in place
+-        addMockFunctionsToSchema({ schema: executableSchema, mocks, preserveResolvers });
++        if (mocks) {
++            addMockFunctionsToSchema({ schema: executableSchema, mocks, preserveResolvers, defaultFieldResolver: fieldResolver });
```
ds300: this is where the weird mocking behaviour gets injected, so now we can avoid it by not passing the `mocks` fn

patches/relay-mock-network-layer+2.0.0.patch
```diff
@@ -0,0 +1,45 @@
...
+ 
+-        return graphql(executableSchema, query, null, null, variableValues).then(
++        return graphql(
++            executableSchema,
++            query,
++            null,
++            null,
++            variableValues,
++            null,
++            fieldResolver
```
ds300: We also need to pass a field resolver into graphql to let us throw errors for missing data, mock __id automatically, and handle aliasing.

src/Apps/Order/Routes/__tests__/Shipping.test.tsx
```diff
@@ -98,27 +98,26 @@ describe("Shipping", () => {
       expect(config.variables.input.orderId).toBe("1234")
     })
 
+    fillAddressForm(component, validAddress)
+
```
ds300: A buuunch of the tests in this file were passing before this PR when they shouldn't have been. Many of them because they forgot to fill in the address form before clicking submit. I haven't had time to look into why that was but it is very surprising indeed.

src/DevTools/createMockNetworkLayer/index.ts
```diff
@@ -12,3 +18,66 @@ export const createMockNetworkLayer = (mockResolvers: IMocks) => {
...
+        complain(info, info.returnType.inspect())
+      }) as GraphQLFieldResolver<any, any>,
+      schema,
+      resolvers: {
+        FormattedNumber: () => FormattedNumber,
+        Query: Object.entries(queryData).reduce(
+          (acc, [k, v]) => ({
+            ...acc,
+            [k]: typeof v === "function" ? v : () => v,
+          }),
```
ds300: here it converts e.g. `{ order: foo, artist: bar }` to `{ order: () => foo, artist: () => bar }` unless you already gave functions in the first place.

alloy:
Solid :sparkles:

patches/relay-mock-network-layer+2.0.0.patch
damassi: @ds300 - should this be PR'd upstream?

src/DevTools/MockRouter.tsx
```diff
@@ -1,6 +1,9 @@
 import { RouterConfig } from "Artsy/Router"
 import { buildClientApp } from "Artsy/Router/buildClientApp"
-import { createMockNetworkLayer } from "DevTools/createMockNetworkLayer"
+import {
+  createMockNetworkLayer,
+  createMockNetworkLayer2,
```
damassi: What is the path forward / level of effort to unify this code around a single file?

ds300: We need to find all usages of `mockResolvers` and replace with `mockData` and `mockMutationResults`. It can be quite fiddly and there may be edge cases we haven't accounted for yet. Also for stories you need to manually check that they still work, though it looks like there's only 5 story files using `mockResolvers` after this gets merged.

So I'd guess it's 1-3 days of work for a focused engineer, depending on how smoothly things go.

ds300:
# Guide to using `mockData` and `mockMutationResults`

## The basics 
If your are testing a component that makes a query to metaphysics, you can use the `renderRelayTree` helper to mock out the data that would be returned by metaphysics. Previously this was done with the `mockResolvers` prop, but now we have an easier and less confusing way: `mockData` and `mockMutationResults`. The idea is that these properties should match the exact data shape that metaphysics would return.

e.g. if you have a component with a query fragment like so:

```graphql
fragment ArtistDetails_artist on Artist {
  name
  birthday
  deathday
}
```

You would do this:

```ts
await renderRelayTree({
  Component: ArtistDetailsFragmentContainer,
  query: graphql`
    query ArtistDetailsTestQuery {
      artist(id: "unused") {
        ...ArtistDetails_artist
      }
    }
  `,
  mockData: {
    artist: {
      name: "Banksy",
      birthday: "unknown",
      deathday: null,
    },
  },
})
```

It handles field aliases just like you'd expect. So if you have

```graphql
fragment ArtistDetails_artist on Artist {
  birthYear: birthday(format: "YYYY")
  birthdayFull: birthday(format: "Do MM YYYY")
}
```

Then your `mockData` can be like

```ts
mockData = {
  artist: {
    birthYear: "1987",
    birthdayFull: "12th April 1987"
  }
}
```

Similarly, if you have a component that does mutations and you want to mock the return result of that mutation, you can do it in the same way but using the `mockMutationResults` prop.

```ts
await renderRelayTree({
  Component: ArtistDetailsFragmentContainer,
  query: graphql`
    query ArtistDetailsTestQuery {
      artist(id: "unused") {
        ...ArtistDetails_artist
      }
    }
  `,
  mockData: {
    artist: {
      name: "Banksy",
      birthday: "unknown",
      deathday: null,
    },
  },
  mockMutationResults: {
    // this mutation doesn't actually exist, it's just an illustration
    setArtistBirthday: { 
      // remember this represents the output that is returned from the mutation request
      artist: {
        name: "Banksy",
        birthday: "2019-01-14T12:38:42.931Z",
      },
    },
  },
})
```

## Dealing with union types

Union types should be disambiguated using the `__typename` property. e.g. for a response from an `/order` app mutation, do something like this:

```ts
mockMutationResults = {
  ecommerceSetOrderShipping: {
    orderOrError: {
      __typename: "OrderWithMutationSuccess",
      order: {
        ...BuyOrderWithShippingDetails,
      },
    },
  },
}
```

or this:

```ts
mockMutationResults = {
  ecommerceSetOrderShipping: {
    orderOrError: {
      __typename: "OrderWithMutationFailure",
      error: {
        type: "validation",
        code: "credit_card_not_found",
        data: '{"credit_card_id":"5b9987f72957190026d0ff54"}',
      },
    },
  },
}
```

## Using functions to return different results

Both `mockData` and `mockMutationResults` can take functions as top-level values, to return different data based on query variables.

```ts
mockMutationResults = {
  ecommerceSetOrderShipping: (_, { input }) => {
    if (input.shipping.country === "US") {
      return {
        orderOrError: {
          __typename: "OrderWithMutationSuccess",
          order: {
            ...BuyOrderWithShippingDetails,
          },
        },
      }
    } else {
      return {
        orderOrError: {
          __typename: "OrderWithMutationFailure",
          error: {
            type: "validation",
            code: "address_outside_usa",
            data: null,
          },
        },
      }
    }
  },
}
```

damassi:
@ds300 - would you mind moving ^ into a README.md file in the `<DevTools>` folder?

PR #1858, opened 2019-01-11
Modal: Support background scroll locking and close via escape key

dblandin:
- Disable scrolling of background content while modal is open 

    This commit disables vertical scrolling of the background content while the modal is open. This covers scrolling while the cursor is positioned off of the modal window and for smaller viewports, scrolling beyond the modal content, which would also cause the background content to scroll.

- Close open modal via escape key

    This commit sets up an event listener while the modal is open to close
the modal if the escape key is pressed.

ticket: https://artsyproduct.atlassian.net/browse/DISCO-674

### DIFF
--- a/src/Components/Modal/ModalWrapper.tsx
+++ b/src/Components/Modal/ModalWrapper.tsx
@@ -33,6 +33,8 @@ injectGlobal`
   }
 `
 
+const KEYBOARD_EVENT = "keyup"
+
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 463ce5cf9eb3161aef438d806abd4c7d82548ff1

https://deploy-preview-1858--artsy-reaction.netlify.com

PR #1860, opened 2019-01-12
Update condition for when to show Artist Insights V2

dblandin:
Instead of the current condition:

Show Artist Insights V2 if V1 or V2 data is present

The condition should be:

Show Artist Insights V2 if V1 data is present

This matches the acceptance criteria of:

> Test should be limited to only our artist pages that currently have Artist Insight V1 badges

ticket: https://artsyproduct.atlassian.net/browse/DISCO-657


artsyit:

Built with commit 1d5a69ec3c23e819f584100598bf7902e116da79

https://deploy-preview-1860--artsy-reaction.netlify.com

PR #1875, opened 2019-01-14
Refactor error modals

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-792 

Finishing up the ErrorModal -> ModalDialog refactor.

Can't wait to clean up those tests 💆‍♂️

### DIFF
--- a/src/Apps/Order/Routes/Counter/index.tsx
+++ b/src/Apps/Order/Routes/Counter/index.tsx
@@ -12,10 +12,10 @@ import {
 import { ShippingSummaryItemFragmentContainer as ShippingSummaryItem } from "Apps/Order/Components/ShippingSummaryItem"
 import { TransactionDetailsSummaryItemFragmentContainer as TransactionDetailsSummaryItem } from "Apps/Order/Components/TransactionDetailsSummaryItem"
 import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
+import { Dialog, injectDialog } from "Apps/Order/Dialogs"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
 import { track } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
-import { ErrorModal } from "Components/Modal/ErrorModal"
 import { CountdownTimer } from "Components/v2/CountdownTimer"
 import { Router } from "found"
 import React, { Component } from "react"
@@ -35,13 +35,11 @@ export interface CounterProps {
   order: Counter_order
   relay?: RelayProp
   router: Router
+  dialog: Dialog
 }
 
 export interface CounterState {
   isCommittingMutation: boolean
-  isErrorModalOpen: boolean
-  errorModalTitle: string
-  errorModalMessage: string
 }
 
 const logger = createLogger("Order/Routes/Counter/index.tsx")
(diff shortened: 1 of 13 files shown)


artsyit:

Built with commit 45ab0bd9b99e6ef4c823c681fb8297f421a3a05a

https://deploy-preview-1875--artsy-reaction.netlify.com

src/Apps/Order/Routes/Counter/index.tsx
```diff
@@ -102,13 +97,11 @@ export class CounterRoute extends Component<CounterProps, CounterState> {
     })
   }
 
-  onMutationError(errors, errorModalTitle?, errorModalMessage?) {
+  onMutationError(errors, title?, message?) {
     logger.error(errors)
+    this.props.dialog.showErrorDialog({ message, title })
```
damassi: 👍

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -240,19 +233,14 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
...
-    errorModalTitle?,
-    errorModalMessage?,
-    errorModalCtaAction?
-  ) {
+  onMutationError(error, title?, message?, onContinue?) {
     logger.error(error)
+    this.props.dialog
+      .showErrorDialog({ message, title })
+      // tslint:disable-next-line:no-empty
+      .then(onContinue || (() => {}))
```
damassi: Ah, now I understand this promise api 😄

sweir27:
This PR looks 🎉 ! One question since I'm coming at these conversations late-- it looks like the `Dialog` component embeds some assumption that it'll be an `error` dialog. Anticipating the eventual desire to move these to [palette](https://github.com/artsy/palette/issues/82), do you feel like we should make any changes now? Or it's in a good place to be eventually refactored?

Apologies if you've already thought about this/commented elsewhere!

ds300:
@sweir27  The `ModalDialog` is good to move to palette as-is. That doesn't have any error-specific code.

Not sure it makes sense to move the state/context management stuff (in `Dialogs.tsx`) to palette at this point. I expect it to be a generally useful pattern, but it feels premature to decide it's the one true way. So I'd prefer to wait and see what requirements emerge in other parts of artsy.net.

src/Apps/Order/Routes/__tests__/Accept.test.tsx
```diff
@@ -207,18 +209,25 @@ describe("Accept seller offer", () => {
...
+      const errorComponent = component.find(ModalDialog)
       expect(errorComponent.props().show).toBe(true)
       expect(errorComponent.text()).toContain("An error occurred")
       expect(errorComponent.text()).toContain(
         "Something went wrong. Please try again or contact orders@artsy.net."
       )
 
       component.find(ModalButton).simulate("click")
-      expect(component.find(ErrorModal).props().show).toBe(false)
+
```
pepopowitz: These bits (lines 223-227) are repeated throughout the tests - we 1. render the component, 2. verify the modal shows up, 3. click the button to dismiss the modal, then 4. verify that the modal is gone. Is there value in repeating steps 3&4 in almost every test, or would it make more sense to write a single test against `injectDialog` that verifies this instead?

damassi:
> state/context management
I agree that this seems a bit premature, might be good just to get the component portion in which we can then populate with data

src/Apps/Order/Routes/__tests__/Accept.test.tsx
ds300: Yeah that would be good. Shall we do it as a follow up or would you prefer to have it taken care of in this PR?

pepopowitz: Follow-up would be great!

pepopowitz:
This looks great!!!! Anything remaining with comments above can be handled as a follow-up.

PR #1883, opened 2019-01-15
Intercom

oxaudo:
Not sure that is a kosher solution here, but I think it will work. Basically if I understood @mzikherman correctly - the idea is to just fire an event on mediator from this side that Force will subscribe to and all the intercom work will be done from there.


artsyit:

Built with commit 455a528fafcc4f81fcc3eee0672027859f819ae9

https://deploy-preview-1883--artsy-reaction.netlify.com

damassi:
@oxaudo - cool, did you see the intercom chat window appear?

PR #1889, opened 2019-01-15
clean up the mediator call for launching intercom and pass artwork to it

oxaudo:
Second try here. I'm a bit confused why CLIENT is not available on force side by default, but seems like for the Intercom we only need is_acquireable and is_offerable. So I think it's ok to just pass those in with the mediator call


oxaudo:
And thanks to @damassi help I actually got to see the intercom locally now!

oxaudo:
Self merging it to test on staging

PR #1897, opened 2019-01-16
[Refactor] Remove <Tabs> and <Toggle> and move to Palette

damassi:
Follow up PR to https://github.com/artsy/palette/pull/191

Moves Tabs and Toggle to Palette. 

@anandaroop and @eessex - tagging y'all because I saw that positron and volt use the Tabs component. 

(For some reason the NPM registry isn't showing that `palette@2.26.2` is available, though it shows that that's them most recent release on npmjs.org...not too sure, will check tomorrow.)


damassi:
(going to merge to keep this moving along)

PR #1900, opened 2019-01-16
Deprecate Grid

eessex:
- Replaces last instance of `Grid` in repo
- Removes `Grid` component and story
- Removes `react-flexbox-grid` library

Use of the `Grid` component has been removed from Force, Positron and this repo -- I believe we are all good to deprecate fully.


damassi:
👍

PR #1924, opened 2019-01-22
remove partner icon for artwork details if artwork is from benefit auction

oxaudo:
Ui change to accomplish https://artsyproduct.atlassian.net/browse/DISCO-456.
MP with addition of `is_benefit` to sale was deployed to production.


artsyit:

Built with commit 12a6537d07ea00777ed826bf483febcd3a440d62

https://deploy-preview-1924--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx
```diff
@@ -81,7 +82,7 @@ export class ArtworkDetailsAboutTheWorkFromPartner extends React.Component<
                   }
                   meta={locationNames}
                   imageUrl={imageUrl}
-                  initials={partner.initials}
+                  initials={showImage && partner.initials}
```
damassi: This is a little confusing to read ("if there's an image, show the initials") -- can we add a comment describing the behavior?

oxaudo: Sure, it is confusing a bit because the default for the Avatar is if either initials or imageUrl is passed - display logo. I think it might be lass confusing if I keep those in a separate variable up top. Will update. One min.

PR #1928, opened 2019-01-23
[WIP] Start on adding view in room to artwork page v2

mdole:
@oxaudo, @zephraph, and myself are working on porting the old View in Room function to the new artwork page. This is a start on the Reaction side of things—there's a sister [WIP Force PR](https://github.com/artsy/force/pull/3388) as well.

There are a few TODOs left here, the biggest of which is figuring out how to actually pass the image to Force since the old View in Room functionality actually expects the full image tag and not just a src url. Justin suggested passing it using the component that wraps ArtworkActions as a prop; I think that's ArtworkActionsFragmentContainer.

One possible hiccup here: the artwork page v2 doesn't actually use an `` tag for the image; it's a `<div>`. Not sure if that means we need to rethink the strategy a bit.

Tracked: [DISCO-465](https://artsyproduct.atlassian.net/browse/DISCO-465)

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
@@ -88,11 +88,6 @@ export class ArtworkActions extends React.Component<
         dimensions,
         image,
       })
-
-    // Still to do:
-    // - Use correct eye button (being updated in palette; we might not even have to change the code if we're lucky)
-    // - Define mediator behavior in Force
-    // - Fix unhappy onClick (line 121)
   }
 
   render() {
@@ -109,14 +104,13 @@ export class ArtworkActions extends React.Component<
               artwork={this.props.artwork}
               render={Save(this.props)}
             />
-            // Remove the admin flag for production
             {artwork.is_hangable &&
               this.isAdmin && (
                 <ContextConsumer>
                   {({ mediator }) => (
                     <UtilButton
                       name="viewInRoom"
-                      onClick={this.openViewInRoom(mediator)}
+                      onClick={() => this.openViewInRoom(mediator)}
                     />
                   )}
                 </ContextConsumer>
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
@@ -22,6 +22,7 @@ import {
   HeartIcon,
   Join,
   Link,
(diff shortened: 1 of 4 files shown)


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
```diff
@@ -77,6 +78,23 @@ export class ArtworkActions extends React.Component<
...
+      mediator.trigger &&
+      mediator.trigger("openViewInRoom", {
+        dimensions,
+        image,
+      })
+
+    // Still to do:
+    // - Use correct eye button (being updated in palette; we might not even have to change the code if we're lucky)
+    // - Define mediator behavior in Force
+    // - Fix unhappy onClick (line 121)
```
oxaudo: What is 'unhappy onClick'?

oxaudo: For the force side - I think we need to look what it actually does with the image. Maybe we can get away with modifying code a bit and based on ImageId - fetch the image and pass it to view in a room.

mdole: Ah sorry for being unspecific about the onClick issue—I'm seeing this error from typescript: [screenshot](https://cl.ly/d6e32ab36ef8)

oxaudo: As discussed in person - it expects a function itself where we are doing a direct function call which just returns void. I thing we got it corrected:) So I think we can merge that once it's all green.

mdole: yep! just pushed a commit with that fix. tests and storybook look good (except for the old eye icon)—note that to see it in the storybook, you'll need to remove the is_hangable check

artsyit:

Built with commit ba32e506a03131aa48f97b31905ae4c49c9a1cdc

https://deploy-preview-1928--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
oxaudo: @mdole - We actually need to add proper test case and update the fixture to show it in storybooks. I'm updating this file slightly so will deal with that as well.

mdole:
@oxaudo sounds good. If y'all are working on resolving I'd love to look over your shoulders and see what's going on!

PR #1929, opened 2019-01-23
Pull `lastOffer` from `OrderOffer` instead of `Order`, due to deprecation

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/PURCHASE-810.

We want to remove the `lastOffer` and `offers` fields from the `Order` type in exchange/metaphysics, as they have been moved to the `OrderOffer` type instead. In order to do this, we need to remove any last references to these fields on the `Order` type. This PR accomplishes that in reaction.


ashkan18:
Awesome! Thanks! 🙌

PR #1930, opened 2019-01-23
pass image url when calling viewInRoom mediator

oxaudo:
Left one of the tests commented out because it needs palette update: https://github.com/artsy/palette/pull/218


artsyit:

Built with commit 15f032eef9db1c3714416666245ca870c16760df

https://deploy-preview-1930--artsy-reaction.netlify.com

oxaudo:
@mdole - So with the icon update - the test works now. I think this is mergeable:)

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
```diff
@@ -162,6 +162,7 @@ export const ArtworkActionsFragmentContainer = createFragmentContainer(
       id
       image {
         id
+        url(version: "larger")
```
zephraph: Are you sure we should use larger here? I guess we might need that version for desktop, but definitely not for mobile... hmm...

mdole: Would the alternative be also requesting a smaller version in the GraphQL query and then specifying which gets passed to Force based on device width?  Agreed that 'larger' is probably not necessary on mobile

oxaudo: It is the one that we are currently using for 'images' on artwork page. So it's helpful that is is the same url that would be a background on the image on the page.

Don't think we are requesting a different one based on the screen size. It just gets resized.

mdole: gotcha, that makes sense to me then. @zephraph good with you if I merge?

zephraph: I think we'll need to just revisit this when we're looking at image optimizations. From my view, the graphql api should'be be responsible for providing the image rendition. That should really be a single URL that can have a specified sized based on the client needs w/ dynamic resizing done via a CDN or service.

PR #1949, opened 2019-01-26
[Bugfix] Prevent apostrophes from line breaking tooltip links

damassi:
Fixes https://artsyproduct.atlassian.net/browse/BUGS-552

Tooltip links that contain apostrophes (not linked in editor) were breaking at punctuation; this fixes that.

linebreak


artsyit:

Built with commit b8fb41bc7c7fea7bb01300f9d356b5c99ea1226d

https://deploy-preview-1949--artsy-reaction.netlify.com

src/Components/Publishing/Sections/Text.tsx
```diff
@@ -85,14 +86,39 @@ export class Text extends Component<Props, State> {
...
-            {text}
-          </LinkWithTooltip>
-        )
+      const linkNode = get(node, n => n.children[0].data && n.children[0], {})
+
+      if (linkNode.data) {
+        const props = { key: href + index, url: href, color }
+        const next = linkNode.parent && linkNode.parent.next
+        const text = linkNode.data
+        const apostropheRe = /[’'][a-zA-Z]/
```
eessex: I believe this also happens with other kinds of punctuation, like commas or periods -- worth double checking these cases and potentially applying to punctuation generally.

damassi: Since I'm not familiar with the general patterns inside of editorial I'd rather wait until we have an explicit need to update (via example), and then we can update here as needed.

eessex: I think you could pretty easily test this by inserting `.` or `,` or `;` rather than adding `'s` in your fixture.

damassi: Updated JIRA ticket description with potential follow-up work (outside of this PR).

PR #1961, opened 2019-01-30
Keep actions and image dots centered for mobile web as well

sweir27:
Missed this nuance when I was matching the designs!

For mobile web, we also want the image browser dots to appear centered instead of floated to the right, where they may overlap with the labels.

image


artsyit:

Built with commit d271d831e0c7733d68fd307d5532ced366f06e70

https://deploy-preview-1961--artsy-reaction.netlify.com

PR #1962, opened 2019-01-30
removes automatic capitalization for collection titles

mdole:
[Jira Ticket](https://artsyproduct.atlassian.net/browse/GROW-1078)

### DIFF
--- a/src/Apps/Collect/Components/Collection/Header/index.tsx
+++ b/src/Apps/Collect/Components/Collection/Header/index.tsx
@@ -198,7 +198,6 @@ const SubtitlesContainer = styled(Box)`
 `
 
 const Title = styled(Serif)`
-  text-transform: capitalize;
   text-shadow: 0 0 15px rgba(0, 0, 0, 0.2);
 `
 
--- a/src/Apps/__stories__/Apps.story.tsx
+++ b/src/Apps/__stories__/Apps.story.tsx
@@ -36,7 +36,7 @@ storiesOf("Apps", module)
     return (
       <MockRouter
         routes={collectRoutes}
-        initialRoute="/collection/street-art-now"
+        initialRoute="/collection/david-hockney-ipad-drawings"
         context={{
           mediator: {
             trigger: x => x,


mdole:
(the second commit is to fix a minor visual bug on storybooks, which you can see here https://cl.ly/a9775584d210 https://cl.ly/4e54de83a683)

artsyit:

Built with commit a266407c0c862022069a849874eba965b741d6a1

https://deploy-preview-1962--artsy-reaction.netlify.com

src/Apps/Collect/CollectApp.tsx
```diff
@@ -61,7 +61,9 @@ export class CollectApp extends Component<CollectAppProps> {
             alignItems="flex-end"
           >
             <Serif size="8">
-              <h1 className="title">Collect art and design online</h1>
+              <h1 className="title" style={h1Style}>
```
eessex: Curious, do you know what the purpose of this className is?

src/Apps/Collect/CollectApp.tsx
```diff
@@ -77,6 +79,14 @@ export class CollectApp extends Component<CollectAppProps> {
   }
 }
 
+const h1Style = {
+  fontStyle: "inherit",
+  fontFamily: "inherit",
+  fontWeight: 400,
```
eessex: fontWeight should be able to inherit from `Serif` as well.

mdole: this is a weird one—when I set it to `"inherit"`, TS gives me this lovely message https://cl.ly/924cf404caf5

seems like it might be an issue with react and not with our code ¯\_(ツ)_/¯

mdole: or rather with TS

src/Apps/Collect/CollectApp.tsx
mdole: I don't—thought it might be for SEO purposes? But on further reflection that doesn't make much sense; I doubt Google cares about classes. Maybe could be removed?

eessex: Yeah classname won't have any affect, the `<h1>` should be enough. You can run a lighthouse audit to raise any SEO red flags.  Also: https://accessibility.iu.edu/creating-content/web-content/titles.html

src/Apps/Collect/CollectApp.tsx
eessex: You could use styled components here rather than passing `style` attributes directly, this is usually the pattern taken in this repo:
```
const H1 = styled.h1`
  font-style: inherit;
  font-family: inherit;
...
`
```

mdole: ah that makes sense! and actually, it looks like Chris has addressed this globally in [another PR](https://github.com/artsy/reaction/pull/1970/files#r252521612), so I should probably go ahead and remove these styles and keep this PR to just the capitalization change

eessex: Great, glad to know this is now global!

src/Apps/Collect/CollectApp.tsx
mdole: Nice, this is a helpful resource. I'll remove the `className="title"`!

PR #1963, opened 2019-01-30
Revert "Refactor tests"

ds300:
Reverts artsy/reaction#1899

see https://github.com/artsy/reaction/pull/1899#issuecomment-459011192

jest's `--forceExit` flag didn't do the trick. Need to figure out how to mitigate this issue.

PR #1964, opened 2019-01-30
[wip] debug jest non-exit issue

ds300:
see #1963


artsyit:

Built with commit 4671f5b5ade1a230beec3534237e2d69611f6ca6

https://deploy-preview-1964--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Status.test.tsx
```diff
@@ -7,269 +7,274 @@ import {
 } from "Apps/__tests__/Fixtures/Order"
 import { TransactionDetailsSummaryItem } from "Apps/Order/Components/TransactionDetailsSummaryItem"
 import { trackPageView } from "Apps/Order/Utils/trackPageView"
-import { MockBoot, renderRelayTree } from "DevTools"
+import { createTestEnv } from "DevTools/createTestEnv"
+import { expectOne } from "DevTools/RootTestPage"
 import { render } from "enzyme"
-import React from "react"
+import { produce } from "immer"
```
damassi: Immer is cool but I worry these sorts of dependencies tend to introduce unnecessary complexity?

ds300: I see where you're coming from. Immer feels like a safe bet to me. It's small, focused, extremely general, and all about transforming immutable data which is quite mundane and unlikely to be a source of complexity.

It does seem weird to me that we have a whole library that's only being used on one line. At the same time, I'm thinking of it like just a lodash function. `_.transformImmutableData(...)`

damassi: Yeah, just thinking about it in terms of tests as documentation, it could be overkill if the same could be achieved in a way without the lib since the library caries with it concepts that a dev down the line would be required to undertstand to read the test. For framework code and the like it makes sense.

PR #1969, opened 2019-01-30
Refactor Input component in preparation for moving to Palette

pepopowitz:
This is ready for review! I'm leaving it as a WIP for now, in case we want to wait until we get more specs from design for the palette version of the inputs.

Collaboration with @jonallured.

## Goal of this PR

We noticed that the "text input" fields were in Reaction, but they feel like something that could belong in palette. Our main goal was to make the text inputs easier to extract into palette. [An issue has been created in palette for the extraction](https://github.com/artsy/palette/issues/253).

While trying to understand how the text inputs worked, we noticed 3 variants of text input - a standard input, a "quick" input where the label for the field is inside the textbox, and a password input where the value is masked rom the user. [Zeplin affirms this](https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5ace5b9019e07a581ca7fa17). We found it easier to understand the code of these variants by breaking them into 3 separate components - [Input](https://github.com/artsy/reaction/pull/1969/files#diff-8034249d0768be96ad9c8106abdd5be2), [QuickInput](https://github.com/artsy/reaction/pull/1969/files#diff-f735be8094a599c1dd8ad5ab9d16cbc5), and [PasswordInput](https://github.com/artsy/reaction/pull/1969/files#diff-8b8ef0b1cfc4b81b3ba43664428b72a8). Those are the meat of this PR.

### Some components were passing props to the Input component that did nothing

[Onboarding/Steps/Artists](https://github.com/artsy/reaction/pull/1969/files#diff-3149367c8afe1b23417b66476758cfd4) and [Onboarding/Steps/Genes](https://github.com/artsy/reaction/pull/1969/files#diff-77a43f31cb80f21de24f6909b31356f3) were passing `leftView` and `rightView` props into the old version of `Input` - but that didn't actually do anything. Those props were only displayed if the `quick` prop was also passed in, and in this case it wasn't.

The intention of these props was to show an `X` to cancel out of the search. This is currently not showing in production; and when we hacked things to make the `X` show up, it didn't look at all correct. For now, we've removed these unused props. We'll raise this issue to our PM, and find a way to address this regression.

### We removed the `inputRef` prop

It appears to have been used for a short period of time by some work on "make offer". With the refactoring @ds300 has done there recently, it doesn't appear to be used anymore. [So, we removed it entirely](https://github.com/artsy/reaction/pull/1969/files#diff-4e60cf6afb78e5c1ef30cdae9d322bd3).

### Not all styles are "palette-friendly" yet

For instance, [font-sizes are still using the reaction font sizes](https://github.com/artsy/reaction/pull/1969/files#diff-8034249d0768be96ad9c8106abdd5be2R48). The existing text input fields are based on Garamond 17px/15px. Palette [doesn't define a 17px or 15px serif font](https://github.com/artsy/palette/blob/master/tokens.json#L171).

We think the answer to this is to change the text inputs to use 16px (size 3). However, this would change the way forms in reaction look. Any forms that have text inputs along with other types of inputs would look wacky - some things in 16px, some in 17px. We'd have to modify other form inputs in reaction to *also* use 16px with those changes.

We think that this dichotomy should be addressed when the inputs have been moved to palette (as 16px), and reaction is updated to point at them.

### DIFF
--- a/src/Apps/Order/Components/OfferInput.tsx
+++ b/src/Apps/Order/Components/OfferInput.tsx
@@ -9,8 +9,6 @@ export interface OfferInputProps {
 }
 
 export class OfferInput extends React.Component<OfferInputProps> {
-  inputRef = React.createRef<HTMLInputElement>()
-
   render() {
     const { id, showError, onFocus } = this.props
 
@@ -20,7 +18,6 @@ export class OfferInput extends React.Component<OfferInputProps> {
         title="Your offer"
         type="text"
         pattern="[0-9]"
-        innerRef={this.inputRef}
         defaultValue={null}
         error={showError ? "Offer amount missing or invalid." : null}
         onFocus={onFocus}
--- a/src/Components/Input.tsx
+++ b/src/Components/Input.tsx
@@ -12,7 +12,6 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
   error?: string
   title?: string
   touchedOnChange?: boolean
-  innerRef?: React.RefObject<HTMLInputElement>
 }
 
 export interface InputState {
--- a/src/Components/Input.tsx
+++ b/src/Components/Input.tsx
@@ -1,6 +1,4 @@
-import { space } from "@artsy/palette"
-import { ClosedEyeIcon, OpenEyeIcon } from "@artsy/palette"
-import { fadeIn, fadeOut, growAndFadeIn } from "Assets/Animations"
+import { growAndFadeIn } from "Assets/Animations"
 import Colors from "Assets/Colors"
 import { garamond, unica } from "Assets/Fonts"
 import React from "react"
@@ -12,54 +10,25 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
   block?: boolean
   description?: string
   error?: string
-  label?: string
   leftView?: JSX.Element
   rightView?: JSX.Element
-  setTouched?: (fields: { [field: string]: boolean }) => void
-  showPasswordMessage?: boolean
   title?: string
-  quick?: boolean
   touchedOnChange?: boolean
   innerRef?: React.RefObject<HTMLInputElement>
 }
 
 export interface InputState {
-  focused: boolean
   value: string
-  showPassword: boolean
 }
 
 /**
- * Configurable input field. It comes in two modes, standard & quick.
- * In standard mode, the `title` and `description` props are rendered above
- * the input. In quick mode, `title` and `description` are ignored, only
- * `label` is rendered inside the input.
+ * Standard input field.
+ * The `title` and `description` props are rendered above the input.
  *
- * @example
- *
- * ```javascript
- *  // Quick mode
- *  <Input
- *    quick
- *    label="Name"
- *    type="text"
- *    placeholder="Enter you name"
- *  />
- *  ```
- *
- * ```
- *  // Standard mode
- *  <Input
- *    title="Name"
- *    description="Your full name."
- *    type="text"
- *  />
  */
 export class Input extends React.Component<InputProps, InputState> {
   state = {
-    focused: false,
     value: (this.props.value as string) || "",
-    showPassword: false,
     touchedOnChange: true,
   }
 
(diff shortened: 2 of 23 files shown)


artsyit:

Built with commit c4443778dafee977b85b5702ddc40717c4ba1b60

https://deploy-preview-1969--artsy-reaction.netlify.com

src/Components/Input.tsx
```diff
@@ -12,276 +10,57 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
+      {description && <Description>{description}</Description>}
+      <StyledInput hasError={!!error} {...rest} />
+      {error && <InputError>{error}</InputError>}
+    </Container>
+  )
 }
 
 const Container = styled.div`
-  padding-bottom: 5px;
+  padding-bottom: ${space(0.5)}px;
```
damassi: Totally cool with ^ but thinking that it might be a little more future-friendly if this was refactored so that it behaved in a more palette-like way -- meaning: 

```jsx
<Box pb={0.5}>
  ...
</Box>
```

Then we can lose the intermediate variable and some indirection.

src/Components/Input.tsx
```diff
@@ -12,276 +10,57 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
   ${garamond("s17")};
 `
 
 const Description = styled.div`
   ${garamond("s15")};
-  color: ${Colors.graySemibold};
-  margin: 3px 0 0;
+  color: ${color("black60")};
+  margin: ${space(0.3)}px 0 0;
 `
```
damassi: Agree this is for a second PR, but wonder what the impact would be if we just went ahead with the new typography and worried about spot-checking issues with 1-2px differences after the fact. I believe everything before the new design system can be considered legacy and thus replaced: 

```jsx
<Description size="3" m={0.3}>
 ...
</Description>
```

src/Components/Input.tsx
pepopowitz: Yeah, we talked about that. We thought those types of changes would fit better with the work in palette - the goal here being to properly identify the seams so that we can dig the inputs out once they're moved, plus a little bit of "easy" palette-izing. 

If you feel strongly that this work should happen here in reaction, we can work on that next week.

src/Components/QuickInput.tsx
```diff
@@ -0,0 +1,178 @@
...
+  margin-right: 0;
+  margin-top: ${space(0.5)}px;
+  margin-bottom: ${space(1)}px;
+  display: flex;
+  position: relative;
+  height: ${p => (p.hasLabel ? `${space(4)}px` : `${space(2)}px`)};
+  flex-direction: row;
+  align-items: center;
+  box-sizing: content-box;
+`
```
damassi: Could this be replaced with `<Flex>`? or 

```js
const InputContainer = styled(Flex)` 
  ...
`
```

damassi:
Looks 👌 @pepopowitz. Great work breaking things apart into discrete components 👍

pepopowitz:
@damassi Each item of your specific feedback is something we're planning on doing when we get into Palette. We'll plan on using Palette-based components to compose the Inputs there, like [this recently-merged PR](https://github.com/artsy/palette/pull/244). 

Is that okay?

src/Components/Input.tsx
damassi: Not  strongly! This is totally fine 👍

damassi:
Sounds good @pepopowitz

artsyit:
:rocket: PR was released in v12.1.19 :rocket:

src/Apps/Order/Components/OfferInput.tsx
```diff
@@ -20,7 +18,6 @@ export class OfferInput extends React.Component<OfferInputProps> {
         title="Your offer"
         type="text"
         pattern="[0-9]"
-        innerRef={this.inputRef}
```
eessex: We are actively using innerRef in Positron right now, curious why this needs to be removed?  Is it incompatible with Native?

damassi: Not sure why it was removed, but do know that it's incompatible with styled-components 4 -- can just use `ref`

jonallured: @eessex damn, we were pretty sure this wasn't used, but didn't CC you - argh!! We removed it mostly because we couldn't find anything using it, clearly we were wrong. I'll reach out on Slack and we can figure out what to do - sorry about that!! ❤️

damassi: Note that this _will_ have to go away soon once I complete upgrades, so if there's a way to avoid using it that would be good.

eessex: @damassi looks like the implementation doesn't go away in `styled-components@4`, it is just renamed to `ref`.

PR #1970, opened 2019-01-31
Fix SSR issues on /collect, grid images, tracking and h1 global tags

damassi:
This PR fixes a bunch of SSR errors I discovered when debugging https://artsyproduct.atlassian.net/browse/GROW-1094: 

1) Noticed that SSR and mobile layout was broken on `/collect` with JS turned off on mobile: 

screen shot 2019-01-30 at 6 02 00 pm

I refactored the view a bit so it's more straightforward and the issue seemed to fix itself. Rather than weaving `<Media at='xs'>` and `<Media greaterThan='xs'>` throughout the same block i made mobile and desktop strictly separate via an acceptable amount of duplication, which allowed us to remove a bunch of intermediate variables for clarity and simplify overall. 

2) Fix SSR rendering on our grid. I've been noticing how pages with grids aways render with grey blocks for a while and tracked that down. With JS turned off: 

Before: 
-----

screen shot 2019-01-30 at 4 52 33 pm

After:
-----

screen shot 2019-01-30 at 4 52 41 pm

3) Noticed that our scroll tracking was buggy due to SSR page thrashing. I had to explicitly add the scroll tracking after the client mounted, which added the class to the artwork grid. 

scroll tracking

4) Fixes the issue with `h1` tags based on [this slack convo](https://artsy.slack.com/archives/C03J4L2KK/p1548866512185100) and [this PR](https://github.com/artsy/reaction/pull/1951/files#diff-69b9332da53df2687332ba3511aa9699R54)

### DIFF
--- a/src/Artsy/Router/Components/Boot.tsx
+++ b/src/Artsy/Router/Components/Boot.tsx
@@ -28,7 +28,15 @@ export interface BootProps {
   headTags?: JSX.Element[]
 }
 
-const { GlobalStyles } = injectGlobalStyles()
+const { GlobalStyles } = injectGlobalStyles(`
+  h1 {
+    font-style: inherit;
+    font-family: inherit;
+    font-weight: inherit;
+    font-size: inherit;
+    margin: 0;
+  }
+`)
 
 // TODO: Do we want to let Force explicitly inject the analytics code?
 @track(null, {
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -4,6 +4,7 @@ import { Mediator } from "Artsy/SystemContext"
 import { isFunction } from "lodash"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { data as sd } from "sharify"
 import styled from "styled-components"
 import colors from "../../Assets/Colors"
 import Metadata from "./Metadata"
@@ -78,12 +79,10 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
       return null
     }
 
-    const width = 300
+    const width = 400
     const height = Math.floor(width / this.props.artwork.image.aspect_ratio)
     const type = this.props.artwork.image.aspect_ratio ? "fit" : "fill" // Either scale or crop, based on if an aspect ratio is available.
-    const url = `https://d7hftxdivxxvm.cloudfront.net/?resize_to=${type}&width=${width}&height=${height}&quality=${IMAGE_QUALITY}&src=${encodeURIComponent(
-      imageURL
-    )}`
+    const url = `${sd.GEMINI_CLOUDFRONT_URL}/?resize_to=${type}&width=${width}&height=${height}&quality=${IMAGE_QUALITY}&src=${encodeURIComponent(imageURL)}` // prettier-ignore
     return url
   }
 
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -1,12 +1,10 @@
 import { Flex, Sans } from "@artsy/palette"
 import { GridItem_artwork } from "__generated__/GridItem_artwork.graphql"
 import { Mediator } from "Artsy/SystemContext"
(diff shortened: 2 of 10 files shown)


artsyit:

Built with commit b6224105e8a14089f267f46176af784cdcdec2ef

https://deploy-preview-1970--artsy-reaction.netlify.com

src/Components/Artwork/GridItem.tsx
```diff
@@ -61,46 +58,29 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
-      // Either scale or crop, based on if an aspect ratio is available.
-      const type = this.props.artwork.image.aspect_ratio ? "fit" : "fill"
-      // tslint:disable-next-line:max-line-length
-      return `${
-        sd.GEMINI_CLOUDFRONT_URL
-      }/?resize_to=${type}&width=${width}&height=${height}&quality=${IMAGE_QUALITY}&src=${encodeURIComponent(
-        imageURL
-      )}`
-    }
+    const width = 300
```
damassi: @mzikherman - Sticking with a concrete value solves a whole host of issues without introducing any apparent degradation.

src/Artsy/Router/Components/Boot.tsx
```diff
@@ -28,7 +28,15 @@ export interface BootProps {
...
 
-const { GlobalStyles } = injectGlobalStyles()
+const { GlobalStyles } = injectGlobalStyles(`
+  h1 {
+    font-style: inherit;
+    font-family: inherit;
+    font-weight: inherit;
+    font-size: inherit;
+    margin: 0;
+  }
```
damassi: @eessex and @mdole - saw the convo earlier. When issues like those encountered with `h1` tags arise you can pass global styles in here. Ideally these then move over [to palette](https://github.com/artsy/palette/blob/master/src/helpers/injectGlobalStyles.tsx#L7).

src/Components/Publishing/Header/Layouts/ClassicHeader.tsx
```diff
@@ -51,14 +51,6 @@ export const Title = styled.div`
...
   ${garamond("s37")};
 
-  h1 {
-    font-style: inherit;
-    font-family: inherit;
-    font-weight: inherit;
-    font-size: inherit;
-    margin: 0;
-  }
-
```
damassi: @eessex - I removed these following the other PR.

src/Components/Artwork/GridItem.tsx
mzikherman: Does this wind up being upscaled when rendered at the `md` breakpoint? (Just looking at the mapping above tha width is 400 vs 300 for the rest, but I'm not sure if that's how wide the columns are). If so, we might want to set this to 400.

src/Artsy/Router/Components/Boot.tsx
eessex: Excellent, glad to see this.

src/Components/Publishing/Header/Layouts/ClassicHeader.tsx
eessex: Cool, thanks for taking this off my todo's

src/Components/Artwork/GridItem.tsx
damassi: Updated.

src/Components/Artwork/GridItem.tsx
```diff
@@ -160,54 +140,49 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
+                this.props.onClick()
+              }
+            }}
+          >
+            <Image src={this.getImageUrl()} />
+          </a>
+
+          {this.renderArtworkBadge(artwork)}
+
+          {this.canHover && (
```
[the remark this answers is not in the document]
alloy: I’d love to. Can you file a ticket for that, though? 🙏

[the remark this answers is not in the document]
damassi: For sure -- will file over in the repo.

PR #1899, opened 2019-01-16
Refactor tests

ds300:
Sorry for the HUGE diff :(

Here's what I've done:

- Abstracted a lot of logic around selecting elements, page interactions, and making assertions into a file called `OrderAppTestPage`. See also `RootTestPage` which has the basic basic stuff.
- Created a `createTestEnv` function which adds support for mounting test pages like the `OrderAppTestPage` and dealing with the mock data that they use.
- Refactored the createMockNetworkLayer to allow for mocking network failures.
- other things to be described later

my personal checklist of DX things to take care of:

- [ ] ~tslint rule to make sure you wait for testpage things that return promises~ wouldn't give in-editor feedback due to the fact that vscode-tslint doesn't use rules that require type info. So no point in having it.
- [ ] ~find out why console.logs in `complain` get swallowed~ still a mystery, but at least we're surfacing those errors now.
- [x] ~do explicit checks for `__typename` in unions and maybe warn if used not in a union position?~ got `__typename` inference working :tada:
- [ ] ~support aliases in root mutation results~ haven't figured this one out, but can't imagine we'll need to send multiple versions of the same mutation at once so not such an issue.
- [x] ~check whether we need a lint rule to make sure you call createTestEnv in describe block~ we do not
- [x] do status page

PR #1979, opened 2019-02-01
Refactor tests (again)

ds300:
Hello! This is a repeat of #1899 

The only change from last time is that we added `--runInBand` to the jest invocation to prevent the nondeterministic fork failure problem. Tests take about 3mins to run on CI now (before it was about 1:30). They still run in parallel locally, where the problem does not occur.

Not ideal, and we should continue debugging the jest/circle problem at some point. But I'd rather have slightly slower CI than lose these changes 😑


artsyit:

Built with commit 22ab7f5c8e34d60c4dbf1f7a9fb1776b267dd327

https://deploy-preview-1979--artsy-reaction.netlify.com

package.json
```diff
@@ -33,7 +33,7 @@
     "storybook": "node verify-node-version.js && concurrently --raw --kill-others 'yarn relay --watch' 'start-storybook --quiet -s ./public -p 9001'",
     "sync-schema": "curl https://raw.githubusercontent.com/artsy/metaphysics/master/_schema.graphql -o data/schema.graphql; yarn prettier --write --parser graphql data/schema.graphql",
     "sync-schema:localhost": "cd ../metaphysics && yarn dump-schema -- ../reaction/data",
-    "test": "node verify-node-version.js && jest",
+    "test": "node verify-node-version.js && jest --forceExit",
```
alloy: Is this still needed besides running in-band?

ds300: ah good catch, no, we should only have it in the circle config file.

ds300:
@alloy should that be in the platform jira project?

alloy:
Yeah let's do that 👍

ds300:
Done https://artsyproduct.atlassian.net/browse/PLATFORM-1155

damassi:
@ds300 - is that the right ticket link? Seems to point to Sidekiq.

ds300:
@damassi Thanks for pointing that out. It was an off-by-one error. Weird.

sweir27:
Cooool! It seems like this is good to go. Gonna merge and 🤞

artsyit:
:rocket: PR was released in v12.1.16 :rocket:

PR #1995, opened 2019-02-06
Add back access to the innerRef for Input

jonallured:
Over on https://github.com/artsy/reaction/pull/1969 @pepopowitz and I removed access to the `innerRef` for `Input` because we thought it wasn't being used - we were wrong!

https://github.com/artsy/positron/blob/010b5d09e27996c997e6b7e6b5d14337dbd59e42/src/client/components/autocomplete2/index.tsx#L205

So here I'm just adding it back to the prop interface so that Positron won't be blocked from bumping Reaction.

Per @damassi, there is some incompatibility with `innerRef` and styled-components so we will have to revisit this at some point. Better to do that work with intention rather than having this PR drop support so suddenly. My thought is to drop that support as we move `Input` over to Palette, but open to other thoughts! ❤️

### DIFF
--- a/src/Components/Input.tsx
+++ b/src/Components/Input.tsx
@@ -10,6 +10,7 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
   block?: boolean
   description?: string
   error?: string
+  innerRef?: React.RefObject<HTMLInputElement>
   title?: string
 }
 


artsyit:

Built with commit 95f2204d6be65770a6984bd1da5de0e6dcaf604e

https://deploy-preview-1995--artsy-reaction.netlify.com

eessex:
Regarding innerRef, would be glad to know in more detail what the changes are -- we are using refs for attaching our autocompletion library to an input, so I'm not sure what the best way for implementing this would be without it.  Is the change a rename, an implementation update, or just removing ability to use refs entirely?

artsyit:
:rocket: PR was released in v12.1.21 :rocket:

PR #2016, opened 2019-02-13
Fix style injection

damassi:
During SC4 upgrade we no longer had a way to directly inject styles into a head that lacks a react tree, so our RRM styles weren't being injected. I added it to the Boot global styles, leading to a little duplication since its already being passed via SSR. Should always be handled by consumers.


artsyit:

Built with commit 9f3918bedae5bb2e4462b8831db13b14e758d387

https://deploy-preview-2016--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v13.0.2 :rocket:

PR #2025, opened 2019-02-14
New payment route

lilyfromseattle:
Build 'Payment' page for resubmitting payment method

### DIFF
--- a/src/Apps/Order/Routes/NewPayment/index.tsx
+++ b/src/Apps/Order/Routes/NewPayment/index.tsx
@@ -206,16 +206,18 @@ export class NewPaymentRoute extends Component<
                 style={isCommittingMutation ? { pointerEvents: "none" } : {}}
               >
                 {order.mode === "OFFER" && (
-                  <Flex>
-                    <CountdownTimer
-                      action="Respond"
-                      note="Expiration will end negotiations on this offer. Keep in mind the work can be sold to another buyer in the meantime."
-                      countdownStart={order.lastOffer.createdAt}
-                      countdownEnd={order.stateExpiresAt}
-                    />
-                  </Flex>
+                  <>
+                    <Flex>
+                      <CountdownTimer
+                        action="Respond"
+                        note="Expiration will end negotiations on this offer. Keep in mind the work can be sold to another buyer in the meantime."
+                        countdownStart={order.lastOffer.createdAt}
+                        countdownEnd={order.stateExpiresAt}
+                      />
+                    </Flex>
+                    <Spacer mb={[2, 3]} />
+                  </>
                 )}
-                <Spacer mb={[2, 3]} />
                 <Join separator={<Spacer mb={3} />}>
                   <Flex flexDirection="column">
                     <Serif mb={1} size="3t" color="black100" lineHeight="1.1em">
--- a/src/Apps/Order/Routes/NewPayment/index.tsx
+++ b/src/Apps/Order/Routes/NewPayment/index.tsx
@@ -1,4 +1,26 @@
-import { Payment_order } from "__generated__/Payment_order.graphql"
+import { NewPayment_order } from "__generated__/NewPayment_order.graphql"
+import { NewPaymentRouteCreateCreditCardMutation } from "__generated__/NewPaymentRouteCreateCreditCardMutation.graphql"
+import { NewPaymentRouteSetOrderPaymentMutation } from "__generated__/NewPaymentRouteSetOrderPaymentMutation.graphql"
+import {
+  Address,
+  AddressChangeHandler,
+  AddressErrors,
+  AddressForm,
+  AddressTouched,
+  emptyAddress,
+} from "Apps/Order/Components/AddressForm"
+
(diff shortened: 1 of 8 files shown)


artsyit:

Built with commit fc1160ba1652677e927670204d017c6a7a969751

https://deploy-preview-2025--artsy-reaction.netlify.com

src/Apps/Order/redirects.tsx
```diff
@@ -183,6 +183,10 @@ export const redirects: RedirectRecord<OrderQuery> = {
         goToShippingIfShippingIsNotCompleted,
       ],
     },
+    {
+      path: "payment/new",
+      rules: [goToShippingIfShippingIsNotCompleted],
```
sweir27: Since you can only access the `payment/new` flow if you've already submitted an offer/order, I _think_ we'll want to change this to:

```
rules: [goToStatusIfNotOfferOrder, goToStatusIfOrderIsNotSubmitted]
```

src/Apps/__stories__/OrderApp.story.tsx
```diff
@@ -65,6 +65,30 @@ storiesOf("Apps/Order Page/Buy Now/Review", module).add("Review", () => (
...
+        stateExpiresAt: null,
+        lastOffer: null,
+      }}
+    />
+  ))
+  .add("With 'Pickup'", () => (
+    <Router
+      initialRoute="/orders/123/payment/new"
+      order={{
+        ...BuyOrderPickup,
```
sweir27: I am pretty sure (will check tomorrow with Stephen) that we only want to allow this page to be accessed if you have an offer-type order. If so we can remove these two stories (otherwise we should update the above fixture to not be `OfferOrderWithShippingDetails`).

src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+            this.createCreditCard({ token: token.id, oneTimeUse: true })
+          }
+        })
+        .catch(e => {
+          this.onMutationError(new ErrorWithMetadata(e))
+        })
+    })
+  }
+
+  @track((props, state, args) => {
```
sweir27: For now we can remove this `track` methods as it'd be instrumenting us against the buy now flow and I'm not sure what the schema is just yet. 😄

src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+      {
+        onCompleted: (data, errors) => {
+          this.setState({ isCommittingMutation: false })
+
+          const {
+            ecommerceSetOrderPayment: { orderOrError },
+          } = data
+
+          if (orderOrError.order) {
+            this.props.router.push(`/orders/${this.props.order.id}/review`)
```
sweir27: We should route them to the status page here.

src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+              <Flex
+                flexDirection="column"
+                style={isCommittingMutation ? { pointerEvents: "none" } : {}}
+              >
+                {order.mode === "OFFER" && (
+                  <>
+                    <Flex>
+                      <CountdownTimer
+                        action="Respond"
+                        note="Expiration will end negotiations on this offer. Keep in mind the work can be sold to another buyer in the meantime."
```
ds300: In zeplin this copy is just "Expired offers end the negotiation process". I think the one you have here is better but we should check with Stephen.

Another question for Stephen, should the action be "Respond by"? Seems like an inappropriate way to describe updating one's payment information. Maybe "Submit by"? Nice and general.

stephen-i-chan: Hey @ds300 - Good question. I believe this countdown messaging applies to the Seller and Buyer, and not all scenarios involve submitting a payment (someone could reject the offer/counter offer as a response), which I imagine that was Brian's reasoning. I think for this particular scenario with 'Add a new payment' page, "Submit by" seems more appropriate since this is specifically geared towards payment submission. Is that a quick text change?

sweir27:
@lilyfromseattle @ds300 going to merge this so we can continue iterating while Lily is on vacation. 😄

There are a few TODOs in the code review comments, but let's address those in the next PR. ✨

artsyit:
:rocket: PR was released in v13.0.7 :rocket:

src/Apps/Order/redirects.tsx
ds300: done

src/Apps/__stories__/OrderApp.story.tsx
ds300: done

src/Apps/Order/Routes/NewPayment/index.tsx
ds300: done

src/Apps/Order/Routes/NewPayment/index.tsx
ds300: done

PR #2028, opened 2019-02-14
Artist search previews

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-717. A 💃collaboration 💃with @jonallured!

This PR adds artist previews for the search bar, and updates the preview for merchandisable artworks. These previews look like this:

### Artist preview for an artist with marketing collections

artist-search-collection

### Artist preview for an artist without marketing collections

artist-search-related-artworks

### Merchandisable artworks preview

artist-search-merchandisable

There is still some fine-tuning to do with styles, but those could get resolved in a follow-up PR.

We also aren't sure about the performance impact of these previews, but will be investigating/measuring.


artsyit:

Built with commit a90d59275e4858c0fde8854d4a805b7936b0c5ee

https://deploy-preview-2028--artsy-reaction.netlify.com

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -0,0 +1,101 @@
...
+        rgba(0, 0, 0, 0.5),
+        rgba(0, 0, 0, 0.5)
+      ),
+      url(${p => p.imageUrl}) center;
+  }
+`
+
+const CollectionTitle = styled(Serif).attrs({ display: "inline" })`
+  color: ${color("white100")};
+`
```
damassi: #minor - it's slightly clearer to move styled-component definitions below the main react components. That way down the line when people view the code they can immediately get to business rather than having to wade through styles.

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import styled from "styled-components"
+
+interface MarketingCollectionsPreviewProps {
+  marketingCollections: MarketingCollectionsPreview_marketingCollections
+}
+
+const CollectionBox = styled(Box)<{ imageUrl: string }>`
+  margin-right: ${space(2)}px;
+  margin-bottom: ${space(2)}px;
```
damassi: Best to keep these common styled-system props inline with the component

```js
<CollectionBox  mr={2} mb={2}>
```

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -0,0 +1,101 @@
...
+        to bottom,
+        rgba(0, 0, 0, 0.5),
+        rgba(0, 0, 0, 0.5)
+      ),
+      url(${p => p.imageUrl}) center;
+  }
+`
+
+const CollectionTitle = styled(Serif).attrs({ display: "inline" })`
+  color: ${color("white100")};
```
damassi: could move the `display: inline` into the main CSS body and lose the `attrs`. (`attrs({ ... })` is good for applying props that can't be directly applied via css, like `ml` or `p` -- things defined by interfaces.)

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -0,0 +1,101 @@
...
+export const CollectionTitles = ({ title }: { title: string }) => {
+  const [first, ...rest] = title.split(": ")
+  const showHeadline = rest.length !== 0
+  const headline = showHeadline && first
+  const subtitle = showHeadline ? rest.join(": ") : first
+
+  return (
+    <Flex justifyContent="flex-end" flexDirection="column" height="80px">
+      {showHeadline && (
+        <Box ml={`${space(1)}px`} mb={`-${space(0.5)}px`}>
```
damassi:
```
<Box ml={1} mb={.5}>
```
These props  are pulled in automatically via our [Theme](https://github.com/damassi/gatsby-starter-docs/blob/master/src/Theme.tsx) and don't need to be translated with helpers.

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <CollectionTitle weight="semibold" size="2">
+            {headline}:
+          </CollectionTitle>
+        </Box>
+      )}
+      <Box ml={`${space(1)}px`} mb={`${space(1)}px`}>
```
damassi: ditto

src/Components/Search/Previews/Grids/ArtistSearch/__tests__/MarketingCollections.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { Serif } from "@artsy/palette"
+import { mount } from "enzyme"
+import React from "react"
+
+import { CollectionTitles } from "../MarketingCollections"
+
+describe("CollectionTitles", () => {
+  it("with no colons", () => {
+    const wrapper = mount(<CollectionTitles title="The best one" />)
+    expect(wrapper.find(Serif).length).toEqual(1)
```
damassi: Maybe add an additional `expect(...)` that checks against the passed title ("with no colons")? Otherwise it might be a little clearer if the test described testing page structure since that seems to be the intent:

```js
it('renders one Serif element if no colon in text', () => { ... })
```

src/Components/Search/Previews/Grids/ArtistSearch/index.tsx
```diff
@@ -1,53 +1,46 @@
...
-  artist: ArtistSearchPreview_artist
+  viewer: ArtistSearchPreview_viewer
 }
 
-export class ArtistSearchPreview extends React.Component<
-  ArtistSearchPreviewProps
-> {
-  render() {
-    const { marketingCollections, artworks_connection } = this.props.artist
+export const ArtistSearchPreview: React.SFC<ArtistSearchPreviewProps> = ({
```
damassi: 👍

src/Components/Search/Previews/Grids/MerchandisableArtworks.tsx
```diff
@@ -23,16 +23,20 @@ const MerchandisableArtworksPreview: React.SFC<
...
-    <PreviewGridItem artwork={artwork} key={i} />
+    <Box width="50%">
+      <PreviewGridItem artwork={artwork} key={i} />
+    </Box>
   ))
 
   return (
     <Box>
-      <Sans size="2" weight="medium">
+      <Sans size="3" weight="medium" color="black100" mb={`${space(2)}px`}>
```
damassi: No need for `space`

src/Components/Search/Previews/Grids/PreviewGridItem.tsx
```diff
@@ -1,27 +1,46 @@
...
+  artwork,
+  emphasizeArtist,
+}) => {
   const imageUrl = get(artwork, x => x.image.cropped.url, "")
   return (
-    <Flex m={2}>
+    <Flex mr={`${space(2)}px`} mb={`${space(2)}px`}>
       <Link href={artwork.href} noUnderline>
-        <Image mr={2} src={imageUrl} />
+        <Image
```
damassi: Should the image component still render if there's no image? 

```
{imageUrl && 
  <Image />
}
```

src/Components/__stories__/Search.story.tsx
```diff
@@ -20,21 +21,27 @@ storiesOf("Components/Search/Suggestions", module).add("Term: Andy", () => (
 
 storiesOf("Components/Search/Previews/Artist", module)
   .add("An artist with collections", () => (
-    <ContextProvider>
-      <SearchPreview entityID="kaws" entityType="Artist" />
-    </ContextProvider>
+    <Box width="450px" p="20px">
```
damassi: `p={2}`

damassi:
Just a couple super minor comments -- this is really coming together nicely 👍 👍

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -0,0 +1,101 @@
...
+  MarketingCollectionsPreviewProps
+> = ({ marketingCollections }) => {
+  const items = marketingCollections.map(
+    ({ headerImage, title, slug }, index) => {
+      const href = `collection/${slug}`
+      const imageUrl = `${
+        sd.GEMINI_CLOUDFRONT_URL
+      }/?resize_to=fill&width=185&height=80&quality=95&src=${encodeURIComponent(
+        headerImage
+      )}`
```
alloy: This type of stuff should be abstracted away and I'm pretty sure there already exists a solution for this in the codebase.

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
pepopowitz: Yeah, I realized way late in our work that we were making these harder than they need to be. I'll circle back and update them!

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
pepopowitz: 👍 I'll see what I can find.

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
pepopowitz: Strangely @damassi....wherever I change these to use an integer instead of an interpolated string, I get an unexpected increment. 

For instance - ``ml={`${space(1)}px`}`` renders as a 10px margin...but converting it to `ml={1}`, I get a 4px margin instead of 10. `ml={2}` gives me 8px; `ml={3}` gives me 16px. It's happening in all of these components.

Everything I've seen convinces me that I should expect 1->10px, 2->20px, and 3->30px. I blew away my node_modules and re-yarned, to make sure it wasn't something with an older version of styled-components, and it still happens.

1. Are my expectations correct? 
2. If this is unexpected behavior, is there anything you can think of that would cause this to happen? Is there any sort of "activation" of the theme that needs to happen, to pull in the correct increments, that we might be missing?

pepopowitz: It does resolve if I wrap a `<Theme>` component around my story. I guess I'm surprised that no other stories do that.

damassi: Oh interesting... I wonder if you're using the right `storiesOf` helper... 👁

pepopowitz: Nope 😀

Good call, and thank you!

src/Components/__stories__/Search.story.tsx
```diff
@@ -1,5 +1,6 @@
+import { Box } from "@artsy/palette"
+import { storiesOf } from "@storybook/react"
```
damassi: @pepopowitz - here's the issue 😄(And it's not totally clear...) 

```
import { storiesOf } from 'storybook/storiesOf' 
```

(https://github.com/artsy/reaction/blob/master/src/__stories__/storiesOf.js#L11)

Which will import `Boot` (and `Theme`) behind the scenes.

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -1,10 +1,9 @@
-import { data as sd } from "sharify"
-
 import { Box, color, Flex, Link, Sans, Serif, space } from "@artsy/palette"
 import { MarketingCollectionsPreview_marketingCollections } from "__generated__/MarketingCollectionsPreview_marketingCollections.graphql"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
+import { crop } from "Utils/resizer"
```
damassi: @pepopowitz - can you make sure this can be run with JS disabled / is SSR friendly?

damassi: (Actually, I think this will be fine since this component will only ever be client-side only! but this is more of an awareness thing that if something can be done in an SSR-friendly way its preferred over something that's client-side only.)

alloy: For my understanding, can you elaborate a bit on what aspect of this worries you re SSR?

damassi: I recall this having some kind of dependency on `react-sizeme` which caused a lot of trouble during the new publishing page roll out but looking at the util everything seems fine.

src/Components/Search/Previews/Grids/PreviewGridItem.tsx
```diff
@@ -1,27 +1,55 @@
...
+  text-overflow: ellipsis;
+  max-width: ${space(12)}px;
+`
+
+export const PreviewGridItem: React.SFC<PreviewGridItemProps> = ({
+  artwork,
+  emphasizeArtist,
+}) => {
+  let imageUrl = get(artwork, x => x.image.cropped.url, "")
+  if (artwork.title === "Untitled") {
```
mzikherman: What's this special case for?

mzikherman:
Looks phenomenal to me! 😎

src/Components/Search/Previews/Grids/PreviewGridItem.tsx
pepopowitz: aaaaggghhhh that is me committing something that I didn't mean to!!! Great catch. I will remove...

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
alloy: Ok thanks 👍

mzikherman: I think the issue was maybe coming from trying to make a responsive image that was 'perfectly' resized for the viewport and layout, which was requiring JS. I believe that was a previous iteration of artwork bricks/images at some point.

Now, we're just resizing images to some fixed dimensions that work for the desired layout and are reasonable (ie- cropped in this case to 185x80). For a more generic artwork grid they are resized to a max width of 3-400px, and similarly for a max height in a rail.

pepopowitz:
I think all things are cleaned up on this one. I'll give y'all a bit longer to dwell on it, and self-merge later this afternoon. Thanks all for the feedback!

artsyit:
:rocket: PR was released in v14.0.1 :rocket:

PR #2030, opened 2019-02-15
DISCO-739: Add Tracking to Artwork Page Filters

jonallured:
This is a bit of a spike after a pairing session with @damassi. First he noticed that we could simplify things quite a bit if we moved the ContextConsumer into the fragment container, so I followed that refactoring through on a9880ff.

Then I added tracking to the selection of categories using the `@track` decorator on 044d02d. I wasn't wild about how this looked because I had to proxy through a factory function in order to get it to work.

That indirection bothered me enough to simplify things with a26e34a. Here I've moved the proxying up into the `onSelect` in the element, which seems better to me.

Finally, with 923cbac I went another direction entirely and switched to tracking the event through the prop that we get from decoration rather than using the decoration function itself. I think I like this best, but it doesn't look like this is the convention in the codebase, so open to thoughts there!!

### DIFF
--- a/src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
@@ -1,6 +1,7 @@
 import { FilterIcon, Toggle } from "@artsy/palette"
 import { ArtworkFilter_artist } from "__generated__/ArtworkFilter_artist.graphql"
 import { FilterState } from "Apps/Artist/Routes/Overview/state"
+import { track } from "Artsy/Analytics"
 import { ContextConsumer, Mediator } from "Artsy/SystemContext"
 import { FollowArtistButtonFragmentContainer as FollowArtistButton } from "Components/FollowButton/FollowArtistButton"
 import React, { Component } from "react"
@@ -33,6 +34,7 @@ interface Props {
   mediator?: Mediator
 }
 
+@track()
 class Filter extends Component<Props> {
   static defaultProps = {
     hideTopBorder: false,
@@ -146,13 +148,9 @@ class Filter extends Component<Props> {
               my={0.3}
               selected={currentFilter === count.id}
               value={count.id}
-              onSelect={({ selected }) => {
-                if (selected) {
-                  return filterState.setFilter(category, count.id, mediator)
-                } else {
-                  return filterState.unsetFilter(category, mediator)
-                }
-              }}
+              onSelect={({ selected }) =>
+                this.handleCategorySelect(selected, category, count)
+              }
               key={index}
               label={count.name}
             />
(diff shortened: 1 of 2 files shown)


alloy:
> Finally, with [923cbac](https://github.com/artsy/reaction/commit/923cbac464ebb17cfb129f8a8ad4df5dc9e6377c) I went another direction entirely and switched to tracking the event through the prop that we get from decoration rather than using the decoration function itself. I think I like this best, but it doesn't look like this is the convention in the codebase, so open to thoughts there!!

I don’t disagree that, in this case, this version reads a bit better; however, there are plenty of cases where inlining analytics code would obfuscate what’s going on. The separation of analytics code into the decorator makes it easier to understand _all_ of the analytics or app code _without_ needing to mentally parse and separate the two–which is good for reading, refactoring, etc. If we would use the decorator on a case-by-case basis, you would add cognitive load for the reader who then needs to be mindful of the possibility of analytics code being inlined _and_ to the writer that needs to decide if they should inline or use the decorator.

In short: I don’t find decorators, in general, the most aesthetically pleasing syntax, but at a larger scale consistency is 👑

alloy:
As for the actual code, I think that perhaps using the object short-hand could already get rid of some syntax noise:

```js
	  @track((_props, _state, [_selected, category, count]) => (
	    { [`changed_${category}`]: count.id }
	  )
	  handleCategorySelect(selected, category, count) {
	    // ...
	  }
```

Or, perhaps there's the possibility to discuss with the data team a slightly different schema that does not encode the verb and the noun into one:

```js
	  @track((_props, _state, [_selected, category, count]) => ({
	    action_name: ActionName.Click,
	    subject: Artwork.Filter.Category,
            artwork_count: count,
	  })
	  handleCategorySelect(selected, category, count) {
	    // ...
	  }
```

(Note that I just made up some constant names and perhaps misunderstood the intent of what needs to be tracked, the main purpose here is to illustrate that things can become a lot more declarative [the React mantra] by having a richer schema.)

jonallured:
Ok, thanks for the feedback @alloy, I've moved this back to using decorators. I'm still fuzzy on the shape of the return there and will work with @mzikherman on figuring out what should be done there.

jonallured:
@mzikherman with 3789657, I think I'm returning the right tracking data, but the compiler doesn't like it - what am I missing?

artsyit:

Built with commit f5e5dbbe50279b1de07e900b7b44902e0df95bce

https://deploy-preview-2030--artsy-reaction.netlify.com

src/Artsy/Analytics/Schema/Values.ts
```diff
@@ -54,6 +54,8 @@ export enum ActionType {
    */
   ClickedReadMore = "Clicked read more",
 
+  ClickedCommercialFilter = "Clicked commercial filter params",
```
mzikherman: This is worth double checking with @cavvia about the desired naming, this differs slightly from what's used for the filter on /collect: https://github.com/artsy/reaction/blob/fd1922d49906c2197834bdc69eebb25a62860007/src/Apps/Collect/FilterState.tsx#L171

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -159,6 +161,23 @@ class Filter extends Component<Props> {
     )
   }
 
+  @track((props: Props, _state, [_selected, category, count]) => {
```
mzikherman: Do we want to instrument this for the 'ways to buy' and price range sections? I think this is just for the radio button style category changes. (fine to come later)

artsyit:
:rocket: PR was released in v14.3.14 :rocket:

PR #2035, opened 2019-02-19
Fail tests on console warn and error logs

alloy:
Related to 
https://github.com/artsy/reaction/pull/1870#discussion_r256152557

Now when a `console.warn` or `console.error` is used during tests, the test will fail with a message like:

screenshot 2019-02-19 at 13 14 44

Obviously the `ALLOW_CONSOLE_LOGS` env variable should only be used for extraordinary circumstances, such as doing large refactors/upgrades of deps.


src/Components/Input.tsx
```diff
@@ -19,21 +19,18 @@ export interface InputProps extends React.HTMLProps<HTMLInputElement> {
...
+    return (
+      <Container>
+        {title && <Title>{title}</Title>}
+        {description && <Description>{description}</Description>}
+        <StyledInput ref={ref} hasError={!!error} {...rest} />
+        {error && <InputError>{error}</InputError>}
+      </Container>
+    )
+  }
+)
```
alloy: @mzikherman This is what needed to be done to deal with the ref that react-autowhataver wants you to forward to your custom input. See https://reactjs.org/docs/forwarding-refs.html

src/setup_jest.ts
```diff
@@ -32,48 +34,38 @@ if (typeof window !== "undefined") {
...
+  beforeEach(done => {
+    ;["error", "warn"].forEach((type: "error" | "warn") => {
+      // Don't spy on loggers that have been modified by the current test.
+      if (console[type] === originalLoggers[type]) {
+        const handler = (...args) => done.fail(logToError(type, args, handler))
+        jest.spyOn(console, type).mockImplementation(handler)
+      }
+    })
+    done() // it is important to call this here or every test will timeout
+  })
```
alloy: This relies on Jasmine being the test runner. This will probably break in a future major Jest version, but we'll deal with it then.

artsyit:

Built with commit 26bc182f5d361690f5d8061283a8463f94d0a0bf

https://deploy-preview-2035--artsy-reaction.netlify.com

src/Components/Input.tsx
mzikherman: Gah, I spent so long trying to figure that out! I'm hopeful this also clears up an error in an edge case I was coming across in testing (from the autosuggest lib, clicking on the custom rendered suggestions container while its open, which only happened when using a custom input renderer) where an input ref was undefined. Thank you!!

mzikherman:
I'll merge this! I think @damassi is out today.

artsyit:
:rocket: PR was released in v14.0.3 :rocket:

src/Components/Input.tsx
alloy: Ah, damn, sorry for not noticing earlier that you were struggling with this… let me know next time!

I made a PR for the test I think you mean https://github.com/artsy/reaction/pull/2036

mzikherman: @damassi this wound up fixing that `ref` issue we were looking at when using a custom render for the input.

damassi: Nice, I haven't seen `React.forwardRef` before. (There's a deluge of new React API's coming down the line I need to get on top of.)

PR #2052, opened 2019-02-21
Collections Rail

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-1112

This PR creates the new Collections Rail to be displayed on Standard, Feature, and News Article pages.

**Desktop:**
screen shot 2019-02-22 at 10 50 08 am

**Mobiile:**
screen shot 2019-02-22 at 10 51 04 am

UI Issues to be addressed:
- [x] There is too much space in between the rails
- [x] There should be left and right margin on the primary rails container in the mobile view
- [x] There needs to be a comma for dollar values higher than 999 (i.e. $1,000)

### DIFF
--- a/src/Components/CollectionsRail/Collection.tsx
+++ b/src/Components/CollectionsRail/Collection.tsx
@@ -1,19 +1,17 @@
 import { Box, color, Sans, Serif } from "@artsy/palette"
+import { CollectionEntity_collection } from "__generated__/CollectionEntity_collection.graphql"
 import React, { SFC } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
 import { resize } from "Utils/resizer"
 
 export interface CollectionProps {
-  headerImage?: string
-  price_guidance?: number
-  slug: string
-  title?: string
+  collection: CollectionEntity_collection
 }
 
 const Background = styled(Box)<{ collectionImage: string }>`
   height: 175px;
   width: 100%;
-  // min-width: 545px;
   background: ${color("black30")};
   background-image: url(${props => props.collectionImage});
   display: inline-flex;
@@ -26,18 +24,31 @@ const CollectionTitle = styled(Serif)`
   width: max-content;
 `
 
-export const CollectionEntity: SFC<CollectionProps> = props => {
+export const CollectionEntity: SFC<CollectionProps> = ({ collection }) => {
   return (
     <Box mb={3} width="100%">
       <Background
-        collectionImage={resize(props.headerImage, {
-          width: 820,
-          height: 240,
+        collectionImage={resize(collection.headerImage, {
+          width: 645,
+          height: 275,
           quality: 80,
         })}
       />
-      <CollectionTitle size="4">{props.title}</CollectionTitle>
-      <Sans size="2">Works from ${props.price_guidance}</Sans>
+      <CollectionTitle size="4">{collection.title}</CollectionTitle>
+      <Sans size="2">Works from ${collection.price_guidance}</Sans>
     </Box>
   )
 }
+
+export const CollectionEntityFragmentContainer = createFragmentContainer(
+  CollectionEntity,
+  graphql`
+    fragment CollectionEntity_collection on MarketingCollection {
+      slug
+      headerImage
+      title
+      price_guidance
+      show_on_editorial
+    }
+  `
+)
--- a/src/Components/CollectionsRail/Collection.tsx
+++ b/src/Components/CollectionsRail/Collection.tsx
@@ -1,5 +1,6 @@
 import { Box, color, Link, Sans, Serif } from "@artsy/palette"
 import { CollectionEntity_collection } from "__generated__/CollectionEntity_collection.graphql"
+import currency from "currency.js"
 import React, { SFC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { data as sd } from "sharify"
@@ -44,7 +45,12 @@ export const CollectionEntity: SFC<CollectionProps> = ({ collection }) => {
           })}
         />
         <CollectionTitle size="4">{collection.title}</CollectionTitle>
-        <Sans size="2">Works from ${collection.price_guidance}</Sans>
+        <Sans size="2">
+          Works from ${currency(collection.price_guidance, {
+            separator: ",",
+            precision: 0,
+          }).format()}
+        </Sans>
       </StyledLink>
     </Box>
   )
--- a/src/Components/CollectionsRail/Collection.tsx
+++ b/src/Components/CollectionsRail/Collection.tsx
@@ -1,0 +1,43 @@
+import { Box, color, Sans, Serif } from "@artsy/palette"
+import React, { SFC } from "react"
+import styled from "styled-components"
+import { resize } from "Utils/resizer"
+
+export interface CollectionProps {
+  headerImage?: string
+  price_guidance?: number
+  slug: string
+  title?: string
+}
+
+const Background = styled(Box)<{ collectionImage: string }>`
+  height: 175px;
+  width: 100%;
+  // min-width: 545px;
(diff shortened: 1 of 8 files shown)


artsyit:

Built with commit dbd4fc1df3b859bbf6b18d85779dca11dcdd824b

https://deploy-preview-2052--artsy-reaction.netlify.com

xtina-starr:
An additional thing I noticed is that when we pass in the size param as 4 we always get the same collections. I was thinking we may want to fetch all the collections and randomly select from what is returned.

eessex:
In regard to the `size` param, this is an excellent point, and we should probably raise this in sprint planning today to figure out an ideal solution. For the article infinite scroll I know we add an `offset` to the params each time additional articles are fetched. I'm not sure what the best tactic for this would be for this component, but it would likely require additional work in Kaws to allow `sort` or `offset`.

src/Components/CollectionsRail/Collection.tsx
```diff
@@ -0,0 +1,43 @@
...
+const CollectionTitle = styled(Serif)`
+  width: max-content;
+`
+
+export const CollectionEntity: SFC<CollectionProps> = props => {
+  return (
+    <Box mb={3} width="100%">
+      <Background
+        collectionImage={resize(props.headerImage, {
+          width: 820,
```
eessex: This width looks quite large for the size of the rail, maybe better to use `crop` to size to the max dimensions of the component display

src/Components/CollectionsRail/Collection.tsx
```diff
@@ -0,0 +1,43 @@
...
+          height: 240,
+          quality: 80,
+        })}
+      />
+      <CollectionTitle size="4">{props.title}</CollectionTitle>
+      <Sans size="2">Works from ${props.price_guidance}</Sans>
```
eessex: Not sure what the best tactic is for this, but maybe we have a library that can parse numbers to price? Worth investigation.

src/Components/CollectionsRail/CollectionsRail.tsx
```diff
@@ -0,0 +1,75 @@
...
+const CollectionsRow = styled(Row)`
+  justify-content: space-between;
+`
+
+export class CollectionsRail extends React.Component<CollectionRailsProps> {
+  getRails = () => {
+    const collections: any[] = this.props.collections.map(
+      (collection, index) => {
+        return (
+          <Col xl={5} lg={5} md={5} sm={12} key={index}>
```
eessex: In the Palette example it looks like you should be able to declare padding on the column, and leave its width at 6 or 50% -- https://palette.artsy.net/tokens/grid/
Would keep investigating this as it would allow you to remove the `CollectionsRow` above, and check with frontend-platform if it is not acting as expected.

damassi: Typically grid props should be reversed, from smallest to largest,  and it will take the lowest match and apply to all larger breakpoints. So, `md={5}` automatically applies 5 to lg and xl. Going in the opposite direction, down to sm and xs, the default is 12 columns. So this can be simplified to: 

```
<Col md={5}>
   ...
</Col>
```

(sorry, this is all kinda confusing! I can explain more if need be)

src/Components/CollectionsRail/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { CollectionsRailQuery } from "__generated__/CollectionsRailQuery.graphql"
+import { ContextConsumer } from "Artsy"
+import { CollectionsRailFragmentContainer as CollectionsRail } from "./CollectionsRail"
+
+interface Props {
+  showOnEditorial?: boolean
+}
+
+export class CollectionsRailContent extends React.Component<Props, null> {
```
damassi: This can be a `React.SFC`, and if no state, don't need to pass `null` as an interface:
```js
React.SFC<Props>
```

src/Components/Publishing/Layouts/Components/CanvasFooter.tsx
```diff
@@ -8,6 +8,7 @@ import { DisplayCanvas } from "Components/Publishing/Display/Canvas"
 import { RelatedArticlesCanvas } from "Components/Publishing/RelatedArticles/Canvas/RelatedArticlesCanvas"
 import React from "react"
 import styled from "styled-components"
+import { CollectionsRailContent } from "../../../CollectionsRail"
```
damassi: If going down further than one level (`../`) the path should be absolute.

src/Components/CollectionsRail/CollectionsRail.tsx
```diff
@@ -0,0 +1,75 @@
...
+    )
+  }
+}
+
+export const CollectionsRailFragmentContainer = createFragmentContainer(
+  CollectionsRail,
+  graphql`
+    fragment CollectionsRail_collections on MarketingCollection
+      @relay(plural: true) {
+      slug
```
eessex: Might make more sense to move the individual fields into the `Collection` component since we don't access any of these fields in this component

src/Components/CollectionsRail/CollectionsRail.tsx
xtina-starr: It looks like that's the case for individual columns so I would only want to do that on index 0 and 2. I guess I could conditionally do that. I'll give that a try.

damassi: Yah, ideally the whole grid can go and we can rely on just flexbox and our responsive styles

eessex: @damassi this @xtina-starr and I were discussing while pairing on this, is it preferable to use `Flex` and `Box` over `Grid` generally?

damassi: Each has its proper place, but I would lean (unless a grid for whatever reason makes total sense) on flex box just because in the end it's one less external dependency to worry about.

src/Components/CollectionsRail/CollectionsRail.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface CollectionRailsProps {
+  collections: CollectionsRail_collections
+}
+
+const RailsWrapper = styled(Flex)`
+  width: 100%;
+  max-width: 1250px;
+  margin: 0 auto;
+  flex-direction: column;
+  ${pMedia.xl`
```
eessex: We are trying to deprecate `pMedia` in favor of Palette's `media` wherever possible, though I'm guessing you are using this to match to further reading or another component that still uses it... probably worth adding a TODO to remove later.

xtina-starr: Yes, that's correct. I'll add a TODO. It also might be good to add as a task to the ticket for closing the test in the event this version wins.

src/Components/Publishing/Layouts/Components/CanvasFooter.tsx
```diff
@@ -18,6 +19,9 @@ export interface CanvasFooterProps {
 
 export const CanvasFooter: React.SFC<CanvasFooterProps> = props => {
   const { article, display, relatedArticles, renderTime } = props
+  const showCollectionRails = ["standard", "feature", "news"].includes(
```
eessex: This can be updated to use the `showCollectionsRail` prop from force, my pr is now updated to include your suggestion to limiting layouts from the top down. https://github.com/artsy/force/pull/3589

xtina-starr:
@eessex the `showCollectionsRail` being handled in Force should impact our ability to QA this right?

eessex:
@xtina-starr we may have to QA in reaction for visuals, we don't want to start setting cookies on the test until it is ready to turn on.  Not sure what the best solution for this is, perhaps running it locally.

artsyit:
:rocket: PR was released in v14.0.15 :rocket:

PR #2060, opened 2019-02-25
Don't dismiss the search results if the user clicked on a descendant component

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-756.

This fixes a timing bug. When a user searches, and they click on a link in the preview pane (right side), the results box closes without going anywhere. This is because the links in the preview pane are removed from the DOM before the browser can follow the link - the `onBlur` event of the search box is triggering a re-render with the clicked link removed.

So, we are grabbing a ref to the container. When the `onBlur` event fires for "leaving" the search box, we check to see if the `relatedTarget` is a descendant of the container ref. If it is, we set a flag indicating to other events that the search results should not dismiss. This allows the browser to actually follow the link.

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -107,6 +107,8 @@ const SuggestionContainer = ({ children, containerProps, preview }) => {
 @track()
 export class SearchBar extends Component<Props, State> {
   public input: HTMLInputElement
+  private containerRef: Node
+  private userClickedOnDescendant: boolean
 
   state = {
     term: "",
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit ccd505e9980bad27ac31b25568d1264d74217225

https://deploy-preview-2060--artsy-reaction.netlify.com

src/Components/Search/SearchBar.tsx
```diff
@@ -107,6 +107,8 @@ const SuggestionContainer = ({ children, containerProps, preview }) => {
 @track()
 export class SearchBar extends Component<Props, State> {
   public input: HTMLInputElement
+  private containerRef: any
```
mzikherman: I _think_ this can be improved from `any` and typed as `React.RefObject<HTMLDivElement>` (or something like that).

pepopowitz: Yes, thank you! Fixing.

pepopowitz: It ended up being just a `Node`. Not `React.RefObject<Node>`, just a plain old `Node`. 🤷‍♀️

jonallured:
Nice one, thanks for the explanation!! 💯

artsyit:
:rocket: PR was released in v14.1.3 :rocket:

PR #2061, opened 2019-02-25
DISCO-752: Add return icon to search suggestions

jonallured:
This PR refactors the way we draw items in the suggestion side of things so that I could ensure consistent height and implement the "enter" icon for active suggestions. Here's a few shots show it at different breaks:

screen shot 2019-02-25 at 1 47 06 pm
screen shot 2019-02-25 at 1 47 18 pm
screen shot 2019-02-25 at 1 47 35 pm

## Pushing Down

In order to feel good about this, I decided to take a step back and push down into the `SuggestionItem` component whatever I could. Prior to this refactoring the `SearchBar` was actually in charge of setting the highlight color on the item. This felt backwards to me so I added a `isHighlighted` prop and now the `SuggestionItem` knows what to do when it is highlighted and `SearchBar` is blissfully ignorant of the juicy details.

In a similar vein, I noticed that `SearchBar` knew more than I'd like about the empty suggestion and placeholder so I pushed those down as well and exposed a simple component for the former (`EmptySuggestion`) and a constant for the latter (`PLACEHOLDER`).

## The SVG

I've added the return icon here right in the `SuggestionItem` file rather than consuming it from Palette which is what I'd like to do. I looked at https://github.com/artsy/palette/issues/283 and was scared off by the amount of work there so this seemed pragmatic. In order to ensure that I don't just leave this here, I've set myself a reminder to come back to this. I'll either take on that palette work or even just move this single component over to Palette. Depends on where things stand at the time.


artsyit:

Built with commit c94d54a494f0b117061547dfbfae6e7771a0358a

https://deploy-preview-2061--artsy-reaction.netlify.com

src/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -1,45 +1,78 @@
...
-        <SuggestionItemContainer>Search "{query}"</SuggestionItemContainer>
+export const SuggestionItem: SFC<Props> = props => {
+  const { label, href, isHighlighted } = props
+
+  const Suggestion = label === "FirstItem" ? FirstSuggestion : DefaultSuggestion
+
+  return (
+    // Note: we've specified a color on the Link below to ensure an underline
+    // is suppressed on hover - technically the value does not matter.
+    <Box bg={isHighlighted ? "#ddd" : "#fff"}>
```
pepopowitz: should these be coming from Palette tokens/color names?

src/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -1,45 +1,78 @@
...
+import { Box, color, Flex, Link, Sans, Serif } from "@artsy/palette"
 import match from "autosuggest-highlight/match"
 import parse from "autosuggest-highlight/parse"
 import React, { SFC } from "react"
-import { SuggestionItemContainer } from "./SuggestionItemContainer"
+
 interface Props {
   display: string
+  href: string
+  isHighlighted: boolean
```
mzikherman: #minor i would expect this to be an optional prop that defaults to false

mzikherman:
Yea, nice refactor!

artsyit:
:rocket: PR was released in v14.1.5 :rocket:

PR #2064, opened 2019-02-25
Lazy-load images in the artwork grid

zephraph:
This adds lazy loading to the artwork grid. Addresses [PLATFORM-1164](https://artsyproduct.atlassian.net/browse/PLATFORM-1164). This demo uses 2G speed.

The placeholder slightly pulses to give the impression of something happening (suggested by Nicole) and fades in at our standard `250ms`. 

### Todo
- [x] Add the ability to specify how many images should be eagerly loaded for each grid
- [x] Setup up appropriate eager loading behaviors for artist, artwork, collect, collection


artsyit:

Built with commit 42b3aa764dad6d09d2194b1e9915bb8bec987f74

https://deploy-preview-2064--artsy-reaction.netlify.com

src/Components/Artwork/GridItem.tsx
```diff
@@ -1,27 +1,44 @@
...
+  50% { background-color: ${color("black5")}; }
+  100% { background-color: ${color("black10")}; }
+`
+
+const pulseAnimation = props =>
+  css`
+    ${pulse} 2s ease-in-out infinite;
+  `
+
+const Image = styled(LazyLoadImage)`
```
damassi: Wondering, should this be configured so that it lazy loads below the fold content? Similar to how we trigger the "Recently Viewed" rail to appear before the user has scrolled to it. Seeing the image jump in in real time isn't too optimial.

src/Components/Artwork/GridItem.tsx
```diff
@@ -34,6 +51,7 @@ interface Props extends React.HTMLProps<ArtworkGridItemContainer> {
 
 interface State {
   isMounted: boolean
+  isImageLoaded: boolean
```
damassi: Should the `<ArtworkGrid>` take a `lazyLoad` configuration prop? I'm thinking of pages like: https://www.artsy.net/collect where the the grid is at the top of the screen and you'd want to be able to see the images on load and with JS disabled.

src/Components/Artwork/GridItem.tsx
damassi: (Ah, I see the `todo` item in the description)

src/Components/Artwork/GridItem.tsx
zephraph: I plan to have a prop for `ArtworkGrid` that accepts a number of images to eagerly load. So for the collect page 9 images on desktop, 6 images on mobile would be ideal. For artist/artwork page 3 to 6 images makes sense to eagerly load on desktop, but you'd probably want none on mobile. 

As per the js disabled bit... that should generally be handled by the lib. I'll look into if it is, if not, I'll PR it.

src/Components/Artwork/GridItem.tsx
zephraph: The library offers a threshold that we can use to set the load distance. It's currently 100px below the fold. Fast scrollers are one case, slow phones are another. We can tweak the setting if needed. Might be worth standardizing on a distance below-the-fold loading threshold.

damassi: Cool, I bet a couple hundred pixels would be sufficient.

src/Components/Artwork/GridItem.tsx
alloy: Is there no way for the lib to just eager load above the fold? (Not sure if the above thread is about that.)

src/Components/Artwork/GridItem.tsx
```diff
@@ -167,7 +193,11 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
               }
             }}
           >
-            <Image src={this.getImageUrl()} />
+            <Image
+              style={{ opacity: isImageLoaded ? "1" : "0" }}
+              src={this.getImageUrl()}
+              onLoad={this.imageLoaded.bind(this)}
+            />
```
alloy: Will there be a `img` tag with `src` attribute coming from the SSR content? If not, does that affect SEO in any way?

cavvia: I had the same question actually... will point nicolette to this for her thoughts.

src/Components/Artwork/GridItem.tsx
zephraph: Nope. I mean, when you load the page and the js initializes it will auto load anything in view (and within the threshold). But the above the fold loading decision has to be made at SSR which falls back to the traditional not knowing the user’s browser size.

src/Components/Artwork/GridItem.tsx
zephraph: From an SSR perspective, no. Based on google’s own recommendations, this should be fine as long as the image does indeed pull in as the bot scrolls. https://developers.google.com/search/docs/guides/lazy-loading. Traditionally lazy-loaded imgs use the `data-src` attribute as a placeholder before loading, but that’s not the case with this library.

NicoletteChao: Would just raise that traditionally `src` and `alt` are used to help Google understand what the image is. Since `src` is typically a cloudfront URL for us without a description of the artwork I think it's okay to forego that, however, can we add `alt` text to the images? They could just be the artwork title. 

I believe we used to have these on our artist pages, or at least, in the snapshot we provided to Google through reflection.

zephraph: I actually noticed we're not really doing a good job of that anymore... I'll try to fix it up. 👍

zephraph:
I made some general updates. I don't use the lazy-load component on preloaded images. That helps to not trigger the lazy loading animation unnecessarily. I also added some `noscript` tags to cover the case where js doesn't load. I added `title` and `alt` tags to the images. I think the biggest thing remaining is to tweak the amount of preloaded images on certain pages (like the Collect page).

src/Components/Artwork/GridItem.tsx
zephraph: I've manually configured the lazy-loading values for each case. For artwork grids completely above the fold (think collect, works-for-you) there are 9 preloaded images. That seems to cover a pretty tall screen.

src/Components/Artwork/GridItem.tsx
zephraph: I've buttoned this up. `title` attributes come from the artwork `title` field and the `alt` attribute is provided by the `image_title` field (that has details such as author and year).

src/Apps/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/ArtistArtworkGrid.tsx
```diff
@@ -45,6 +45,7 @@ export class ArtistArtworkGrid extends React.Component<ArtistArtworkGridProps> {
         <ArtworkGrid
           artworks={artist.artworks_connection}
           columnCount={[2, 3, 4]}
+          preloadImageCount={0}
```
damassi: Should this be a default prop?

src/Apps/Collect/Components/Base/CollectArtworkGrid.tsx
```diff
@@ -86,6 +86,7 @@ class CollectArtworkGrid extends Component<Props, LoadingAreaState> {
                     <ArtworkGrid
                       artworks={artworks as any}
                       columnCount={columnCount}
+                      preloadImageCount={9}
```
damassi: Or should _this_ be a default prop 😄

src/Apps/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/ArtistArtworkGrid.tsx
zephraph: This definitely shouldn't be! That'd mean they always lazy load by default... which, I mean might be good sometimes, but might not others.

src/Components/Artwork/GridItem.tsx
```diff
@@ -73,6 +101,12 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
     }
   }
 
+  imageLoaded() {
```
damassi: `imageLoaded = () =>` and you can loose the `.bind` down below.

src/Components/Artwork/GridItem.tsx
```diff
@@ -167,7 +202,28 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
+                alt={artwork.image_title}
+                src={this.getImageUrl()}
+              />
+            ) : (
+              <LazyImage
+                title={artwork.title}
+                alt={artwork.image_title}
+                style={{ opacity: isImageLoaded ? "1" : "0" }}
+                src={this.getImageUrl()}
+                onLoad={this.imageLoaded.bind(this)}
```
damassi: See above

src/Apps/Collect/Components/Base/CollectArtworkGrid.tsx
zephraph: It _could_ be. Right now I have the default as 6 which seems to cover most cases.

src/Components/Artwork/GridItem.tsx
```diff
@@ -167,7 +202,28 @@ class ArtworkGridItemContainer extends React.Component<Props, State> {
...
+              />
+            )}
+            <noscript>
+              <Image
+                title={artwork.title}
+                alt={artwork.image_title}
+                src={this.getImageUrl()}
+              />
+            </noscript>
```
damassi: SEO?

zephraph: Neither. Just the no js use case. I mean, it _might_ be unnecessary... so much of our site wouldn't work without js anyway. But.. yeah. SEO effectively ignores this and I doubt it'd be used for accessibility... though I'm not 100% sure of that.

damassi: OH duh, for no JS -- but that said, our new pages are totally browsable / usable without JS turned on so we want to keep things fluid where possible -- this seems like a good solution 👍 .

damassi:
Pulled this down and scrolled and the fade-in distance timings felt very natural / expected. Turned off JS and also looked great 💯

Mergin

artsyit:
:rocket: PR was released in v15.2.0 :rocket:

PR #2072, opened 2019-02-26
Make @sentry/browser a peer dependency of reaction

zephraph:
This removes `@sentry/browser` from being a direct dependency of reaction. Libraries like this that are utilized by the parent project shouldn't be direct dependencies. This was causing a duplication in our upstream webpack dependency in force. 

I'll go through and ensure positron, volt, and whoever else depends on reaction has this dependency specified in their projects.  I'm using [this link](https://github.com/search?q=org%3Aartsy+%22%40artsy%2Freaction%22+filename%3Apackage.json&type=Code) to look through the org for things that depend on reaction and ensure they have sentry too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "react": "^16.5.0",
     "react-dom": "^16.5.0",
     "react-router": "^4.2.0",
-    "styled-components": "^4"
+    "styled-components": "^4",
+    "@sentry/browser": "^4.2.3"
   },
   "scripts": {
     "clean": "rm -rf dist",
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 869cbcf0675ad311f60995af93e653f3913796b8

https://deploy-preview-2072--artsy-reaction.netlify.com

damassi:
@zephraph - I _think_ that this sentry code only applies to our new Apps and is constrained to places that use `<Boot>`, meaning [this right here](https://github.com/artsy/reaction/blob/master/src/Artsy/Router/Components/Boot.tsx#L48-L64).

damassi:
Also, now that I'm thinking about it, if we've already registered Sentry in force this bit of code could be removed completely.

artsyit:
:rocket: PR was released in v14.1.14 :rocket:

PR #2077, opened 2019-02-27
View-in-room: Select default image when "view in room" is clicked

lilyfromseattle:
Slider moves to default image before view in room


artsyit:

Built with commit f255ade971757960c7c09954ebe65bec0305c638

https://deploy-preview-2077--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -108,6 +117,15 @@ export class SmallArtworkImageBrowser extends React.Component<
...
   ArtworkBrowserState
 > {
+  slider: Slider
+
+  setSliderRef = slider => {
+    this.slider = slider
+    if (this.props.sliderRef) {
+      this.props.sliderRef(slider)
+    }
+  }
```
ds300: It turns out that `SmallArtworkImageBrowser` doesn't need to maintain a reference to the slider so we can delete these two things (`slider` and `setSliderRef`) and just pass `this.props.sliderRef` directly to the `ref` property on the `Slider` element.

src/Apps/Artwork/Components/ArtworkImageBrowser/index.tsx
```diff
@@ -5,28 +5,52 @@ import { ArtworkImageBrowser_artwork } from "__generated__/ArtworkImageBrowser_a
...
+    const { images, image } = this.props.artwork
     if (!images.length) {
       return null
     }
+
+    const defaultImageIndex = images
+      .map(e => {
+        return e.id
+      })
+      .indexOf(image.id)
```
ds300: #superminor you could also do `images.findIndex(e => e.id === image.id)`

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -17,16 +17,17 @@ import {
...
+export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
   return (
     <>
       <Media at="xs">
-        <SmallArtworkImageBrowser images={props.images} />
+        <SmallArtworkImageBrowser {...props} />
       </Media>
       <Media greaterThan="xs">
-        <LargeArtworkImageBrowser images={props.images} />
+        <LargeArtworkImageBrowser {...props} />
```
damassi: Why use the spread here?

ds300: These two components take the same props as the wrapper component, so why not pass them all along.

sweir27:
Agree with comments above ☝️ but otherwise looks good to me! ✨ 

Would you mind adding a screen cap (I use `LiceCAP` but there are probably other, fancier solutions) that shows the new behavior?

In this case, since all this PR does is explicitly select the default image when "View in room" is clicked and sets the `data-is-default` attribute, can we update the description to reflect that? Maybe something like `View-in-room: Select default image when "view in room" is clicked`?

ds300:
> other, fancier solutions

https://getkap.co/ :heart:

artsyit:
:rocket: PR was released in v14.1.28 :rocket:

PR #2085, opened 2019-02-28
Artist Collections Rail initial setup

eessex:
Related to https://artsyproduct.atlassian.net/browse/GROW-1140

- Scaffolds collections rail components for artist page test
- Sets up display of collections rail in artist page
- Adds initial test config for a/b test in artist page

This is not the complete implementation, which is blocked by changes to the Kaws API before we can implement Relay.  This step will be addressed separately once fields are available. This PR sets up the basic component display and formatting, and a mechanism to display it in artist pages based on test params from force. 

For now I've added the rail component to `Components/Artist/CollectionsRail`, we may want to reevaluate where the different collections components live when we start the next placement.  Wanted to keep separate from `Components/CollectionsRail` until the editorial test is launched to avoid slowing down that process.

screen shot 2019-02-28 at 12 08 42 pm


artsyit:

Built with commit 5abc76daff6f2884f3e2f0cbac815a1c670e28f5

https://deploy-preview-2085--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -52,8 +55,28 @@ class OverviewRoute extends React.Component<OverviewRouteProps, State> {
     return showGenes
   }
 
+  trackingCollectionsRailTest() {
```
dblandin: I don't see the `trackingCollectionsRailTest` function invoked elsewhere in this file. If we want the `ExperimentViewed` event recorded, I think that we'll want to bring in a `TrackingProp` and and use `tracking.trackEvent()`.

I removed similar code after the Artist Insights A/B test ended: https://github.com/artsy/reaction/pull/1919/files

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -143,6 +166,14 @@ class OverviewRoute extends React.Component<OverviewRouteProps, State> {
 
         {!hideMainOverviewSection && <Spacer mb={4} />}
 
+        {showCollectionsRail && ( // TODO: remove after CollectionsRail a/b test
+          <div>
+            <Separator mb={3} />
+            <ArtistCollectionsRail />
+            <Spacer mb={3} />
+          </div>
```
dblandin: I remember learning a cool shorthand for `<div>` from @jonallured, I think. Unless you're passing properties to the tag, you can use:

```js
<>
  <Separator mb={3} />
  <ArtistCollectionsRail />
  <Spacer mb={3} />
</>
```

src/Components/Artist/CollectionsRail/CollectionsRail.tsx
```diff
@@ -0,0 +1,44 @@
+import { Box, Sans, Spacer } from "@artsy/palette"
+import { ArrowButton, Carousel } from "Components/v2"
+import React, { SFC } from "react"
+import styled from "styled-components"
+import { ArtistCollectionEntity } from "./CollectionEntity"
+
+interface ArtistCollectionsRailProps {
+  collections?: any
```
dblandin: Is there a `Collection` typescript type we can use here?

dblandin:
I think that this should be tagged as a `Version: Minor` change in the GitHub labels since we're adding a new feature.

src/Apps/Artist/Routes/Overview/index.tsx
eessex: Unfortunately the fragment doesn't seem to work well as a wrapper for the carousel

src/Components/Artist/CollectionsRail/CollectionsRail.tsx
eessex: This is an incremental PR, but we will have a typing soon -- we are in the process of adding new fields to kaws and the schema has not yet been defined. This will be updated as soon as the API supports data to fetch with relay.

src/Apps/Artist/Routes/Overview/index.tsx
eessex: This is just a placeholder until we are ready to implement the test, but maybe it is better to comment out for now, or to call it and comment that line out.

src/Apps/Artist/Routes/Overview/index.tsx
jonallured: Oops, misunderstanding somewhere, the short hand syntax you're showing here would result in a Fragment, see also:

https://reactjs.org/docs/fragments.html#short-syntax

Looks like <div></div> is what @eessex is after here.

src/Apps/Artist/Routes/Overview/index.tsx
jonallured: FWIW, I'm pretty sure using the tracking prop is an anti-pattern in reaction and the guideline is to use a decorator. My work with this is still WIP, but I got this feedback over on #2030.

l2succes: yup, btw we should use the `@track` decorator for analytics events instead of calling `tracking.trackEvent()` directly whenever possible

src/Components/Artist/CollectionsRail/CollectionEntity.tsx
```diff
@@ -0,0 +1,68 @@
...
+  collection,
+}) => {
+  const { price_guidance, slug, title, headerImage } = collection
+  const formattedTitle = title.split(": ")[1] || title
+
+  return (
+    <Box width="100%" pr={2}>
+      <StyledLink href={`${sd.APP_URL}/collection/${slug}`}>
+        <ImgWrapper pb={1}>
+          <ImgPlaceholder headerImage={headerImage} />
```
xtina-starr: I know we haven't finalized the data, but I wonder if it's worth switching the structuring of the headerImage to be some sort of collection/array property on the `collection` object (i.e. `collection.railImages`). These will be the icon images or curated in some cases, correct? It may not be worth tackling in this PR but is something to keep in mind.

xtina-starr:
This is looking good so far.

xtina-starr:
This may be out of scope for this initial PR, but I did notice that these rails are above the fold and I believe the Artist page is SSR. I think we will run into issues with the `randomize` argument and the double render.

src/Apps/Artist/Routes/Overview/index.tsx
eessex: sounds good -- based this on the [artist test ](https://github.com/artsy/reaction/pull/1919/files#diff-bfc57d8ce9d60a71675e59322d85c6f3L68) but can def edit to use the decorator pattern.

src/Components/Artist/CollectionsRail/CollectionEntity.tsx
eessex: If you think necessary to change will just remove the image and leave a gray box for now, not sure what the field name or shape will look like so don't want to over-engineer.

eessex:
@xtina-starr the randomize arg is an interesting point that I hadn't thought of -- was not planning to use randomize since these are explicitly selected for display, but potentially would be of interest. Will check with Jun for clarification.

src/Components/Artist/CollectionsRail/CollectionEntity.tsx
xtina-starr: Yeah that makes sense. I think it's fine to leave it as is for now.

eessex: Went ahead and removed to remove any ambiguity, added a comment that background URL should be updated when fields are available.

eessex:
Updated to address comments, and also added more context to the PR description to indicate that this is a partial step towards a new feature.  Agreed that we should do a minor release when this component is ready to go, but this PR is not a complete implementation and would rather wait till it is ready for use.

src/Apps/Artist/Routes/Overview/index.tsx
dblandin: Ahhh, gotcha. :+1:

eessex:
@dblandin ok to merge? This will render nothing in the Artist page until we enable the test in Force.

artsyit:
:rocket: PR was released in v14.2.1 :rocket:

PR #2097, opened 2019-03-01
Measure performance of search bar

dblandin:
This pull request reports the duration between entering a term in the new search bar and rendering search suggestions to the user.


artsyit:

Built with commit 626ae6ab1bf736709e3ee5203f7b756e49c2b12a

https://deploy-preview-2097--artsy-reaction.netlify.com

src/Components/Search/SearchBar.tsx
```diff
@@ -175,8 +178,33 @@ export class SearchBar extends Component<Props, State> {
...
+  }
+
+  shouldMeasurePerformance = () => {
+    return typeof performance !== "undefined" && sd.VOLLEY_ENDPOINT
+  }
+
   searchTextChanged = (_e, { newValue: term }) => {
     this.setState({ term })
+
+    if (this.shouldMeasurePerformance()) {
```
mzikherman: Why not put these lines (and the reporting) right in the Relay refetch call? Here it looks like they are lumped into some render/callbacks/side-effect-y places.

I think it can just go in https://github.com/artsy/reaction/blob/6e3144e252f646cf1d45ea2b492b87752948c0b8/src/Components/Search/SearchBar.tsx#L148 and then you probably don't need the `private performanceStart` declaration either (can just be a local in there).

src/Components/Search/SearchBar.tsx
```diff
@@ -21,7 +21,9 @@ import {
   QueryRenderer,
   RelayRefetchProp,
 } from "react-relay"
+import { data as sd } from "sharify"
 import styled from "styled-components"
+import request from "superagent"
```
mzikherman: interesting- didn't realize we had this in Reaction!

src/Components/Search/SearchBar.tsx
```diff
@@ -175,8 +178,33 @@ export class SearchBar extends Component<Props, State> {
...
+
+    const payload = {
+      type: "timing",
+      name: "autocomplete-search-response",
+      timing: duration,
+      tags: [`device-type:${deviceType}`, "design:rich"],
+    }
+
+    request.post(sd.VOLLEY_ENDPOINT).send({
+      serviceName: "force",
```
mzikherman: #minor this is sort of hardcoded (which I think is fine!), but possibly `VOLLEY_SERVICE_NAME` should be an env var.

dblandin: Good point. Is that something that needs to be added to Force first? Maybe a reasonable follow-up PR?

artsyit:
:rocket: PR was released in v14.2.11 :rocket:

PR #2098, opened 2019-03-01
Hide artist insights on artwork page behind button

lilyfromseattle:
Button toggles artist insights. Button is hidden if artist insights have no data.

screen shot 2019-03-06 at 3 41 46 pm

screen shot 2019-03-06 at 3 43 56 pm

__Artwork page with multiple artists:__
Screen Shot 2019-03-13 at 10 37 19 AM

Screen Shot 2019-03-13 at 10 37 27 AM


artsyit:

Built with commit b04b74badbf9455c9a3e31b87fc23d3fe512a03d

https://deploy-preview-2098--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -114,24 +142,39 @@ export class ArtistInfo extends Component<ArtistInfoProps> {
                   />
                 </>
               )}
+              {showArtistInsightsButton && (
+                <Button
+                  onClick={this.toggleArtistInsights}
+                  width={130}
```
ds300: I think instead of specifying a static width here, we should be making sure the button hugs its inner text (as opposed to stretching to fit it's parent container).

I'd probably do that be wrapping the button in a 

    <Flex flexDirection="column" alignItems="flex-start">

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -114,24 +142,39 @@ export class ArtistInfo extends Component<ArtistInfoProps> {
                   />
                 </>
               )}
+              {showArtistInsightsButton && (
```
ds300: Should the button disappear after having been clicked?

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -159,6 +208,36 @@ export const ArtistInfoFragmentContainer = createFragmentContainer(
       exhibition_highlights(size: 3) {
         ...SelectedExhibitions_exhibitions
       }
+      collections
```
mzikherman: It might be worth a comment here explaining that all this stuff has been added to the fragment for the exist-y check for the button. This winds up being consolidated and deduped with the data requirement from the fragment here:  https://github.com/artsy/reaction/blob/6e3144e252f646cf1d45ea2b492b87752948c0b8/src/Components/Artist/MarketInsights/MarketInsights.tsx#L219

If one of the fragments is then subsequently changed, it could lead to overfetching so it might be worth a comment!

sweir27:
@lilyfromseattle would you mind also adding a gif or screenshot of the behavior here?

src/Apps/Artwork/Components/ArtistInfo.tsx
lilyfromseattle: Nicole said animation wasn't necessary

src/Apps/Artwork/Components/ArtistInfo.tsx
sweir27: The desire is to have it remain with an updated message (show vs. hide).

src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -114,35 +146,59 @@ export class ArtistInfo extends Component<ArtistInfoProps> {
...
-              totalExhibitions={this.props.artist.counts.partner_shows}
-              exhibitions={this.props.artist.exhibition_highlights}
-              ViewAllLink={
-                <a href={`${sd.APP_URL}/artist/${this.props.artist.id}/cv`}>
-                  View all
-                </a>
-              }
-              Container={Container}
-            />
+            {this.state.showArtistInsights && (
```
sweir27: One thing I noticed is that this version loses the bottom border in the case of multiple artists. (Possibly because of the extra fragment container here messing with the `StackableBorderBox`?).

See https://artsy-reaction.netlify.com/?path=/story/apps-artwork-page--artwork-with-multiple-artists (current state)
vs. https://deploy-preview-2098--artsy-reaction.netlify.com/?selectedKind=Apps%2FArtwork%20Page&selectedStory=Artwork%20with%20multiple%20artists&full=0&addons=0&stories=1&panelRight=0

image

vs. 

image

sweir27:
@lilyfromseattle I think this one is good but like @ds300 said, can you merge in master? I think we updated our circle checks, so that should get this one in a 🍏 state 😄 .

artsyit:
:rocket: PR was released in v15.8.4 :rocket:

PR #2102, opened 2019-03-04
Disco 786 xs placeholder

pepopowitz:
Currently, we are rendering no placeholder for the search bar at xs sizes. We probably did this initially because the usual placeholder took up too much space. 

This PR uses "Search Artsy" as the placeholder at xs.

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -10,6 +10,7 @@ import { SearchPreviewWrapper as SearchPreview } from "Components/Search/Preview
 import {
   EmptySuggestion,
   PLACEHOLDER,
+  PLACEHOLDER_XS,
   SuggestionItem,
 } from "Components/Search/Suggestions/SuggestionItem"
 import { throttle } from "lodash"
@@ -298,7 +299,7 @@ export class SearchBar extends Component<Props, State> {
       onChange: this.searchTextChanged,
       onFocus: this.onFocus.bind(this),
       onBlur: this.onBlur,
-      placeholder: xs ? "" : PLACEHOLDER,
+      placeholder: xs ? PLACEHOLDER_XS : PLACEHOLDER,
       value: term,
       name: "term",
     }
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -432,7 +432,7 @@ export const SearchBarQueryRenderer: React.SFC = () => {
                   <Input
                     name="term"
                     style={{ width: "100%" }}
-                    placeholder={PLACEHOLDER}
+                    placeholder={PLACEHOLDER_XS}
                   />
                 )
               }
--- a/src/Components/Search/Suggestions/SuggestionItem.tsx
+++ b/src/Components/Search/Suggestions/SuggestionItem.tsx
@@ -47,6 +47,7 @@ const InnerWrapper = styled(Flex)`
 `
 
 export const PLACEHOLDER = "Search by artist, gallery, style, theme, tag, etc."
+export const PLACEHOLDER_XS = "Search Artsy"
 
 export const EmptySuggestion = () => (
   <SuggestionWrapper>{PLACEHOLDER}</SuggestionWrapper>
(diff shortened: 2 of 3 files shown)


artsyit:

Built with commit 9f8d5a4cc133cd70a27741e0e85cd66f1899e36b

https://deploy-preview-2102--artsy-reaction.netlify.com

src/Components/Search/SearchBar.tsx
```diff
@@ -427,13 +427,7 @@ export const SearchBarQueryRenderer: React.SFC = () => {
...
                 )
               } else {
-                return (
-                  <Input
-                    name="term"
-                    style={{ width: "100%" }}
-                    placeholder={PLACEHOLDER}
-                  />
-                )
+                return null
```
pepopowitz: This branch executes when you're loading up storybooks. I _think_ that's the only time, but let me know if you think otherwise.

I removed the `<Input>` that was rendered because it doesn't seem to add much value, and it was another place we were using `PLACEHOLDER` but probably didn't need to.

Also, `return null` seems to be a pattern we use for these branches, [as seen here](https://github.com/pepopowitz/artsy/blob/9f8d5a4cc133cd70a27741e0e85cd66f1899e36b/src/Components/Onboarding/Steps/Genes/GeneSearchResults.tsx#L213).

mzikherman: This is the code path taken before hydration, so I think by adding this `return null`, if you link to Force and load up a page locally w/ JS disabled, you won't see any search bar.

So the 'base case' of just returning an `<input>` (which won't generate any suggestions/previews), but which still allows the user to type into and hit 'enter' to get to search results should still be supported.

For the SSR render, we won't have any screen size/mobile device flags accessible (there is a `res.locals.sd.IS_MOBILE` in Force (https://github.com/artsy/force/blob/7c78fb679885745008a0c0bdb41a538bab88fb21/src/lib/middleware/locals.coffee#L46) that might need to be passed in as a prop where the component is initialized: https://github.com/artsy/force/blob/7c78fb679885745008a0c0bdb41a538bab88fb21/src/desktop/components/react/stitch_components/SearchBar.tsx#L12

But, I would skip that and possibly just go with 'Search Artsy' here.

mzikherman: In terms of the `return null` pattern- that does make sense when you need to fetch data and then render it, since most likely before the data is fetched you just dont want to render anything. That certainly applies to a lot of our components. However, for some, instead of not rendering anything you may want to render a loading state/spinner or some other default state, so we wind up using both patterns.

pepopowitz: Regarding SSR, this doesn't seem to be the case - when I linked force to reaction & ran without JavaScript, I still saw a search bar. It came from here - https://github.com/artsy/force/blob/84267a709e8c2976f587ef3dcb5d2c2ecf208fa6/src/desktop/components/search_bar/templates/index.jade#L10.

I'll put the branch back with the xs placeholder, though.

mzikherman: That's actually the old search bar you linked to, so I wouldn't expect to ever see that if you have the lab feature on.

I double checked my suspicions and turns out I was incorrect! I got my wires crossed and mis remembered. This branch winds up getting rendered (briefly) during the async lifecycle of the query renderer (the initial 'empty' request which is returned early: https://github.com/artsy/reaction/blob/d4fbda80cbb7bb1f5022a42d151cbe3303247c11/src/Artsy/Relay/createRelaySSREnvironment.ts#L81-L82).

In reality it's so brief I haven't been able to visually see it even refreshing a bunch of times, but if you put a `debugger` into the `return null` part of the branch, you'll see that it _is_ rendered at some point, however briefly.

Thank you for pushing back on my SSR comment- it has nothing to do with that. I also think that possibly a `return null` is fine here, since the Relay environment returns early/immediately from that initial query- even though it's async its actually imperceptible.

mzikherman: Anyways, that all being said, I think it's probably better to keep an input component, rather than a null/nothing. In the unlikely event _something_ errors in the client-side JS and the user is left looking at this branch, instead of a disappearing search bar there will be an input box.

artsyit:
:rocket: PR was released in v14.2.9 :rocket:

PR #2111, opened 2019-03-05
Artwork page - Some artist names are taking up two lines in sidebar

lilyfromseattle:
Some artist names are taking up two lines in sidebar

Before:
screen shot 2019-03-06 at 1 30 52 pm

After:
screen shot 2019-03-06 at 1 59 00 pm

screen shot 2019-03-06 at 1 59 11 pm


artsyit:

Built with commit 3eb2ac6fd0af7aa688e07138323c7292cc354c3c

https://deploy-preview-2111--artsy-reaction.netlify.com

sweir27:
👍 Cool!

Can you post a few screenshots showing what it looks like now? I'm also interested in the case where an artist name is very long (since in that case it should wrap).

oxaudo:
I'm also wondering how that might affect the cases of Multiple artists. I think now they will be pretty compact when the name can be in one row and last name is on the next one. But with this change they will most likely take one row each. That might be a desired behavior actually but just bringing it up here so we make this change consciously.

See multiple artist spec here: https://app.zeplin.io/project/5b113f413ede4bb10dbfdc42/screen/5b114000413156335e285483

lilyfromseattle:
Updated my changes to account for multiple artists and added some screenshots - PR is no longer WIP

ds300:
I think `white-space: pre` is not a good solution for this, since it prevents word wrapping entirely. This would mean artists with long names might have part of their name cut off. 

Having a quick look at a random 10,000 artists, there are some artists with really long names, and also more than a few cases where galleries have put multiple artist names in the 'name' field.

image

Having normal word wrapping would be ideal. I think it should be a simple dom structure change, but figuring out where might be tricky. Happy to pair on this! :+1:

artsyit:
:rocket: PR was released in v15.3.4 :rocket:

PR #2114, opened 2019-03-06
Refactor spinner function

jonallured:
While working with @pepopowitz on some styling improvement for the new search bar, we noticed that it's pretty hard to control how the spinner renders because you don't have access to the wrapper. This PR adds the ability to pass one in should you want that sort of control.

In order to feel good about the state of this file, I started by removing commented out code and some old todos. My philosophy is that todos like these are lies we tell ourselves with the best of intentions but are ultimately noise. Further, since this code exists in git, commenting it out is unnecessary. If or when we come back to the concept of retry, we can review the history of this file and get what we need from there. If returning to retry is important and something we don't want to forget to do, then I would argue that a ticket in Jira is far superior to the todos and comments. I would happily work with someone to get this ticket added!!

With that done, the next thing I noticed was that there was a lot of noise around handling errors, so I extracted a function for that part of this and that made it a lot easier to find the heart of this function.

Finally, I added a third argument to the function for the wrapper. It defaults to the existing `SpinnerContainer` so no calling code has to be changed. I played with typing it something other than `any`, but nothing I tried would work - open to ideas!! ❤️

### DIFF
--- a/src/Artsy/Relay/renderWithLoadProgress.tsx
+++ b/src/Artsy/Relay/renderWithLoadProgress.tsx
@@ -17,32 +17,36 @@ const SpinnerContainer = styled.figure`
 
 export const LoadingClassName = "relay-loading"
 
+const handleError = error => {
+  // In tests we want errors to clearly bubble up.
+  if (typeof jest !== "undefined") {
+    throw error
+  }
+
+  if (error.message) {
+    console.error(error.message)
+  }
+
+  const networkError = error as any
+  if (networkError.response && networkError.response._bodyInit) {
+    let data = networkError.response._bodyInit
+    if (data) {
+      try {
+        data = JSON.parse(data)
+        console.error(`Metaphysics Error data:`, data)
+        // tslint:disable-next-line:no-empty
+      } catch (e) {}
+    }
+  }
+}
+
 export function renderWithLoadProgress<P>(
   Container: RelayContainer<P>,
   initialProps: object = {}
 ): (readyState: ReadyState<P>) => React.ReactElement<RelayContainer<P>> | null {
   return ({ error, props, retry }) => {
     if (error) {
-      // In tests we want errors to clearly bubble up.
-      if (typeof jest !== "undefined") {
-        throw error
-      }
-
-      const networkError = error as any
-      if (error.message) {
-        console.error(error.message)
-      }
-      if (networkError.response && networkError.response._bodyInit) {
-        let data = networkError.response._bodyInit
-        if (data) {
-          try {
-            data = JSON.parse(data)
-            console.error(`Metaphysics Error data:`, data)
-            // tslint:disable-next-line:no-empty
-          } catch (e) {}
-        }
-      }
-
+      handleError(error)
       return null
     } else if (props) {
       return <Container {...initialProps} {...props as any} />
--- a/src/Artsy/Relay/renderWithLoadProgress.tsx
+++ b/src/Artsy/Relay/renderWithLoadProgress.tsx
@@ -21,10 +21,6 @@ export function renderWithLoadProgress<P>(
   Container: RelayContainer<P>,
   initialProps: object = {}
 ): (readyState: ReadyState<P>) => React.ReactElement<RelayContainer<P>> | null {
-  // TODO: We need design for retry-ing, or the go-ahead to re-use the design of
-  //       the iOS app.
-  //
-  // let retrying = false
   return ({ error, props, retry }) => {
     if (error) {
       // In tests we want errors to clearly bubble up.
@@ -47,18 +43,6 @@ export function renderWithLoadProgress<P>(
         }
       }
 
-      // if (retrying) {
-      //   retrying = false
-      //   // TODO: Even though this code path is reached, the retry button keeps spinning. iirc it _should_ disappear when
-      //   //      `onRetry` on the instance is unset.
-      //   //
-      //   // This will re-use the native view first created in the renderFailure callback, which means it can
-      //   // continue its ‘retry’ animation.
-      //   return <LoadFailureView style={{ flex: 1 }} />
-      // } else {
-      //   retrying = true
-      //   return <LoadFailureView onRetry={retry} style={{ flex: 1 }} />
-      // }
       return null
     } else if (props) {
       return <Container {...initialProps} {...props as any} />
--- a/src/Artsy/Relay/renderWithLoadProgress.tsx
+++ b/src/Artsy/Relay/renderWithLoadProgress.tsx
@@ -2,6 +2,7 @@ import { Spinner } from "@artsy/palette"
 import React from "react"
 import { ReadyState, RelayContainer } from "react-relay"
 import styled from "styled-components"
+import createLogger from 'Utils/logger'
 
 /**
  * WARNING: Do _not_ change this element to something common like a div. If the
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 370695aec83ff19555e0ef89bbd75b8f23cf2c48

https://deploy-preview-2114--artsy-reaction.netlify.com

pepopowitz:
This comment was placed right above the definition of `SpinnerContainer` - 

```
/**
 * WARNING: Do _not_ change this element to something common like a div. If the
 * element of this container is the same as the element used in the RelayContainer
 * then rehydration can fail and cause the RelayContainer to receive styles
 * from the SpinnerContainer and Spinner.
 */
```

I don't really understand how that works, but by allowing the user to pass their own `SpinnerContainer, I worry that we're providing them a pretty big foot-gun. There's nothing to stop them from passing in a `div` now, and falling victim to what that comment describes.

This makes me feel like it's safer to allow them to pass in props or styles that we'd apply to the `SpinnerContainer`, but never let them control what type of element it is.

jonallured:
Yeah, I had this same thought. I too don't fully understand that comment, but maybe clever typing could help us here? What I was trying to do with that wrapper argument was do something like:

```
React.ReactElement<FigureProps>
```

But I couldn't really figure out how to do it and it wasn't seeming to play nice with styled components so I punted.

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -17,56 +17,45 @@ const SpinnerContainer = styled.figure`
...
+      } catch (e) {}
+    }
+  }
+}
+
 export function renderWithLoadProgress<P>(
   Container: RelayContainer<P>,
-  initialProps: object = {}
+  initialProps: object = {},
+  Wrapper: any = SpinnerContainer,
```
ds300: `Wrapper` feels a little nonspecific to me, since it's only used for wrapping the loading spinner. How about `SpinnerWrapper`? Or why not even just let consumers specify the entire thing by making this `LoadingComponent: React.ComponentType<{}> = ...`

and then instead of returning

        <Wrapper className={LoadingClassName}>
          <Spinner />
        </Wrapper>

you just return

        <LoadingComponent />

Then you could have

    const DefaultSpinner = () => (
        <SpinnerContainer className={LoadingClassName}>
          <Spinner />
        </SpinnerContainer>
    )

and 

`LoadingComponent: React.ComponentType<{}> = DefaultSpinner`

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -17,56 +17,45 @@ const SpinnerContainer = styled.figure`
...
+    let data = networkError.response._bodyInit
+    if (data) {
+      try {
+        data = JSON.parse(data)
+        console.error(`Metaphysics Error data:`, data)
+        // tslint:disable-next-line:no-empty
+      } catch (e) {}
+    }
+  }
+}
```
alloy: ✨ 👌

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -17,56 +17,45 @@ const SpinnerContainer = styled.figure`
...
+    console.error(error.message)
+  }
+
+  const networkError = error as any
+  if (networkError.response && networkError.response._bodyInit) {
+    let data = networkError.response._bodyInit
+    if (data) {
+      try {
+        data = JSON.parse(data)
+        console.error(`Metaphysics Error data:`, data)
```
damassi:
```js
`import { createLogger } from 'Utils/logger'`

const logger = createLogger('Artsy/Relay/renderWithLoadProgress')
logger.error(data) 
```

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -17,56 +17,45 @@ const SpinnerContainer = styled.figure`
...
+        // tslint:disable-next-line:no-empty
+      } catch (e) {}
```
damassi: If `JSON.parse` fails it will fail silently

jonallured:
Ok, I switched gears and took the path of least resistance and just switched to accepting a props object for that wrapper. This should give us the control I'm after but make it harder to break things.

@alloy I opened this one: https://artsyproduct.atlassian.net/browse/PLATFORM-1272 with almost no context, feel free to update!

> ...leaving these breadcrumb in place for people to be informed of them as they work in that are of the code-base could surface them at the right time for them to get prioritised.

I want to believe this too! But evidence tells us that this does not happen. This code has been here since October of last year. Again, I think with good intentions, but these TODOs are just noise. 🤷‍♂️

src/Artsy/Relay/renderWithLoadProgress.tsx
jonallured: Added a generic log message with da360a3 so at least something shows up!

src/Artsy/Relay/renderWithLoadProgress.tsx
jonallured: Wasn't planning on addressing this, but since you did all the work for me, I got this in with da360a3. ❤️ 🏆 💯

src/Artsy/Relay/renderWithLoadProgress.tsx
jonallured: This is good feedback and a solid approach, but I ended up going with the easier solution of just accepting a props object for the container. Over time we'll see if that's good enough or if we end up wanting more control.

damassi:
> This code has been here since October of last year

That means its very new code, at least according to how we've been working while also being spread fairly thin across different teams. For platform-like and / or global things it's best not to remove comments because it often just takes a while, and they have proven useful in the past.

jonallured:
"New" is in the eye of the beholder I guess, haha! 😝 

@damassi being spread thin is even more reason to get work items into Jira and out of TODOs! I put a TODO back to leave a breadcumb for our future selves, but I don't see any value in putting commented code back. That stuff is in version control and can be easily resurrected if needed.

alloy:
> I want to believe this too! But evidence tells us that this does not happen. This code has been here since October of last year. Again, I think with good intentions, but these TODOs are just noise. 🤷‍♂️

I have to, respectfully, disagree again. More than once has an inline TODO comment helped me save time, some of those written by past me, some even multiple years old. There’s a good balance to strike in what to comment about and in what detail (link to more authoritative place for tracking), but this is certainly not a black and white matter.

Thanks for the ticket! I’ve added a bit more detail 👍 

> You probably did this also @jonallured, but I verified that this approach will indeed enable us to vertically center the spinner within the autosuggest preview pane.

Hmm, so was the real need to center the spinner on both axes? If so, I think that should just be applied to the spinner defined in the `renderWithLoadProgress` module, because I think that’s a better default and would also have expected the existing code to already work like.

jonallured:
Ok, I feel heard about TODOs and recognize that people see this differently so I'm happy to move on from that point. It's ok for friends to disagree!! ❤️ 🤗 

This morning I've addressed the feedback from @damassi re: that unparseable error so that the raw error object will show up in the logs. I also updated the styles so that the spinner will default to vertical and horizontal center - that does indeed seem like a smart default.

So what else? Do you feel good about the state of this PR @alloy?

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -10,61 +11,56 @@ import styled from "styled-components"
  * from the SpinnerContainer and Spinner.
  */
 const SpinnerContainer = styled.figure`
-  width: 100%;
-  height: 100px;
+  align-items: center;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
```
alloy: Great!

alloy:
Perfect! I’m feelin’ real good 👌 Let’s do this

artsyit:
:rocket: PR was released in v14.3.5 :rocket:

src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -38,7 +38,7 @@ const handleError = error => {
       console.error(`Metaphysics Error data:`, data)
       logger.error(data)
     } catch (e) {
-      logger.error('Metaphysics Error could not be parsed.')
+      logger.error("Metaphysics Error could not be parsed.", error)
```
damassi: @jonallured - This should return the error from the `catch(e)`, not from up above, as that won't give us the correct  infor as to why the try branch failed.

jonallured: Ugh, nice catch! I will open a PR for this. 👍

alloy: Good catch, thanks @damassi !

PR #2115, opened 2019-03-06
Set static preview height & use flex to vertically center 'no results'

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-758.

1. We noticed that the way we were getting the "no results" preview vertically centered was with a whole bunch of vertical margin. It's now centered via flexbox.

2. The height of the suggestions/preview box was dependent on the height of the content, so you'd get some dancing around that was distracting & looked not so "quality worthy of art." This PR adds a static height to the preview box, eliminating the dancing, and improving the QWOA factor.

What it looked like before (keep your eye on the bottom of the suggestion/preview box!):

dancing-box-before

What it looks like now:

dancing-box-after

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -70,6 +70,12 @@ const SuggestionsWrapper = styled(Box)`
   border-right: 1px solid ${colors.grayRegular};
 `
 
+const PreviewWrapper = styled(Box)`
+  > div {
+    height: 100%;
+  }
+`
+
 const SuggestionContainer = ({ children, containerProps, preview }) => {
   return (
     <AutosuggestWrapper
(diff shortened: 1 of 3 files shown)


src/Components/Search/SearchBar.tsx
```diff
@@ -252,9 +263,6 @@ export class SearchBar extends Component<Props, State> {
     if (!focused) {
       return null
     }
-    if (xs && !query) {
```
jonallured: Should this have been removed? I seem to remember us being confused that this was in the diff.

artsyit:

Built with commit 371cb52cbc73c8dabb1fca5c0e585dd2a7146ac5

https://deploy-preview-2115--artsy-reaction.netlify.com

src/Components/Search/SearchBar.tsx
pepopowitz: No! Thank you for catching that. I'm not sure why this keeps happening, but I'll remove it.

artsyit:
:rocket: PR was released in v14.3.6 :rocket:

PR #2121, opened 2019-03-06
add tabIndex={0} to preview links to get Safari to specify e.relatedTarget on blur

pepopowitz:
Inspired by this stackoverflow answer: https://stackoverflow.com/a/42764495

### DIFF
--- a/src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
+++ b/src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
@@ -92,7 +92,7 @@ const renderItems = (
         mb={2}
         itemsPerRow={itemsPerRow}
       >
-        <Link href={href} color="black100" noUnderline>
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 51919b908d3308af7b60f6f0ee1369e1ef724884

https://deploy-preview-2121--artsy-reaction.netlify.com

mzikherman:
I guess I'm surprised this was an issue since aren't `<Link>` elements all `<a>` tags with an `href`, and so are thus 'selectable' and should provide that target? Thanks for digging!

artsyit:
Yeah, I am also surprised. It doesn't seem like this should have done anything, but it definitely seems to have convinced Safari that these links were really links.

pepopowitz:
Yeah, I am also surprised. It doesn't seem like this should have done anything, but it definitely seems to have convinced Safari that these links were really links.

artsyit:
:rocket: PR was released in v14.3.3 :rocket:

PR #2125, opened 2019-03-07
Artist insights 2.0 - Some artists with 1 badge don't have insights displayed

lilyfromseattle:
__Problem:__
Blek le rat has 1 badge and it displays, but Dustin Yellin also has one badge (reviewed by ArtForum) and his Artist Insights isn't displayed.

__Solution:__
Check to display the SelectedCareerAchievements component included a check on artist.insights.length, but there was a return inside the component's render that didn't include this check and was returning for artists with insights, so I added it there as well.

__Before:__
screen shot 2019-03-07 at 11 27 24 am

__After:__
screen shot 2019-03-07 at 11 26 16 am


sweir27:
🎉 And thanks for including screenshots!!

Could you add a little description of what the change was/why you had to make it? It's not totally clear to me! (for reference @ashkan18 has some good PR description formats for this 😄i.e. https://github.com/artsy/vortex/pull/28)

src/Components/v2/SelectedCareerAchievements.tsx
```diff
@@ -89,7 +89,10 @@ export class SelectedCareerAchievements extends React.Component<
   }
 
   render() {
-    if (!hasSections(this.props.artist)) {
+    if (
+      !hasSections(this.props.artist) &&
+      this.props.artist.insights.length === 0
```
sweir27: Can we add a test that would have caught this case?

artsyit:

Built with commit e4576e23754668de25b40ed92b5f67493e55879e

https://deploy-preview-2125--artsy-reaction.netlify.com

src/Components/v2/__tests__/SelectedCareerAchievements.test.tsx
```diff
@@ -40,4 +40,29 @@ describe("SelectedCareerAchievements", () => {
     expect(text).toContain("Reviewed by a major art publication")
     expect(text).toContain("Participated in a major biennial")
   })
+
+  it("renders selected career achievements if no auction results or partner highlights", async () => {
+    ArtistFixture.auctionResults = null
+    ArtistFixture.highlights.partners = null
+    wrapper = await getWrapper(ArtistFixture)
```
ds300: For good data hygiene we should avoid mutating the `ArtistFixture` in these tests. Although the mutations can't escape outside of the bounds of this file (because of the way jest executes test files in isolation) it could cause confusing problems if people add or modify tests here in the future.

The idiomatic and safe thing to do would be to construct a new version of the artist fixture using object spread syntax, e.g.

```ts
wrapper = await getWrapper({
   ...ArtistFixture,
   auctionResults: null,
   highlights: {
    ...ArtistFixture.highlights,
    partners: null
   }
})
```

src/Components/v2/SelectedCareerAchievements.tsx
```diff
@@ -89,7 +89,10 @@ export class SelectedCareerAchievements extends React.Component<
   }
 
   render() {
-    if (!hasSections(this.props.artist)) {
+    if (
+      (!hasSections(this.props.artist) && !this.props.artist.insights) ||
+      this.props.artist.insights.length === 0
```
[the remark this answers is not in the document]
lilyfromseattle: Had my parentheses in the wrong place my bad!

artsyit:
:rocket: PR was released in v15.2.2 :rocket:

PR #2131, opened 2019-03-08
WIP AutoSuggest: Remove unstated & replace it with a React Context/ContextConsumer wrapper

pepopowitz:
**Update:** I _think_ this PR is going to be closed without merging. We are looking at removing the "previews" from the search bar later today (3/19), and I think that will also mean removing all of this state management. 

**WIP** because I still want to a bit more testing. I don't intend on making more code changes, unless I find something while testing or you request changes.

---

This PR removes the `unstated` dependency from the src/Components/Search folder, though it is still a dependency in other areas of reaction.

We're replacing [state.ts](https://github.com/artsy/reaction/compare/master...pepopowitz:replace-unstated-with-hooks#diff-af1dfdfbba7bfa878f43674670ce1490) with [SearchBarContext.ts](https://github.com/artsy/reaction/compare/master...pepopowitz:replace-unstated-with-hooks#diff-b4c7af9e3fbffc81864f1c0821cb6b75), which is a component based on a React Context.

To inject the context into the components in our tree with minimal churn on this PR, we're replacing all `<Subscribe>` wrappers with a new `<SearchBarConsumer>` wrapper ([example](https://github.com/artsy/reaction/compare/master...pepopowitz:replace-unstated-with-hooks#diff-1ae621323dbe87da0a101e12f5c58328R88)). This injects the `SearchBarContext` into each component, and the components can interact with it in similar ways that they were interacting with `state.ts`. 

This is less ambitious than what I originally had in mind, which was to replace all state management and mounting within the src/Components/Search folder with `useState` and `useEffect`. That goal proved to be _too_ ambitious, hard to untangle, and probably not a great use of my time. Instead, I focused on simply removing the `unstated` dependency, which is one of our stated goals in the front-end practice.


artsyit:

Built with commit 3020f84d038b2d9425997a515ee4927d37ceac6a

https://deploy-preview-2131--artsy-reaction.netlify.com

src/Components/Search/AutosuggestManager.tsx
```diff
@@ -64,7 +60,7 @@ export const AutosuggestManager: React.ExoticComponent<
   React.HTMLProps<HTMLDivElement>
 > = React.forwardRef(({ children }, ref: React.RefObject<any>) => {
   return (
-    <Subscribe to={[SearchBarState]}>
+    <SearchBarConsumer>
```
damassi: With the latest changes to reaction and force we can use hooks here. So...
```js
import { SearchBarContext } from './SearchBarContext'

const { searchState } = useContext(SearchBarContext)
... 
```

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -109,7 +111,7 @@ export class MarketingCollectionsPreview extends React.Component<
       return { href: `${sd.APP_URL}/collection/${slug}` }
     })
 
-    this.props.searchState.registerItems(
+    this.props.searchState.registerPreviewItems(
```
damassi: 👍

src/Components/Search/Previews/Grids/ArtistSearch/MarketingCollections.tsx
```diff
@@ -143,7 +145,7 @@ export const MarketingCollectionsPreviewFragmentContainer = createFragmentContai
       <Responsive>
         {({ xs, sm, md }) => {
           return (
-            <Subscribe to={[SearchBarState]}>
+            <SearchBarConsumer>
```
damassi: I wonder if hooks can be used as a child? I _think_ so, but not totally sure. 

That, said, even if we couldn't do ^, we can flatten this thing out nicely between the consumer and responsive.

If we add an export [here](https://github.com/artsy/reaction/blob/master/src/Utils/Responsive/DeprecatedResponsive.tsx#L12) then it should be possible to use the responsive component like so: 

```js
import { ResponsiveComponents } from 'Utils/Responsive/DeprecatedResponsive'

const { xs, sm, md } = useContext(ResponsiveComponents) 
const { searchState } = useContext(SearchBarContext)

...
```

src/Components/Search/Previews/Grids/ArtistSearch/RelatedArtworks.tsx
```diff
@@ -97,7 +97,7 @@ export const RelatedArtworksPreviewFragmentContainer = createFragmentContainer(
       <Responsive>
         {({ xs, sm, md }) => {
           return (
-            <Subscribe to={[SearchBarState]}>
+            <SearchBarConsumer>
```
damassi: If possible (not sure how much effort it would be to move out of a Class, so not totally critical) this could use a hook. If it's a pain to refactor then all good 👍but if it seems painless then it should also apply to `Responsive`, similar to the comment above.

src/Components/Search/SearchBarContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+    hasEnteredPreviews,
+    selectedPreviewIndex,
+    incrementSelectedPreviewIndex,
+    reset,
+    enterPreview,
+    enterPreviewWithoutSelection,
+    leavePreviewIfNoSelection,
+    decrementOrLeavePreview,
+    clearPreviewItems,
+  }
```
damassi: Very nicely done 👍

src/Components/Search/SearchBarContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export const SearchBarConsumer: React.FC<{
+  children: (searchBarState: SearchBarState) => JSX.Element
+}> = ({ children }) => {
+  const searchBarState = useContext(SearchBarContext)
+
+  const rendered = children(searchBarState)
+
+  return rendered
+}
```
damassi: Seeing this part it might not be so simple to quickly swap out with with `useContext` elsewhere, so we can def iterate. This tackles the "remove unstated" need well.

src/Components/Search/SearchBarContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+  decrementOrLeavePreview: () => void
+  clearPreviewItems: () => void
+  reset: () => void
+}
+
+export const SearchBarContext = React.createContext<SearchBarState>(
+  {} as SearchBarState
+)
+
+export const SearchBarProvider: React.SFC<any> = ({ children }) => {
```
damassi: `React.FC`

damassi:
@pepopowitz - After doing some experimentation this weekend with [`easy-peasy`](https://github.com/ctrlplusb/easy-peasy) I think it could be a good formal state management lib that would replace unstated and serve our future needs really well, particularly as we would have access to Redux's entire middleware ecosystem. This has common hooks patterns baked in, would scale well outside of custom solutions, and it doesn't lock us into any particular patterns since it's just Redux under the hood -- meaning, we've got an "eject" button, should we need it.

We're using Redux in numerous other Artsy apps (including in Force and Positron) so there's nothing new there, and no new dependencies that would need to be added. 

Thoughts?  

We have a formal RFC process for new libs which I would be happy to shepard, but I'd definitely like to get you input because I think that this Search bar would be a most excellent first candidate.

pepopowitz:
First off, an update on this PR - it has stalled, as it looks like our auto-suggest experiment did not bring the outcomes we were looking for. We're holding off on work on this search bar stuff, including this PR, and I could see us pulling this code back out at some point.

Regarding `easy-peasy` - In this example, I think things would be greatly improved if we were to complete a refactor with `useContext`, `useState`, and maybe `useReducer`. I'd personally like to see what things look like using built-in hooks before moving directly to another state management library. I _do_ like the idea of a wrapper on Redux that removes boilerplate; but I also don't think this particular feature is one that warrants Redux (or easy-peasy). 

I tend to think of Redux as being useful for application-level state, and React Context as being useful for state used exclusively by a subtree of components. This is an example of the latter, and Context feels like the right approach to me.

damassi:
Thanks @pepopowitz, this is the feedback I was wondering about 👍I'll hold off on filing RFC.

pepopowitz:
Closing, as auto-suggest previews have been removed.

PR #2138, opened 2019-03-11
explicitly add redirectTo property for fb auth flow

xtina-starr:
Addresses: [GROW-1147](https://artsyproduct.atlassian.net/browse/GROW-1147)

When a user logs in with Facebook after clicking the make offer button on an artwork page, they get redirected to the homepage instead of back to the artwork page. I noticed that we weren't passing in a `redirectTo` property to the `open:auth` trigger and that it wasn't included as part of the [`authQueryData`](https://github.com/artsy/reaction/blob/fd1922d49906c2197834bdc69eebb25a62860007/src/Components/Authentication/FormSwitcher.tsx#L200-L203) that gets sent through to Facebook. I'm not able to verify this locally due to Facebook's [requirement of https for login](https://developers.facebook.com/blog/post/2018/06/08/enforce-https-facebook-login/). I think we can add the property and test it in staging.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -230,7 +230,10 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
         }
       })
     } else {
-      mediator.trigger("open:auth", { mode: "login" })
+      mediator.trigger("open:auth", {
+        mode: "login",
+        redirectTo: location.href,
+      })
     }
   }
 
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit a6514ae02cc43c4112f46ce6c10a2580bc26ccd1

https://deploy-preview-2138--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -310,7 +310,10 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
         }
       })
     } else {
-      mediator.trigger("open:auth", { mode: "login" })
+      mediator.trigger("open:auth", {
+        mode: "login",
+        redirectTo: location.href,
```
mzikherman: Does the login flow not redirect you back to where you were when you started (which can then be overridden by passing in a `redirectTo`)?

artsyit:
:rocket: PR was released in v14.3.12 :rocket:

PR #2151, opened 2019-03-12
DISCO-738: Conditionally open mobile sign up when saving artwork

jonallured:
I paired with @xtina-starr on this and after some poking around, we figured out that this `SaveButton` component's `handleSave` function could be improved to include a check for `IS_MOBILE` and use that experience when true.

We started by extracting a function called `openDesktopAuth` so that we could keep our logic in `handleSave` nice and neat. From there we investigated how to get access to `sd.IS_MOBILE` and found that it was an import away, so added sharify to the file. That set us up to add a check for that flag and open the mobile sign up by adding a `openMobileAuth` function.

The mobile signup flow encodes details in the URL, including a redirect url, so that's where we'll land after signing up/in.

### DIFF
--- a/src/Components/Artwork/Save.tsx
+++ b/src/Components/Artwork/Save.tsx
@@ -16,6 +16,9 @@ import styled from "styled-components"
 import colors from "../../Assets/Colors"
 import Icon from "../Icon"
 
+import { stringify } from "qs"
+import { data as sd } from "sharify"
+
 const SIZE = 40
 
 export interface SaveTrackingProps {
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 470ba1c8eb6a383075177c2005dd57412e4cc2f0

https://deploy-preview-2151--artsy-reaction.netlify.com

mzikherman:
Nice!

artsyit:
:rocket: PR was released in v15.2.1 :rocket:

PR #2153, opened 2019-03-12
GROW-1135: resize full-width images to 2000

xtina-starr:
Addresses: [GROW-1135](https://artsyproduct.atlassian.net/browse/GROW-1135)

This increases the resize width to be 2000 instead of 1200 to prevent blurred images.

### DIFF
--- a/src/Components/Publishing/Header/Layouts/Components/FeatureFullscreenHeader.tsx
+++ b/src/Components/Publishing/Header/Layouts/Components/FeatureFullscreenHeader.tsx
@@ -16,9 +16,7 @@ export const FeatureFullscreenHeader: React.SFC<FeatureHeaderProps> = props => {
   const url = (hero_section && hero_section.url) || ""
   const isVideo = url.includes("mp4")
   const src =
-    !isVideo &&
-    url.length &&
-    resize(url, { width: !sd.IS_MOBILE ? 2000 : 1600 })
+    !isVideo && url.length && resize(url, { width: !sd.IS_MOBILE ? 2000 : 600 })
   const hasNav = seriesArticle || super_article || is_super_article
 
   return (
--- a/src/Components/Publishing/Header/Layouts/Components/FeatureFullscreenHeader.tsx
+++ b/src/Components/Publishing/Header/Layouts/Components/FeatureFullscreenHeader.tsx
@@ -1,6 +1,7 @@
 import { space } from "@artsy/palette"
 import { PartnerInline } from "Components/Publishing/Partner/PartnerInline"
 import React from "react"
+import { data as sd } from "sharify"
 import styled from "styled-components"
 import { resize } from "../../../../../Utils/resizer"
 import { pMedia } from "../../../../Helpers"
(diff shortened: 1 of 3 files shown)


src/Components/Publishing/Sections/Image.tsx
```diff
@@ -31,7 +31,10 @@ export const Image: React.SFC<ImageProps> = props => {
     width,
   } = props
   const caption = image.caption || ""
-  const src = resize(image.url, { width: 1200, quality: GLOBAL_IMAGE_QUALITY })
+  const src = resize(image.url, {
+    width: layout === "feature" ? 2000 : 1200,
```
xtina-starr: I took a stab at this and am guessing this is where we want to make the change, but wondering if there is anywhere else I might have missed.

Also I conditionally change the resize width for articles with the feature layout because I'm assuming this gets used by other article layouts where we may not want to increase the width. Maybe you can confirm?

artsyit:

Built with commit 79c5696778d1598258951f6e9b9cc073bd07666f

https://deploy-preview-2153--artsy-reaction.netlify.com

src/Components/Publishing/Sections/Image.tsx
eessex: We should use the image section's layout for this instance rather than the article layout, to avoid using large images when the layout is not wide.  For the fullscreen images the `ImageCollection` component should have prop `sectionLayout: "fillwidth".`

eessex: Additionally it would be good to resize this conditionally based on the screen size, we should use `media` to resize to 2000 only if the screen size is larger than 1200px

eessex:
Not sure what is up with this snapshot, ran into the same issue yesterday -- posted something here: https://artsy.slack.com/archives/C2TQ4PT8R/p1552489136287400

eessex:
This change should also be applied to the fullscreen feature header: https://github.com/artsy/reaction/blob/master/src/Components/Publishing/Header/Layouts/Components/FeatureFullscreenHeader.tsx#L17

src/Components/Publishing/Sections/Image.tsx
xtina-starr: So the change is still in this file, but should be conditionally based on the sectionLayout instead of the article layout, correct?

eessex: Yep exactly

src/Components/Publishing/Header/Layouts/Components/FeatureFullscreenHeader.tsx
```diff
@@ -14,7 +15,10 @@ export const FeatureFullscreenHeader: React.SFC<FeatureHeaderProps> = props => {
   } = props
   const url = (hero_section && hero_section.url) || ""
   const isVideo = url.includes("mp4")
-  const src = !isVideo && url.length && resize(url, { width: 1600 })
+  const src =
+    !isVideo &&
+    url.length &&
+    resize(url, { width: !sd.IS_MOBILE ? 2000 : 1600 })
```
eessex: Since we are toggling for mobile we can probably bump the lower number down to something more like a phone size, potentially 600 would work.

xtina-starr: That makes sense.

eessex:
Looks good to me! Merge on green.

artsyit:
:rocket: PR was released in v15.3.2 :rocket:

PR #2159, opened 2019-03-14
updating artsy for galleries link

oxaudo:
The follow up on this one: https://github.com/artsy/force/pull/3716
to address reaction rendered footer.

Will also follow up with one more link update on force. Missed that this link also appears on Fair partner pages.


artsyit:
:rocket: PR was released in v15.3.3 :rocket:

PR #2160, opened 2019-03-14
Improve error message for failed charge in BN flow

lilyfromseattle:
At the moment, if you successfully add a credit card during the payment step but then the card is declined when you submit at the final step, the error message is the generic "Something went wrong". We have enough information from the request to be able to specify that the payment failed, so it seems like a shame not to pass that context on to the user.

__Before:__
Screen Shot 2019-03-14 at 2 39 49 PM

__After:__
Screen Shot 2019-03-14 at 2 39 54 PM


artsyit:

Built with commit 84c52ebe6b47d64e65c5fcb17fe13263001d1440

https://deploy-preview-2160--artsy-reaction.netlify.com

src/Apps/Order/Routes/__tests__/Review.test.tsx
```diff
@@ -111,6 +115,30 @@ describe("Review", () => {
       )
     })
 
+    it("shows a modal with a helpful error message if the user's card is declined", async () => {
+      window.location.assign = jest.fn()
```
ds300: Do we need this mock? (and same Q for test below?)

src/Apps/Order/Routes/__tests__/Review.test.tsx
```diff
@@ -201,6 +229,30 @@ describe("Review", () => {
       expect(window.location.assign).toBeCalledWith("/artwork/artworkId")
     })
 
+    it("shows a modal with a helpful error message if the user's card is declined", async () => {
```
ds300: Offer-mode orders will not try to charge the card when submitting, so it's OK to not have these two tests.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -196,6 +196,27 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
           )
           break
         }
+        case "charge_authorization_failed": {
+          let data = {} as any
+          if (error.data) {
+            data = JSON.parse(error.data)
```
ds300: This was probably my oversight from the `/accept` page, but I'd love to see this wrapped in a try/catch (no need to do anything in the catch block), just in case we get something other than well-formed json back from exchange.

artsyit:
:rocket: PR was released in v15.8.3 :rocket:

src/Apps/Order/Routes/__tests__/Review.test.tsx
xtina-starr: @lilyfromseattle @ds300 popping in to ask a quick question. I don't have enough context to know whether this needs to be mocked but I was wondering if the mock should be in the individual test or in the [jest tests setup](https://github.com/artsy/reaction/blob/4c6d2e125dd32a90ee780a14d2f207c847defc43/src/setup_jest.ts#L29-L35). I'm more so wondering when should it be in the individual test vs the the test setup for my own knowledge.

ds300: That seems like a philosophical question 🙂 Here's my take: It makes sense to put a mock in the global setup file if it needs to be mocked in a particular way most of the time. Good examples are libs that perform network requests like `react-tracking`. Mocking too many things might lead to surprising behaviour for people writing new tests or refactoring old tests. That kind of thing can cost people a lot of time and frustration in debugging. Explicit local mocking can also make tests easier to read than implicit global mocking.

In other words... 🤷‍♂️:D

PR #2171, opened 2019-03-15
Add generic item component for search results page

dblandin:
This commit introduces a generic component for search result items when a richer component is not needed.

**Next steps**

- Add a new fallback image provided by design
- Add `description` attribute to `SearchableItem` type from Metaphysics and use in component

**Questions**

- ~Should we translate entity types here or should that happen upstream?
  For example `PartnerShow` -> `Show`~ Putting this in the reaction component
- ~Should we add any information we might need to the SearchableItem
  type, or will we need some attributes from entity-specific endpoints?~ Using entity-specific types when available. Will need to add more support in Metaphysics.

**Screenshots**

2019-03-15-154722_954x328_scrot
2019-03-15-154758_1027x396_scrot

### DIFF
--- a/src/Apps/Search/Components/GenericSearchResultItem.tsx
+++ b/src/Apps/Search/Components/GenericSearchResultItem.tsx
@@ -1,46 +1,60 @@
-import { Flex, Image, Link, Sans, Serif, Spacer } from "@artsy/palette"
-import React, { SFC } from "react"
+import { Box, Flex, Image, Link, Sans, Serif, Spacer } from "@artsy/palette"
+import React, { FC } from "react"
 
 interface GenericSearchResultItemProps {
-  item: any
+  imageUrl: string
+  name: string
+  description?: string
   index: number
+  href: string
+  entityType: string
 }
 
-export const GenericSearchResultItem: SFC<
+export const GenericSearchResultItem: FC<
   GenericSearchResultItemProps
 > = props => {
-  const { item, index } = props
+  const { imageUrl, href, name, description, index, entityType } = props
+
+  const translateEntityType = anEntityType => {
+    switch (anEntityType) {
+      case "PartnerShow":
+        return "Show"
+        break
+      default:
+        return anEntityType
+    }
+  }
 
   return (
     <>
       <Flex flexDirection="row">
-        <div>
-          <Link href={item.href}>
-            <Image
-              width={70}
-              height={70}
-              mr={20}
-              src={
-                item.imageUrl || `https://picsum.photos/70/70/?random=${index}`
-              }
-            />
-          </Link>
-        </div>
-        <div>
+        <Box>
+          <Image
+            width={70}
+            height={70}
+            mr={20}
+            src={imageUrl || `https://picsum.photos/70/70/?random=${index}`}
+          />
+        </Box>
+        <Box>
           <Sans color="black100" size="2" weight="medium">
-            {item.searchableType}
+            {translateEntityType(entityType)}
           </Sans>
           <Spacer mb={0.5} />
-          <Serif color="black100" size="3">
-            {item.displayLabel}
-          </Serif>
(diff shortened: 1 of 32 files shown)


artsyit:

Built with commit 01a222f7ffa499f0f31790306c2543566c0475ee

https://deploy-preview-2171--artsy-reaction.netlify.com

mzikherman:
> Should we add any information we might need to the SearchableItem
type, or will we need some attributes from entity-specific endpoints?

Up to you! I'm not sure...if there's data that's available/accessible to fully fill in the falllback item we can add it to `SearchableItem`, otherwise we can just grab attributes from the entity-specific schema on each tab. That triggers the N+1 query thing (each object on a tab would be fetched separately)- but those can be straightforwardly batched, so it won't be a performance issue.

I think I would advocate for just reaching in to the entity specific ones, those should all have images/titles/descriptions at the ready! Then w/ batching we'll be good. It will likely be more straightforward than trying to get everything from just the searchable data.

> Should we translate entity types here or should that happen upstream?
For example PartnerShow -> Show

I don't follow...do you mean that the show specific tab would need a `... on Show` fragment? I might be missing something but feel free to hardcode/translate anything that's needed, either in components that need them or in MP (if it makes sense).

src/Apps/Search/Components/GenericSearchResultItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const GenericSearchResultItem: SFC<
+  GenericSearchResultItemProps
+> = props => {
+  const { item, index } = props
+
+  return (
+    <>
+      <Flex flexDirection="row">
+        <div>
```
mzikherman: Rather than `<div>`, we should try to use things from our design system. So `<Box>` maybe? (or `<>` if it doesn't even need its own container).

src/Apps/Search/Routes/Articles/SearchResultsArticles.tsx
```diff
@@ -71,7 +72,9 @@ export class SearchResultsArticlesRoute extends React.Component<
         {articles.map((article, index) => {
           return (
             <Box key={index}>
-              {article.displayLabel}
+              <GenericSearchResultItem item={article} index={index} />
+              <Spacer mb={3} />
+              <Separator />
```
mzikherman: #minor nit, we probably don't need `<Separator />` for the _last_ element to be rendered.

src/Apps/Search/Components/GenericSearchResultItem.tsx
```diff
@@ -0,0 +1,47 @@
+import { Flex, Image, Link, Sans, Serif, Spacer } from "@artsy/palette"
+import React, { SFC } from "react"
+
+interface GenericSearchResultItemProps {
+  item: any
+  index: number
+}
+
+export const GenericSearchResultItem: SFC<
```
damassi: This should be of type `React.FC` -- `SFC` has been deprecated due to react hooks.

dblandin:
> I don't follow...do you mean that the show specific tab would need a ... on Show fragment? I might be missing something but feel free to hardcode/translate anything that's needed, either in components that need them or in MP (if it makes sense).

Translating with the reaction component sounds like a reasonable next step. I was talking about the specific value of `show.entityType // PartnerShow` which we want to display as "Show".

mzikherman:
> Translating with the reaction component sounds like a reasonable next step. I was talking about the specific value of show.entityType // PartnerShow which we want to display as "Show".

Got it- yea I'll leave that up to you! As long as pick one and are consistent...

dblandin:
@mzikherman Just updated this PR! I'll have to switch over to Metaphysics to add entity-specific support for more types but this should be good to merge as an iteration.

artsyit:
:rocket: PR was released in v15.5.3 :rocket:

PR #2179, opened 2019-03-19
Refactor commitMutation usage

ds300:
Problems and solutions:

- We had some gnarly spaghetti logic because we were using the raw version of `commitMutation` which is built on callbacks, and things evolved organically over several months. I created a promise-based wrapper for `commitMutation` and did as much as I could to linearize the logic around invoking mutations and handling their results.
- Every page in the order app (except the status page) independently managed the state of whether or not a mutation is being committed. This state is used to show loading spinners and so forth. Here I abstracted that state management out into context land and made the `isCommittingMutation` state an injected prop.

This order app code was all quite well-tested and made it easy to work on this! Huge shout out to @sweir27 for keeping up that level of testing :pray:

I also manually tested all of the user flows in force and everything was looking :+1: 👌 🙆‍♂️ 🏄‍♀️

### DIFF
--- a/src/Apps/Order/Routes/Accept/index.tsx
+++ b/src/Apps/Order/Routes/Accept/index.tsx
@@ -117,7 +117,7 @@ export class Accept extends Component<AcceptProps> {
       case "insufficient_inventory": {
         await this.props.dialog.showErrorDialog({
           title: "Not available",
-          message: "Sorry, the work is no longer available",
+          message: "Sorry, the work is no longer available.",
         })
         this.routeToArtistPage()
         break
--- a/src/Apps/Order/Routes/Accept/index.tsx
+++ b/src/Apps/Order/Routes/Accept/index.tsx
@@ -11,21 +11,19 @@ import {
   OrderStepper,
 } from "../../Components/OrderStepper"
 
-import {
-  commitMutation,
-  createFragmentContainer,
-  graphql,
-  RelayProp,
-} from "react-relay"
+import { createFragmentContainer, graphql, RelayProp } from "react-relay"
 
 import { AcceptOfferMutation } from "__generated__/AcceptOfferMutation.graphql"
 import { ConditionsOfSaleDisclaimer } from "Apps/Order/Components/ConditionsOfSaleDisclaimer"
 import { ShippingSummaryItemFragmentContainer as ShippingSummaryItem } from "Apps/Order/Components/ShippingSummaryItem"
 import { TransactionDetailsSummaryItemFragmentContainer as TransactionDetailsSummaryItem } from "Apps/Order/Components/TransactionDetailsSummaryItem"
 import { Dialog, injectDialog } from "Apps/Order/Dialogs"
+import {
+  CommitMutation,
+  injectCommitMutation,
+} from "Apps/Order/Utils/commitMutation"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
 import { CountdownTimer } from "Components/v2/CountdownTimer"
-import { ErrorWithMetadata } from "Utils/errors"
 import { get } from "Utils/get"
 import createLogger from "Utils/logger"
 import { ArtworkSummaryItemFragmentContainer as ArtworkSummaryItem } from "../../Components/ArtworkSummaryItem"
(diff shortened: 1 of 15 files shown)


artsyit:

Built with commit 464c4cb40fe1f26266074d59bba1797e2b52bc84

https://deploy-preview-2179--artsy-reaction.netlify.com

src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -293,7 +254,7 @@ export class Accept extends Component<AcceptProps, AcceptState> {
 }
 
 export const AcceptFragmentContainer = createFragmentContainer(
-  injectDialog(trackPageViewWrapper(Accept)),
+  injectCommitMutation(injectDialog(trackPageViewWrapper(Accept))),
```
damassi: Maybe 
```js
import { compose } from 'lodash'

export const AcceptFragmentContainer = createFragmentContainer( 
  compose(
     injectCommitMutation, 
     injectDialog, 
     trackPageViewWrapper
  )(trackPageViewWrapper)
)
```
or some variant thereof (since this is getting kinda windy, though not sure if this creates typing headaches).

Either way just a pattern thought, not a blocker.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -435,7 +406,9 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
 }
 
 export const PaymentFragmentContainer = createFragmentContainer(
-  injectStripe(trackPageViewWrapper(injectDialog(PaymentRoute))),
+  injectCommitMutation(
+    injectStripe(trackPageViewWrapper(injectDialog(PaymentRoute)))
+  ),
```
damassi: Or something like

```js
const decorate = compose(
  injectCommitMutation, 
  injectStripe, 
  injectDialog, 
  trackPageViewWrapper
) 

export const PaymentFragmentContainer = createFragmentContainer(decorate(PaymentRoute), ...)
```

src/Apps/Order/Utils/commitMutation.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+export function injectCommitMutation<Props extends CommitMutationProps>(
+  Component: React.ComponentType<Props>
+): React.ComponentType<
+  Pick<Props, Exclude<keyof Props, keyof CommitMutationProps>>
+> {
+  return props => {
+    return (
```
damassi:
```js
import { SystemContext } from 'Artsy' 
...
const { isCommittingMutation, commitMutation } = useContext(MutationContext)
.... 
```

src/DevTools/createTestEnv.tsx
```diff
@@ -154,17 +155,24 @@ class TestEnv<MutationNames extends string, TestPage extends RootTestPage> {
...
-          headTags={this.headTags}
-        >
-          <Component
-            {...props}
-            router={{ push: this.routes.mockPushRoute }}
-            route={{ onTransition: this.routes.mockOnTransition }}
-          />
-          <ConnectedModalDialog />
-        </MockBoot>
+        <SystemContext.Consumer>
```
damassi: `useContext(SystemContext)`

src/Apps/Order/Routes/Payment/index.tsx
damassi: Ideally we could do 

```js
@injectCommitMutation
@injectStripe
@injectDialog
@trackPageViewWrapper
export class Foo extends Component { ... }
```

Which a few different devs and I have looked at in the past, but couldn't figure out the type properly, and so we tabled it. Since we already use the `@track` decorator all over, and top-level decorators can be simple HOCs, it makes sense to keep these things together if possible. But also don't want to increase complexity by introducing intermediary layers _just_ to decorate since wrapping is the same thing. 

:silverbullet:

ds300:
Thanks Chris! :pray: I'm going to leave the injections as-is. I agree annotations would be better, but like you say it's tough to type them. That's because annotations are supposed to return a class with a structure that extends the input class's structure, but doing that while modifying the props requirements is a challenge I haven't figured out yet. Should theoretically be doable though.

artsyit:
:rocket: PR was released in v15.5.10 :rocket:

src/Apps/Order/Routes/Payment/index.tsx
sweir27: whoaaaa

src/Apps/Order/Utils/commitMutation.tsx
```diff
@@ -0,0 +1,104 @@
+import { SystemContext } from "Artsy"
+import { GraphQLError } from "graphql"
+import React, { useContext } from "react"
+import {
+  commitMutation as relayCommitMutation,
+  GraphQLTaggedNode,
+} from "react-relay"
+import { Environment, OperationBase } from "relay-runtime"
+jest.unmock("react-relay")
```
mzikherman: Should this be in here? This isn't a test file.

ds300: 😱 good catch. no it should not.

PR #2180, opened 2019-03-19
Add back artwork structured data

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-850

This PR adds structured data to the artwork pages. I used [the old Force implementation](https://github.com/artsy/force/blob/bfc6156feef90d6ee625a7d70d094f198a93ab33/src/desktop/apps/artwork/helper.js) as a guide, but there were a couple things that changed:

* Dimensions in metaphysics are concatenated into one string. I added code to parse them back out, and populate the `width`, `height`, and `depth` fields of the json-ld.
* A comma-separated list of artist names is available in metaphysics, so I'm using that instead of combining them manually.
* When we have a price range instead of a single price, we emit a `priceSpecification` property with a `minPrice` and `maxPrice`. I made this decision based on [this spec doc](https://docs.google.com/document/d/1mdxdBWshmYv0XWwylrX8f87b5i1hg6xcQoKtGaFtiIw/edit#), linked from [this ticket](https://artsyproduct.atlassian.net/browse/GROW-878). 
* I added a seller image, because the Google structured data tool was complaining about not having one.

Here is a screenshot of the page run through the Google structured data test page. This artwork has a price range, so we can see the `priceSpecification` element:

image

You can see that it is still complaining about one thing - it doesn't like that we didn't specify a `price`, even though we specified a `priceSpecification`. I _think_ this is fine, but I defer to anyone who thinks this is something worth investigating.


artsyit:

Built with commit 1ff2138c41c58d7c55a9a9c50429bafaafe2a317

https://deploy-preview-2180--artsy-reaction.netlify.com

src/Apps/Artwork/Components/Seo/SeoDataForArtwork.tsx
```diff
@@ -0,0 +1,184 @@
+import { trim } from "lodash"
+import React from "react"
+
+import { SeoDataForArtwork_artwork } from "__generated__/SeoDataForArtwork_artwork.graphql"
+import { CreativeWork } from "Components/v2/Seo/CreativeWork"
+import { Product } from "Components/v2/Seo/Product"
+import { createFragmentContainer, graphql } from "react-relay"
+import { get } from "Utils/get"
+
+const { APP_URL } = process.env
```
mzikherman: This should actually come from Sharify (that's our way of sharing those configs between Force and Reaction components), vs using `process.env`.

For storybooks, we have a [custom webpack plugin](https://github.com/artsy/reaction/blob/86e9ac1793333e1f9edec4b54815bec04ae9a219/.storybook/webpack.config.js) that will load values from `process.env` _into_ sharify (so that it's consistent and we can always just use sharify in components).

src/Apps/Artwork/Components/Seo/__tests__/SeoDataForArtwork.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+        .props().data
+      expect(data).toEqual({
+        brand: {
+          "@type": "Person",
+          name: "Artist McArtist",
+        },
+        description: "artwork description",
+        image: "artwork-image",
+        name: "artwork title",
+        url: "undefined/artwork/an-artwork",
```
mzikherman: #minor you can mock sharify in this file and provide a `{ data: { APP_URL: "..." } }` (once you switch to using it for `APP_URL`), and then here you can test the full URL!

mzikherman:
Looking great! Very thorough 👌 

Since this is certainly better than what we have now (which is nada), I'm certainly in favor of merging (and seeing how things look in results in a few days, iterate, etc.).

To be fully transparent we can link this PR in #seo on Slack and ask the price question there, and/or once this is merged and on staging, maybe they can run the page thru their Google tools and come back with their thoughts on the results?

pepopowitz:
Per [a slack conversation](https://artsy.slack.com/archives/C02BDPZPC/p1553098482034400), I think we're good to go with the `price` error when there is a range. 

If anyone else has feedback, I'll wait until tomorrow to merge!

artsyit:
:rocket: PR was released in v15.7.2 :rocket:

PR #2182, opened 2019-03-19
DISCO-853: Yank autosuggest previews

jonallured:
When I started working on this I was getting console errors about an invalid entity type, so I posted on Slack and got some context from @mzikherman:

https://artsy.slack.com/archives/CAZG82X4G/p1553006313582300

Turns out over in https://github.com/artsy/metaphysics/pull/1593 we added `PartnerShow` and that's not really compatible for autosuggest.

So, my first step here was to specify the entity types and exclude that particular one.

Once those errors went away, I looked at the SearchBar file and yanked previews. Once that was green, then I searched for lingering files and nuked those as well.

Looks a little something like this:
Screen Shot 2019-03-19 at 3 30 20 PM

/cc @damassi


artsyit:

Built with commit d7f01ea0933ff341bdfd8da27e2b3cb09b3097dc

https://deploy-preview-2182--artsy-reaction.netlify.com

pepopowitz:
I _think_ the AutoSuggestManager code was written to manage keyboard navigation in the previews, and should be unnecessary now. Did you look into removing that also?

jonallured:
Ok, I changed tactics here given some feedback from @mzikherman - instead of having reaction know about the entities that are suitable for autosuggest, I'm pushing that up into MP: https://github.com/artsy/metaphysics/pull/1607.

@pepopowitz thanks for the reminder - I think I got rid of it all and I think that means https://github.com/artsy/reaction/pull/2131 is probably stale. Does that sound right to you?

pepopowitz:
Yes, that sounds right! Once this is merged, I'll close #2131.

mzikherman:
Woo! This should also make the keyboard accessibility stuff saner. I think the library handled all that perfectly out of the box (for the suggestions list itself).

I also think this still has the logic for maintaining the suggestions list when you click on a suggestion (since there's sometimes a lag as your browser navigates). So that's good!

Locally it did look a tad funky, but that might be something you're planning on cleaning up...or I may need a `rm -rf node_modules/ && yarn install --check-files` due to recent upgrades and linking and what not.

Screen Shot 2019-03-20 at 11 44 59 AM

jonallured:
@mzikherman strange! I am seeing this locally too so I'll dig into which change caused that style regression.

jonallured:
@mzikherman found the styling issue with @pepopowitz and pushed 142fa63 to reset the styles. They were formerly handled here:

https://github.com/artsy/reaction/blob/7cd9f7fc78f00af48bf842344d4ebea322e3fe84/src/Components/Search/AutosuggestManager.tsx#L16

mzikherman:
Oh yea! 😎 

Consider the previews....yanked.

artsyit:
:rocket: PR was released in v15.5.13 :rocket:

PR #2190, opened 2019-03-20
Render a gray box for search result items without an image

dblandin:
This gray box will serve as a fallback for search result items that might have an image, but don't currently. We're also discussing new fallback images or icons for entity types that don't currently support an image at all, like City.

2019-03-20-144508_1370x426_scrot

### DIFF
--- a/src/Apps/Search/Components/GenericSearchResultItem.tsx
+++ b/src/Apps/Search/Components/GenericSearchResultItem.tsx
@@ -5,7 +5,6 @@ interface GenericSearchResultItemProps {
   imageUrl: string
   name: string
   description?: string
-  index: number
   href: string
   entityType: string
 }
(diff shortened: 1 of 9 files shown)


artsyit:

Built with commit 3c88e970952da0a62fe429f32b79e868c9861f7e

https://deploy-preview-2190--artsy-reaction.netlify.com

src/Apps/Search/Components/GenericSearchResultItem.tsx
```diff
@@ -1,19 +1,26 @@
...
-  index: number
   href: string
   entityType: string
 }
 
+const ImageBox = styled(Box)`
+  height: 70px;
+  width: 70px;
+  margin-right: 20px;
+  background-color: #f8f8f8;
```
mzikherman: This should probably inline the styles and use the design system for the margin and color.

Something like, `<Box height={70} width={70} mr={2} bg="black5">...` or whatever

dblandin: Ahh, I was wondering how to set background using a palette defined color. Nice :100:

artsyit:
:rocket: PR was released in v15.6.0 :rocket:

PR #2194, opened 2019-03-21
[Router] Add queryParam support

damassi:
Updates our `queryMiddleware` to support query in-router query params. See [documentation](https://github.com/4Catalyzer/farce#querymiddleware-and-createquerymiddleware) for more info. 

Also adds a `preload` boolean prop to `PreloadLink` which, if `false`, means that we bypass upfront loading in favor of an immediate route transition, then load, then display. It's not as nice UX but it's the only way we can get around the inability to access `prepareVariables` in an async fashion. See https://graphql.slack.com/archives/C0BEXJLKG/p1553118312208200. 

- [x] Ensure legacy query params still work on /collect and /artist/id
- [x] Ensure that things work SSR-wise, with JS disabled
- [x] Ensure route transitions work

search

search

### DIFF
--- a/src/Apps/Search/Components/NavigationTabs.tsx
+++ b/src/Apps/Search/Components/NavigationTabs.tsx
@@ -30,7 +30,7 @@ export class NavigationTabs extends React.Component<Props> {
     const { exact } = options
 
     return (
-      <RouteTab to={to} exact={exact}>
+      <RouteTab to={to} exact={exact} preload={false}>
         {text}
       </RouteTab>
     )
--- a/src/Apps/Search/Routes/Artworks/SearchResultsArtworks.tsx
+++ b/src/Apps/Search/Routes/Artworks/SearchResultsArtworks.tsx
@@ -3,19 +3,37 @@ import { SearchResultsArtworks_viewer } from "__generated__/SearchResultsArtwork
 import { SearchResultsFilterFragmentContainer as ArtworkGrid } from "Apps/Search/Routes/Artworks/Components/Filter/SearchResultsFilterContainer"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { Provider } from "unstated"
+import { FilterState } from "../../FilterState"
 
 export interface Props {
   viewer: SearchResultsArtworks_viewer
+  location: {
+    query: {
+      term: string
+    }
+  }
 }
 
 export class SearchResultsArtworksRoute extends React.Component<Props> {
   render() {
-    const { viewer } = this.props
+    const {
+      viewer,
+      location: { query },
+    } = this.props
 
     return (
-      <Box>
-        <ArtworkGrid viewer={viewer} />
-      </Box>
+      <Provider
+        inject={[
+          new FilterState({
+            keyword: query.term,
+          }),
+        ]}
+      >
+        <Box>
+          <ArtworkGrid viewer={viewer} />
+        </Box>
+      </Provider>
     )
   }
 }
--- a/src/Apps/Search/routes.tsx
+++ b/src/Apps/Search/routes.tsx
@@ -1,3 +1,6 @@
+import { RouteConfig } from "found"
+import { graphql } from "react-relay"
+
 import { SearchResultsArticlesRouteRouteFragmentContainer as SearchResultsArticlesRoute } from "Apps/Search/Routes/Articles/SearchResultsArticles"
 import { SearchResultsArtistsRouteFragmentContainer as SearchResultsArtistsRoute } from "Apps/Search/Routes/Artists/SearchResultsArtists"
 import { SearchResultsArtworksRouteFragmentContainer as SearchResultsArtworksRoute } from "Apps/Search/Routes/Artworks/SearchResultsArtworks"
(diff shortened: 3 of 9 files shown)


artsyit:

Built with commit ff577add3a0f653b47573ef69dd8177738e07dfa

https://deploy-preview-2194--artsy-reaction.netlify.com

src/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -21,7 +21,7 @@ export class NavigationTabs extends React.Component<Props> {
     const { exact } = options
 
     return (
-      <RouteTab to={to} exact={exact}>
+      <RouteTab to={to} exact={exact} preload={false}>
```
damassi: If we set this to `false` it will fall back to `<Link>` (from `found`) under the hood, and skip all preload behavior. For routes that depend on `prepareVariables` this is a must.

damassi:
Here's an example of the Artist page without preloading: 

nopreload

Not as friendly.

mzikherman:
Overall I'm not that crazy about this, b/c you're changing existing UX behavior (and personally I do prefer the current UX, although I agree with you it's fairly minor and my opinion is not strong on this). The reason for changing the existing behavior....is to get around a limitation of a library, and we're only trying to use this specific library functionality (the query middleware stuff), because it's more 'idiomatic' or 'correct' to try to use it, vs the, admittedly not as pretty, manual parsing of `window.location`.

Can you provide other justification for using this library? I'm just wondering why the need to switch to it at all, vs our own implementation of `prepareVariables`, which was working w/o needing to adjust the UX. Basically, adjust the UX to account for a limitation of a library, which it's not clear _why_ we need- feels off to me.

I do acknowledge that the routing in these apps wouldn't be portable to RN most likely (cause of `window.location`, right?), but that would still feel like an over optimization. Thus far we only really share Palette stuff, and that pattern seems to be solidifying. I don't think there's any issue with writing code in Reaction that's not RN portable, personally (in case that was a concern).

damassi:
> is to get around a limitation of a library

It's not a limitation in the library, it's a limitation in my custom roll-your-own `PreloadLink` implementation. 

> Can you provide other justification for using this library?

You mean, why we need the SSR router?

mzikherman:
@damassi the query middleware. As far as i can tell, the functionality is the same as it was before (but w/o our custom query param parsing), but with some UX adjustments.

I'm fine with merging if you are, but I'm wondering what the motivation is for using it at all? I'm not aware of what the issues are with the current approach, besides wondering if there was a way to do it without some hacky query param parsing.

mzikherman:
Also, I pulled this branch down and I'm not seeing the query params being correctly handled in storybooks as you navigate among tabs. So I'm still wondering about whether we can actually remove all our hacky `prepareVariables` code. That's there precisely to be able to work in Force via linking and in storybooks.

damassi:
The queryMiddleware is used so that the router state and the history state stay in sync. Farce is history state management for Found; farce has middleware plugins similar to other libs like react-router, that deals with external communication with with history. This is important for the SSR pass (deserializing query params on the server and then routing to the correct place, for rehydration on client), as well as consistency; `window.history` calls are a side-effect, meaning they go outside of the router. It's like calling `domElement.innerHTML = "foo"` instead of using the render function in a react component. Same thing, both leading to inconsistency. 

See [history.js](https://github.com/ReactTraining/history) for a similar implementation used in react-router. 

I'll look at removing `prepareVariables`.

damassi:
@mzikherman - This is ready to. I've got the go-ahead to remove `PreloadLink` behavior from the Artist page in favor if defaults (as seen in screenshots above) but I'll do that in a separate PR, and remove PreloadLink completely. 

Note that in storybooks there's some weird double loader thing going on related  to the tab nav spinner conflicting with the refetch container spinner, but it isn't seen in force. I'll tackle that in the other PR, seems harmless ATM.

src/Apps/Search/Routes/Artworks/SearchResultsArtworks.tsx
```diff
@@ -3,19 +3,37 @@ import { SearchResultsArtworks_viewer } from "__generated__/SearchResultsArtwork
 import { SearchResultsFilterFragmentContainer as ArtworkGrid } from "Apps/Search/Routes/Artworks/Components/Filter/SearchResultsFilterContainer"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { Provider } from "unstated"
+import { FilterState } from "../../FilterState"
 
 export interface Props {
   viewer: SearchResultsArtworks_viewer
+  location: {
```
mzikherman: This can be typed as `Location` from `found` ?

src/Apps/Search/routes.tsx
```diff
@@ -7,38 +10,11 @@ import { SearchResultsCollectionsRouteFragmentContainer as SearchResultsCollecti
...
 
-  const allParams = {
-    ...paramsFromUrl,
-    ...params,
-  }
+import { SearchAppFragmentContainer as SearchApp } from "./SearchApp"
 
-  allParams.keyword = allParams.term
-  return allParams
+const prepareVariables = (_params, { location }) => {
```
mzikherman: Cool, this looks like what I would expect!

src/Apps/Search/Routes/Artworks/SearchResultsArtworks.tsx
```diff
@@ -3,19 +3,37 @@ import { SearchResultsArtworks_viewer } from "__generated__/SearchResultsArtwork
...
+    const {
+      viewer,
+      location: { query },
+    } = this.props
 
     return (
-      <Box>
-        <ArtworkGrid viewer={viewer} />
-      </Box>
+      <Provider
```
mzikherman: Cool yea with the easy handling of the term it makes sense here, vs `routes`.

src/Apps/Search/routes.tsx
```diff
@@ -7,38 +10,11 @@ import { SearchResultsCollectionsRouteFragmentContainer as SearchResultsCollecti
...
-import qs from "qs"
-import React from "react"
-import { graphql } from "react-relay"
-import { Provider } from "unstated"
-import { FilterState } from "./FilterState"
-import { SearchAppFragmentContainer as SearchApp } from "./SearchApp"
-
-// FIXME: The initial render includes `location` in props, but subsequent
-// renders (such as tabbing back to this route in your browser) will not.
-const prepareVariables = (params, props) => {
```
mzikherman: This code was uber hacky, really glad it's gone!

mzikherman:
Nice @damassi this locally works great in storybooks and in Force!! I _did_ see the double spinner on the first tab nav (maybe my system is slower than yours?), but we can tackle that separately, see how it is on staging/prod, etc.

artsyit:
:rocket: PR was released in v15.8.1 :rocket:

PR #2200, opened 2019-03-21
[Refactor] Remove PreloadLink

damassi:
Follow up from https://github.com/artsy/reaction/pull/2194

Due to the synchronous nature of our router there's no way that we could preserve the existing preload behavior while supporting rich queryParams. This was discovered during the new `/search2` work. In an effort to reduce a future maintainability burden / confusion we're gonna kill `PreloadLink` completely, and fall back to default `<Link>` behavior as seen in found. This change in UX has been ok'd by design. 

See #2194 2194 for more detailed info, including links. 

loader2


artsyit:

Built with commit ea13bbf354aedb821d4f6d2edec37e77d7c1e86b

https://deploy-preview-2200--artsy-reaction.netlify.com

src/Apps/Artist/ArtistApp.tsx
```diff
@@ -49,33 +45,17 @@ export class ArtistApp extends React.Component<ArtistAppProps> {
...
-            */}
-
-              <Subscribe to={[PreloadLinkState]}>
-                {({ state: { isLoading } }: PreloadLinkState) => {
-                  return (
-                    <LoadingArea isLoading={isLoading}>{children}</LoadingArea>
-                  )
-                }}
-              </Subscribe>
+              <Box minHeight="30vh">{children}</Box>
```
damassi: This `30vh` is arbitrary, just enough to prevent the layout from shifting too much between tab transitions.

src/Apps/Artist/ArtistApp.tsx
```diff
@@ -49,33 +45,17 @@ export class ArtistApp extends React.Component<ArtistAppProps> {
...
-                  <RecentlyViewed />
-                </Col>
-              </Row>
-            </LazyLoadComponent>
-          )}
+          <Row>
+            <Col>
+              <RecentlyViewed />
+            </Col>
+          </Row>
```
damassi: Removed lazy loading on this; we were seeing double preloaders as this is loading in. Also, it would frequently cause our layouts to jump as we scroll. Not sure if it's necessary -- user is logged in, and it's behind a query renderer.

src/Apps/Search/SearchApp.tsx
```diff
@@ -40,19 +39,15 @@ export class SearchApp extends React.Component<ArtistAppProps> {
               <span id="jumpto--searchResultTabs" />
               <NavigationTabs term={term} searchableConnection={search} />
               <Spacer mb={3} />
-              {this.props.children}
+              <Box minHeight="30vh">{this.props.children}</Box>
```
damassi: Same as the `/artist/id` page

src/Artsy/Relay/createRelaySSREnvironment.ts
```diff
@@ -106,9 +107,18 @@ export function createRelaySSREnvironment(config: Config = {}) {
...
+        // preloader updates
+        if (environment.toggleFetching) {
+          environment.toggleFetching(true)
+        }
+
         const response = await next(req)
 
         if (!checkStatus || (response.status >= 200 && response.status < 300)) {
+          if (environment.toggleFetching) {
+            environment.toggleFetching(false)
```
damassi: Toggles state on the SystemContext so components can tap into loading states of relay. Similar, but simpler than the old `PreloadLink`

src/Artsy/Router/Utils/Route.tsx
```diff
@@ -50,11 +55,15 @@ function createRender({ prerender, render, renderFetched }) {
 
     // This should only ever show when doing client-side routing.
     if (!props) {
-      return (
-        <SpinnerContainer className={LoadingClassName}>
-          <Spinner />
-        </SpinnerContainer>
-      )
+      if (showPreloader) {
```
damassi: If for whatever reason a dev doesn't want to show the default preloader on route change (in favor of some kind of custom loader via SystemContext) they can pass `showPreloader: false` into their route.

src/Artsy/SystemContext.tsx
```diff
@@ -57,12 +59,22 @@ export const ContextProvider: SFC<ContextProps<any>> = ({
...
     user: _user,
     relayEnvironment,
+    isFetchingData,
   }
 
+  // TODO: Not sure that this is the best way to assign this. We have a few
+  // different ways in which we create a relay environment, and while direct
+  // assignment is most direct it feels a bit hacky.
+  // @ts-ignore
+  relayEnvironment.toggleFetching = toggleFetching
```
damassi: This is certainly simplest, but wondering what you think @alloy?

src/Artsy/SystemContext.tsx
```diff
@@ -57,12 +59,20 @@ export const ContextProvider: SFC<ContextProps<any>> = ({
...
   const providerValues = {
     ...props,
     user: _user,
     relayEnvironment,
+    isFetchingData,
   }
 
+  // Attach a loading state toggle to the relay environment which is triggered
+  // from within the middleware when fetches begin / complete.
+  relayEnvironment.toggleFetching = toggleFetching
```
damassi: This is certainly the most direct way to get this into the middleware, but wondering what you think @alloy? We have many different ways to create a network layer across a few different pathways, but in the end everything ends up here.

src/Apps/Search/SearchApp.tsx
```diff
@@ -9,15 +9,14 @@ import {
 } from "Components/v2"
 import { Location } from "found"
 import React from "react"
-import { LazyLoadComponent } from "react-lazy-load-image-component"
 import { createFragmentContainer, graphql } from "react-relay"
 
-export interface ArtistAppProps {
+export interface SearchAppProps {
```
alloy: 🍝

src/Apps/Artist/ArtistApp.tsx
alloy: Are we still using the dep elsewhere?

damassi: It is; on other pages without tabbed navigation it's still in place.

src/Artsy/Relay/createRelaySSREnvironment.ts
```diff
@@ -40,6 +40,7 @@ interface Config {
 
 export interface RelaySSREnvironment extends Environment {
   relaySSRMiddleware: RelayClientSSR | RelayServerSSR
+  toggleFetching?: (isFetching) => void
```
alloy: The parameter is missing its type annotation. I assume it’s a boolean?

```suggestion
  toggleFetching?: (isFetching: boolean) => void
```

src/Artsy/Router/Utils/Route.tsx
alloy: Do we actually have a use-case for this right now?

src/Artsy/SystemContext.tsx
```diff
@@ -28,9 +24,15 @@ export interface SystemProps {
    * If none is provided to the `ContextProvider` then one is created, using
    * the `user` if available.
    */
-  relayEnvironment?: Environment
+  relayEnvironment?: Environment & { toggleFetching?: (isFetching) => void }
```
alloy:
```suggestion
  relayEnvironment?: Environment & { toggleFetching?: (isFetching: boolean) => void }
```

src/Artsy/SystemContext.tsx
alloy: Hmm. Yeah, I can see how this makes sense. It’s not the cleanest to extend upstream objects like this, but also not unheard of. I wonder mostly, though, how are we using this? Is that in another PR, because I can’t find any usage in this PR.

If we’re not using it right now then I’d prefer we don’t add this right now either. It’s good to know _how_ we’d easily solve it if the need arises, though 👍

damassi: Cool, that seems reasonable. Initially it was placed to wrestle a bit more control over the spinner experience but that turned out to be unnecessary because the `<Route>` component sufficiently handles it. I can remove.

src/Artsy/Router/Utils/Route.tsx
damassi: Not at the moment. Similar to the other comment, can remove until this is needed 👍

damassi:
@alloy - Updated `SystemContext` and removed references to `isFetchingData` for now.

alloy:
Great, thanks!

Sorry, I should have said that you should feel free to merge after addressing comments, now this needs a rebase… but def do merge when that’s done! 👍

artsyit:
:rocket: PR was released in v15.11.0 :rocket:

PR #2206, opened 2019-03-23
Ensure SEO parsing doesn't fail if no dimensions

zephraph:
On artworks without dimensions this code was failing. I provided a default value for `get` so it wouldn't return undefined. 

Something to look into later: Why didn't the types catch this?

PR #2207, opened 2019-03-25
Use ts-get for better type coverage of our get function 

zephraph:
There was an issue where a bug was introduced largely because the typing for `get` wasn't respecting its types. Specifically it wouldn't return a result as undefined. 

Here's the PR detailing the fix: https://github.com/artsy/reaction/pull/2206.

This replaces our implementation of `get` with `ts-get` which has overall better type coverage.


zephraph:
Updating this brought up some type issues that we weren't catching before.

src/Utils/get.ts
```diff
@@ -1,26 +1 @@
...
-      return result
-    } else {
-      throw new Error()
-    }
-    return
-  } catch (error) {
-    return valueIfFailOrUndefined
-  }
-}
+export { default as get } from "ts-get"
```
damassi: I know this is a draft, but is the plan to find and replace?

damassi:
closing in favor of https://github.com/artsy/reaction/pull/2214

PR #2214, opened 2019-03-26
[Experiment] Add incremental migration to strict null checks

zephraph:
This is an experiment in determining if it's feasible to add a mechanism for incrementally transitioning to strict null checks. 

This adds a step in our build that runs ts with `strictNullChecks` enabled if it has `// @ts-strict` at the top of the file.


alloy:
Hah, smart.

I could imagine that TS is considering module dependencies when performing its type-checking, although I only know for a fact of this when it comes to ambient declarations (i.e. libraries) with `skipLibCheck=false`.

src/Utils/getUser.ts
```diff
@@ -1,7 +1,7 @@
-export function getUser(user: User | null | undefined): User | null {
+export function getUser(user?: User): User | undefined {
   let _user = user
 
-  if (_user === undefined) {
+  if (!_user) {
```
alloy: Is this correct or did the code previously allowed for explicitly setting the value to `null`?

src/Components/v2/FollowArtistPopover/FollowArtistPopoverRow.tsx
```diff
@@ -104,7 +106,7 @@ class FollowArtistPopoverRow extends React.Component<Props, State> {
     const { artist: originalArtist } = this.props
     const { swappedArtist } = this.state
     const artist = swappedArtist || originalArtist
-    const imageUrl = get(artist, a => a.image.cropped.url)
+    const imageUrl = get(artist, a => a!.image!.cropped!.url, "")
```
damassi: Would be be able to modify the `get` function so as to avoid `!`? Not sure if its the best idea, but one nice characteristic about the get function is being able to throw whatever you want inside of it (including complex logical operations), and assuming there's a default fallback things are fine.

zephraph: Hmm... This is actually the thing [`ts-get` does](https://github.com/RIP21/ts-get/blob/master/src/ts-get/index.ts#L34). There's just [the issue](https://github.com/RIP21/ts-get#known-issues) with the return type. Maybe @alloy can poke it.

alloy: Yeah I don’t see any conditional for function properties. Nonetheless, the commented on code does include function calls and so should already work, no?

alloy:
@zephraph What’s the status on this?

zephraph:
I haven't poked it in a bit. I'll need to update w/ master.

alloy:
@zephraph What do you want to do here, is it ready to merge (after another update)?

zephraph:
Potentially. There are still a lot of fixes needed. While it seems to scope it to a single file, we get errors from all imported files of that file. So depending where you add it it spiderwebs out quite a bit. 

Still a bit better than just turning it on wholesale.

I'll put a issue in the platform backlog.

PR #2215, opened 2019-03-26
Sync metaphysics schema

ds300:
We made a couple of breaking changes to vortex's schema. Just propagating the change to here.


artsyit:
:rocket: PR was released in v15.10.4 :rocket:

PR #2222, opened 2019-03-27
Put pricing context on artwork page

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-927

Note that you need to add yourself to the 'Pricing Context' lab feature in order to see this. You also need to find an artwork that we have enough data to show pricing context for. Here's an example: https://staging.artsy.net/artwork/alex-katz-night-william-dunas-dance-3-pamela

Desktop: 

image

Mobile:

image


artsyit:

Built with commit 117f3ae73a39038bfd210f99a965db8e800a39da

https://deploy-preview-2222--artsy-reaction.netlify.com

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -141,6 +142,7 @@ export class ArtworkApp extends React.Component<Props> {
                 <ArtworkImageBrowser artwork={artwork} />
                 <ArtworkSidebar artwork={artwork} />
                 <ArtworkDetails artwork={artwork} />
+                <PricingContext artwork={artwork} />
```
sweir27: I think we'll want to add some logic here-- the lab feature check exists in metaphysics, but here we need to verify:
- Artwork is for sale
- Artwork is _not_ in an auction

Otherwise we don't want to show the graph. I guess we could also put that in metaphysics, but might make more sense here.

ds300: Let's do it in metaphysics. That way we don't end up fetching the histogram for works that it won't be shown for.

ds300:
Ah yeah we need to promote MP first! Good catch.

sweir27:
metaphysics has been promoted..... merging!

artsyit:
:rocket: PR was released in v15.11.7 :rocket:

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -230,6 +233,7 @@ export const ArtworkAppFragmentContainer = createFragmentContainer(
       ...ArtworkDetails_artwork
       ...ArtworkImageBrowser_artwork
       ...OtherWorks_artwork
+      ...PricingContext_artwork
```
mzikherman: This winds up adding a potentially expensive stitched fetch (plus bandwidth) on all requests, even those that don't have the lab feature.

Can this be guarded with `...PricingContext_artwork @include(if: $includePricingFeature)`? That variable, `includePricingFeature`, can then be injected into the context by Force, which allows you to refer to it here.

Then, we can incrementally roll this out and test it w/o adding to load for everyone else. If/when the feature is rolled out to everyone it can just be referred to normally here.

LMK if you want some help setting that up! Here's some examples: https://github.com/artsy/force/blob/203673e3ce5c0c936d51ac3c66e19cbf471e8c70/src/desktop/apps/artist/server.tsx#L36-L39 on the server and https://github.com/artsy/force/blob/203673e3ce5c0c936d51ac3c66e19cbf471e8c70/src/desktop/apps/artist/client.tsx#L16-L19

ds300: @mzikherman, good thinking!

Happily the way it's already implemented in metaphysics means that no extra fetching will be done for people who don't have the lab feature, so any perf impact should be negligible.

mzikherman: I checked it out, and it looks like MP is fetching the user and doing a lab feature check: https://github.com/artsy/metaphysics/blob/9b5a3d892ac06f4e18223c3778cc64e8da144065/src/lib/stitching/vortex/stitching.ts#L93-L98

This is still a superfluous request. Force (and thus Reaction) _already_ has access to the current user serialized from the cookie, and so you can determine whether to fetch anything at this point...no need for that MP resolver to make a redundant request.

ds300: Ah yeah, good catch! It's an extra fetch for logged-in users 🤔 I'll add a ticket to the purchase sprint. Thanks for raising this!

PR #2223, opened 2019-03-27
[WIP] Fix chrome lazy loading issue by removing overflow

zephraph:
Fixes [PLATFORM-1313](https://artsyproduct.atlassian.net/browse/PLATFORM-1313).

For reasons I don't fully comprehend, having a overflow hidden on a parent container of an absolute element causes the observer worker in chrome to intermittently not correctly detect that the image is visible. 

It's a weird issue. This fixes the lazy loading issue that was seen in prod. I'll dig through and try to find a more robust solution though.

### DIFF
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -15,7 +15,6 @@ const Placeholder = styled.div`
   background-color: ${color("black10")};
   position: relative;
   width: 100%;
-  overflow: hidden;
 `
 
 const Image = styled(BaseImage)`


src/Components/Artwork/GridItem.tsx
```diff
@@ -15,7 +15,6 @@ const Placeholder = styled.div`
   background-color: ${color("black10")};
   position: relative;
   width: 100%;
-  overflow: hidden;
```
damassi: Has this been tested across a variety of different images / pages? There have been issues in the past where images have overflowed, leading to textual content being pushed down in the brick. (I wish I had something more concrete to point to.)

artsyit:

Built with commit 21fe7d77d6b323c8de3849d59bc07ab1c2ed7643

https://deploy-preview-2223--artsy-reaction.netlify.com

oxaudo:
OOfff.... That is a strange one for sure. 

With my totally unscientific testing on Chrome - I was getting this bug very consistently on Artwork page (those grids below the fold) and very intermittently on Collect pages.

In either case resizing the window to either a tiny bit bigger or smaller would fix the issue. Definitely seems like whatever causes the image to be considered 'in view' is not triggered for certain images:(

PR #2236, opened 2019-03-29
Artist recommendations 

pepopowitz:
A collaboration with @jonallured.

Addresses https://artsyproduct.atlassian.net/browse/DISCO-870.

This PR hooks up the artist recommendation section for users who have access to a specific lab feature.

We're using a separate QueryRenderer for the ArtistRecommendations component, for now, because it was the easiest way to make Relay not query for that data unless the user has access to the recommendations feature.

Here's what the page looks like for someone with access to the feature: 

image

Not super useful or usable yet, but it emits a label that reads "Related to __artist name__" where the recommendations will go.

To prove that this won't do anything for most users, here are two more screenshots.

1. what the page looks like for someone logged in, but who doesn't have access to the feature:

image

2. what the page looks like for someone not logged in:

image

### DIFF
--- a/src/Apps/Artist/ArtistApp.tsx
+++ b/src/Apps/Artist/ArtistApp.tsx
@@ -3,18 +3,18 @@ import { ArtistApp_artist } from "__generated__/ArtistApp_artist.graphql"
 import { NavigationTabsFragmentContainer as NavigationTabs } from "Apps/Artist/Components/NavigationTabs"
 import { AppContainer } from "Apps/Components/AppContainer"
 import { HorizontalPadding } from "Apps/Components/HorizontalPadding"
+import { ContextConsumer } from "Artsy"
 import { track } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
-import React from "react"
-import { createFragmentContainer, graphql } from "react-relay"
-import { ArtistHeaderFragmentContainer as ArtistHeader } from "./Components/ArtistHeader"
-
-import { ContextConsumer } from "Artsy"
 import {
   Footer,
   RecentlyViewedQueryRenderer as RecentlyViewed,
 } from "Components/v2"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
 import { userHasLabFeature } from "Utils/getUser"
+import { ArtistHeaderFragmentContainer as ArtistHeader } from "./Components/ArtistHeader"
+import { ArtistRecommendations } from "./Components/ArtistRecommendations"
 
 export interface ArtistAppProps {
   artist: ArtistApp_artist
@@ -101,7 +101,3 @@ export const ArtistAppFragmentContainer = createFragmentContainer(ArtistApp, {
     }
   `,
 })
-
-function ArtistRecommendations() {
-  return <div>recommendations goes here </div>
-}
--- a/src/Apps/Artist/ArtistApp.tsx
+++ b/src/Apps/Artist/ArtistApp.tsx
@@ -14,7 +14,7 @@ import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { userHasLabFeature } from "Utils/getUser"
 import { ArtistHeaderFragmentContainer as ArtistHeader } from "./Components/ArtistHeader"
-import { ArtistRecommendationsFragmentContainer as ArtistRecommendations } from "./Components/ArtistRecommendations"
+import { ArtistRecommendationsQueryRenderer as ArtistRecommendations } from "./Components/ArtistRecommendations"
 
 export interface ArtistAppProps {
   artist: ArtistApp_artist
@@ -98,7 +98,6 @@ export const ArtistAppFragmentContainer = createFragmentContainer(ArtistApp, {
       id
       ...ArtistHeader_artist
       ...NavigationTabs_artist
-      ...ArtistRecommendations_artist
     }
   `,
 })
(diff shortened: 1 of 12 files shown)


src/Apps/Artist/ArtistApp.tsx
```diff
@@ -77,6 +98,7 @@ export const ArtistAppFragmentContainer = createFragmentContainer(ArtistApp, {
       id
       ...ArtistHeader_artist
       ...NavigationTabs_artist
+      ...ArtistRecommendations_artist
```
mzikherman: This winds up adding a potentially heavy data load to every request, even if its for an unauthenticated user or a user who doesn't have the lab feature. We can actually prevent against that, and only fetch this fragment if the user has the lab feature.

That concept (does this request need/have access to the recommendations part), can be handled higher up, and here can be referred to as a variable: `...ArtistRecommendations_artist @include(if: $includeRecommendations)`.

For purposes of storybooks and ease of this initial development, you can possibly hard-code that to true (or we can have it read from ENV, or something).

Then, when it comes to actually using it in Force w/ it being actually set properly, you can pass it in via context.

Basically like: https://github.com/artsy/force/blob/203673e3ce5c0c936d51ac3c66e19cbf471e8c70/src/desktop/apps/artist/server.tsx#L33-L36 for the server and https://github.com/artsy/force/blob/203673e3ce5c0c936d51ac3c66e19cbf471e8c70/src/desktop/apps/artist/client.tsx#L13-L16 in the client.

Let me know if you have questions or need help setting that up!

mzikherman: Realizing something similar was happening for the 'pricing' lab feature on the artwork page, so just made a similar comment there: https://github.com/artsy/reaction/pull/2222/files#r270593368

pepopowitz: 👍 Thank you! I'll work on this today and reach out with questions.

src/Apps/Artist/Components/ArtistRecommendations.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const ArtistRecommendationsFragmentContainer = createFragmentContainer(
+  ArtistRecommendations,
+  graphql`
+    fragment ArtistRecommendations_artist on Artist {
+      name
+    }
+  `
+)
+
+export const ArtistRecommendationsQueryRenderer: React.FC<{
```
pepopowitz: @mzikherman Let us know if this looks different than you'd have expected!

mzikherman: Yep looks great!

artsyit:

Built with commit 16719dd3ec407bbf3495d6c8e35b8ea8d0f4a084

https://deploy-preview-2236--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
```diff
@@ -0,0 +1,57 @@
...
+  ArtistRecommendations,
+  graphql`
+    fragment ArtistRecommendations_artist on Artist {
+      name
+    }
+  `
+)
+
+export const ArtistRecommendationsQueryRenderer: React.FC<{
+  artist: Overview_artist
```
mzikherman: It might be better not to pass the whole artist object here, but just `artistID`. Then the invocation is `<ArtistRecommendations artistID=...`, which might make it easier to reuse this component outside the overview tab.

Although, that might be a premature optimization since I don't believe there's any plans for that, so feel free to disregard.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
```diff
@@ -0,0 +1,57 @@
...
+}> = ({ artist }) => {
+  const { relayEnvironment } = useContext(SystemContext)
```
mzikherman: You should be able to also destructure `user` here as well (and have lab feature access). What do you think about moving the `showRecommendations && ...` logic [from here](https://github.com/artsy/reaction/pull/2236/files#diff-bfc57d8ce9d60a71675e59322d85c6f3R182) into this query renderer so that it's all just in one place (`return null` if the user doesn't have access).

It might be easier to then keep track/change that condition later.

pepopowitz: Yeah, good call! I will update that.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
pepopowitz: I did feel a little weird about pulling in an object from a different component (`Overview_artist`)...so I'll make this change.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
pepopowitz: Looking into this, I prefer it where it's at. The `<ArtistRecommendations>` component [gets wrapped in a `<Row><Col>`](https://github.com/artsy/reaction/pull/2236/files#diff-bfc57d8ce9d60a71675e59322d85c6f3R182). I'd rather the Row/Col be in the `Overview` than in the `ArtistRecommendations`, and I don't think they should be emitted unless the recommendations are going to be emitted.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
pepopowitz: Done!

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
mzikherman: Cool, makes sense!

mzikherman:
Oh yea, looking great! 😎

artsyit:
:rocket: PR was released in v15.11.21 :rocket:

PR #2237, opened 2019-04-01
Border on BNMO status page is too thick (mobile only)

lilyfromseattle:
The border between the transaction summary and this shipping summary on mobile is too thick. This is because the 'StackableBorderBox' components need to share the same parent to get the correct interstitial border thickness. Can be worked around with a media query too.

__Before:__
image-2019-02-22-14-01-20-636

__After:__
Screen Shot 2019-04-01 at 11 17 56 AM

### DIFF
--- a/src/Apps/Order/Components/ShippingSummaryItem.tsx
+++ b/src/Apps/Order/Components/ShippingSummaryItem.tsx
@@ -1,5 +1,4 @@
 import { Serif } from "@artsy/palette"
-import { media } from "@artsy/palette"
 import { ShippingSummaryItem_order } from "__generated__/ShippingSummaryItem_order.graphql"
 import {
   StepSummaryItem,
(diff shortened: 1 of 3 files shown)


src/Apps/Order/Components/ShippingSummaryItem.tsx
```diff
@@ -38,10 +40,17 @@ const ShippingSummaryItem = ({
...
+    </StyledStepSummaryItem>
   )
 }
 
+const StyledStepSummaryItem = styled(StepSummaryItem)`
+  ${media.xs`
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+  `};
+`
```
ds300: This is the right approach, but these components are used in different places and not always together, so these border changes are appearing in places they shouldn't. e.g. here's how the payment page looks with these changes:

image

We can move these `media.xs` styles up into the status page by styling the `ShippingSummaryItem` and `TransactionDetailsSummaryItem` in `Status/index.tsx`. 

e.g.

```ts
const StyledShippingSummaryItem = styled(ShippingSummaryItem)`
  ${media.xs`
    border-top-left-radius: 0;
    border-top-right-radius: 0;
  `};
`
```

artsyit:

Built with commit 1c6e135c101ec862d0a412287f619f09e51bb8a3

https://deploy-preview-2237--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v15.11.19 :rocket:

PR #2240, opened 2019-04-02
Gate pricing context

ds300:
This was a neat little solution 👌 Thanks again for the suggestion, @mzikherman

### DIFF
--- a/src/Apps/Artwork/routes.tsx
+++ b/src/Apps/Artwork/routes.tsx
@@ -1,9 +1,9 @@
 import { graphql } from "react-relay"
-import sd from "sharify"
 import { ArtworkAppFragmentContainer as ArtworkApp } from "./ArtworkApp"
 
 // @ts-ignore
 import { ComponentClass, StatelessComponent } from "react"
+import { get } from "Utils/get"
 
 // TODO: Investigate better error boundaries for runtime errors
 
@@ -21,11 +21,12 @@ export const routes = [
         }
       }
     `,
-    prepareVariables: params => ({
+    prepareVariables: (params, { context }) => ({
       ...params,
-      // @ts-ignore
-      enablePricingContext: sd.CURRENT_USER.lab_features.includes(
-        "Pricing Context"
+      enablePricingContext: get(
+        context,
+        ctx => ctx.user.lab_features.includes("Pricing Context"),
+        false
       ),
     }),
   },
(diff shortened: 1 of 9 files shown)


artsyit:

Built with commit 36d07a381710d7ab7e0493b7f803ad935a6dc792

https://deploy-preview-2240--artsy-reaction.netlify.com

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -67,6 +67,7 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
             routeConfig: createRouteConfig(routes),
             resolver,
             render,
+            matchContext: { user },
```
mzikherman: So dope! @pepopowitz and I were trying to do this exact thing yesterday for a module on the artist page, and were fumbling trying to get the user/lab feature accessible to a route/query. Even with explicitly passing in context from Force we couldn't figure out how to access it inside `prepareVariables`.

So we decided to just make that new component its own query renderer as well, which was sort of a workaround since this exact setup was our initial thought! Although, having that be its own query renderer _might_ actually be a good thing, since it's way below the fold and will likely be an intensive fetch on its own.

src/Apps/Artwork/routes.tsx
```diff
@@ -11,11 +12,22 @@ export const routes = [
...
         artwork(id: $artworkID) {
           ...ArtworkApp_artwork
         }
       }
     `,
+    prepareVariables: (params, { context }) => ({
+      ...params,
+      enablePricingContext: get(
+        context,
+        ctx => ctx.user.lab_features.includes("Pricing Context"),
```
sweir27: This ends up being so clean!

Is it possible that `user.lab_features` will return `nil` here? Or `ctx.user`?

src/Artsy/Relay/createRelaySSREnvironment.ts
```diff
@@ -79,6 +79,7 @@ export function createRelaySSREnvironment(config: Config = {}) {
       // or further refactoring to extract `addMiddlewareToEnvironment(environment)`,
       // to be used in the SearchBar QueryRenderer (for example).
       next => req => {
+        console.log({ req })
```
ds300: oops

sweir27: (extra)

ds300: haha just beat you to it by 44 seconds 

image

Also just fixed it

artsyit:
:rocket: PR was released in v15.11.13 :rocket:

PR #2245, opened 2019-04-02
Force search result items of City type to use FallbackIcon

dblandin:
The image returned from Metaphysics isn't desired by the design comps so always render the FallbackIcon for City entity types.

We could also make a change upstream but that would affect both old and new search result pages. We can follow up with that upstream change to Metaphysics once this new page fully rolls out.


artsyit:

Built with commit 6850d88d13d608bb67b6b5b302134d422397e829

https://deploy-preview-2245--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v15.11.16 :rocket:

PR #2246, opened 2019-04-02
Remove the SystemContext around an Artwork metadata because it wasn't being used

orta:
So, when diving through the code we spotted that the `user` aspect of the `ContextConsumer` wasn't being used - this was likely removed in a refactor but not cleaned up. Removing this kinda drastically simplified this component. 

Tested it out in Storybooks and I could still save/unsave and artwork, so I don't think it's related to that ( and I'd be surprised if `ContextConsumer` had side-effects. But open to this being a bad refactor.


artsyit:

Built with commit 69fe3050927856d6741e2f25e01660c051e074b6

https://deploy-preview-2246--artsy-reaction.netlify.com

orta:
Maybe I'm just confused, I found another one of these:

https://github.com/artsy/reaction-force/blob/e23fa05d5db19bec8b6b37aabc3dae2f59dd14a8/src/Components/Artwork/Details.tsx#L189-L191

Should this also get removed?

damassi:
@orta, just now seeing this - LGTM 👍 merge when ready

artsyit:
:rocket: PR was released in v16.5.7 :rocket:

PR #2247, opened 2019-04-02
DISCO-886: Search bar QA

jonallured:
This is a QA PR so here's the ticket:

https://artsyproduct.atlassian.net/browse/DISCO-886

And here's my visual proof:

Screen Shot 2019-04-02 at 3 52 30 PM

The only code thing that might be worth considering is the way I split up the rendering of the first suggestion from the rest. I ended up adding the `totalCount` to the first suggestion item and then split up the `SuggestionItem` into a `FirstSuggestionItem` so that I could more easily render these things differently.

That leave us with some duplication in `src/Components/Search/Suggestions/SuggestionItem.tsx`, but it's the sort of duplication that offends me least - open to other ideas though!

Finally, I did not address adding a comma to the total count - is there a MP helper for this? /cc @damassi

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -357,7 +357,7 @@ export const SearchBarRefetchContainer = createRefetchContainer(
           term: { type: "String!", defaultValue: "" }
           hasTerm: { type: "Boolean!", defaultValue: false }
         ) {
-        search(query: $term, mode: AUTOSUGGEST, first: 5)
+        search(query: $term, mode: AUTOSUGGEST, first: 7)
           @include(if: $hasTerm) {
           edges {
             node {
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -8,6 +8,7 @@ import colors from "Assets/Colors"
 import Input from "Components/Input"
 import {
   EmptySuggestion,
+  FirstSuggestionItem,
   PLACEHOLDER,
   PLACEHOLDER_XS,
   SuggestionItem,
@@ -272,10 +273,29 @@ export class SearchBar extends Component<Props, State> {
     return displayLabel
   }
 
-  renderSuggestion = (
-    { node: { displayLabel, displayType, href } },
-    { query, isHighlighted }
-  ) => {
+  renderSuggestion = (edge, rest) => {
+    const renderer = edge.node.isFirstItem
+      ? this.renderFirstSuggestion
+      : this.renderDefaultSuggestion
+    const item = renderer(edge, rest)
+    return item
+  }
+
+  renderFirstSuggestion = (edge, { query, isHighlighted }) => {
+    const { displayLabel, displayType, href } = edge.node
+    return (
+      <FirstSuggestionItem
+        display={displayLabel}
+        href={href}
+        isHighlighted={isHighlighted}
+        label={displayType}
+        query={query}
+      />
+    )
+  }
+
+  renderDefaultSuggestion = (edge, { query, isHighlighted }) => {
+    const { displayLabel, displayType, href } = edge.node
     return (
       <SuggestionItem
         display={displayLabel}
@@ -304,6 +324,7 @@ export class SearchBar extends Component<Props, State> {
 
     const firstSuggestionPlaceholder = {
       node: {
+        isFirstItem: true,
         displayType: "FirstItem",
         displayLabel: term,
         href: `/search?q=${term}`,
--- a/src/Components/Search/Suggestions/SuggestionItem.tsx
+++ b/src/Components/Search/Suggestions/SuggestionItem.tsx
@@ -30,11 +30,6 @@ export const SuggestionItem: SFC<Props> = props => {
           >
             <Suggestion {...props} />
           </InnerWrapper>
-          {isHighlighted && (
-            <Flex flexGrow="0" px={2}>
-              <HighlightIcon />
-            </Flex>
-          )}
         </SuggestionWrapper>
       </Link>
     </Box>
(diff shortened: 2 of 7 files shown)


artsyit:

Built with commit 0c6b96c99fc58363ad5b194dea7941fd74ba9504

https://deploy-preview-2247--artsy-reaction.netlify.com

src/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -12,29 +12,57 @@ interface Props {
...
+          <InnerWrapper
+            flexDirection="column"
+            flexGrow="1"
+            justifyContent="center"
+          >
+            See full results for "{query}"
+          </InnerWrapper>
+          <Flex flexGrow="0" px={2}>
+            <Serif size="2" color="black30">
+              {totalCount} results
```
mzikherman: Try `totalCount.toLocaleString()` for the comma.

jonallured: TY!! ❤️

src/Components/Search/SearchBar.tsx
```diff
@@ -357,8 +381,9 @@ export const SearchBarRefetchContainer = createRefetchContainer(
           term: { type: "String!", defaultValue: "" }
           hasTerm: { type: "Boolean!", defaultValue: false }
         ) {
-        search(query: $term, mode: AUTOSUGGEST, first: 5)
+        search(query: $term, mode: AUTOSUGGEST, first: 7)
           @include(if: $hasTerm) {
+          totalCount
```
mzikherman: Technically this isn't the correct count to use. This returns the total count for an auto suggest style query, whereas the search results page does keyword search, and so the actual count you'll see when you click on this item and go to the search results page is likely to be different.

You can check out how we're getting the count for search results here: https://github.com/artsy/reaction/blob/0aba9cc92bc99464bf1e4209301bd78eba7f39be/src/Apps/Search/SearchApp.tsx#L35-L45

Not sure how much of a blocker it is to get this number to exactly match what you'd see on search results!

jonallured: Great callout @mzikherman, thanks! This requirement came from @cavvia, so I'm tagging him here and would love some direction about what was envisioned here. I just reached for the closest thing, but yeah, would probably be best to have this number match the number they see when they go to the results page!

mzikherman: I'm assuming that was the intent and it's probably safe to go ahead and implement that. I was just calling out why `totalCount` as-is wouldn't get you there.

jonallured: @mzikherman reading through this again this morning with fresh eyes, I think this is what we're discussing here:

The total count being used in this PR is wrong - what I've got here is the total number of suggestions. Were the user to follow the first item to the results page, the totals wouldn't match. In order for these to line up, I'd have to adjust my query like so:

```graphql
{
  suggestionSearch:search(query: $term, mode: AUTOSUGGEST, first: 7) {
    ...
  }
  totalCountSearch:search(query: $term, first: 1, aggregations: [TYPE]) {
    ...
  }
  filter_artworks(keyword: $term, size: 0, aggregations: [TOTAL]) {
    ...
  }
}
```

The `suggestionSearch` gives us the results we need to draw the suggestions, the `totalCountSearch` gives us the aggregations for non-artwork results and the `filter_artworks` gives us the results for artworks. I could then use the code snippet you quoted above to compute the same total from the results page.

Did I get this right? Is this what we want to do? Are there performance implications here that I'm not seeing? This seems like an expensive operation to send to MP, but maybe I'm missing something!

mzikherman: That is definitely correct, and that is definitely expensive!!

It's up to design (cc @cavvia ) to decide if the performance hit (both to the user and to ES which is going to see 3 simultaneous queries for every new autosuggest fetch), if it's worth it or not. I don't think performance is _necessarily_ a blocker (this is only on interactions with search, and not on page load)...but it's worth bringing up.

I was wondering how much of a blocker it is to get this number to exactly match what you'd see on the search results page. Because if it is...it's 2 additional ES calls _just_ for that count (on top of the autosuggest results).

cavvia: Yes, this is expensive - we should only consider this if we can guarantee that fetching the counts will not block the rendering of the auto-suggest results in any way, those need to load as quickly as possible... it's crucial that this interaction be very snappy.

jonallured: Ok, just talked this through with @jeffreytr and we brainstormed some ways to move forward here:

1. drop the count from the first item in suggestions
2. use the total count from suggestions even though it could potentially not
   match results page
3. use the expensive operation and hope for the best in terms of performance
4. delay the rendering of the count so that suggestions render first, then we
   fetch the count

We agreed that options 2 and 3 are not viable. That left us with a choice between dropping the count altogether for this AB test or exploring another approach of splitting up the queries to not block the rendering of the suggestions.

In the end, we've opted for dropping the count and creating a ticket in Jira to explore option 4 in a follow up. We felt that holding up this AB test anymore is not worth the minor UX improvement the count gets us, but that as we iterate on the auto suggest, this is an important topic to return to. That ticket is here:

https://artsyproduct.atlassian.net/browse/DISCO-896

So that means I'll get a commit in that removes the count and then I think this would be ready to merge.

jonallured:
Ok, I rebased out the total count part of this and it's 🍏good to go - so calling all final thoughts before I merge, thanks!

jonallured:
Merging!

artsyit:
:rocket: PR was released in v15.11.23 :rocket:

PR #2248, opened 2019-04-02
Adds the bid count to Artist/Artwork page Artwork grids

orta:
This makes the number of bids show up on artwork grids.

Screen Shot 2019-04-02 at 4 36 14 PM

Things to debate:
 - [x] Do we need the paddle?
   - see design comment below -- no we don't
 - [x] Should it be bold?
   - see design comment below -- should not be bold


artsyit:

Built with commit 478cc7a295db7ac8c05475d42a3364235fe55feb

https://deploy-preview-2248--artsy-reaction.netlify.com

src/Components/Artwork/__tests__/Details.test.tsx
```diff
@@ -101,20 +119,25 @@ const artworkInAuction = {
...
   },
   sale_artwork: {
-    highest_bid: { display: "$2,600", __id: null },
+    highest_bid: { display: "$2,600" },
     opening_bid: { display: "$2,400" },
-    __id: "U2FsZUFydHdvcms6Z2VyaGFyZC1yaWNodGVyLXR1bGlwcy1wMTctMTQ=",
+    counts: { bidder_positions: 0 },
   },
-  __id: "QXJ0d29yazpnZXJoYXJkLXJpY2h0ZXItdHVsaXBzLXAxNy0xNA==",
 }
```
orta: These `__id`s are probably be added by Relay at runtime, but aren't represented in the type system. Is this gonna break stuff ( /cc @ds300 ) ?

src/Components/Artwork/__tests__/Details.test.tsx
```diff
@@ -27,44 +28,36 @@ describe("Details", () => {
...
     it("shows sale 'timely' data", async () => {
-      const data = cloneDeep(artworkInAuction)
-      const wrapper = await getWrapper(data)
-      const html = wrapper.html()
-      expect(html).toContain("ends in 14d")
-    })
-
-    it("shows sale 'timely' data when auction is in a preview state", async () => {
-      const data = cloneDeep(artworkInAuction)
-      data.sale.is_open = false
```
orta: `is_open` wasn't used in the component, so I removed it from the query, and that makes these two tests the exact same code. So we nuked it.

erikdstock: 🔫

williardx:
Thanks for the quick turnaround @orta and @dleve123!

Before we merge I think it's worth quickly consulting with design on the two questions in the PR + whether or not we want the bid count to be in a separate set of parentheses or the same set as the sale open/close time.

dleve123:
@williardx 

> Before we merge I think it's worth quickly consulting with design on the two questions in the PR + whether or not we want the bid count to be in a separate set of parentheses or the same set as the sale open/close time.

Let me know what you think the best process if here for sharing with design here (or if you want to handle those comms) -- happy either way.

src/Components/Artwork/__tests__/Details.test.tsx
```diff
@@ -101,20 +119,25 @@ const artworkInAuction = {
...
 }
+
+// So, we make changes to the fixtured data in tests
+// you'll need to use a version that's Any'd so that
+// tsc passes. Trade-off of any in the test vs guarantee
+// of the data in the fixture.
+//
+// Maybe we could use something like a
+// Muatable<Partial<Details_artwork>>
+// instead?
```
alloy: Nope, that would only work if that fragment has no fragments of child containers spread in. That’s what the first item in this issue is about https://github.com/facebook/relay/issues/2597

src/Components/Artwork/__tests__/Details.test.tsx
alloy: They are added by Relay at compile (everything is done at compile time). _However_, unless you explicitly asked for them in your fragment, they won’t show up in the types emitted for your fragment. Nonetheless, they should be included.

ds300: If you omit them, random uuids will be generated in their place. [Here, in the case of these tests](https://github.com/apollographql/graphql-tools/blob/b85137ba81c9bdc5667bced9cad9b1d226ea83fc/src/mock.ts#L57). Or [here, in the case of newer tests using the `mockData` option of `renderRelayTree`](https://github.com/artsy/reaction/blob/b7533dfcfe0b28df4ec8de0d472e5e0e4d44e0f5/src/DevTools/createMockNetworkLayer/index.ts#L94-L111)

That is normally fine, but might break tests which rely on disparate bits of mock data belonging to the same graphql node (they'll be stored in different places in the cache, potentially producing inconsistent views). I think as long as your tests still pass then deleting the `__id`s is OK, but might as well keep them just in case?

src/Components/Artwork/__tests__/Details.test.tsx
orta: Thank you sire

williardx:
@orta @dleve123 I spoke with @jeffreytr and he's going to take a quick look into how to resolve the old/new styles for the bid counts on artwork bricks.

jeffreytr:
Hey folks, just took a spin on this and this is what I came up with:
Screen Shot 2019-04-04 at 12 11 04 PM

Changes being made:
-Time status added to the "Bid" label in parenthesis ...same format as current, just appended on the end
-Bid count is in parenthesis next to price (if applicable).. Show nothing if no bids. Note the price is in medium weight and the bid count is in regular weight, both at black100

Zeplin specs for the artwork brick variations can be found here: zpl.io/aBBlL5e


Any questions, feel free to ask :)

orta:
OK, this is ready for review

Screen Shot 2019-04-15 at 5 18 20 PM
Screen Shot 2019-04-15 at 5 18 06 PM
Screen Shot 2019-04-15 at 5 17 05 PM
Screen Shot 2019-04-15 at 5 13 49 PM

artsyit:
:rocket: PR was released in v15.15.0 :rocket:

PR #2255, opened 2019-04-03
Artist recommendation carousel

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-869.

This PR builds out the carousel for a recommended artist.

Here's what it looks like:

image

### DIFF
--- a/src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
@@ -5,6 +5,7 @@ import { SystemContext } from "Artsy"
 import { renderWithLoadProgress } from "Artsy/Relay/renderWithLoadProgress"
 import React, { useContext } from "react"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import { RecommendedArtistFragmentContainer as RecommendedArtist } from "./RecommendedArtist"
 
 interface ArtistRecommendationsProps {
   artist: ArtistRecommendations_artist
@@ -14,14 +15,19 @@ export const ArtistRecommendations: React.FC<
   ArtistRecommendationsProps
 > = props => {
   const {
-    artist: { name },
+    artist: { name, related },
   } = props
+
+  const relatedArtists = related.artists.edges.map(edge => (
+    <RecommendedArtist artist={edge.node} key={edge.node.__id} />
+  ))
 
   return (
     <div>
       <Serif size="8" color="black100">
         Related to {name}
       </Serif>
+      {relatedArtists}
     </div>
   )
 }
@@ -31,6 +37,16 @@ export const ArtistRecommendationsFragmentContainer = createFragmentContainer(
   graphql`
     fragment ArtistRecommendations_artist on Artist {
       name
+      related {
+        artists(first: 1) {
+          edges {
+            node {
+              __id
+              ...RecommendedArtist_artist
+            }
+          }
+        }
+      }
     }
   `
 )
--- a/src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
@@ -5,22 +5,20 @@ import { SystemContext } from "Artsy"
 import { renderWithLoadProgress } from "Artsy/Relay/renderWithLoadProgress"
 import React, { useContext } from "react"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import { get } from "Utils/get"
 import { RecommendedArtistFragmentContainer as RecommendedArtist } from "./RecommendedArtist"
 
 interface ArtistRecommendationsProps {
   artist: ArtistRecommendations_artist
 }
 
-export const ArtistRecommendations: React.FC<
-  ArtistRecommendationsProps
-> = props => {
-  const {
-    artist: { name, related },
-  } = props
-
-  const relatedArtists = related.artists.edges.map(edge => (
-    <RecommendedArtist artist={edge.node} key={edge.node.__id} />
-  ))
+export const ArtistRecommendations: React.FC<ArtistRecommendationsProps> = ({
+  artist,
+}) => {
+  const { name } = artist
+  const relatedArtists = get(artist, a => a.related.artists.edges, []).map(
+    edge => <RecommendedArtist artist={edge.node} key={edge.node.__id} />
+  )
 
   return (
     <div>
--- a/src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
@@ -19,9 +19,6 @@ const HEIGHT = 150
 
 // TODO: Remaining tasks
 /*
- * Move handleOpenAuth/openMobileAuth/openDesktopAuth to shared location
- * Buy now/make offer labels
- * Possibly extract <followArtistButton> to a shared location (it was a lot of copypasta that just worked)
  * Tests
  */
 
--- a/src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
@@ -1,5 +1,6 @@
 import { EntityHeader, Sans, Spacer } from "@artsy/palette"
 import { RecommendedArtist_artist } from "__generated__/RecommendedArtist_artist.graphql"
+import { track } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
 import { SystemContext } from "Artsy/SystemContext"
 import { FillwidthItem } from "Components/Artwork/FillwidthItem"
@@ -18,7 +19,7 @@ const HEIGHT = 150
 
 // TODO: Remaining tasks
 /*
- * Handle tracking (& find out constants to use)
+ * find out tracking constants to use
  * Move handleOpenAuth/openMobileAuth/openDesktopAuth to shared location
  * Buy now/make offer labels
  * Possibly extract <followArtistButton> to a shared location (it was a lot of copypasta that just worked)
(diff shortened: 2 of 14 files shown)


src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -0,0 +1,188 @@
...
+      action: "follow",
+      objectId: artist.id,
+    },
+  })
+}
+
+@track({
+  context_module: Schema.ContextModule.RecommendedArtists,
+})
+class RecommendedArtistWithTracking extends React.Component<
```
pepopowitz: Any thoughts on this "pattern"? I pulled the tracking stuff out of the `RecommendedArtist` component below (https://github.com/artsy/reaction/pull/2255/files#diff-b7e8e47a9fac963f0852e544e7e80e51R92), because the `@track` calls were the only thing that required a class. 

This allowed me to keep the `RecommendedArtist` component functional, and to use `useContext` instead of `ContextConsumer`/render props. It's kind of a "container & presentational" breakdown, I guess.

(And maybe there's a way to apply the `@track` calls to a function, instead of a class with methods, but it didn't look like it to me. @damassi what do you think?)

artsyit:

Built with commit 201dd953cca2f06ae73a3cef077eaacae8285987

https://deploy-preview-2255--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -0,0 +1,188 @@
...
+      <RecommendedArtist
+        {...this.props}
+        onArtworkClicked={this.trackArtworkClicked.bind(this)}
+      />
+    )
+  }
+}
+
+const RecommendedArtist: FC<
+  RecommendedArtistProps & { onArtworkClicked: () => void }
```
pepopowitz: And then this "extension" of RecommendedArtistProps is a result of me splitting the component into a slim class for tracking + a function that does most of the stuff. I can be convinced that I should create a new "type" for the props here, but this was quick, easy, and made sense.

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
mzikherman: Looks cool to me! I don't have a strong opinion, I possibly would have just gone with a class overall. Personally I pretty much always go w/ a React class over FC/SFC component, since a lot of times I wind up wanting to add a bit of local state. That's been much more useful (to me), vs having more functional components. Tests wind up being more integration-y (or snapshots) anyway, and I haven't needed to do any fancy functional....stuff.

So while I probably would have just had everything in one React class, splitting a piece of it out as you've done is 👍 from me nonetheless!

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -0,0 +1,188 @@
...
+interface RecommendedArtistProps {
+  artist: RecommendedArtist_artist
+}
+const HEIGHT = 150
+
+// TODO: Remaining tasks
+/*
+ * find out tracking constants to use
+ * Move handleOpenAuth/openMobileAuth/openDesktopAuth to shared location
+ * Buy now/make offer labels
```
mzikherman: It looks like the 'badges' have only been added to `GridItem`: https://github.com/artsy/reaction/blob/686719aa1c8594a8a6baa73312e7c13c77b2c7ca/src/Components/Artwork/GridItem.tsx#L183 , so perhaps the work item is to extract that out into its own component and include that in `FillwidthItem` as well?

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -0,0 +1,188 @@
...
+        }
+      }
+      artworks_connection(first: 20) {
+        edges {
+          node {
+            __id
+            image {
+              aspect_ratio
+            }
+            ...FillwidthItem_artwork @relay(mask: false)
```
mzikherman: Was this needed? At first glance this _seems_ correct but everything should work w/o the mask directive.

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -0,0 +1,188 @@
+import { EntityHeader, Sans, Spacer } from "@artsy/palette"
+import { RecommendedArtist_artist } from "__generated__/RecommendedArtist_artist.graphql"
+import { track } from "Artsy/Analytics"
+import * as Schema from "Artsy/Analytics/Schema"
+import { SystemContext } from "Artsy/SystemContext"
+import { FillwidthItem } from "Components/Artwork/FillwidthItem"
```
mzikherman: Oh yea, since you're doing the `...FillwidthItem_artwork` thing below, this has to be a fragment container, but you're importing the React component. If you change this to `import FillwidthItem` (it's the [default export](https://github.com/artsy/reaction/blob/686719aa1c8594a8a6baa73312e7c13c77b2c7ca/src/Components/Artwork/FillwidthItem.tsx#L133) for some reason) , then you won't need the `@relay(mask: false)` anymore and it should all work.

mzikherman:
This looks 👍 to me, merge whenever you feel ready!

Since this is behind a lab feature, also feel free to iterate over multiple PR's and not nail everything down in one if you feel that this winds up starting to creep a bit.

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
pepopowitz: Now that we've introduced hooks to Reaction, my plan is to use them whenever possible moving forward. I'm avoiding classes as much as I can.

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
pepopowitz: I cargo-culted the FillwidthItem stuff from the RecentlyViewed carousel - I'll investigate whether I need it or not!

pepopowitz:
Here is a screenshot of the metaphysics calls being made for one artist. Does any of this look surprising @mzikherman? It doesn't feel unreasonable to me, but I want to verify with others. (Note that I don't have the buy now/make offer badges hooked up yet, and I'm not sure if/how those will affect this.)

image

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
pepopowitz: Thanks for the tip! This worked.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
```diff
@@ -14,14 +15,19 @@ export const ArtistRecommendations: React.FC<
   ArtistRecommendationsProps
 > = props => {
   const {
-    artist: { name },
+    artist: { name, related },
   } = props
 
+  const relatedArtists = related.artists.edges.map(edge => (
```
mzikherman: I'm thinking about how this component handles when there are no related artists available, and realizing you'll see a `Related to ...` header (and corresponding section), but then it'll be empty.

So I think technically you'll want a fragment one level up, in the `Overview` that fetches the count of related artists.

Then, you can do `const showRecommendations = userHasLabFeature && hasRelatedArtists`, and so forth from there. Then in this component, you don't have to worry about if there are no related artists coming back, since it'll only be rendered when there are, and you won't have an 'empty' section being rendered.

pepopowitz: Yeah, I definitely assumed there would be at least one related artist. I'll work on this.

pepopowitz: I pushed some changes, including tests for the Overview component. Let me know what you think!

mzikherman: Very nice! 👍

pepopowitz:
The only red on this is from a timed-out netlify deployment; I'm going to merge.

artsyit:
:rocket: PR was released in v15.12.4 :rocket:

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
damassi: @alloy - thoughts on a `useTracking` hook? It would be nice to move away from classes completely.

src/Apps/Artist/Routes/Overview/__tests__/Index.fixture.tsx
```diff
@@ -0,0 +1,67 @@
...
+    partners: {
+      edges: [],
+    },
+  },
+  insights: [],
+  " $fragmentRefs": null,
+  " $refType": null,
+}
+
+export const artistWithRelatedArtists: OverviewRouteArtist = {
```
damassi: Fixtures have conventionally all [lived together here](https://github.c/om/artsy/reaction/tree/master/src/Apps/__tests__/Fixtures)

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
alloy: https://artsyproduct.atlassian.net/browse/PLATFORM-1216

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -0,0 +1,179 @@
...
+class RecommendedArtistWithTracking extends React.Component<
+  RecommendedArtistProps
+> {
+  @track({
+    type: Schema.Type.Thumbnail,
+    action_type: Schema.ActionType.Click,
+  })
+  trackArtworkClicked() {
+    // noop
+  }
```
alloy: Methods shouldn’t be called “track” something. Even when we define no-ops right now, the fact that we attach tracking hooks to a lifecycle callback should remain an implementation detail. In this case the method should probably just be `handleArtworkClick`.

PR #2261, opened 2019-04-05
User can see top three recommended artists

pepopowitz:
Addresses

This is a small change to return 3 recommended artists instead of 1. 

Here's what it looks like now! (It's hard to get all three artists in one screenshot, sorry...)

image

Here are the metaphysics calls for the "ArtistRecommendations" component now: 

image


mzikherman:
That was easy!

jonallured:
Nice one!! 🎉

artsyit:
:rocket: PR was released in v15.13.0 :rocket:

PR #2270, opened 2019-04-08
[Toolchain] Fix CI deployment output

damassi:
Noticed that builds were failing on Netflify, but wasn't sure why due to output being truncated by extraneous progress bar output. 

This updates our Webpack build config to storybooks' [full control mode](https://storybook.js.org/docs/configurations/custom-webpack-config/#full-control-mode) and away from the deprecated 'extend mode'. In full control mode we can silence the default progress bar.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -40,6 +40,7 @@ const {
 } = process.env
 
 const isCI = CI || NETLIFY
+console.log("--------------------------- isCI", isCI, CI, NETLIFY)
 const notOnCI = value => (isCI ? [] : [value])
 
 /**
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -3,13 +3,11 @@
 const env = require("dotenv")
 const path = require("path")
 const sharify = require("./sharify")
+const webpack = require("webpack")
 
 const ForkTsCheckerNotifierWebpackPlugin = require("fork-ts-checker-notifier-webpack-plugin")
 const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin")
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 16f925d30aecf3c328d880cc78b025061b8aaf97

https://deploy-preview-2270--artsy-reaction.netlify.com

src/Artsy/index.tsx
```diff
@@ -1,7 +1,2 @@
 export * from "./Analytics"
-export {
-  ContextProvider,
-  ContextConsumer,
-  SystemContext,
-  ContextProps,
-} from "./SystemContext"
+export * from "./SystemContext"
```
damassi: This _may_ also fix the build issues on Netlify. For some reason we were getting some odd circular-ref like issues when compiling that are fixed when we export named exports in batch. (This will also fix the warning that appears in the console when booting Storybooks.)

.storybook/webpack.config.js
```diff
@@ -96,69 +94,66 @@ if (USER_ID && USER_ACCESS_TOKEN) {
...
-        include: [/src/],
-        exclude: [/node_modules/, new RegExp(package.jest.testRegex)],
-        use: [
-          {
-            loader: "cache-loader",
-            options: {
-              cacheDirectory: path.join(cacheDirectory),
-            },
+  }
+  config.plugins = plugins
```
eessex: I notice that the below `rules` array uses push, are we sure that `plugins` does not use this as well?

damassi: We're replacing it outright in this case, but that's a good catch.

artsyit:
:rocket: PR was released in v15.13.8 :rocket:

PR #2282, opened 2019-04-09
Fixes OtherAuction component copy

ashfurrow:
This is hard to do because the storybook won't load: https://artsy-reaction.netlify.com/?path=/story/apps-artwork-page-components-otherauctions--other-auctions The is caused by a Relay container containing a non-Relay container, which itself contains Relay container child components. The non-Relay component breaks context that Relay injects into child components.

This PR is a WIP because Relay isn't fetching the sales defined in the new fragment – I haven't had a chance to look too deeply yet. Maybe there's some limitation here that caused us to avoid making `OtherAuctions` a fragment container to being with? @damassi? 

Fixes [DISCO-857](https://artsyproduct.atlassian.net/browse/DISCO-857) /cc @jonallured


ashfurrow:
I've opened a metaphysics PR that blocks this one. I still want to figure out why Relay isn't working as I expect, but we can continue with the formatting changes until then.

ashfurrow:
Okay, this is mostly done! It's blocked on that MP PR, and I want some tests for this component. I've added a new story for showing 40 auctions – staging gets polluted with outdated auctions and so having a bunch lets us see them all and verify they work. Here's a screenshot:

Screen Shot 2019-04-10 at 3 57 41 PM

(Some are missing cover images, that's fine.)

artsyit:

Built with commit 6b11a336046f53418e757103c2f580dbe110d894

https://deploy-preview-2282--artsy-reaction.netlify.com

src/Apps/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextAuction.tsx
```diff
@@ -98,6 +98,8 @@ export const ArtworkContextAuctionFragmentContainer = createFragmentContainer<{
         ...RelatedWorksArtworkGrid_artwork
       }
       sales(size: 4, sort: TIMELY_AT_NAME_ASC) {
+        ...OtherAuctions_sales
+        # Relay isn't transitively expanding the child fragments for some reason, so we'll expand manually.
         ...AuctionCard_sale
```
ashfurrow: I've read [the docs](https://facebook.github.io/relay/docs/en/graphql-in-relay.html) again but still can't figure out why this isn't working. I'm guessing it might be due to limitations on how complex the root query can be, but this same pattern is used elsewhere in Reaction without any problems. (This is also causing the `as any` casts in the above props.)

Any ideas, @alloy?

mzikherman: The problem is that `<OtherAuctions sales={sales as any} />` lines above isn't importing the right component.

If you look at the import: https://github.com/artsy/reaction/pull/2282/files#diff-e8d6463d830fb8f6a30d2867e8da582cL9 , that's importing the vanilla React component , whereas you should be importing the Relay fragment container. That _should_ clear up the issues (`any` and essentially needing to do `@relay(mask: false)` in your fragments).

src/Apps/Artwork/Components/__stories__/OtherAuctions.story.tsx
```diff
@@ -1,46 +1,45 @@
...
-  "Other Auctions",
-  () => {
+  return (
+    <QueryRenderer<OtherAuctionsQuery>
+      environment={relayEnvironment}
+      query={graphql`
+        query OtherAuctionsQuery($size: Int!) {
+          sales(size: $size, sort: TIMELY_AT_NAME_ASC) {
+            ...OtherAuctions_sales
+            # Relay isn't transitively expanding the child fragments for some reason, so we'll expand manually.
```
mzikherman: Here, because you are importing and using the fragment container (and not the React component), for `OtherAuctions`, you shouldn't need the `AuctionCard_sale` or the `as any` below.

src/Components/v2/AuctionCard.tsx
```diff
@@ -16,32 +16,47 @@ import { createFragmentContainer, graphql } from "react-relay"
...
+  const time = moment(timeIn, "YYYY-MM-DD").tz("America/New_York")
+  const now = moment()
+  if (time.diff(now, "days") >= 1) {
+    return `${time.diff(now, "days")}d`
+  } else if (time.diff(now, "hours") >= 1) {
+    return `${time.diff(now, "hours")}h`
+  } else if (time.diff(now, "minutes") >= 1) {
+    return `${time.diff(now, "minutes")}m`
+  }
+  return `${time.diff(now, "seconds")}s`
```
alloy: This looks like the type of time related code we’d want in MP, no?

src/Apps/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextAuction.tsx
ashfurrow: Haha, I feel like every six months or so, I make this mistake 😅 Thanks Matt!

src/Components/v2/AuctionCard.tsx
ashfurrow: Yeah, I agree in principle, but there are complications. Consider the difference between a specific time, and an interval: clients that have a countdown timer, for example, wouldn't be able to use an interval specified in MP (because, as time passes, the interval changes). So I think this kind of real-time, constantly-changing value doesn't make sense to model on MP. /cc @kierangillen

alloy: Ah gotcha 👍

What about that hardcoded NYC TZ? Why is that in there?

ashfurrow: I included it to match the previous behaviour, but I'll take another look to see. We don't actually _format_ the time in this function, so it can probably be removed.

kierangillen: We are hardcoding many timestamps into NYC TZ currently in Force as well. This had been the status quo. I'm interested in reevaluating all of these fields when we take inventory.

damassi:
@ashfurrow - All of the artwork grids below the fold are loaded client-side only (there can be a lot of different grids at once, depending on the artwork), so we kept things inside of `QueryRenderer`s for performance reasons. But it looks like things are ok now and that it was simply a wrong import issue?

ashfurrow:
@damassi yup, that’s right. Thanks for the context!

ashfurrow:
Okay! This is ready for review 👍

mzikherman:
Possibly a real ❌

ashfurrow:
Yeah, I needed to add explicit timezones to the tests; I also made sure to be explicit about what "now" means in the test context, to avoid intermittent test failures (the difference between `1s` and `0s` due to test execution timing)

artsyit:
:rocket: PR was released in v15.14.8 :rocket:

PR #2286, opened 2019-04-10
Update reaction to route users to the correct place

lilyfromseattle:
Creates the collect url and passes an onClick to the palette component that opens it on a new tab


artsyit:

Built with commit 5fc9b1af79213e16cae448f2e02cb26d2206f947

https://deploy-preview-2286--artsy-reaction.netlify.com

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,46 @@
+import { data as sd } from "sharify"
+
+export const createCollectUrl = (
```
ds300: How about refactoring this to use `"qs"`? It's already in our bundle, and it's a bit more maintainable than string concatenation :)

Here's an example usage: https://github.com/ds300/reaction/blob/f51af697e0871022d8754e558b82826e10b5a50b/src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromPartner.tsx#L65

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,46 @@
...
+  minCents,
+  maxCents,
+  category,
+  widthCm,
+  heightCm,
+  artistId
+) => {
+  const categoryHref = "/collect/" + category.toLowerCase()
+  const acquirableHref =
+    "?page=1&sort=-decayed_merch&acquireable=true&offerable=true&inquireable_only=true"
```
ds300: Why do we need `inquirable_only=true`? Is `acquirable=true` too broad?

src/Apps/Artwork/Utils/__tests__/createCollectUrl.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+      "Painting",
+      76.2,
+      101.6,
+      "evelyn-walg"
+    )
+
+    expect(result).toEqual(
+      "test-url/collect/painting?page=1&sort=-decayed_merch&acquireable=true&offerable=true&inquireable_only=true&price_range=4400-4455&height=28-*&width=28-*&artist_id=evelyn-walg"
+    )
+  })
```
ds300: How about some test cases for the different size options here?

src/Apps/Artwork/Utils/createCollectUrl.tsx
lilyfromseattle: Wow I had no idea that was a thing! Will do!

src/Apps/Artwork/Utils/createCollectUrl.tsx
lilyfromseattle: I'm not sure exactly. I was going off of the example url in the jira ticket

src/Apps/Artwork/Utils/__tests__/createCollectUrl.test.tsx
lilyfromseattle: Yes, good idea.

src/Apps/Artwork/Utils/createCollectUrl.tsx
sweir27: `acquireable=true` will return all artworks that have "Buy now" enabled. `inquireable_only=true` will return all artworks that are inquireable (but not "Buy now"-able).

With `offerable`, `acquireable`, and `inquireable_only`, we'll capture these three checkboxes, which is what we want (since it maps to what we pull in vortex).

image

ds300: Ahhh I see. My brain interpreted the filters as doing a set intersection rather than a set union. 👍

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,68 @@
...
+  "Drawing, Collage or other Work on Paper": "drawing",
+  "Fashion Design and Wearable Art": "wearable art",
+  Installation: "installation",
+  Jewelry: "jewelry",
+  "Mixed Media": "",
+  Other: "",
+  Painting: "painting",
+  "Performance Art": "performance-art",
+  Photography: "photography",
+  Posters: "",
```
sweir27: This is another place where we may want to iterate as we see this in-the-wild! I'm not sure if this page will be useful if we can't slice by certain categories.

We can talk about ways to fix in the future, though. 👍

Nice work @ds300 and @lilyfromseattle !!

artsyit:
:rocket: PR was released in v16.0.1 :rocket:

PR #2287, opened 2019-04-10
Show 0 instead of null for lowest bucket

lilyfromseattle:
__Before:__
Screen Shot 2019-04-10 at 1 19 48 PM

__After:__
Screen Shot 2019-04-10 at 1 23 07 PM

### DIFF
--- a/src/Apps/Artwork/Components/PricingContext.tsx
+++ b/src/Apps/Artwork/Components/PricingContext.tsx
@@ -19,6 +19,7 @@ function PricingContext({ artwork }: PricingContextProps) {
     return null
   }
 
+  // TODO: Investigate why metaphysics is returning null instead of zero for minPrice
   return (
     <BorderBox mb={2} flexDirection="column">
       <Sans size="2" weight="medium">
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit b4043d4295f9a2a7784340d68f27d0710cf45f18

https://deploy-preview-2287--artsy-reaction.netlify.com

src/Apps/Artwork/Components/PricingContext.tsx
```diff
@@ -32,14 +32,19 @@ function PricingContext({ artwork }: PricingContextProps) {
       </Flex>
       <Spacer mb={[2, 3]} />
       <BarChart
-        minLabel={artwork.pricingContext.bins[0].minPrice}
+        minLabel={
+          artwork.pricingContext.bins[0].minPrice != null
```
sweir27: It looks like from vortex, we're returning a `0` correctly:
image

From metaphysics, we're returning a `null`:
image

So it seems in metaphysics we should be converting that `0` cents value into `$0` but aren't doing so. I remembered encountering this issue before: https://github.com/artsy/metaphysics/pull/1325 but I didn't end up revisiting like I said I would, and now it's biting us again!

Could you add a `TODO` here so we can revisit? I'll add an item to our sprint to revisit the metaphysics work, but that can be a separate work-item.

lilyfromseattle: Will do

artsyit:
:rocket: PR was released in v15.14.4 :rocket:

PR #2288, opened 2019-04-10
Adjustments to bin labels and bar height for bins with 1 or 0 works

lilyfromseattle:
If an artwork falls into a bucket that has no works in it, the placeholder bar should be the same height as other bars that signify "1 work"

If bin has 0 works but is highlighted, on hover it should say "1 work"

__Before:__
Screen Shot 2019-04-10 at 2 07 23 PM

Screen Shot 2019-04-10 at 2 07 28 PM

__After:__

Screen Shot 2019-04-10 at 2 40 09 PM

Screen Shot 2019-04-10 at 2 49 33 PM


artsyit:

Built with commit 5526e74c5138809e0741a89aff7871a0c571bd52

https://deploy-preview-2288--artsy-reaction.netlify.com

src/Apps/Artwork/Components/PricingContext.tsx
```diff
@@ -43,11 +43,23 @@ function PricingContext({ artwork }: PricingContextProps) {
             const artworkFallsInThisBin =
               artwork.priceCents.min >= bin.minPriceCents &&
               artwork.priceCents.min < bin.maxPriceCents
+            let labelDescription
+            let binValue = bin.numArtworks
+
+            if (bin.numArtworks > 1) {
+              labelDescription = bin.numArtworks + " works"
+            } else if (artworkFallsInThisBin && bin.numArtworks === 0) {
+              labelDescription = "1 work"
```
sweir27: Mind adding some tests for these new cases?

Also, I think we can make this a little simpler:
```js
const binValue = artworkFallsInThisBin && bin.numArtworks === 0 ? 1 : bin.numArtworks
const labelSuffix = bin.numArtworks > 1 ? " works" : " work"

...

description: binValue + labelSuffix
```

alloy: @sweir27 I probably lack context, but shouldn't that be this?

```js
const labelSuffix = bin.numArtworks === 1 ? " work" : " works"
```

ds300: @alloy that makes sense, although it wouldn't make a difference in this case since in the 0 case, the bar will have 0px height and therefore there is nowhere for you to hover to see the '0 works' label.

alloy: Gotcha, thanks for the context 👍 I’d probably still suggest that change, for in case the design changes in the future.

src/Apps/Artwork/Components/PricingContext.tsx
```diff
@@ -43,16 +43,23 @@ function PricingContext({ artwork }: PricingContextProps) {
             const artworkFallsInThisBin =
               artwork.priceCents.min >= bin.minPriceCents &&
               artwork.priceCents.min < bin.maxPriceCents
+
+            const binValue =
+              artworkFallsInThisBin && bin.numArtworks === 0
```
sweir27: Cool! Mind adding a few test cases for this? Then it should be good to go 👍

src/Apps/Artwork/Components/__tests__/PricingContext.test.tsx
```diff
@@ -118,12 +118,36 @@ describe("PricingContext", () => {
...
+    const highlightedBar = wrapper.find(".Bar__BarBox-vx0va3-0").at(2)
+
+    highlightedBar.simulate("mouseenter")
+    expect(wrapper.text()).not.toContain("0 works")
+    expect(wrapper.text()).toContain("1 work")
+  })
+
+  it("displays 'work' singular not 'works' plural in label when there is only 1 artwork in a bin", async () => {
+    const wrapper = await getWrapper()
+    const secondBar = wrapper.find(".Bar__BarBox-vx0va3-0").at(1)
```
ds300: This selector could be brittle since that little hash at the end is a content hash of the BarBox's css properties, which are just an implementation detail and subject to change.

Luckily `wrapper.find()` is quite flexible. In addition to regular CSS selectors you can use a a react component, or even a react component's `displayName` .  I think `wrapper.find("Bar")` will work in this case because `Bar` is a functional component and therefore the function's `.name` property will be used as the `displayName`.

src/Apps/Artwork/Components/__tests__/PricingContext.test.tsx
```diff
@@ -118,12 +118,36 @@ describe("PricingContext", () => {
...
   })
 
-  it("displays $0 as the minimum price label is the minimum price is null", async () => {
+  it("displays $0 as the minimum price label if the minimum price is null", async () => {
     const wrapper = await getWrapper()
     expect(wrapper.text()).not.toContain("null")
     expect(wrapper.text()).toContain("$0")
   })
 
+  it("displays $0 as the minimum price label if the minimum price is null", async () => {
```
ds300: Is this test the same as the one above?

artsyit:
:rocket: PR was released in v16.0.1 :rocket:

PR #2294, opened 2019-04-12
Preserve search term in SearchBar, based on window.location.search

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-925.

* I'm checking against window.location for the current query string, instead of using the `found` router, because this component is stitched in via Force. As far as I can tell, it is not a child of the router, so it doesn't have access to its location. I don't feel great about this approach, but I'm hopeful it is only temporary until @damassi gets the NavBar rewritten.
* I've added what I think is adequate test coverage to the function that parses the term from the query string...let me know if you think any cases are missing.
* I left the `getSearchTerm` function as part of `SearchBar.tsx` because I didn't want to glorify it with its own file. This function feels like a sub-optimal way to identify the current route, and hiding it in this file feels like the least I can do to discourage people using it.

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -14,6 +14,7 @@ import {
   SuggestionItem,
 } from "Components/Search/Suggestions/SuggestionItem"
 import { throttle } from "lodash"
+import qs from "qs"
 import React, { Component, useContext } from "react"
 import Autosuggest from "react-autosuggest"
 import {
@@ -90,7 +91,7 @@ export class SearchBar extends Component<Props, State> {
   private userClickedOnDescendant: boolean
 
   state = {
-    term: "",
+    term: getSearchTerm(window.location),
     entityID: null,
     entityType: null,
     focused: false,
@@ -427,4 +428,16 @@ export const SearchBarQueryRenderer: React.FC = () => {
       }}
     />
   )
+}
+
+export function getSearchTerm(location: Location): string {
+  const term = get(
+    qs,
+    querystring => querystring.parse(location.search.slice(1)).term,
+    ""
+  )
+  if (Array.isArray(term)) {
+    return term[0]
+  }
+  return term
 }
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 32ee55e1bc60e47790b8cf04405225767dba849d

https://deploy-preview-2294--artsy-reaction.netlify.com

src/Components/Search/__tests__/SearchBar.test.tsx
```diff
@@ -112,3 +115,57 @@ describe("SearchBar", () => {
...
+  it("returns the term if there is a single one", () => {
+    const location = buildLocationWithQueryString("?term=monkey")
+
+    const result = getSearchTerm(location)
+
+    expect(result).toEqual("monkey")
+  })
+
+  it("returns the first term if there are multiple", () => {
+    // This is not really a concern in real-life, but TypeScript says
```
mzikherman: Ah cool, was wondering with the array check but this makes sense, someone could do `?term[]=...&term[]=...` and it might get parsed as an array.

mzikherman:
Sweet, nice and simple!

artsyit:
:rocket: PR was released in v15.14.5 :rocket:

src/Components/Search/SearchBar.tsx
```diff
@@ -90,7 +91,7 @@ export class SearchBar extends Component<Props, State> {
   private userClickedOnDescendant: boolean
 
   state = {
-    term: "",
+    term: getSearchTerm(window.location),
```
damassi: Are there any SSR / UI implications to this approach? Meaning: 
- Will it error out (`window is undefined`)
- If there's a `q=` in the url bar, will there be a jump when the client-side mounts?

src/Components/Search/SearchBar.tsx
```diff
@@ -428,3 +429,15 @@ export const SearchBarQueryRenderer: React.FC = () => {
...
+
+export function getSearchTerm(location: Location): string {
+  const term = get(
+    qs,
+    querystring => querystring.parse(location.search.slice(1)).term,
+    ""
+  )
+  if (Array.isArray(term)) {
+    return term[0]
+  }
```
damassi: This logic could also be moved right inside of the `get` function. It's useful for any old return value :)

damassi:
Oh lol, this has already been merged -- just got a Jasper update due to deleted branch.

PR #2295, opened 2019-04-15
When displaying a histogram for a work that displays a price range, highlight the bin representing the middle of the range

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-1003

Here's an artwork I rigged up to show this case, so you can test local vs staging https://staging.artsy.net/artwork/alex-katz-diana-6

Local: 
image

Staging:
image


artsyit:

Built with commit d6d9c3ae3aa8308dadfe4e982cc16e8a5f291f8e

https://deploy-preview-2295--artsy-reaction.netlify.com

src/Apps/Artwork/Components/PricingContext.tsx
```diff
@@ -19,6 +19,10 @@ function PricingContext({ artwork }: PricingContextProps) {
     return null
   }
 
+  const priceCents = artwork.priceCents.max
```
sweir27: Do we have to worry about `artwork.priceCents.min` being `null` here as well?

ds300: Nope :) If there's no 'min' price then there's no price whatsoever and we don't return histogram data from metaphysics.

artsyit:
:rocket: PR was released in v15.14.7 :rocket:

PR #2303, opened 2019-04-17
Unify logic to require authentication for following an artist, following a partner, and saving an artwork

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-901.

While working on artist recommendations (#2261, #2255, and #2236), I copy/pasted some code that forces a user to authenticate when they want to follow an artist. I noticed that this code was also copy/pasted several other places throughout Reaction. This PR moves all of that duplication into one location, so we don't have to copy/paste it anymore.

Some notes:

* The most significant change is the addition of [the `openAuthModal` utility](https://github.com/artsy/reaction/compare/master...pepopowitz:DISCO-901-extract-open-auth-modal#diff-b7bdcf43b23b670bb9781092c518e1f9). This is where all the different calls got consolidated to. 

* I did not write any unit tests to guide this refactoring, because it seemed unlikely to be fun in enzyme. I _did_ write a handful of Cypress e2e tests to guide it, though. They aren't sturdy enough to commit anywhere, and I'm not really sure we have a place for them. If you're interested in seeing them, let me know, and I'll push out my repo somewhere.

* The large amount of relay churn is because [Save.tsx was referencing `artwork.title`](https://github.com/artsy/reaction/compare/master...pepopowitz:DISCO-901-extract-open-auth-modal#diff-aa8af9f4ec65394f68ce324ec459a6a6L151), but not actually including it in the fragment. I added it to the fragment, and about ~50 other fragments regenerated to include it.

* Another side-effect of Save.tsx not including `artwork.title` in its fragment is that we were never actually sending the title through in the mobile page redirect! With these changes, we now are sending the title through as the `entityName` property.

* My goal was to not change _any_ of the values being passed through to authentication, and I almost met that goal. There was one property that I changed intentionally - I'll comment on it inline to bring awareness.


artsyit:

Built with commit 0aec6779dd28426ae744ba283562c3cc1228f834

https://deploy-preview-2303--artsy-reaction.netlify.com

src/Utils/openAuthModal.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+function getMobileIntentToFollowPartner({
+  contextModule,
+  entity,
+}: AuthReason): MobileIntent {
+  return {
+    action: "follow",
+    contextModule,
+    intent: "follow partner",
```
pepopowitz: This `intent` (and the `signUpIntent` on line 104) [_used_ to be "follow gallery"](https://github.com/artsy/reaction/pull/2303/files#diff-070e554f536ad96b2b31ad0e4544b513L68), not "follow partner". 

I changed it to "follow partner", because [this code in Force](https://github.com/artsy/force/blob/74fb00704a7c33b0a7b02c5bf328159990b65a9b/src/desktop/apps/auth2/routes.ts#L62-L63) indicates that it is expecting "follow partner". 

A side-effect of this change is that the mobile auth screen now says "Sign in to follow partners" instead of "Sign in to Artsy" when trying to follow a partner.

src/Utils/openAuthModal.ts
```diff
@@ -0,0 +1,192 @@
...
+
+function getDesktopIntentToFollowPartner({
+  entity,
+}: AuthReason): DesktopIntent {
+  return {
+    mode: "signup",
+    copy: `Sign up to follow ${entity.name}`,
+    signupIntent: "follow partner",
+    afterSignUpAction: {
+      kind: "profile", // should this be "partner"?
```
pepopowitz: I wasn't confident enough to change this line, but I _suspect_ it should be `kind: "partner"` instead of `kind: "profile"`. I don't know what effect this change would have, so I didn't make it.

src/Utils/openAuthModal.ts
```diff
@@ -0,0 +1,51 @@
...
+    window.location.href = "/login"
+  }
+}
+
+const openMobileAuth = ({ entity, contextModule }) => {
+  const params = stringify({
+    action: "follow",
+    contextModule,
+    intent: "follow artist",
+    kind: "artist",
```
xtina-starr: Right now this seem to only take in account artist entity types. Is that intentional? I ask because the PR title includes following a partner and saving an artwork but that wouldn't be compatible with the hardcoded values.

xtina-starr: Ahh this is an older commit sorry.

artsyit:
:rocket: PR was released in v16.3.2 :rocket:

PR #2309, opened 2019-04-18
Updated copy for pricing context

lilyfromseattle:
__Title copy:__ “Price in context” (currently says “Price”)

__Example:__ https://cl.ly/ca1dfa681cc9

__FAQ modal copy:__

__title:__ Price in context

__body:__
This feature aims to provide insight into the range of prices for an artist's works and allow buyers to discover other available works by the artist at different price points.

The graph displays current and past list prices for works that are similar in size and category to the work you're viewing. The prices included in the graph are only from galleries and dealers on Artsy.

Artwork prices are affected by a variety of objective and subjective factors including the artist's relative position in the art market and the artwork's size, condition, rarity, and subject matter. These factors are unique to every artwork. As such, this feature is not intended to provide pricing guidance for the artwork being viewed.


artsyit:

Built with commit 03925ecfeacacaa9ddc6a991821e0e518edd5178

https://deploy-preview-2309--artsy-reaction.netlify.com

src/Apps/Artwork/Components/PricingContextModal.tsx
```diff
@@ -36,12 +36,14 @@ export class PricingContextModal extends React.Component<State> {
   }
 
   render() {
+    const linkHref =
+      "https://www.artsy.net/article/artsy-editorial-artworks-prices"
```
ashkan18: should we make this a configuration? or even maybe a prop passed to this component? 

Artsy article urls are technically editable, i _think_ we would still support old link but maybe having it as a prop and have the consumer set it may make more sense. @ds300 what do you think?

ds300: I think this is as good a place as any to specify it. The only reason to parameterize it would be if we decide to use the Modal in more than one place or have it point to different URLs in different contexts.

One suggestion I would make would be to remove the protocol and the host from the url so it works on staging and local.

ds300:
@lilyfromseattle just one test to fix and this will be good to go!

artsyit:
:rocket: PR was released in v16.5.8 :rocket:

PR #2310, opened 2019-04-18
Add ability to “show more” artist recommendations

pepopowitz:
*Update*: I've removed the WIP. I'm ready for this one to go!

Addresses https://artsyproduct.atlassian.net/browse/DISCO-872

This PR adds a "Show More" button at the bottom of the artist recommendations. When you click it, it loads three more artists!

This is what it looks like!

Kapture 2019-04-18 at 17 41 30

### DOM Weight

@kierangillen did not have problems with lag, and I no longer have problems. I might have been in desperate need of a restart, or something.

~I'm worried about the weight of all these DOM elements. When viewing in storybooks, it seems a bit laggy when I load more artists; I haven't looked at it in force yet. We should keep our eye on this once it gets to staging (it's behind a feature flag), and figure out if we need to think about virtualizing the scroll or some other approach.~

### Relay errors in VS Code

Per Eloy's [comment below](https://github.com/artsy/reaction/pull/2310#issuecomment-484760418), these are ignorable errors coming from the Apollo VS Code extension.

~Running `yarn relay` passes, but there are a couple red squigglies in VS Code that I'm not sure what to do about. Once I switched from a fragment container to a pagination container, VS Code started complaining that I wasn't accounting for the `$showFollowSuggestions` variable.~

image

~I'm not sure why I need to account for this. It is part of a fragment container that is included in my query/query renderer via a few hops ([`ArtistRecommendations_artist`](https://github.com/artsy/reaction/compare/master...pepopowitz:DISCO-872-show-more-recommendations#diff-851a427551b166f833a018066feed83cR114) -> [`RecommendedArtist_artist`](https://github.com/pepopowitz/reaction/blob/295700d3759b6a60d58b8a12fa34b2aa32ce33d3/src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx#L178) -> [`FollowArtistButton_artist`](https://github.com/artsy/reaction/blob/295700d3759b6a60d58b8a12fa34b2aa32ce33d3/src/Components/FollowButton/FollowArtistButton.tsx#L195))...but there is a `defaultValue` defined for the `showFollowSuggestions` argument, so I'm not sure why this pagination container needs to account for it. I am mostly cargo-culting this pagination container stuff, as I've never worked with it before, so there's also a good chance I messed something up. @damassi @mzikherman @alloy maybe you have some thoughts on this?~


artsyit:

Built with commit 295700d3759b6a60d58b8a12fa34b2aa32ce33d3

https://deploy-preview-2310--artsy-reaction.netlify.com

alloy:
These are not Relay errors, those would show in your terminal where you run the relay-compiler. What I _think_ is going on is that you have the Apollo vscode extension enabled which we know to complain about variables as it doesn’t take into account that with Relay you can have ‘local’ fragment variables defined as ‘fragment arguments’. I.e. this is a false positive that should be ignored. I have a ticket on the platform backlog about fixing the extension, but haven’t gotten around to it yet.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
```diff
@@ -1,53 +1,155 @@
...
+
+export const ArtistRecommendationsPaginationContainer = createPaginationContainer(
   ArtistRecommendations,
   {
+    // TODO: can we remove endCursor????
     artist: graphql`
-      fragment ArtistRecommendations_artist on Artist {
+      fragment ArtistRecommendations_artist on Artist
+        @argumentDefinitions(
+          count: { type: "Int", defaultValue: 3 }
```
mzikherman: #minor Since we explicitly use `first`/`after`, it might be worth it to just use that everywhere and not use the more generic `count` and `cursor` names

mzikherman:
This is looking great! Nice use of pagination container 😎

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -147,6 +147,7 @@ const RecommendedArtist: FC<
               user={user}
               mediator={mediator}
               onClick={onArtworkClicked}
+              lazyLoad
```
mzikherman: Oo nice esp for the off screen rail items this should be great!

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
```diff
@@ -1,53 +1,153 @@
...
 interface ArtistRecommendationsProps {
   artist: ArtistRecommendations_artist
+  relay: RelayPaginationProp
 }
 
+const PAGE_SIZE = 3
+
+// TODO:
+// * get help with relay errors
+// * uncomment test & get it working & add coverage for loading more
```
ashfurrow: You mentioned in knowledge share that you wanted to remove these.

src/Apps/Artist/Routes/Overview/Components/ArtistRecommendations.tsx
ashfurrow: Yeah, I think we should probably tackle this all at once across Reaction and Emission. They're functionally the same but both get copied when creating new pagination containers. Maybe we could also deprecate the non-canonical ones in the MP schema?

src/Apps/Artist/Routes/Overview/Components/__tests__/ArtistRecommendations.test.tsx
```diff
@@ -52,22 +52,89 @@ describe("ArtistRecommendations", () => {
...
+    artists: {
+      pageInfo: {
+        hasNextPage: false,
+        endCursor: "QXJ0aXN0OnBhYmxvLXBpY2Fzc28=",
+      },
+      edges: [
+        {
+          node: {
+            id: "QXJ0aXN0OnBhYmxvLXBpY2Fzc28=",
+            _id: "QXJ0aXN0OnBhYmxvLXBpY2Fzc28=",
```
ashfurrow: We can remove this `_id` to match the change in the component. The `id` should have the `=` removed so it looks more like the Gravity ID that this field returns.

src/Apps/Artist/Routes/Overview/Components/__tests__/ArtistRecommendations.test.tsx
```diff
@@ -52,22 +52,89 @@ describe("ArtistRecommendations", () => {
...
+            ...artistFields,
           },
-        ],
+          cursor: "QXJ0aXN0OnBhYmxvLXBpY2Fzc28=",
+        },
+      ],
+    },
+  },
+  id: "QXJ0aXN0Omp1YW4tZ3Jpcw==",
+  _id: "QXJ0aXN0Omp1YW4tZ3Jpcw==",
```
ashfurrow: Same here with `_id` and `id`, as well as in `pagedArtist`.

src/Apps/Artist/Routes/Overview/Components/__tests__/ArtistRecommendations.test.tsx
pepopowitz: Oh, this is something I forgot to mention in knowledge share. When I moved from using `mockResolvers` to `mockData` in the `<MockRelayRenderer>`, it required hydrating my fixture data with all fields that any component in the tree would need. There is a [downstream component that is referencing `_id`](https://github.com/pepopowitz/reaction/blob/46e6672438af72fd34eecaa94f62073e12891d43/src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx#L109), so it needs to be there. 

This is also why I ended up creating [this object](https://github.com/pepopowitz/reaction/blob/46e6672438af72fd34eecaa94f62073e12891d43/src/Apps/Artist/Routes/Overview/Components/__tests__/ArtistRecommendations.test.tsx#L68) to spread into all of the fixture nodes.

pepopowitz:
@ashfurrow I cleaned up what I could; let me know if you'd like any other changes!

artsyit:
:rocket: PR was released in v16.4.0 :rocket:

PR #2298, opened 2019-04-15
Fixes search bar button press getting redirected to homepage

ashfurrow:
When tapping the magnifying glass icon, users were getting directed to the homepage instead of the search results page. What was happening was: [react-autosuggest](https://github.com/moroshko/react-autosuggest) was intercepting `onBlur` from the input and calling `onSuggestionsClearRequested`, which cleared the search bar of text. _Then_ the button press would submit the form, directing to `/search?term=`, which Force redirects back to `/`. 

Because of how `SearchInputContainer` and `Autosuggest` refs work, checking the `form` property was the easiest way to make ignore button-press events (this piggy-backed on existing code to ignore other click events). I'm open to suggestions, though!

Fixes [DISCO-934](https://artsyproduct.atlassian.net/browse/DISCO-934).

PR #2312, opened 2019-04-19
Prevents search bar from submitting with empty query.

ashfurrow:
I was hoping to be able to specify form-level enabled-ness, but the form itself is housed in Force, so instead we have to add the following logic:

- When blurring from search because the icon button was clicked, only mark a descendant as being clicked if the search query is non-empty.
- When pressing enter to submit the form, only submit the search form if the query is non-empty.
- When handling click events on the icon button, only submit the search form if the query is non-empty.

This kind of DOM manipulation is kind of tricky to test, or at least I'm unfamiliar with it and there weren't any existing tests at this level of granularity. If tests would be appropriate for this PR, maybe we could pair on them? 

Follow-up from #2298. Fixes DISCO-934.

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -13,6 +13,7 @@ import {
   PLACEHOLDER_XS,
   SuggestionItem,
 } from "Components/Search/Suggestions/SuggestionItem"
+import { isEmpty } from "lodash"
 import { throttle } from "lodash"
 import qs from "qs"
 import React, { Component, useContext } from "react"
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 072c0bc091918af82bdf76d38bc8ec5666ff78a2

https://deploy-preview-2312--artsy-reaction.netlify.com

ashfurrow:
@dblandin Thanks for confirming locally!

artsyit:
:rocket: PR was released in v16.1.2 :rocket:

mzikherman:
I wonder if it's better to have the `form` here also at this point? I don't think there was any particular reason to keep that in Force (although when we were doing the AB test it was shared between the old search bar and this one). Might be a good refactor now that we've settled on this as our new search bar to bring that form in too.

ashfurrow:
Makes sense! That would simplify the logic added in this PR.

ashfurrow:
I've opened [DISCO-968](https://artsyproduct.atlassian.net/browse/DISCO-968) to follow-up on this from Matt's suggestion.

PR #2313, opened 2019-04-19
Fixes mapping between category and medium filter on /collect page to match gene names

sweir27:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-1025

The `medium` filter on the /collect page is actually filtering by _gene_, not by the exact category that was set in gravity. The mapping between category and mapping is [here](https://github.com/artsy/gravity/blob/c2599496bd098c489bb884f41e415bf4ff6739db/app/models/util/provisional_genome.rb#L29-L58), so I updated our mapping to match this logic.

Note that this is one more reason for why the number of artworks represented in the bar will not match exactly the number that are returned on the /collect page, but in this case it's only likely to show _more_ works here, not fewer.


artsyit:

Built with commit a280eca96deb7f02fa10e896b7fe053e27046cc5

https://deploy-preview-2313--artsy-reaction.netlify.com

ashkan18:
these mappings are pretty fragile in general... but 🙌

artsyit:
:rocket: PR was released in v16.1.4 :rocket:

PR #2325, opened 2019-04-23
User can not click into histogram bars

lilyfromseattle:
NOTE: I am reconsidering removing createCollectUrl as it can be modified to support the new link to browse similar works. This is why I added WIP. 

__User story__
As a user on an artwork page with a histogram
When I hover over the bars my cursor is an arrow (not the finger)
And I am not able to click into the bars

__Spec__
https://gallery.io/projects/MCHbtQVoQ2HCZbeSGvTZys3i/files/MCEJu8Y2hyDSce7zzddRvM87T4H3IF-rmBs


artsyit:

Built with commit eca98ce02261be9e64cf2fe4484c4cdfc2da70c0

https://deploy-preview-2325--artsy-reaction.netlify.com

ashkan18:
Nice! yes, i agree, i think it makes sense to have the new link as part of the same change so we can reuse existing method.

ds300:
As for re-using createCollectUrl in the same PR, I don't think it matters too much. Especially if it means this will PR will remain unmerged for more than a day. I'd recommend just merging now and bringing those files back in the next PR.

sweir27:
I agree we should merge this! We won't be able to add the link anyways until we're properly passing `size_score` through metaphysics (so that PR needs to be merged/deployed first).

Let's not block ourselves unnecessarily. 😄

artsyit:
:rocket: PR was released in v16.4.2 :rocket:

PR #2329, opened 2019-04-23
Replaces Slick for Flickity in Carousel component

kierangillen:
- https://artsyproduct.atlassian.net/browse/DISCO-947
- Replaces Slick for Flickity in our carousels
- For reasons for changing to Flickity: https://github.com/artsy/reaction/issues/2343

## With Javascript disabled for SSR

Screen Shot 2019-04-26 at 5 01 41 PM


artsyit:

Built with commit 984cf8ff44bf22acc8cec3e3af199bc0c519967f

https://deploy-preview-2329--artsy-reaction.netlify.com

src/Components/v2/Carousel.tsx
```diff
@@ -46,28 +45,35 @@ export class Carousel extends React.Component<Props> {
...
-  const [currentSlideIndex, setSlideIndex] = useState(0)
+  const [isMounted, toggleMounted] = useState(false)
+  useEffect(() => {
+    toggleMounted(true)
+  }, [])
+  const flcktyRef = useRef(null)
+
+  let flickity = null
+  if (isMounted) {
+    flickity = flcktyRef.current.flkty
```
damassi: 👍

src/Components/Artist/ArtistCollectionsRail/ArtistCollectionsRail.tsx
```diff
@@ -54,9 +54,9 @@ export class ArtistCollectionsRail extends React.Component<
 
           <Carousel
             height={200}
-            settings={{
-              slidesToScroll: 1,
-              infinite: true,
+            flickitySettings={{
```
damassi: If thinking in declarative terms, one shouldn't need to know that it's flickity underneith, and keep things general. By encoding `flickity` into the prop name we're leaking implementation details. Devs can refer to the inner component docs to know where to look for settings details.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+/*
+ * Returns null if no window for SSR
+ */
+const Flickity =
+  typeof window !== "undefined"
+    ? require("react-flickity-component")
+    : () => null
```
damassi: 👍

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+export const LargeCarousel = (props: Props) => {
+  const [isMounted, toggleMounted] = useState(false)
+  const [currentSlideIndex, setSlideIndex] = useState(0)
+  const [lastItemVisible, setLastItemVisible] = useState(false)
+
+  useEffect(() => {
+    toggleMounted(true)
+  }, [])
+  const flcktyRef = useRef(null)
```
damassi: typo?

src/Components/v2/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  accessibility?: boolean
+  setGallerySize?: boolean
+  resize?: boolean
+  cellAlign?: "left" | "center" | "right"
+  contain?: boolean
+  percentPosition?: boolean
+  rightToLeft?: boolean
+  prevNextButtons?: boolean
+  pageDots?: boolean
+}
```
damassi: Since the lib use is basically a one-off and not a larger framework-like-thing this doesn't need to be in its own file in the root, but rather live inside of the carousel file directly. Devs updating / making changes to the carousel then have a nice list of options immediately available to them, right at the top of the file.

kierangillen: Ok I'll move it there.

src/Apps/Artist/Routes/Overview/Components/RecommendedArtist.tsx
```diff
@@ -120,20 +119,12 @@ const RecommendedArtist: FC<
...
-          const { Arrow, currentSlideIndex, slidesToScroll } = arrowProps
-          return (
-            <ArrowWrapper>
-              {currentSlideIndex < artistData.length - slidesToScroll && (
-                <Arrow />
-              )}
-            </ArrowWrapper>
-          )
+          const { Arrow } = arrowProps
+          return <Arrow />
```
damassi: These render functions necessary anymore?

src/Components/Artist/ArtistCollectionsRail/ArtistCollectionsRail.tsx
```diff
@@ -54,9 +54,9 @@ export class ArtistCollectionsRail extends React.Component<
 
           <Carousel
             height={200}
-            settings={{
-              slidesToScroll: 1,
-              infinite: true,
+            carouselSettings={{
```
damassi: What the settings are for is implied, since the component name is Carousel.

src/Components/Artist/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
```diff
@@ -25,12 +25,15 @@ describe("CollectionsRail", () => {
...
   it("Renders expected fields", () => {
     const component = getWrapper()
-    expect(component.text()).toMatch("Browse by iconic collections")
-    expect(component.find(ArtistCollectionEntity).length).toBe(8)
-    expect(component.text()).toMatch("Flags")
-    expect(component.text()).toMatch("From $1,000")
-    expect(component.text()).toMatch("Street Art Now")
-    expect(component.text()).toMatch("From $200")
+    // Settimeout needed here for carousel render
+    setTimeout(() => {
```
damassi: Did the `renderUntil` enzyme helper not work that I linked yesterday? The idea is that enzyme will render and then wait until a certain class is on the dom before returning the wrapper that one then uses to make assertions against. 

If you do a global search in Reaction for `renderUntil` there are a lot of examples, [such as this](https://github.com/damassi/reaction/blob/a5e1c0b47091dc595a4cf557c5618d870992e593/src/Apps/Search/Routes/Artworks/Components/Filter/__tests__/SearchResultsArtworkGrid.test.tsx#L13). 

As an aside, `setTimeout` can often be swapped with `setImmediate` -- which is the same thing as `setTimeout` with a time set to 0. This means "on the next run loop, execute this function" which is often all that's needed. But avoiding this sort of run-loop hack is specifically why `renderUntil` was written. Let me know if the examples in Reaction aren't clear.

kierangillen: I would have preferred to use the `renderUntil` but I didn't have any luck getting it to work properly, so I used the `setTimeout` which fixed the issue.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
+import { Box, ChevronIcon, Flex } from "@artsy/palette"
+import React, { ReactNode, useEffect, useRef, useState } from "react"
+import styled from "styled-components"
+import { left, LeftProps, right, RightProps } from "styled-system"
+import { Media } from "Utils/Responsive"
+
+export interface FlickitySettings {
```
damassi: Screen Shot 2019-04-26 at 10 21 42 AM

These are available through http://definitelytyped.org/. 

`yarn add @types/flickity` -D`

damassi: Screen Shot 2019-04-26 at 10 23 45 AM

These types are available through http://definitelytyped.org/, which has Types for most JS libs.

`yarn add -D @types/flickity`

src/Components/Artist/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
damassi: I'm down to pair, and others in the office have used it; they'll be able to help if need be.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+  }
+}
+
+/*
+ * Returns null if no window for SSR
+ */
+const Flickity =
+  typeof window !== "undefined"
+    ? require("react-flickity-component")
+    : () => null
```
damassi: @kierangillen - can you post a screen-cap of this new work running (linked) in force with JS disabled, showing that the layout isn't broken during the SSR pass?

(Also, we need a screen cap of it running regularly in the description, too, along with some content from the now-closed RFC describing the _why_ around why this switch was necessary).

damassi: @kierangillen - can you post a screen-cap of this new work running (linked) in force with JS disabled, showing that the layout isn't broken during the SSR pass?

(Also, we need a screen cap of it running regularly in the description, too, along with some content from the now-closed RFC describing the _why_ around why this switch was necessary).

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+export const LargeCarousel = (props: Props) => {
+  const [isMounted, toggleMounted] = useState(false)
+  const [currentSlideIndex, setSlideIndex] = useState(0)
+  const [lastItemVisible, setLastItemVisible] = useState(false)
+
+  useEffect(() => {
+    toggleMounted(true)
+  }, [])
```
damassi: Since this is a somewhat common requirement, it might be good to move this to a custom hook `Utils/hooks/useDidMount.ts`. Something like:

```js
export function useDidMount() {
  const [isMounted, toggleMounted] = useState(false)

  useEffect(() => {
    toggleMounted(true)
  }, [])

  return isMounted
}
```

And then later, in this carousel, can do 

```
const isMounted = useDidMount()
```

It would also be good to announce in the #front-end that we now have a folder under `Utils/hooks` where common hooks are stored and that folks should contribute there.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  const settings = {
+    draggable: false,
+    freeScroll: false,
+    wrapAround: false,
+    cellAlign: "left",
+    pageDots: false,
+    prevNextButtons: false,
+    groupCells: true,
+    contain: true,
```
damassi: Are each of these default settings relevant to Flickity, or are they related to Slick?

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+    ...props.settings,
+  }
+
+  const LeftArrow = () => {
+    const getFlickity = flickity
```
damassi: Why add an intermediary variable? Seems to introduce some unnecessary indirection.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+    return (
+      <ArrowButton
+        onClick={() => {
+          getFlickity && getFlickity.previous ? getFlickity.previous() : null // check existence for tests
```
damassi:
```js
if (flickity) {
  flickity.previous()
} 
```
No need to return null on an `onClick` event handler; tests should be able to accommodate. The above form is also a bit clearer.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+      </ArrowButton>
+    )
+  }
+
+  const RightArrow = () => {
+    const getFlickity = flickity
+    return (
+      <ArrowButton
+        onClick={() => {
+          getFlickity && getFlickity.next ? getFlickity.next() : null // check existence for tests
```
damassi: Same feedback above applies here

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+  }
+  return (
+    <Flex
+      flexDirection="row"
+      position="relative"
+      justifyContent="space-around"
+      alignItems="center"
+      height={props.height}
+    >
+      {(currentSlideIndex !== 0 || settings.wrapAround === true) && (
```
damassi: if `currentSlideIndex !== 0 || settings.wrapAround === true` means "showArrow", then best to assign this to a variable just above the `return` so that future devs don't need to parse logic to understand behavior at a glance.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+        <ArrowWrapper left={-38}>
```
damassi: `left={space('4')}` so that things are constrained within our design system -- will need to figure out how to work in those 2px tho 😄

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+        {Flickity && (
+          <Flickity options={settings} ref={flicktyRef}>
+            {props.data.map((slide, index) => {
+              return <Box key={index}>{props.render(slide)}</Box>
+            })}
+          </Flickity>
+        )}
+      </CarouselContainer>
+
+      {(!lastItemVisible || settings.wrapAround === true) && (
```
damassi: `const showRightArrow = ...`

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+}
+
+export const SmallCarousel = (props: Props) => {
+  const settings = {
+    draggable: true,
+    freeScroll: false,
+    wrapAround: false,
+    cellAlign: "left",
+    pageDots: false,
+    prevNextButtons: false,
```
damassi: These settings are all specific to Flickity, yes? (Wondering if any are left over from slick)

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+            <LeftArrow />
+          )}
+        </ArrowWrapper>
+      )}
+
+      <CarouselContainer height={props.height}>
+        {Flickity && (
+          <Flickity options={settings} ref={flicktyRef}>
+            {props.data.map((slide, index) => {
+              return <Box key={index}>{props.render(slide)}</Box>
```
damassi: This means that this isn't comparable with SSR and will break the layout. If there's no flickity (no `window`), then `renderSlide` is never called. If that's not called then there will be a big gap when JS is disabled and google indexes content for SEO.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  return (
+    <Flex justifyContent="space-around" alignItems="center">
+      <CarouselContainer height={props.height}>
+        {Flickity && (
+          <Flickity options={settings}>
+            {props.data.map((slide, index) => {
+              return <Box key={index}>{props.render(slide)}</Box>
+            })}
+          </Flickity>
```
damassi: Same here

src/Components/v2/__stories__/CarouselV2.story.tsx
```diff
@@ -0,0 +1,186 @@
...
+  SmallCarousel,
+} from "Components/v2/CarouselV2"
+import { RelayStubProvider } from "DevTools/RelayStubProvider"
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { Section } from "Utils/Section"
+
+storiesOf("Styleguide/Components", module).add("CarouselV2", () => {
+  return (
+    <React.Fragment>
```
damassi: Can also do `<>` and `</>`

src/Components/v2/__stories__/CarouselV2.story.tsx
```diff
@@ -0,0 +1,186 @@
...
+              return (
+                <Image
+                  px={5}
+                  src={props.resized.url}
+                  width={props.resized.width}
+                  height={props.resized.height}
+                />
+              )
+            }}
+            renderLeftArrow={({ getFlickity }) => {
```
damassi: Since this is no longer a function call or a getter, the API can be updated to simply `flickity`. 

Once this PR merged, and a second PR is opened swapping out the old carousel for the new (removing V2) will need to search reaction and update the API where these render callbacks are being used (I think just one spot).

src/Components/Artist/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
kierangillen: Just resolved this.

src/Components/v2/CarouselV2.tsx
kierangillen: Added!

src/Components/v2/CarouselV2.tsx
kierangillen: They're all flickity settings.

src/Components/v2/CarouselV2.tsx
kierangillen: All Flickity settings.

src/Components/v2/CarouselV2.tsx
kierangillen: Do you think each hook should have it's own file or just one file for all hooks...`hooks.ts`

damassi: Own file -- small files are ok :) That way devs can scan their folder directory to see whats available, vs having to dig into files.

src/Components/Artist/ArtistCollectionsRail/ArtistCollectionsRail.tsx
```diff
@@ -55,8 +55,8 @@ export class ArtistCollectionsRail extends React.Component<
           <Carousel
             height={200}
             settings={{
-              slidesToScroll: 1,
-              infinite: true,
+              groupCells: 1,
+              wrapAround: true,
```
damassi: wondering -- should these two settings be in the defaults?

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,241 @@
...
+) => React.ReactNode
+
+interface Props {
+  height?: number
+  data: object[] // This is designed to handle any shape of data passed, as long as its an array
+  render: (slide) => ReactNode
+  renderLeftArrow?: Arrow
+  renderRightArrow?: Arrow
+  onArrowClick?: () => void
+  settings?: Flickity.Options
```
damassi: `import { Options as FlickityOptions } from 'flickity'`

src/Components/Artist/ArtistCollectionsRail/ArtistCollectionsRail.tsx
kierangillen: These two settings are unique to this carousel alone.

src/Components/v2/CarouselV2.tsx
damassi: (`space` is [a helper](https://github.com/artsy/palette/blob/master/packages/palette/src/helpers/space.ts) exported from `@artsy/palette`, fyi)

src/Components/v2/CarouselV2.tsx
damassi: Also, since `flickity` uses "Options" language, might be worth renaming `settings` to `options` so that things are 1:1.

src/Components/Artist/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
```diff
@@ -23,8 +23,12 @@ describe("CollectionsRail", () => {
...
   })
 
-  it("Renders expected fields", () => {
-    const component = getWrapper()
+  it("Renders expected fields", async () => {
+    const component = await mount(
+      <ArtistCollectionsRail {...props} />
+    ).renderUntil(n => {
+      return n.html().search("is-selected") > 0
+    })
```
damassi: 👍

src/Components/Artist/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
```diff
@@ -64,13 +68,15 @@ describe("CollectionsRail", () => {
...
         .simulate("click")
-
-      expect(dispatch).toBeCalledWith({
-        action_type: "Click",
-        context_module: "CollectionsRail",
-        context_page_owner_type: "Artist",
-        subject: "clicked next button",
-        type: "Button",
+      // Settimeout needed here for carousel render
+      setTimeout(() => {
```
damassi: This still needed with the `await` up above?

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+
+const renderSlides = props => {
+  return props.data.map((slide, index) => {
+    return <Box key={index}>{props.render(slide)}</Box>
+  })
+}
+
+const CarouselServer = (props: Props) => {
```
damassi: A small comment describing what's happening here would be good

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,260 @@
...
+    if (!flickity.slides.length) {
+      return
+    }
+    const isLastItemVisible = flickity.selectedElements.includes(
+      flickity.getLastCell().element
+    )
+    setLastItemVisible(isLastItemVisible)
+  }
+
+  if (isMounted) {
```
damassi: A little comment here to about what's going on / why this is needed; don't be afraid of comments 😄

src/Components/Artist/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
kierangillen: This one I couldn't get to work with the await :(

kierangillen:
@damassi This should be good to go at this point. Going to wait until Monday morning to merge.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,272 @@
...
+            <Media at="xs">
+              <SmallCarousel {...this.props} />
+            </Media>
+            <Media greaterThan="xs">
+              <LargeCarousel {...this.props} />
+            </Media>
+          </Box>
+        )}
+        {!isClient && (
+          <Box key={Math.random()}>
```
zephraph: I don't think you should leave the `Math.random` here. These keys just need to be different from each other so the `client` and `server` strings _should_ work. It's just that with `Math.random` there's a _slight_ chance that they may end up with the same values with would cause a really hard to reproduce bug.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,272 @@
...
+        <Flex>{renderSlides(props)}</Flex>
+      </CarouselContainer>
+    </Flex>
+  )
+}
+
+/*
+ * Returns null if no window for SSR
+ */
+const Flickity = isClient ? require("react-flickity-component") : () => null
```
zephraph: Hmmmmmmm.... This is interesting. Might be worth a TODO to make this an `import` when we can do proper bundle splitting.

src/Components/v2/__tests__/Carousel.test.tsx
```diff
@@ -46,7 +46,10 @@ describe("Carousel", () => {
           }}
         />
       </MockBoot>
-    )
+    ).renderUntil(n => {
+      return n.html().search("is-selected") > 0
```
zephraph:
```suggestion
      return n.html().includes("is-selected")
```

It's a minor thing but `includes` has a boolean api.

src/Components/v2/CarouselV2.tsx
kierangillen: the `client` and `server` strings weren't working but i think we could do `client + math.random()` or an equivalent so they'll never match.

kierangillen:
@zephraph Not all of `Slick` is removed, there are still two other components using it but should be easy to remove once we merge this in a follow up PR.

src/Components/v2/CarouselV2.tsx
damassi: Is a problem to use consistent keys? A `isServer ? 'keyA' : 'keyB'` sort of thing?

Otherwise, i don't know. Theoretically its possible for `Math.random()` to return two identical numbers two times in a row, but in that case the person seeing the missing slider probably wasn't meant to see the slider... That said, it's kinda a 🤔 to have to go the `Math.random()` route -- might want to play with it a bit.

zephraph: Seems weird to me that the two different strings aren't working. Random with a prefix would be fine. 👍

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -122,6 +122,8 @@ export class LargeArtistHeader extends Component<Props> {
             <Spacer my={2} />
           </Fragment>
         )}
+        <Spacer my={2} />
+        <Spacer my={2} />
```
damassi: Why both `<Spacers>`?

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+export const LargeCarousel = (props: Props) => {
+  const [currentSlideIndex, setSlideIndex] = useState(0)
+  const [lastItemVisible, setLastItemVisible] = useState(false)
+  const flicktyRef = useRef(null)
+  const isMounted = useDidMount()
```
damassi: Rather than using this `isMounted` can access via props passed down from the component's state up above. Can also delete this `useDidMount` file since we're no longer using it.

src/Apps/Artist/Components/ArtistHeader.tsx
kierangillen: For some reason after rerender the single spacer was collapsing on itself and not rendering any margin. Having two fixes the issue and their margins collapse together so there's always 20px.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,282 @@
...
+    setLastItemVisible(isLastItemVisible)
+  }
+
+  /*
+   * We need to wait for the component to mount before using the flickity ref.
+   * Flickity will only initialize after the component has rendered.
+   */
+  if (isMounted) {
+    if (flicktyRef.current !== null) {
+      flickity = flicktyRef.current.flkty
```
damassi: `flickity` !== `flkty` -- better to always keep naming 1:1 as that reduces refactor burdens later on.

src/Components/v2/CarouselV2.tsx
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    flickity.on("select", index => {
+      setSlideIndex(index)
+      checkLastItemVisible()
+    })
+  }
+  const options = {
```
damassi: Line break between 134 and 135

src/Components/v2/__stories__/CarouselV2.story.tsx
```diff
@@ -0,0 +1,186 @@
...
+                    getFlickity && getFlickity.next()
+                  }}
+                >
+                  Next
+                </Box>
+              )
+            }}
+          />
+        </Box>
+      </Section>
```
damassi: If it's easy, can we add another entry here that _only_ renders `CarouselServer`?

src/Components/v2/CarouselV2.tsx
kierangillen: I renamed the `flicktyRef` to `flickityRef`, Flickity uses `flkty` a lot in their docs and this is the object from the library which uses `flckty` here

src/Utils/hooks/useDidMount.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function useDidMount() {
+  const [isMounted, toggleMounted] = useState(false)
+
+  useEffect(() => {
+    toggleMounted(true)
+  }, [])
+
+  return isMounted
+}
```
damassi: This file can be removed since we're now tracking `isMounted` in the parent.

src/Apps/Artist/Components/ArtistHeader.tsx
damassi: Def a weird code smell -- maybe something like `<Box m={2} />` or `<Box m={2}>{"}</Box>`?

src/Components/v2/CarouselV2.tsx
damassi: However it's used in the docs is how it should be referenced in local assignments, so 1:1.

src/Apps/Artist/Components/ArtistHeader.tsx
damassi: Though it is weird that this is happening (haven't seen this before)

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -102,7 +102,7 @@ export class LargeArtistHeader extends Component<Props> {
     return (
       <Box width="100%">
         {hasImages && (
-          <Fragment>
+          <section>
```
damassi: Ahh yeah, i've seen weirdness with fragments in the past (just in general, not involving SSR)

damassi:
Nice work on this -- thanks for working through all the back and forth 👍

artsyit:
:rocket: PR was released in v16.6.0 :rocket:

kierangillen:
🥇

PR #2351, opened 2019-04-30
User can click browse link to go to filtered collect page

lilyfromseattle:
__User story__
As a user on an artwork page with a histogram I can see a link under the title that says "Browse works in this category" and when I click that link then I go to a collect page filtered by the same parameters as the histogram which can either be size, category, artist or category, artist

__Spec__
https://gallery.io/projects/MCHbtQVoQ2HCZbeSGvTZys3i/files/MCEJu8Y2hyDSce7zzddRvM87T4


artsyit:

Built with commit 505ed8478dd0022cca36c6795a2e976b5615fe15

https://deploy-preview-2351--artsy-reaction.netlify.com

artsyit:

Built with commit 2705a205cd21212e00418d5c9562436ae70f49c0

https://deploy-preview-2351--artsy-reaction.netlify.com

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,60 @@
...
+  dimension,
+  category,
+  sizeScore,
+  artistId,
+}) => {
+  let dimensionRange
+
+  if (dimension === null) {
+    dimensionRange = "*-*"
+  } else if (sizeScore < 1600) {
```
ds300: Can we use the `dimension` parameter instead of checking the value of `sizeScore`?

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,60 @@
+import * as qs from "qs"
+import { data as sd } from "sharify"
+
+export const createCollectUrl = ({
+  dimension,
+  category,
+  sizeScore,
+  artistId,
+}) => {
```
ds300: Would love to see these parameters typed.

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return `${path}?${query}`
+}
+
+// these come from MediumFilter.tsx
+const filterCategories = {
+  Architecture: "",
+  "Books and Portfolios": "",
+  "Design/Decorative Art": "design",
+  "Drawing, Collage or other Work on Paper": "drawing",
+  "Fashion Design and Wearable Art": "wearable art",
```
ds300: where did `"wearable art"` come from? I don't see it as an option in `MediumFilter.tsx`

src/Apps/Artwork/Components/PricingContext.tsx
```diff
@@ -50,14 +51,37 @@ export class PricingContext extends React.Component<PricingContextProps> {
...
   render() {
     const { artwork } = this.props
-
     if (!artwork.pricingContext) {
       return null
     }
 
+    let sizeScore
+
+    if (artwork.sizeScore != null) {
```
ds300: I don't think we need to be concerned with sizeScore in reaction yet. As mentioned elsewhere in this review I think we can lean on MP and vortex to return us a meaningful `appliedFilters.dimension`.

data/schema.graphql
```diff
@@ -48,19 +48,15 @@ type AggregationCount {
 type AnalyticsArtworksPublishedStats {
   percentageChanged: Int!
   period: AnalyticsQueryPeriodEnum!
-
-  # Time series data on number of artworks published
-  timeSeries(
-    samplingFrequency: AnalyticsSamplingFrequencyEnum
-  ): [AnalyticsArtworksPublishedTimeSeriesStats!]
+  timeSeries: [AnalyticsArtworksPublishedTimeSeriesStats!]
```
ds300: Might need to rebase, I think these schema changes are already in master.

src/Apps/Artwork/Utils/createCollectUrl.tsx
lilyfromseattle: Good idea

src/Apps/Artwork/Utils/createCollectUrl.tsx
lilyfromseattle: Idk I pulled these from the previous createCollectUrl for the clickable bars

ds300: haha turns out I added this :joy: https://github.com/artsy/reaction/commit/fec57ec0a8a2f7b50a3f7c234e28208dca6de533#diff-a829f088eac28b6207af9063873fc2ec

Let's change `"wearable art"` to `""` for now. `"wearable art"` is not a thing that is handled anywhere, and there doesn't seem to be an analogous filter.

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,59 @@
+import * as qs from "qs"
+import { data as sd } from "sharify"
+
+export const createCollectUrl = (
+  dimension: string,
+  category: string,
```
ds300: this type can be `keyof typeof filterCategories`

src/Apps/Artwork/Utils/__tests__/createCollectUrl.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    expect(result).toMatchInlineSnapshot(
+      `"https://staging.artsy.net/collect/drawing?page=1&sort=-decayed_merch&acquireable=true&offerable=true&inquireable_only=true&dimension_range=0-1600&artist_id=banksy"`
+    )
+  })
+
+  it("formats the collect page url correctly when not filtering by dimension", async () => {
+    const result = createCollectUrl(
+      null,
+      "Drawing, Collage or other Work on Paper",
+      "banksy"
```
ds300: This is a #nonblocking comment :)

A personal preference of mine is that functions with >= 3 arguments should almost always have those arguments grouped into a single object argument to make code easier to read and modify.

Here's a decent discussion of why https://stackoverflow.com/a/12827204

In this case, to understand the `null` in `createCollectUrl(null, ...)` you need to have in your mind which arguments the function takes and what order they are in. Whereas if this was written `createCollectUrl({size: null, ...})` you get extra information inline to help you understand what the `null` means.

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,65 @@
...
+  artistId,
+}: {
+  dimension: "SMALL" | "MEDIUM" | "LARGE" | null
+  category: FilterCategory
+  artistId: string
+}) => {
+  let dimensionRange
+
+  if (dimension === "SMALL") {
+    dimensionRange = "0-1600"
```
ds300: Steve just merged the PR to update vortex with proper size_score bands, so might as well fix these quickly before merging.

lilyfromseattle: Won't the dimension range stay the same?

ds300: Ah yes, sorry. I was under the mistaken impression that dimension_range was based on size_score rather than width_cm and height_cm.

src/Apps/Artwork/Utils/createCollectUrl.tsx
```diff
@@ -0,0 +1,65 @@
...
+    .filter(Boolean)
+    .join("/")
+
+  return `${path}?${query}`
+}
+
+export type FilterCategory = keyof typeof filterCategories
+
+// these come from MediumFilter.tsx
+const filterCategories = {
```
sweir27: It looks like this is using the outdated mapping that I fixed in https://github.com/artsy/reaction/pull/2313-- mind updating to match?

lilyfromseattle: Will do!

artsyit:
:rocket: PR was released in v16.8.4 :rocket:

PR #2356, opened 2019-04-30
Removes slick from ArtworkImageBrowser

kierangillen:
- Updates ArtworkImageBrowser to use `Flickity`
- Adds page dots to small carouselV3 

## Page dots

Screen Shot 2019-05-08 at 4 43 20 PM

## Client

Screen Shot 2019-05-01 at 4 53 58 PM

Screen Shot 2019-05-01 at 4 53 18 PM

## SSR
Screen Shot 2019-05-01 at 4 53 40 PM

Screen Shot 2019-05-01 at 4 53 29 PM

### DIFF
--- a/src/Apps/Artist/Components/ArtistHeader.tsx
+++ b/src/Apps/Artist/Components/ArtistHeader.tsx
@@ -106,7 +106,7 @@ export class LargeArtistHeader extends Component<Props> {
             <Carousel
               height="200px"
               data={carousel.images as object[]}
-              render={(slide: Image, index: number) => {
+              render={(slide: Image) => {
                 return (
                   <a href={slide.href} onClick={() => this.onClickSlide(slide)}>
                     <Image
--- a/src/Apps/Artist/Components/ArtistHeader.tsx
+++ b/src/Apps/Artist/Components/ArtistHeader.tsx
@@ -104,9 +104,9 @@ export class LargeArtistHeader extends Component<Props> {
         {hasImages && (
           <section>
             <Carousel
-              height={200}
+              height="200px"
               data={carousel.images as object[]}
-              render={(slide: Image) => {
+              render={(slide: Image, index: number) => {
                 return (
                   <a href={slide.href} onClick={() => this.onClickSlide(slide)}>
                     <Image
@@ -190,7 +190,7 @@ export class SmallArtistHeader extends Component<Props> {
           <Fragment>
             <Carousel
               data={carousel.images as object[]}
-              height={200}
+              height="200px"
               render={slide => {
                 return (
                   <a href={slide.href} onClick={() => this.onClickSlide(slide)}>
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
@@ -10,7 +10,7 @@ import { Box, ChevronIcon, Col, color, Flex, space } from "@artsy/palette"
 interface ArtworkBrowserProps {
   imageAlt: string
   images: ArtworkImageBrowser_artwork["images"]
-  flickityRef?: (flickityRef: Flickity) => void
+  setFlickityRef: (flickityRef: Flickity) => void
 }
 
 export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
@@ -40,7 +40,7 @@ export class LargeArtworkImageBrowser extends React.Component<
 
   render() {
     const hasMultipleImages = this.props.images.length > 1
-    const { imageAlt, images } = this.props
+    const { imageAlt, images, setFlickityRef } = this.props
 
     // FIXME: During SSR pass want to hide other images. Work around for lack
     // of SSR support in Flickity.
@@ -53,6 +53,7 @@ export class LargeArtworkImageBrowser extends React.Component<
           options={this.options}
           oneSlideVisible
           height="60vh"
+          setFlickityRef={setFlickityRef}
           data={carouselImages}
           renderLeftArrow={({ flickity }) => (
             <Col sm={1}>
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
@@ -1,7 +1,7 @@
 import { ArtworkImageBrowser_artwork } from "__generated__/ArtworkImageBrowser_artwork.graphql"
 import { Lightbox } from "Components/v2"
+import { Options as FlickityOptions } from "flickity"
 import React from "react"
-import Slider, { Settings } from "react-slick"
 import styled from "styled-components"
 import { Media } from "Utils/Responsive"
 
@@ -10,8 +10,19 @@ import { Box, ChevronIcon, Col, color, Flex, Row } from "@artsy/palette"
 interface ArtworkBrowserProps {
   imageAlt: string
   images: ArtworkImageBrowser_artwork["images"]
-  sliderRef?(slider: Slider): void
+  flickityRef?: (flickityRef: Flickity) => void
 }
+
+interface ArtworkBrowserState {
+  isLocked?: boolean
+  mounted: boolean
+}
+
+/*
+ * Returns null if no window for SSR
+ */
+const isClient = typeof window !== "undefined"
+const Flickity = isClient ? require("react-flickity-component") : () => null
 
 export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
   return (
@@ -27,29 +38,72 @@ export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
 }
 
 export class LargeArtworkImageBrowser extends React.Component<
-  ArtworkBrowserProps
+  ArtworkBrowserProps,
+  ArtworkBrowserState
 > {
-  slider: Slider
+  state = { mounted: false }
 
-  setSliderRef = slider => {
-    this.slider = slider
-    if (this.props.sliderRef) {
-      this.props.sliderRef(slider)
+  flickity = null
+
+  options = {
+    prevNextButtons: false,
+    wrapAround: true,
+    pageDots: true,
+    cellAlign: "left",
+    draggable: false,
+    lazyLoad: true,
+  } as FlickityOptions
+
+  componentDidMount() {
+    this.setState({
+      mounted: true,
+    })
+    const { flickityRef } = this.props
+    if (this.flickity && flickityRef) {
+      flickityRef(this.flickity)
     }
   }
 
-  get settings(): Settings {
-    return {
-      arrows: false,
-      customPaging: () => <PageIndicator />,
-      dots: true,
-      infinite: false,
-      lazyLoad: "ondemand",
-      speed: 0,
-      swipe: false,
-      slidesToShow: 1,
-      slidesToScroll: 1,
-    }
+  renderSlide = (image, hasMultipleImages) => {
+    return (
+      <Flex
+        flexDirection="column"
+        justifyContent="center"
+        width="100%"
+        px={hasMultipleImages ? [2, 2, 0] : 0}
+        key={isClient + image.id}
+      >
+        <Lightbox
+          deepZoom={image.deepZoom}
+          enabled={image.is_zoomable}
+          isDefault={image.is_default}
+        >
+          <DesktopImage src={image.uri} width="100%" />
+        </Lightbox>
+      </Flex>
+    )
+  }
+
+  renderCarousel = hasMultipleImages => {
+    return (
+      <Flickity options={this.options} flickityRef={c => (this.flickity = c)}>
+        {this.props.images.map(image => {
+          return this.renderSlide(image, hasMultipleImages)
+        })}
+      </Flickity>
+    )
+  }
+
+  renderServerCarousel = hasMultipleImages => {
+    return (
+      <Flex>
+        {this.props.images.map((image, index) => {
+          const isFirstImage =
+            index === 0 ? this.renderSlide(image, hasMultipleImages) : null
+          return isFirstImage
+        })}
+      </Flex>
+    )
   }
 
   render() {
@@ -62,7 +116,7 @@ export class LargeArtworkImageBrowser extends React.Component<
             <Col sm={1}>
               <ArrowButton
                 direction="left"
-                onClick={() => this.slider.slickPrev()}
+                onClick={() => this.flickity.previous(false, true)}
               />
             </Col>
           )}
@@ -93,7 +147,7 @@ export class LargeArtworkImageBrowser extends React.Component<
             <Col sm={1}>
               <ArrowButton
                 direction="right"
-                onClick={() => this.slider.slickNext()}
+                onClick={() => this.flickity.next(false, true)}
               />
             </Col>
           )}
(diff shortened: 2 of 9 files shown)


artsyit:

Built with commit de0f9866c6529efeb37ca43d15f730a946d2302d

https://deploy-preview-2356--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -33,73 +39,104 @@ export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
...
+    lazyLoad: true,
+  } as FlickityOptions
+
+  componentDidMount() {
+    this.setState({
+      mounted: true,
+    })
+    const { setFlickity } = this.props
+    if (this.flickity && setFlickity) {
+      setFlickity(this.flickity)
```
damassi: Can you elaborate on this a bit?

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -33,73 +39,104 @@ export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
   )
 }
 
+/*
+ * Returns null if no window for SSR
+ */
+const isClient = typeof window !== "undefined"
+const Flickity = isClient ? require("react-flickity-component") : () => null
```
damassi: Rather than duplicating this logic, it might make more sense to move it inside of a factory function inside of `Carousel`, something like: 

```js
const { Flickity } = getFlickity() 
```
and inside of there it handles this bit of duplicated logic. 

The ideal is being able to just use `Carousel` (or Flickity if its separate from the Carousel component) with as little overhead as possible.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -33,73 +39,104 @@ export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
...
+  renderServerCarousel = hasMultipleImages => {
+    return (
+      <Flex>
+        {this.props.images.map((image, index) => {
+          const isFirstImage =
+            index === 0 ? this.renderSlide(image, hasMultipleImages) : null
+          return isFirstImage
+        })}
+      </Flex>
+    )
```
damassi: Same here -- we shouldn't have to scaffold components in order to use Flickity. Behind the scenes this kind of setup can happen so that at the consumer level its usable without needing explicit setup instructions.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -33,73 +39,104 @@ export const ArtworkImageBrowser = (props: ArtworkBrowserProps) => {
...
-                    >
-                      <DesktopImage src={image.uri} width="100%" />
-                    </Lightbox>
-                  </Flex>
-                )
-              })}
-            </Slider>
+            {this.state.mounted
+              ? this.renderCarousel(hasMultipleImages)
+              : this.renderServerCarousel(hasMultipleImages)}
```
damassi: Also stuff that should happen behind the scenes

src/Apps/Artwork/Components/ArtworkImageBrowser/index.tsx
```diff
@@ -28,13 +30,12 @@ export class ArtworkImageBrowserContainer extends React.Component<
 
     return (
       <>
-        <ArtworkImageBrowser
-          images={images}
-          sliderRef={slider => (this.slider = slider)}
-        />
+        <ArtworkImageBrowser images={images} setFlickity={this.setFlickity} />
```
damassi: `flickityRef={flickityRef => this.setFlickity = flickityRef}`

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
damassi: A good way to think about this is a new file -- `Components/Flickity` -- serves as a factory function (or something) handling setup and so on. Inside of there is where implementation details live, rather than within consumers, which serves to centralize / hide complexity. If things need to change, we change them there rather than at each call-site.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
kierangillen: I'm not sure this is possible with all of the rehydration and ref passing but I can timebox looking into it.

damassi: If needed we can work through some designs to help keep things as simple as we can. 

One strategy I often employ is to first write out exactly what I would _like_ to use, like the ideal component / api as it would eventually be used, and then after that feels right I then work backwards and retrofit the underlying code to make everything work according to the ideal. This approach sometimes takes a bit more upfront planning (not necessarily so, tho) but sometimes working backwards helps one work forwards in a better way.

damassi: (I can also take a look at the new Flickity scroller to see if there's anything we can do to make sure the implementation details are contained a bit more.)

src/Components/v2/CarouselV2.tsx
```diff
@@ -220,7 +220,7 @@ export const LargeCarousel = (props: Props) => {
 export const SmallCarousel = (props: Props) => {
   const options = {
     draggable: true,
-    freeScroll: false,
+    freeScroll: true,
```
kierangillen: Many images are wider than the mobile viewport so having `freescroll` enabled is definitely a better UX than having images lock when scrolled.

src/Components/v2/CarouselV3.tsx
```diff
@@ -128,6 +134,7 @@ export class BaseCarousel extends React.Component<
     FlickityCarousel: Flex as any,
     currentSlideIndex: 0,
     lastItemVisible: false,
+    mounted: false,
```
damassi: `isMounted`

src/Components/v2/CarouselV3.tsx
```diff
@@ -11,10 +11,15 @@ interface CarouselProps {
    */
   data: any
 
+  /**
+   * If this carousel contains only one image on render set to true (for SSR)
+   */
+  isOneSlide?: boolean
```
damassi: This is an implementation detail rather than a consumer API prop. Stuff related to SSR should be entirely hidden, rather than creating a scaffolding in order to work around a problem.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -190,10 +195,14 @@ export class SmallArtistHeader extends Component<Props> {
           <Fragment>
             <Carousel
               data={carousel.images as object[]}
-              height={200}
+              height="200px"
```
damassi: Styled system will infer numerics values, or a string. Why make this change?

src/Components/v2/CarouselV3.tsx
```diff
@@ -63,7 +68,8 @@ type ArrowProps = (props: {
 
 export class Carousel extends React.Component<CarouselProps> {
   static defaultProps = {
-    height: 300,
+    height: "300px",
+    isOneSlide: false,
```
damassi: Also, how many slides a carousel has can be inferred from `data.length` vs a declaration

src/Components/v2/CarouselV3.tsx
```diff
@@ -333,14 +347,14 @@ const CarouselContainer = styled.div<{ height?: number; mounted: boolean }>`
...
-        height: ${props.height}px;
+        height: ${props.height};
       `
     }
   }};
 `
 
 export const ArrowButton = styled(Flex)<
-  LeftProps & RightProps & { height?: number }
+  LeftProps & RightProps & { height?: string }
```
damassi: `HeightProps`

src/Components/v2/__stories__/CarouselV3.story.tsx
```diff
@@ -44,6 +45,7 @@ storiesOf("Styleguide/Components/CarouselV3", module)
                 src={props.resized.url}
                 width={props.resized.width}
                 height={props.resized.height}
+                key={props.resized.url}
```
damassi: Did adding this key fix the flash?

src/Components/v2/__tests__/Carousel.test.tsx
```diff
@@ -13,8 +13,8 @@ describe("Carousel", () => {
       <MockBoot breakpoint="xs">
         <Carousel
           data={[{ name: "foo" }]}
-          render={props => {
-            return <div />
+          render={(_, index) => {
+            return <div key={index} />
```
damassi: Since the render method simply returns a value, the key can be added inside of Carousel directly within the iterator. No need to pass additional data up through props.

src/Components/v2/CarouselV3.tsx
kierangillen: ahh ok thanks

src/Components/v2/CarouselV3.tsx
kierangillen: It's the visible slides not the length of the carousel

kierangillen: Maybe I can make that more explicit in the naming of this variable

src/Apps/Artist/Components/ArtistHeader.tsx
kierangillen: I was going to only allow for strings so we can pass `"60vh"` so to not get `"60vhpx"` elsehwere.

src/Components/v2/CarouselV3.tsx
damassi: Ahh i see; yup, just a naming issue 👍 That said, this sort of thing is pretty low-level; sure there's not something within Flickity that we can piggyback on?

src/Components/v2/__tests__/Carousel.test.tsx
kierangillen: I didnt want to return any wrapper element surrounding the Carousel mapped objects because that wrapper takes specific stylings.

src/Components/v2/__stories__/CarouselV3.story.tsx
kierangillen: :( nope

src/Apps/Artist/Components/ArtistHeader.tsx
damassi: 👍 using `HeightProps` in the interface should allow for more flexibility

src/Components/v2/CarouselV3.tsx
kierangillen: We need to know before flickity is initialized so to only render one image for SSR, so we wouldn't be able to use flickity's API.

src/Components/v2/__stories__/CarouselV3.story.tsx
damassi: Unless it's needed (errors in the console) then we can omit keys

src/Components/v2/__tests__/Carousel.test.tsx
damassi: https://github.com/artsy/reaction/blob/b186231b2b67d0c8b94b8151777b2b47c1ef5f02/src/Components/v2/CarouselV3.tsx#L316

can be wrapped with a 

`<Fragment key={index}>`

kierangillen: Cool

src/Components/v2/CarouselV3.tsx
damassi: I guess what i mean is something like [`slidesToShow`](https://react-slick.neostack.com/docs/api#slidesToShow) as seen in `react-slick`; this kind of detail can quickly lead to complexity within the wrapper so I think it might be worth thinking about a bit before adding. We're working around an SSR issue, rather than defining behaviors for the consumer of the carousel. If we _didn't_ need to SSR render then this prop would be unnecessary. 

Let's give this a think... the snippet sent last night was a bit of a hack 🤔 🤔

src/Apps/Artist/Components/ArtistHeader.tsx
kierangillen: `HeightProps` isn't working so leaving this alone for now.

damassi: 👍

src/Components/v2/CarouselV3.tsx
kierangillen: ```const carouselImages = typeof window === "undefined" && oneSlideVisible ? [data[0]] : data```

We need to do something like this otherwise if the carousel has more than one image it will only render the first image on page load.

damassi: Thoughts on https://github.com/metafizzy/flickity/issues/526?

kierangillen: We can't use Flickity's API, because this is pre-flickity loading. This is just for SSR so we don't have that double image in one slide like we were seeing in SSR.

damassi:
A PR has been opened upstream to address the server / client flicker issue: https://github.com/theolampert/react-flickity-component/pull/64

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -1,16 +1,20 @@
 import { ArtworkImageBrowser_artwork } from "__generated__/ArtworkImageBrowser_artwork.graphql"
 import { Lightbox } from "Components/v2"
 import React from "react"
-import Slider, { Settings } from "react-slick"
 import styled from "styled-components"
 import { Media } from "Utils/Responsive"
+import { BaseCarousel as Carousel } from "../../../../Components/v2/CarouselV3"
```
damassi: Anything more distant that 1 level deep should be an absolute path

src/Components/v2/CarouselV3.tsx
```diff
@@ -9,12 +9,22 @@ interface CarouselProps {
...
   /**
    * The height of the carousel
    */
-  height?: number
+  height?: string
+
+  /**
+   * The width of the carousel
+   */
+  width?: number
```
damassi: If `height` is a string we should be sure `width` matches too

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -104,9 +104,9 @@ export class LargeArtistHeader extends Component<Props> {
         {hasImages && (
           <section>
             <Carousel
-              height={200}
+              height="200px"
               data={carousel.images as object[]}
-              render={(slide: Image) => {
+              render={(slide: Image, index: number) => {
```
damassi: this `index` being used anywhere?

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -1,16 +1,20 @@
...
 
 interface ArtworkBrowserProps {
   imageAlt: string
   images: ArtworkImageBrowser_artwork["images"]
-  sliderRef?(slider: Slider): void
+  flickityRef?: (flickityRef: Flickity) => void
+}
+
+interface ArtworkBrowserState {
+  isLocked?: boolean
```
damassi: This being used?

src/Components/v2/CarouselV3.tsx
```diff
@@ -24,7 +34,7 @@ interface CarouselProps {
   /**
    * The render callback returns an image to display
    */
-  render: (slide) => React.ReactNode
+  render?: (slide, index) => React.ReactNode
```
damassi: I believe this `render` is required

kierangillen: And only takes a `slide` param looks like

artsyit:
:rocket: PR was released in v16.9.0 :rocket:

PR #2358, opened 2019-05-01
[NavBar] Fix notification count

damassi:
#trivial 

We can't access per-request data via sharify in a static way; needed to pass things in via `SystemContext` from force and access them that way.

### DIFF
--- a/src/Artsy/SystemContext.tsx
+++ b/src/Artsy/SystemContext.tsx
@@ -35,6 +35,11 @@ export interface SystemContextProps {
    * and `USER_ACCESS_TOKEN` environment variables if available.
    */
   user?: User
+
+  /**
+   * FIXME: Ask alloy how to pass one-off props like this in from force
+   */
+  notificationCount?: number
 }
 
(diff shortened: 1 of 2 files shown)


src/Artsy/SystemContext.tsx
```diff
@@ -35,6 +35,11 @@ export interface SystemContextProps {
    * and `USER_ACCESS_TOKEN` environment variables if available.
    */
   user?: User
+
+  /**
+   * FIXME: Ask alloy how to pass one-off props like this in from force
+   */
+  notificationCount?: number
```
damassi: @alloy - when you get back wondering if we can discuss approaches to passing data into the `SystemContext` from force. Now that we removed free-form access this will error out and I'm not sure how to extend it from the perspective of `<SystemContextProvider {...otherProps}>`.

artsyit:

Built with commit 44809e17bb701d2574a432fdff0aed3d5c71d294

https://deploy-preview-2358--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v16.6.6 :rocket:

PR #2364, opened 2019-05-02
Update <Carousel /> with BaseCarousel

damassi:
This continues on work done in CarouselV2, where @kierangillen replaced Slick with Flickity. We were noticing some issues with SSR rendering which required some round about ways of handling it, so this takes the work done there and slightly rearranges it inside of a Class, which seemed to fix the problem. 

Also broke the Storybook entry out into individual stories so that we can select items from the sidebar vs scrolling. 

#### Todo:
- [x] Update tests


artsyit:

Built with commit 790ac9c6849ed361f1204f41d9c2b2d8a8600aa5

https://deploy-preview-2364--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v16.8.0 :rocket:

PR #2365, opened 2019-05-02
DISCO-920 - Artwork page images do not use <img> with alt text

kierangillen:
- Artwork images are now using the `` element and they're using the artwork's `image_alt` field as their alt text.
- https://artsyproduct.atlassian.net/browse/DISCO-920

Screen Shot 2019-05-02 at 1 14 36 PM

Screen Shot 2019-05-02 at 1 14 42 PM

PR #2371, opened 2019-05-06
Uses <Img> instead of <div> for SSR artwork image

kierangillen:
- Followup to https://github.com/artsy/reaction/pull/2365
- Artwork SSR images were still rendering as `<div>`, this PR renders them in `` elements for SEO purposes and adds an `alt` to the image.
- https://artsyproduct.atlassian.net/browse/DISCO-920

## SSR (JS disabled rendering)

Screen Shot 2019-05-06 at 11 31 07 AM

Screen Shot 2019-05-06 at 11 32 05 AM

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
@@ -5,15 +5,7 @@ import Slider, { Settings } from "react-slick"
 import styled from "styled-components"
 import { Media } from "Utils/Responsive"
 
-import {
-  Box,
-  ChevronIcon,
-  Col,
-  color,
-  Flex,
-  ResponsiveImage,
-  Row,
-} from "@artsy/palette"
+import { Box, ChevronIcon, Col, color, Flex, Image, Row } from "@artsy/palette"
 
 interface ArtworkBrowserProps {
   imageAlt: string
@@ -89,8 +81,11 @@ export class LargeArtworkImageBrowser extends React.Component<
                       deepZoom={image.deepZoom}
                       enabled={image.is_zoomable}
                       isDefault={image.is_default}
+                      src={image.uri}
                     >
-                      <DesktopImage src={image.uri} width="100%" />
+                      <Flex justifyContent="center" alignItems="center">
+                        <DesktopImage alt={imageAlt} src={image.uri} />
+                      </Flex>
                     </Lightbox>
                   </Flex>
                 )
@@ -138,10 +133,11 @@ export class SmallArtworkImageBrowser extends React.Component<
   }
 
   render() {
+    const { sliderRef, imageAlt, images } = this.props
     return (
       <Container>
-        <Slider {...this.settings} ref={this.props.sliderRef}>
-          {this.props.images.map(image => {
+        <Slider {...this.settings} ref={sliderRef}>
+          {images.map(image => {
             return (
               <Flex
                 flexDirection="column"
@@ -150,12 +146,13 @@ export class SmallArtworkImageBrowser extends React.Component<
                 key={image.id}
               >
                 <Lightbox
-                  imageAlt={this.props.imageAlt}
+                  imageAlt={imageAlt}
                   deepZoom={image.deepZoom}
                   enabled={!this.state.isLocked && image.is_zoomable}
                   isDefault={image.is_default}
+                  src={image.uri}
                 >
-                  <ResponsiveImage src={image.uri} width="100%" />
+                  <Image alt={imageAlt} src={image.uri} width="100%" />
                 </Lightbox>
               </Flex>
             )
@@ -209,8 +206,9 @@ const Container = styled(Box)`
   }
 `
 
-const DesktopImage = styled(ResponsiveImage)`
-  padding-bottom: 60vh; /* Responsive max height */
+const DesktopImage = styled(Image)`
+  max-height: 60vh;
+  max-width: 100%;
 `
 
 const PageIndicator = styled.span`
@@ -221,5 +219,4 @@ const PageIndicator = styled.span`
 
 DesktopImage.displayName = "DesktopImage"
 // @ts-ignore
-ResponsiveImage.displayName = "ResponsiveImage"
 PageIndicator.displayName = "PageIndicator"
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
@@ -5,7 +5,7 @@ import Slider, { Settings } from "react-slick"
 import styled from "styled-components"
 import { Media } from "Utils/Responsive"
 
-import { Box, ChevronIcon, Col, color, Flex, Image, Row } from "@artsy/palette"
+import { Box, ChevronIcon, Col, color, Flex, Row } from "@artsy/palette"
 
 interface ArtworkBrowserProps {
   imageAlt: string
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit cdec5f2094cc09b4ae6a8075b4c00002f64efa43

https://deploy-preview-2371--artsy-reaction.netlify.com

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -278,7 +279,7 @@ export class Lightbox extends React.Component<LightboxProps, LightboxState> {
   }
 
   render() {
-    const { children, enabled, isDefault, imageAlt } = this.props
+    const { children, enabled, isDefault, imageAlt, src } = this.props
```
pepopowitz: If I'm understanding correctly, this means that a `<Lightbox>` component will apply its `src` to all `<Image>` elements within it. I don't know much about the context of these changes, but that seems like a bug. 

From reading this change, I think this code:

```
<Lightbox src="image-1.jpg">
  <Image src="image-2.jpg" />
  <Image src="image-3.jpg" />
  <Image src="image-4.jpg" />
</Lightbox>
```

would result in `image-1.jpg` showing up three times, instead of seeing `image-2`, `image-3`, and `image-4`. 


Am I misunderstanding this change?

kierangillen: Each `Lightbox` component renders only one image. Happy to look through it via a call if you want to delve into it. It's a complicated component.

pepopowitz: Summarizing a call between @kierangillen and I, for permanence: 

The `<Lightbox>` component is currently giving conflicting hints to devs. It iterates over all the children passed in, but (a) we are never actually giving it more than one, and (b) it doesn't seem like the second one is displayed very well (or at all, we aren't sure). With the changes in this PR, it now actually would iterate over all of them, but show the `src` applied to the `<Lightbox>` element each time.

We're working on finding out which is more accurate about the `<Lightbox>` - does it need to display more than one, in which case we need to remove these assumptions of only displaying one? Or does it only ever need to display one, in which case we can make it impossible to give this component more than one child image?

kierangillen: Via Slack we've agreed to remove the `.map()` and render only one image per `Lightbox` component for now.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -89,8 +81,11 @@ export class LargeArtworkImageBrowser extends React.Component<
                       deepZoom={image.deepZoom}
                       enabled={image.is_zoomable}
                       isDefault={image.is_default}
+                      src={image.uri}
                     >
-                      <DesktopImage src={image.uri} width="100%" />
+                      <Flex justifyContent="center" alignItems="center">
+                        <DesktopImage alt={imageAlt} src={image.uri} />
```
damassi: why is is `src` in `<Lightbox>` + `<DesktopImage>` needed? Can't one be derived from the other?

kierangillen: `DesktopImage` is rendered only on the SSR pass and the `Lightbox` renders a separate ``

damassi: If you imagine `<Lightbox>` as a self contained thing, then piggybacking off of the image as is passed into `src` makes it configuration free, and all behavior can be contained within.

Inside of `<Lightbox>`, where it was mapping over children, the image can be rendered, with the container if need be. And then from an API perspective all that’s needed is

```js
<Lightbox src="path/to/image" ... />
```

which makes everything nice and portable. It also removes the need to duplicate `image.uri` in two separate places.

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -278,42 +279,31 @@ export class Lightbox extends React.Component<LightboxProps, LightboxState> {
...
-                src={src}
-                alt={imageAlt}
-                data-type="artwork-image"
-                data-is-default={isDefault}
-              />
-            </Flex>
-          )
-        )
-      }
-    )
```
damassi: Mapping over children to derive props is ok, and is a a common approach when writing lib-like code (such as `Lightbox`. 

The way to enforce that only one child is present is via the `React.Children.only`: https://reactjs.org/docs/react-api.html#reactchildrenonly. 

That way this all of this stuff can be self-contained, and the api remain configuration-less (vs moving the container and so on _outside_ of the component.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
damassi: That said, the currently-existing behavior as described [here](https://github.com/artsy/reaction/pull/2371#discussion_r281344381) is actually correct.

zephraph: Originally, `Lightbox` was just intended to be a HOC that could consume an Image component of some sort and decorate it with the things it needed to invoke a Lightbox.

From just reading over the PR this API seems a little deceptive to me. You're allowing children to be passed but *only* rendering the children when it's not SSR. This would be a really easy way for folks to unknowingly introduce a rehydration issue. 

I'm just worried it'd be confusing for folks. It should either always take a child (and render that appropriately based on context) or never take a child.

damassi: 👍 We were just chatting on a call and that's what we settled on; sticking to a simple `src` and ommitting children completely. That way it removes the duplication and the ambiguity.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -209,8 +202,9 @@ const Container = styled(Box)`
   }
 `
 
-const DesktopImage = styled(ResponsiveImage)`
-  padding-bottom: 60vh; /* Responsive max height */
+const DesktopImage = styled(Image)`
```
damassi: Is `DesktopImage` still being used?

src/Components/v2/__stories__/Lightbox.story.tsx
```diff
@@ -27,7 +27,11 @@ storiesOf("Styleguide/Components", module).add("Lightbox", () => {
   return (
     <React.Fragment>
       <Section title="Lightbox">
-        <Lightbox imageAlt="Kieran Gillen, Untitled, 1984" deepZoom={deepZoom}>
+        <Lightbox
+          src={image("large")}
+          imageAlt="Arya Stark, Untitled, 1984"
+          deepZoom={deepZoom}
+        >
           <Image width={width} height={height} src={image("large")} />
```
damassi: This can probably be removed, yah?

kierangillen: Yup!

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -278,42 +280,36 @@ export class Lightbox extends React.Component<LightboxProps, LightboxState> {
   }
 
   render() {
-    const { children, enabled, isDefault, imageAlt } = this.props
+    const { enabled, isDefault, imageAlt, src, initialHeight } = this.props
+    const height = initialHeight ? initialHeight : "auto"
```
damassi:
```js 
const height = initialHeight || "auto" 
```

damassi:
This ended up being a nice refactor 👍

kierangillen:
@damassi Yeah, I think it's much easier to understand what's going on now.

artsyit:
:rocket: PR was released in v16.8.7 :rocket:

PR #2379, opened 2019-05-07
Pass dimension_range through the CollectApp viewer + update dimension_range to match size_score

sweir27:
Noticed while QAing that we weren't actually filtering the /collect page by `dimension_range`. 

There were two issues:
- `dimension_range` indexes based on `size_score`, so we need to update the numbers to match those.
- We were ingesting `dimension_range` at the route level, but weren't passing it through to the `CollectApp_viewer` or the refetch container. This fixes that, so now we're correctly filtering on initial load and on subsequent loads if you change the other filters.


artsyit:

Built with commit 90f51035c5510cf45361702e15d39f4547f52b3a

https://deploy-preview-2379--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v16.8.10 :rocket:

PR #2380, opened 2019-05-07
Use the current URL to determine the paginated URL in a shared component

pepopowitz:
There was a small bug in #2369. It made the assumption in `<CollectArtworkGrid>` that the URL was `/collect`...but that component is used by both the `/collect` and `/collection/:collectionId` routes. The result was that paging through results on https://staging.artsy.net/collection/art-of-polka-dots would change the URL to https://staging.artsy.net/collect?page=X. 

This change pulls the path from the current location, instead of a hardcoded value.

For the record, #2369 has been merged & deployed to staging. This fix would ideally also get merged before our next production deploy of Force.


artsyit:

Built with commit 19bf8e52d369e16e9544419459e720c70de3c95f

https://deploy-preview-2380--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v16.8.11 :rocket:

PR #2384, opened 2019-05-07
DISCO-1005: preserve pagination on collections page

pepopowitz:
This PR populates the browser history with the appropriate URLs during pagination of a "collection/XXX" page, and makes the Collection app respect params passed in for any filters when you first hit the page. 

It is similar to #2369, but for a different page/"app". 

While doing this work, I also noticed that there were a couple other filters that weren't being respected when passed as querystring params: `offerable` and `color`. I corrected those also.


artsyit:

Built with commit 30270792fdf06e4d94e1bccd5e004e7650a9a76d

https://deploy-preview-2384--artsy-reaction.netlify.com

src/Apps/Collect/Components/Base/CollectArtworkGrid.tsx
```diff
@@ -55,7 +55,8 @@ class CollectArtworkGrid extends Component<Props, LoadingAreaState> {
         const { state } = filters
         const urlFragment = urlFragmentFromState(state, { page })
 
-        // TODO: Look into using router push w/ query params.
+        // Using window.history.pushState instead of router.push, because
+        //  we just want to add to the history, not navigate to another route.
```
damassi: #so-absurdly-minor: there's an extra space before the `we`

src/Apps/Collect/Components/Base/CollectRefetch.tsx
```diff
@@ -64,11 +64,14 @@ export class CollectRefetch extends Component<CollectRefetchProps> {
             console.error(error)
```
damassi: This should be using `Utils/logger`

src/Apps/Collect/Components/Base/CollectArtworkGrid.tsx
ashfurrow: That makes sense – I don't have a strong opinion on starting subsequent comment lines with spaces, personally. We could use two spaces would make it more obvious that this is done with intention – a single space kind of jumps out to me, like it's in the uncanny valley of whitespace.

pepopowitz: Resolved with Ash's suggestion of adding _more_ space, to escape the uncanny valley.

artsyit:
:rocket: PR was released in v16.8.14 :rocket:

src/Apps/Collect/Components/Base/CollectArtworkGrid.tsx
pepopowitz: Ha, that extra space is intentional. I prefer my multi-line comments to have that indentation after the first line, so you can clearly identify paragraphs. I think it helps readability.

I'm happy to change this if you think it's a big deal, but it was done with intention.

PR #2385, opened 2019-05-08
[NavBar] Implement tracking

damassi:
Implements tracking for the new `<NavBar />` component. 

Todo: 
- [x] Wire up tracking code
- [x] Add tests

Tracking matrix: https://docs.google.com/spreadsheets/d/17oqFli15x1SmNBu_-fPZhFifMw2BNSPJe9VdOgs1aKs/edit#gid=797068217

### DIFF
--- a/src/Artsy/Analytics/Schema/Values.ts
+++ b/src/Artsy/Analytics/Schema/Values.ts
@@ -242,6 +242,7 @@ export enum ContextModule {
   Header = "Header",
   NavigationTabs = "NavigationTabs",
   RecentlyViewedArtworks = "recently_viewed_artworks",
+  HeaderMoreDropdown = "HeaderMoreDropdown",
   HeaderUserDropdown = "HeaderUserDropdown",
 
   /**
--- a/src/Artsy/Analytics/Schema/Values.ts
+++ b/src/Artsy/Analytics/Schema/Values.ts
@@ -225,6 +225,14 @@ export enum Subject {
   HistogramBar = "Histogram Bar",
   BrowseWorks = "Browse works in this category",
   QuestionMarkIcon = "Question Mark Informational Icon",
+
+  /**
+   * Header
+   */
+  NotificationBell = "Notification Bell",
+  Login = "Log In",
+  Signup = "Sign Up",
+  SmallScreenMenuSandwichIcon = "Small Screen Menu Sandwich Icon",
 }
 
 /**
@@ -234,6 +242,7 @@ export enum ContextModule {
   Header = "Header",
   NavigationTabs = "NavigationTabs",
   RecentlyViewedArtworks = "recently_viewed_artworks",
+  HeaderUserDropdown = "HeaderUserDropdown",
 
   /**
    * Artist page
@@ -305,6 +314,7 @@ export enum Flow {
   Auctions = "Auctions",
   BuyNow = "Buy now",
   MakeOffer = "Make offer",
+  Header = "Header",
 }
 
 export enum Label {
--- a/src/Components/NavBar/Menus/MoreNavMenu.tsx
+++ b/src/Components/NavBar/Menus/MoreNavMenu.tsx
@@ -1,9 +1,9 @@
 import { Menu, MenuItem } from "@artsy/palette"
-import React, { useContext } from "react"
-import { NavbarContext } from "../NavBar"
+import React from "react"
+import { useTracking } from "../Utils/useTracking"
 
 export const MoreNavMenu: React.FC = () => {
-  const { tracking, Schema } = useContext(NavbarContext)
+  const { tracking, Schema } = useTracking()
 
   const trackClick = event => {
     event.preventDefault()
--- a/src/Components/NavBar/Menus/MoreNavMenu.tsx
+++ b/src/Components/NavBar/Menus/MoreNavMenu.tsx
@@ -1,23 +1,55 @@
 import { Menu, MenuItem } from "@artsy/palette"
-import React from "react"
+import React, { useContext } from "react"
+import { NavbarContext } from "../NavBar"
 
 export const MoreNavMenu: React.FC = () => {
+  const { tracking, Schema } = useContext(NavbarContext)
+
+  const trackClick = event => {
+    event.preventDefault()
+    const link = event.currentTarget
+    const text = link.innerText
+    const href = link.getAttribute("href")
+
+    tracking.trackEvent({
+      flow: Schema.Flow.Header,
+      context_module: Schema.ContextModule.HeaderUserDropdown,
+      subject: text,
+      destination_path: href,
+    })
+  }
+
   return (
     <Menu title="More">
       {/*
         Hide nav items at md / lg as they appear in the top nav
       */}
-      <MenuItem href="/galleries" display={["block", "block", "none"]}>
+      <MenuItem
+        href="/galleries"
+        display={["block", "block", "none"]}
+        onClick={trackClick}
+      >
         Galleries
       </MenuItem>
-      <MenuItem href="/fairs" display={["block", "block", "block", "none"]}>
+      <MenuItem
+        href="/fairs"
+        display={["block", "block", "block", "none"]}
+        onClick={trackClick}
+      >
         Fairs
       </MenuItem>
-
-      <MenuItem href="/artists">Artists</MenuItem>
-      <MenuItem href="/shows">Shows</MenuItem>
-      <MenuItem href="/institutions">Museums</MenuItem>
-      <MenuItem href="https://partners.artsy.net">Artsy for Galleries</MenuItem>
+      <MenuItem href="/artists" onClick={trackClick}>
+        Artists
+      </MenuItem>
+      <MenuItem href="/shows" onClick={trackClick}>
+        Shows
+      </MenuItem>
+      <MenuItem href="/institutions" onClick={trackClick}>
+        Museums
+      </MenuItem>
+      <MenuItem href="https://partners.artsy.net" onClick={trackClick}>
+        Artsy for Galleries
+      </MenuItem>
     </Menu>
   )
 }
--- a/src/Components/NavBar/Menus/MoreNavMenu.tsx
+++ b/src/Components/NavBar/Menus/MoreNavMenu.tsx
@@ -6,50 +6,32 @@ export const MoreNavMenu: React.FC = () => {
   const { tracking, Schema } = useTracking()
 
   const trackClick = event => {
-    event.preventDefault()
-    const link = event.currentTarget
+    const link = event.target
     const text = link.innerText
-    const href = link.getAttribute("href")
+    const href = link.parentNode.parentNode.getAttribute("href")
 
     tracking.trackEvent({
-      flow: Schema.Flow.Header,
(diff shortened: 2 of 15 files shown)


artsyit:

Built with commit ef45d3bb1a5f75d29ad857f54384a91d32f12f4d

https://deploy-preview-2385--artsy-reaction.netlify.com

src/Artsy/Analytics/Schema/Values.ts
```diff
@@ -36,263 +36,134 @@ export enum OwnerType {
...
-  /**
-   * A/B Test Experiments
-   */
   ExperimentViewed = "Experiment Viewed",
-
-  /**
-   * Moving the mouse pointer over a UI element or, when browsing on a mobile
-   * device, by first tapping the UI element once making it switch into
-   * continuous hover mode.
-   */
```
damassi: I'm not one to delete comments, but these aren't useful within the context of how we work with data-team implementing analytics. They pass us a matrix, we drop the values in. What is useful however is alphabetization, so that we can quickly add a new field or see if something already exists. 

I preserved `TODO`'s as those seem important in terms of future schema work.

src/Components/NavBar/NavBar.tsx
```diff
@@ -28,18 +32,31 @@ import {
...
 
 interface NavBarProps {
-  user?: User
+  tracking?: TrackingProp
 }
 
-export const NavBar: React.FC<NavBarProps> = () => {
+export const NavbarContext = React.createContext<{
+  tracking?: TrackingProp
+  Schema?: typeof Schema
```
damassi: This `typof Schema` isn't actually enforcing anything, but I imagine we could use `Trackables` here

src/Artsy/Analytics/Schema/Values.ts
alloy: I think this is more a result of things haven gotten messy and there haven’t been time yet for the data team and us to sit down and tighten the schema. The ones with comments look right, some of the newer ones don’t (e.g. `SubmittedOffer`, which was probably a click with the action `submitted_offer`).

Please revert this for the time being.

src/Artsy/Analytics/Schema/Values.ts
```diff
@@ -36,263 +36,136 @@ export enum OwnerType {
...
-  BidFlowPlaceBid = "placeBid",
-  BidFlowSaveBillingAddress = "saveBillingAddress",
-
-  /**
-   * Generic
-   */
-  ReadMoreExpanded = "readMoreExpanded", // TODO: This differs from old event
   InSale = "In current auction", // TODO: Old schema
   InShow = "In featured show", // TODO: Old schema
+  PasswordNextButton = "passwordNextButton",
```
alloy: (Another example of what I mentioned above, this seems more like a ‘subject’.)

alloy:
As I mentioned in Slack and on the JIRA ticket, I don’t think we should be inlining analytics code into business logic. Until we settle on our next function/hook-friendly iteration, I think we should stay consistent with the rest of the code-base and use classes/decorators.

src/Components/NavBar/NavBar.tsx
damassi: Actually, `typeof Schema` did work.

src/Artsy/Analytics/Schema/Values.ts
damassi: Done

src/Artsy/Analytics/Schema/Values.ts
damassi: Not sure what this relates to -- this change is just from alpha; not part of the tracking matrix that data-team passed over. (See latest diff for Header-only related changes.)

alloy: Yeah, no, I know you didn’t add this. I meant this in relation to my comment that said:

> some of the newer ones don’t (e.g. `SubmittedOffer`, which was probably a click with the action `submitted_offer`).

src/Artsy/Analytics/Schema/Values.ts
alloy: Thanks!

src/Components/NavBar/NavBar.tsx
```diff
@@ -28,18 +32,31 @@ import {
...
   const { mediator, user } = useContext(SystemContext)
   const [showMobileMenu, toggleMobileNav] = useState(false)
   const isLoggedIn = Boolean(user)
 
+  const navBarContextValue = useMemo(() => {
+    return {
+      tracking,
+      Schema,
+    }
+  }, [])
```
alloy: Was this actually needed to avoid re-renders? Is it the `tracking` value that changes?

src/Components/NavBar/NavBar.tsx
```diff
@@ -86,7 +103,20 @@ export const NavBar: React.FC<NavBarProps> = () => {
...
+                <NavItem
+                  href="/works-for-you"
+                  Menu={NotificationsMenu}
+                  Overlay={NotificationsBadge}
+                  onClick={() => {
+                    tracking.trackEvent({
+                      flow: Schema.Flow.Header,
+                      context_module: Schema.ContextModule.Header,
+                      subject: Schema.Subject.NotificationBell,
+                      new_notification_count: cookie.get("notification-count"),
```
alloy: Is there a way to decouple this and have the `<NavBar … />` call-site pass this in somehow? It feels a bit magical and unclear where/how this comes from.

src/Components/NavBar/Menus/MoreNavMenu.tsx
```diff
@@ -1,23 +1,55 @@
...
+  const { tracking, Schema } = useContext(NavbarContext)
+
+  const trackClick = event => {
+    event.preventDefault()
+    const link = event.currentTarget
+    const text = link.innerText
+    const href = link.getAttribute("href")
+
+    tracking.trackEvent({
+      flow: Schema.Flow.Header,
```
alloy: This one only needs to be defined once on the root header component.

src/Components/NavBar/Menus/MoreNavMenu.tsx
```diff
@@ -1,23 +1,55 @@
...
   return (
     <Menu title="More">
       {/*
         Hide nav items at md / lg as they appear in the top nav
       */}
-      <MenuItem href="/galleries" display={["block", "block", "none"]}>
+      <MenuItem
+        href="/galleries"
+        display={["block", "block", "none"]}
+        onClick={trackClick}
```
alloy: Rather than defining this prop on each individual item, can’t you rely on DOM events bubbling up the tree and define the handler once on the `Menu` component? The usage of `event.currentTarget` should iirc already provide you with the actual clicked link.

src/Components/NavBar/Menus/UserMenu.tsx
```diff
@@ -11,24 +11,42 @@ import {
...
+      context_module: Schema.ContextModule.HeaderUserDropdown,
+      subject: text,
+      destination_path: href,
+    })
+  }
+
   return (
     <Menu>
-      <MenuItem href="/user/saves">
+      <MenuItem href="/user/saves" onClick={trackClick}>
```
alloy: Same re defining the handler once on the `Menu` component.

src/Components/NavBar/NavItem.tsx
```diff
@@ -15,24 +18,42 @@ export const NavItem: React.FC<NavItemProps> = ({
...
       p={1}
+      className={className}
       display={display}
       position="relative"
       style={{ cursor: "pointer" }}
       onMouseEnter={() => toggleHover(true)}
       onMouseLeave={() => toggleHover(false)}
-      onClick={onClick}
+      onClick={event => {
+        event.preventDefault
```
alloy:
```suggestion
        event.preventDefault()
```

src/Components/NavBar/NavBar.tsx
alloy: I think that wrapping the components with `track()` and importing the schema is better as it’s consistent with other places and doesn’t require the definition of a new context nor an unclear need [to a casual reader] for memoization.

src/Components/NavBar/__tests__/NavBarTracking.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    wrapper.find("NavItem").simulate("click")
+
+    expect(tracking.trackEvent).toBeCalledWith({
+      flow: Schema.Flow.Header,
+      context_module: Schema.ContextModule.Header,
+      subject: "Fairs",
+      destination_path: "/art-fairs",
+    })
+  })
+})
```
alloy: Great tests! ✨

src/Components/NavBar/NavBar.tsx
damassi: Since nothing ever changes `[]` will return the same object over and over.

src/Components/NavBar/NavBar.tsx
damassi: I agree this is a bit weird. Here's the flow: If the cookie has been set from a previous visit then it's possible to pluck this value from React's context, as the data is passed over from the server and injected into into it. But if it's a first visit, then the most direct way of getting this number is via cookie, because it's set right after relay finishes its fetch (which leaves some space open for a race condition should the user click, however unlikely, right when they land for the first time on the page.)

src/Components/NavBar/Menus/MoreNavMenu.tsx
damassi: Will try this 👍

src/Components/NavBar/NavBar.tsx
alloy: I see. Would it be possible to insert the data into the Relay store, server-side, so that either way this can always pull from the store and either get the SSR value or the newly fetched value? That should also ensure it always automatically correctly reflects the count.

damassi: Ah, that's interesting - let me think about this a bit... Seems possible!

src/Components/NavBar/NavBar.tsx
damassi: I removed the `useMemo` for now (since we're not yet concerned with changing values, and relocating tracking-specific stuff to a `useTracking` hook in [this commit](https://github.com/artsy/reaction/pull/2385/commits/f7bcc57ebcbf875efa1775ec0887d7934488fbec). In a follow up PR I'll spike on moving this pattern to SystemContext, and out of NavBar's util's folder. This will allow us to cleanly use tracking in FC components without needing to HOC wrap and import and mock.

damassi:
@alloy - I've addressed the feedback above. I have a Q about how best to handle updating the relay store with a tree of connection data in the comment above, but that can take place in a follow-up (if that's ok, and is indeed the way to go).

src/Components/NavBar/NavBar.tsx
damassi: As we're currently deriving the notification count by the number of returned edges, would a good approach be to update Metaphysics the notifications type with a new `count` field, which I can then set via `setLinkedRecord` before passing the environment into `<SystemContext>`?

artsyit:
:rocket: PR was released in v16.9.8 :rocket:

PR #2386, opened 2019-05-08
Auth footer updates

eessex:
Related to work adding reCAPTCHA to auth forms, this PR adds an optional google disclaimer text.  Currently this is only available on the signup form, and uses a prop from Force to be turned on when we are ready.

Some updates were made to the footer text generally (design point-person: Jeffrey), with some minor refactoring to replace fonts and colors from `/Assets` with their Palette equivalents. 

Props for `onTwitterLogin` have been removed, because we no longer have the links to allow login/signup with this method.

Also added `ModalType` in a few places, which kills some errors that appear when linking Reaction and Force.  

Screen Shot 2019-05-08 at 2 55 54 PM
Screen Shot 2019-05-08 at 2 54 54 PM

### DIFF
--- a/src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
+++ b/src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
@@ -69,7 +69,7 @@ export class ForgotPasswordForm extends Component<
                 handleTypeChange={() =>
                   this.props.handleTypeChange(ModalType.login)
                 }
-                mode="forgot"
+                mode={"forgot" as ModalType}
               />
             </Form>
           )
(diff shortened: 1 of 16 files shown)


artsyit:

Built with commit 87b60552c4e2039b726695b84bae843c90326648

https://deploy-preview-2386--artsy-reaction.netlify.com

eessex:
I don't see the difference in font-size you reference, as far as I can see in the code they are using the same size. Jeffrey has already approved these screenshots.

eessex:
Designs were revised to add an additional margin in response to @l2succes comments:
Screen Shot 2019-05-09 at 1 19 50 PM
Screen Shot 2019-05-09 at 1 20 03 PM

artsyit:
:rocket: PR was released in v16.9.4 :rocket:

PR #2387, opened 2019-05-08
If a user has multiple saved credit cards, returning to the Payment page on an order should automatically choose the card associated with the order

lilyfromseattle:
__Desired behavior__
As a user with multiple saved credit cards,
When I select a card on the "Payment" page of an order,
And then advance to the "Review" page and click "Change" to change my payment method...
I expect to see the card associated with the order to be selected.

__Current behavior__
Returning to the "Payment" page after clicking "Change" shows the first card as being selected.


artsyit:

Built with commit ed3b6513458b55e4a7c707e1370161a35e6878e0

https://deploy-preview-2387--artsy-reaction.netlify.com

artsyit:

Built with commit b2426a2db4be34d89299bfeb6a399ca6de793f17

https://deploy-preview-2387--artsy-reaction.netlify.com

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -197,6 +197,16 @@ export class PaymentPicker extends React.Component<
...
+    const orderCard = this.props.order.creditCard
+
+    let defaultCard
+
+    if (creditCardSelection.type === "new") {
+      defaultCard = "new"
+    } else {
+      defaultCard = orderCard ? orderCard.id : creditCardSelection.id
+    }
+
```
ds300: This looks good, but I think it makes more sense to put this logic in the `getInitialCreditCardSelection` method near the top of this file.

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -199,12 +203,18 @@ export class PaymentPicker extends React.Component<
...
     const orderCard = this.props.order.creditCard
 
-    let defaultCard
+    const creditCardsArray: any = []
 
-    if (creditCardSelection.type === "new") {
-      defaultCard = "new"
-    } else {
-      defaultCard = orderCard ? orderCard.id : creditCardSelection.id
+    creditCards.edges.map(e => {
```
ds300: You could do this when initializing the array:

`const creditCardsArray = creditCards.edges.map(e => e.node)`

That way it also retains the type information instead of making it an `any`

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -197,6 +201,22 @@ export class PaymentPicker extends React.Component<
...
+
+    const creditCardsArray: any = []
+
+    creditCards.edges.map(e => {
+      creditCardsArray.push(e.node)
+    })
+
+    if (
+      orderCard != null &&
+      creditCardsArray.filter(card => card.id === orderCard.id).length === 0 &&
```
ds300: You can use the new array method `some` for this :)

```suggestion
      creditCardsArray.some(card => card.id === orderCard.id) &&
```

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -197,6 +201,22 @@ export class PaymentPicker extends React.Component<
...
+      orderCard.brand
```
ds300: Would be worth adding a comment about why this is required.

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -404,6 +424,10 @@ export const PaymentPickerFragmentContainer = createFragmentContainer(
           state
           country
           postal_code
+          expiration_month
+          expiration_year
+          last_digits
+          brand
```
ds300: Unrelated to this PR, A nice follow-up PR might be to move the radio buttons inside a fragment container so that the main component doesn't need to care about these fields.

src/Apps/Order/Components/__tests__/PaymentPicker.test.tsx
```diff
@@ -370,22 +376,69 @@ describe(PaymentPickerFragmentContainer, () => {
         expiration_year: 2019,
       },
     ]
-    function getPage(_cards: typeof cards) {
```
ds300: weird diff, github :duck:

src/Apps/Order/Components/PaymentPicker.tsx
lilyfromseattle: For some reason some isn't working for me here. We can either merge as is or pair on this on Monday if you're up for it.

ds300: :grimacing: That's my bad, in the suggestion it should have been negated, i.e.

```ts
      !creditCardsArray.some(card => card.id === orderCard.id) &&
```

But it's no biggie feel free to keep as-is.

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -205,10 +205,10 @@ export class PaymentPicker extends React.Component<
 
     const creditCardsArray = creditCards.edges.map(e => e.node)
 
-    // # only add the unsaved card to the cards array if its not already there and has the necessary properties
+    // only add the unsaved card to the cards array if its not already there and has the necessary properties
```
ds300: In which situations would the credit card not have the necessary properties?

ds300:
@lilyfromseattle there's some failing tests here now. Need to update the `createCreditCard` mock fixtures to include the new fields being selected.

artsyit:
:rocket: PR was released in v16.12.20 :rocket:

PR #2391, opened 2019-05-09
hides new ads behind admin and env feature flags

ashleyjelks:
This PR adds additional checks to the externally served ads so they are hidden behind a feature flag. Ads are now made viewable to all admins and never render in a production environment.


artsyit:

Built with commit 06bbe1a9457b4024d36f614f4ab57ec0810c3952

https://deploy-preview-2391--artsy-reaction.netlify.com

src/Components/Publishing/Ads/AllowedUsers.ts
```diff
@@ -0,0 +1,29 @@
...
+ *  2) When a user is an Admin
+ *  3) When a user has been added to the allowlist set via HASHTAG_LAB_ADS_ALLOWLIST
+ */
+
+export const getAllowedUsers = () => {
+  const allowedUsers = (sd.HASHTAG_LAB_ADS_ALLOWLIST || "")
+    .split(",")
+    .filter(Boolean)
+  const currentUser = get(sd, "CURRENT_USER.email", "")
+  const env = sd.HASHTAG_LAB_ADS_ENABLED
```
damassi: This is a bit ambiguous naming-wise; things like "enabled", "disabled", "on", "off" all all reduced in practice to booleans values -- true / false. 

Since ENV vars are set on the server _per environment_ it's totally possible to set this to  `true` on staging and `false` production, and then simplify things down below with 

```js
if (sd.HASHTAG_LAB_ADS_ENABLED) {
  ...
}
```

src/Components/Publishing/Ads/AllowedUsers.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (env === "production") {
+    return false
+  }
+  if (allowedUsers.includes(currentUser)) {
+    return true
+  }
+  if (isAdmin) {
+    return true
+  }
+  return false
```
damassi: The return-types here, combined with the function name, is also a bit ambiguous language-wise. Thinking: similar to the comment above, it's common to encode the type of data into the name of something -- so

 ```js
const isEnabled = true // => true / false == boolean 
```
or
```js
const username = 'chris' // name => string
```
So thinking about the name of the function `getAllowedUsers`, it encodes a few assumptions:
1) `get` -> return something
1) `Allowed` -> there's some sort of filtering logic going on internally
1) `Users` -> a collection of things

Since it's only returning a single boolean value, the name should reflect that: something like `isUserAllowed` (it's common to prefix things that are booleans with "is".)

damassi: Also, another possibility for evaluating a set of conditions is the array method [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some). (Though the use of `if` conditions is totally fine and clear!):

```js
const isAllowed = [
  isAdmin,
  sd.HASHTAG_LAB_ADS_ENABLED,
  allowedUsers.includes(currentUser),
].some(condition => condition === true) // or .some(Boolean)

return isAllowed
```

src/Components/Publishing/Display/Canvas/index.tsx
```diff
@@ -56,19 +56,13 @@ export class DisplayCanvas extends React.Component<DisplayCanvasProps> {
...
     const disclaimer = (
       <Disclaimer layout={unit.layout}>{unit.disclaimer}</Disclaimer>
     )
 
     // TODO: Remove the allowlist and env checks after externally served ads are implemented
-    if (
-      allowedUsers.includes(currentUser) &&
-      process.env.NODE_ENV !== "production"
-    ) {
+    if (isHTLAdEnabled) {
```
damassi: Maybe just `if (isHTLAdEnabled()) { ... }`?

src/Components/Publishing/Display/__tests__/DisplayCanvas.test.tsx
```diff
@@ -131,6 +132,11 @@ describe("unit", () => {
     ])
   })
 
+  it("checks for enabled ads", () => {
+    const isHTLAdEnabled = getAllowedUsers()
+    expect(isHTLAdEnabled).toBe(false)
```
damassi: `expect(isHTLAdEnabled()).toBe(false)`

src/Components/Publishing/Ads/EnabledAd.ts
```diff
@@ -0,0 +1,27 @@
+import { get } from "lodash"
+import { data as sd } from "sharify"
+
+/**
+ * For now ads, served by Hashtag Labs are only to viewable in the following instances:
+ *  1) When the environment is NOT Production (HASHTAG_LAB_ADS_ENABLED set to false)
+ *  2) When a user is an Admin
+ *  3) When a user has been added to the allowlist set via HASHTAG_LAB_ADS_ALLOWLIST
+ */
```
damassi: Thanks for adding the comment describing the flow 👍

src/Components/Publishing/Ads/EnabledAd.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const isHTLAdEnabled = () => {
+  const allowedUsers = (sd.HASHTAG_LAB_ADS_ALLOWLIST || "")
+    .split(",")
+    .filter(Boolean)
+  const currentUser = get(sd, "CURRENT_USER.email", "")
+  const isAdminUser =
+    get(sd, "CURRENT_USER.type", "") === "Admin" ? true : false
+  const isAllowedUser = allowedUsers.includes(currentUser)
```
damassi: #minor - since `isAllowedUser`'s value is derived from `currentUser`, might want to swap places with `isAdminUser` so that it sits just underneath where `currentUser` is defined and things read more linearly.

src/Components/Publishing/Display/__tests__/DisplayCanvas.test.tsx
```diff
@@ -131,6 +132,10 @@ describe("unit", () => {
     ])
   })
 
+  it("checks for enabled ads", () => {
+    expect(isHTLAdEnabled()).toBe(false)
+  })
```
damassi: Rather than test inside of `DisplayCanvas.test`, it would be better to create a `isHTLAdEnabled.test` file and test expectations there. (Which will also allow for the testing of a `toBe(true)` case.)

src/Components/Publishing/Display/__tests__/DisplayPanel.test.tsx
```diff
@@ -106,6 +107,10 @@ describe("units", () => {
       ])
     })
 
+    it("checks for enabled ads", () => {
+      expect(isHTLAdEnabled()).toBe(false)
+    })
```
damassi: Same here -- this can be moved into it's own test file.

What should be tested inside of `DisplayPanel` is that an ad is injected, and that insertion works correctly.

src/Components/Publishing/Ads/EnabledAd.ts
```diff
@@ -0,0 +1,27 @@
...
+ *  1) When the environment is NOT Production (HASHTAG_LAB_ADS_ENABLED set to false)
+ *  2) When a user is an Admin
+ *  3) When a user has been added to the allowlist set via HASHTAG_LAB_ADS_ALLOWLIST
+ */
+
+export const isHTLAdEnabled = () => {
+  const allowedUsers = (sd.HASHTAG_LAB_ADS_ALLOWLIST || "")
+    .split(",")
+    .filter(Boolean)
+  const currentUser = get(sd, "CURRENT_USER.email", "")
```
damassi: I think it's ok to add `CURRENT_USER` value to `sharify.d.ts` for now. When this test is removed we can delete it.

damassi:
```js
import { get } from 'Utils/get'

const currentUserEmail = get(sd, => sharifyData.CURRENT_USER.email, '')
```

ashleyjelks: Can I do this in a separate PR?

ashleyjelks: Trying to get these ads available on Staging so they can be tested by the folks at HTL.

damassi: Yup totally 👍

src/Components/Publishing/Ads/__tests__/isHTLAdEnabled.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(allowedUser).toHaveLength(2)
+    expect(allowedUser).toEqual([
+      "alloweduser@email.com",
+      "alloweduser2@email.com",
+    ])
+  })
+
+  it("checks for enabled ads", () => {
+    expect(isHTLAdEnabled()).toBe(false)
+  })
```
damassi: Can we add a case where this is `true`?

artsyit:
:rocket: PR was released in v16.9.5 :rocket:

damassi:
(Merged, but lets loop back around to the tests once the hashtag folks see staging)

PR #2392, opened 2019-05-09
DISCO-1004: Preserve pagination in browser history on artist page

pepopowitz:
This PR addresses https://artsyproduct.atlassian.net/browse/DISCO-1004. It solves 3 problems with artist pagination:

1. Individual artist pages (e.g. /artist/kaws) were not respecting the `page` querystring variable, so using the Back/Forward buttons in your browser would have unpredictable results when you used the pager.
2. It cleans up the URLs that end up in browser history - removing any default values for filters that weren't actually selected by a user.
3. We were using `mediator` to handle pushing to browser history; this PR shifts that work into Reaction, instead.

This work is based on work from #2369 and #2384, with a slightly different direction. In #2369/#2384 we are depending on the [child components to explicitly push a URL to the browser history](https://github.com/artsy/reaction/pull/2369/files#diff-9f168adbb6f7152f905d1a0f67188475R68). In this PR, we avoid forcing the child components to call the helper, and instead our state management component pushes URLs to history whenever the state changes. This feels like the history management is in the right place - consolidated alongside the state management, instead of scattered throughout the App. Depending on how people feel about this, I'd definitely be up for going back and modifying the work from those previous PR's to match this approach.

There is one thing in this PR I'm not totally thrilled with. Because of a possible bug in a dependency (rc-slider) where the `onAfterChange` event gets fired twice every time you interact with the PriceRangeFilter, we added code to ensure the URL has actually changed whenever we attempt to push to the browser history. I'll point this bit out with an inline comment.


src/Apps/Artist/Routes/Overview/state.ts
```diff
@@ -40,6 +41,48 @@ export const initialState = {
   showZeroState: false,
 }
 
+// Returns a string representing the query part of a URL.
+// It removes default values.
+export const urlFragmentFromState = (state: State, extra?: Partial<State>) => {
```
pepopowitz: This function, and `isDefaultFilter` just below it, are really similar to functions in the Collect/Search Apps. I think the duplication is correct - while the functions are similar, they aren't identical, and each App seems like it might need to do each of these things slightly differently.

artsyit:

Built with commit f3dabb884d616c5c42bbd47b1432dc2a9456b8ca

https://deploy-preview-2392--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/state.ts
```diff
@@ -71,15 +114,30 @@ export class FilterState extends Container<State> {
     }
   }
 
-  setPage(page, mediator) {
+  previousQueryString = ""
+  pushHistory() {
+    const currentQueryString = urlFragmentFromState(this.state)
+    // PriceRangeFilter's onAfterChange event fires twice; this ensures
+    //   we only push that history event once.
+    if (this.previousQueryString !== currentQueryString) {
```
pepopowitz: The comment pretty well describes it, but here's where we have to maintain the previous query string, so we can compare the current one. Without this check, we end up with duplicate history items when you interact with the Price Range slider. I _believe_ this is because the `rc-slider` dependency is unexpectedly firing an event twice, but I didn't chase it down much further than that. This felt like a reasonable workaround to me.

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/PriceRangeFilter.tsx
```diff
@@ -21,7 +19,7 @@ export const PriceRangeFilter: React.FC<{
         const minStr = min === FilterState.MIN_PRICE ? "*" : min
         const maxStr = max === FilterState.MAX_PRICE ? "*" : max
 
-        filters.setFilter("price_range", `${minStr}-${maxStr}`, mediator)
```
damassi: Nice cleanup getting rid of mediator throughout.

pepopowitz: I plan to go back and do the same for the aforementioned PR's.

src/Apps/Artist/Routes/Overview/state.ts
```diff
@@ -71,15 +114,30 @@ export class FilterState extends Container<State> {
...
+      window.history.pushState(
+        {},
+        null,
+        `${window.location.pathname}?${currentQueryString}`
+      )
+      this.previousQueryString = currentQueryString
+    }
+  }
```
damassi: Should `pushHistory` exist on the `State` container? Since it's not explicitly manipulating state, seems like it could exist just outside of the class (similar to `urlFragmentFromState`), and below be called without need for `this`.

damassi: This def does seem like the right approach though -- keeps the little engine spinning outside of view 😄

pepopowitz: The advantages we get out of putting it there are (1) we don't need to pass in `this.state` every time we call it (admittedly a #minor advantage), and (2) to [avoid double-stuffing the browser history](https://github.com/artsy/reaction/pull/2392/files/f3dabb884d616c5c42bbd47b1432dc2a9456b8ca#r282675767), `pushHistory` needs to maintain state of what it just pushed to the history.

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/PriceRangeFilter.tsx
damassi: Unrelated (but also very much related), it would be so nice to consolidate all of our `ArtworkFilters` into one; Matt and I have talked about this a bunch as we now have three duplicated instances in the Reaction.

src/Apps/Artist/Routes/Overview/state.ts
pepopowitz: Er...point 2 is incomplete. It's an advantage in the sense that we'd need to make `pushHistory` maintain state _anyway_, so why not use the mechanism we already have to maintain state?

damassi: Sounds good!

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/PriceRangeFilter.tsx
pepopowitz: That sounds like a fun thing to pair on some time. 😃🍐

src/Apps/Artist/Routes/Overview/state.ts
```diff
@@ -40,6 +41,48 @@ export const initialState = {
...
+const isDefaultFilter = (filter, value): boolean => {
+  if (filter === "major_periods" || filter === "attribution_class") {
+    return value.length === 0
+  }
+
+  if (filter === "sort") {
+    return value === "-decayed_merch"
+  }
+
+  if (filter === "price_range" || filter === "height" || filter === "width") {
```
ashfurrow: For code like this, where we're checking to see if one variable (`filter`) is one of a few options, I'd recommend something like this:

```js 
if (["price_range", "height", "width"].includes(filter)) {
```

But that also breaks parallelism with the other `if` statements, hmm... 

I'm fine with this as-is. A `switch` might be an idiomatic solution, but I'll leave it up to you 👍

src/Apps/Artist/Routes/Overview/state.ts
ashfurrow: This code, and the comment, makes total sense – thanks for the clarity! Does this appear to be a bug in `rc-slider`? Could we open an issue reporting it to the authors? Or, if one already exists, could we link to it from this comment?

pepopowitz: I am not totally sure. I _suspect_ that it is, but I haven't taken the time to isolate the actual issue. 

There are a couple issues in the rc-slider repo that are closed but have comments on them indicating they are not actually fixed ([466](https://github.com/react-component/slider/issues/466) and [361](https://github.com/react-component/slider/issues/361)). I suspect those are related but again...I need to invest more time to isolate the issue.

src/Apps/Artist/Routes/Overview/state.ts
pepopowitz: I agree that there are probably ways to cut down the number of lines of code in this function, but I'm not personally offended by its current state. I find it pretty readable, and I think trying to remove the `if` statements would result in some trickery that required more thinking to understand. I'm planning to leave this as-is.

artsyit:
:rocket: PR was released in v16.9.9 :rocket:

PR #2393, opened 2019-05-09
Add reCAPTCHA events to Auth forms

eessex:
**HOLD: This PR depends on https://github.com/artsy/force/pull/4015/files**
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-1391

- Adds global `grecaptcha` to window
- Adds `sd.RECAPTCHA_KEY`, already set in staging and production
- Fires `grecaptcha` actions when submitting auth forms
- Updates tests to cover all steps in auth wizard

This PR adds reCAPTCHA actions to all auth forms. We are not verifying or blocking based on these captcha events, the current implementation fires an action, creating a score for each user and sends it to our analytics dashboard: https://www.google.com/u/0/recaptcha/admin/site/346113690

One open question is how to name these actions. Because we are firing these on form submission, I used the names `login_submit`, `signup_submit` and `forgot_submit`.  This gives us the option to retain the existing event names when we extend the action to use server validation. Another option would be to fire these on loading the form, and fire something named for an impression. I'm open to all suggestions, perhaps @cavvia or the data team would have suggestions on this front.


artsyit:

Built with commit de3bf734fbe24fc7f1a99120c16d8790f12c23f7

https://deploy-preview-2393--artsy-reaction.netlify.com

src/Components/Authentication/Desktop/LoginForm.tsx
```diff
@@ -26,11 +27,18 @@ export class LoginForm extends Component<FormProps, LoginFormState> {
     error: this.props.error,
   }
 
+  onSubmit = (values: InputValues, formikBag: FormikProps<InputValues>) => {
+    window.grecaptcha.ready(() => {
+      window.grecaptcha.execute(sd.RECAPTCHA_KEY, { action: "login_submit" })
+      this.props.handleSubmit(values, formikBag)
+    })
```
damassi: It seems like maybe these grecaptcha calls can live inside of a util function: 

```js
export function showRecaptcha(action: string) {
  window.grecaptcha.ready(() => {
    window.grecaptcha.execute(sd.RECAPTCHA_KEY, { 
      action
    })
  })
}
```

And then functionality tests can be localized here, while simple assertions about whether `showRecaptcha` has been called take place elsewhere.

src/Components/Authentication/__tests__/Desktop/ForgotPasswordForm.test.tsx
```diff
@@ -2,55 +2,77 @@ import { ForgotPasswordForm } from "Components/Authentication/Desktop/ForgotPass
...
-    formik.submitForm()
+      setTimeout(() => {
+        expect(props.handleSubmit).toBeCalledWith(
+          {
+            email: "foo@bar.com",
+          },
+          formik.getFormikActions()
+        )
+        done()
+      })
```
damassi: Something like
```js
import { showRecaptcha } from 'Utils/showRecaptcha'
...
jest.mock('Utils/showRecaptcha')
...
expect(showRecaptcha).toHaveBeenCalledWith(...)
```

joeyAghion:
You said that this depends on https://github.com/artsy/force/pull/4015, but isn't it the other way around? I'm also not quite sure what's meant by:

> These are not captcha events, the current implementation fires an action...

Code looks 👍 from what I can tell.

One other question: unsetting the `RECAPTCHA_KEY` value will be a convenient way to enable or disable this feature. Would Force or Reaction tolerate a missing key, or would that break things and require a revert to disable?

src/Components/Authentication/__tests__/Desktop/ForgotPasswordForm.test.tsx
eessex: The async behavior relates to the Formik submit rather than the recaptcha functionality.

eessex:
This PR has been updated to move `grepcaptcha` calls into the `/Utils` folder, and add an enum typing allowed action names. 

The description has been updated for clarity regarding @joeyAghion's comment. We are firing captcha events, but we are not currently making any evaluations of the user based on the result, so actual auth actions still proceed as before.

Regarding artsy/force#4015, this PR turns on the google terms and conditions text, which I want to leave off until this is merged and we are firing actions from those forms.

As far as missing `sd.RECAPTCHA_KEY`, this fails gracefully at the moment -- if it is not present, the `window.recaptcha.execute` action will fail and console.log the error, but login will proceed as usual.

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
```diff
@@ -23,11 +24,16 @@ export class ForgotPasswordForm extends Component<
     error: this.props.error,
   }
 
+  onSubmit = (values: InputValues, formikBag: FormikProps<InputValues>) => {
+    Grepcaptcha(GrepcaptchaAction.ForgotSubmit)
```
damassi: This capitalization is a bit unidiomatic. Wondering why?

eessex: Ah yes I see now that only components use capitalization while looking closer... wasn't totally clear on the rules, at a glance most things look capitalized in this repo.

damassi: Maybe something like 

```ts
triggerRecaptcha(RecaptchaAction.ForgotSubmit)
```

Another thing is, with TS importing and using enums in this way isn't really necessary due to the type-checker. Something like this: 

```
type RecaptchaAction = 'foo' | 'bar' | 'baz | 'bam'


function triggerRecaptcha(action: RecaptchaAction) {
  ..
}
```
As soon as the user types the opening paren of the function intellisense will be triggered, showeing the available options according to the `type`. This will simplify things. 

In normal JS this imported / exported enum pattern would work nicely! But in TS it's not necessary.

src/Components/Authentication/Desktop/LoginForm.tsx
```diff
@@ -16,6 +16,7 @@ import PasswordInput from "Components/PasswordInput"
 import QuickInput from "Components/QuickInput"
 import { Formik, FormikProps } from "formik"
 import React, { Component } from "react"
+import { Grepcaptcha, GrepcaptchaAction } from "Utils/Grepcaptcha"
```
damassi: Can probably drop the "g" from the recaptcha -- seems to be encoding implementation details ("recaptcha" is a universal thing, g + recaptcha is some other thing, plus the universal). If we wanted to use some other recaptcha implementation, we should be able to swap it out silently in the background. ("grecapta" is also a bit weird to type 😄)

src/Components/Authentication/__tests__/Desktop/ForgotPasswordForm.test.tsx
damassi: No prob 👍

src/Components/Authentication/__tests__/Mobile/SignUpForm.test.tsx
```diff
@@ -16,14 +27,120 @@ describe("MobileSignUpForm", () => {
...
+              "recaptcha-api-key",
+              {
+                action: "signup_submit",
+              }
+            )
+            done()
+          })
+        })
+      })
+    })
```
damassi: @eessex - have you explored Jest's async / await / promise features? (`resolves`, etc.) All of these setTimeouts seem like a bit of a flag, (though totally understand needing to use them from time to time).

eessex: Yes I am familiar with the api but as stated above, the async behavior is not related to `recaptcha`. I'm wary to mock the actual form behavior because I don't know it very well and could not see an obvious place that the async originates, but did add a comment that this should be fixed.  As far as the tests I fixed the ones that failed, and extended the existing ones to make sure we got to the last step where the captcha is called.

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
```diff
@@ -25,7 +25,7 @@ export class ForgotPasswordForm extends Component<
   }
 
   onSubmit = (values: InputValues, formikBag: FormikProps<InputValues>) => {
-    Grepcaptcha(GrepcaptchaAction.ForgotSubmit)
+    repcaptcha("forgot_submit")
```
damassi: 👍 nice and simple, while also safe. `snake_case` is kinda ugly but /shrug, things are 1:1.

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
eessex: Changed this to a string to avoid over-engineering at the moment. We have discussed the idea of firing actions in many contexts, both for validation and for analytics purposes -- for example, during purchase and bid flows. As our actions and contexts grow I think we should revisit this, I could see a use of enum's similar to Analytics being useful.

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
eessex: Snake case is required by google's API.

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
damassi: Once we need to scale things out lets def revisit 👍

src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
damassi: cool yeah, that's what I figured

artsyit:
:rocket: PR was released in v16.11.0 :rocket:

PR #2405, opened 2019-05-10
Sketch out new hooks-friendly tracking pattern

damassi:
Sketching out a possible pattern for dealing with react-tracking in the hooks era. 

**NOTE:** This is meant as a small iteration to enable the forward-progression of hooks by not forcing users onto classes due to tracking. Other ideas can and should be explored.

#### General Overview: 

- Tracking is managed via `TrackingContext`. Rather than importing `track` for every file and having to wrap things with an HOC and needing to stub out instances for tests, we can pull things from a context, via a hook. This simplifies things across the board.
- Since react-tracking calls are composable, we need a way to register a new tracking instance for a given tree. How this happens is we wrap the top of the tree with `provideTracking`, which injects `trackEvent` into our context:

```tsx
import { provideTracking, useTracking, AnalyticsSchema } from 'Artsy/Analytics'

const App = provideTracking({
  flow: AnalyticsSchema.Flow.Header,
})(props => {

  const trackClick = () => {
    trackEvent({
      ...
    })
  }

  return (
    <div onClick={() => trackClick()}>
      ...
    </div>
  )
})
```

Once injected into the context, child components can pull it right out:

```tsx
const Child = props => {
  ...
}
```

#### Types:

In general, props passed into the `trackEvent` function will be validated against `Trackables`, which is a collection of all of the possible properties that we can dispatch to our tracking service. 

If there's an instance where a particular piece of trackable data doesn't belong in the schema, we can wire up one-off validations via the `AnalyticsProp` type:

```ts
import { AnalyticsProp } from 'Artsy/Analytics' 

trackEvent<AnalyticsProp<{ someRandomProp: string }>>({ 
  someRandomProp: "foo" 
})
```

#### Testing:

Testing is as easy as mocking `useTracking` and `TrackingContext`: 

```tsx
import { useTracking } from 'Artsy/Analytics/useTracking' 

jest.mock("Artsy/Analytics/useTracking")
jest.mock("Artsy/Analytics/TrackingContext", () => {
  return {
    provideTracking: () => Component => Component,
  }
})

describe('SomeTest', () => {
  const trackEvent = jest.fn() 
 
  beforeEach(() => {
    useTracking.mockImplementation(() => {
      return tracking
    }) 
  })

  afterEach(() => jest.clearAllMocks())

  it('should track', () => {
    ...
    expect(tracking).toHaveBeenCalledWith(...)
  })
})
```

Feedback appreciated!


src/Artsy/SystemContext.tsx
```diff
@@ -28,18 +36,20 @@ export interface SystemContextProps {
    */
   relayEnvironment?: Environment
 
+  /**
+   * An instance of react-tracking. When
+   */
+  tracking?: TrackingProp
+  initTracking?: (tracking: TrackingProp) => void
```
damassi: A bit unsure of the naming here; this function is also _only_ called from within `useTracking()`, so it's effectively private.

src/Artsy/SystemContext.tsx
```diff
@@ -19,6 +22,11 @@ export interface SystemContextProps {
    */
   mediator?: Mediator
 
+  /**
+   * FIXME: Ask alloy how to pass one-off props like this in from force
+   */
```
damassi: (Ignore this `FIXME`; spoke over slack today and going to investigate something in another PR.)

src/Utils/hooks/useTracking.ts
```diff
@@ -0,0 +1,24 @@
...
+  newTrackingInstance?: TrackingProp
+): {
+  tracking: TrackingProp
+  AnalyticsSchema: typeof AnalyticsSchema
+} {
+  const { tracking, setTracking } = useContext(SystemContext)
+
+  useEffect(() => {
+    if (newTrackingInstance) {
+      setTracking(newTrackingInstance)
```
ds300: What if another subtree with a different base tracking prop is registered under the same system context provider? Won't that destroy the context for the parent/sibling trees?

src/Artsy/SystemContext.tsx
```diff
@@ -57,11 +66,17 @@ export const SystemContextProvider: SFC<SystemContextProps> = ({
...
+
+  const providerValues = useMemo(() => {
+    return {
+      ...props,
+      relayEnvironment,
+      setTracking,
+      tracking,
+      user,
+    }
+  }, [props.relayEnvironment, props.user, props.tracking])
```
ds300: why only these three values? Shouldn't any context value change cause cascading updates?

src/Components/NavBar/NavBar.tsx
```diff
@@ -2,11 +2,24 @@ import cookie from "cookies-js"
...
+  SoloIcon,
+  Spacer,
+} from "@artsy/palette"
+
 import { SystemContext } from "Artsy/SystemContext"
 import { SearchBarQueryRenderer as SearchBar } from "Components/Search/SearchBar"
 
 import { track } from "Artsy/Analytics"
-import * as Schema from "Artsy/Analytics/Schema"
+import * as AnalyticsSchema from "Artsy/Analytics/Schema"
```
ds300: :+1: good name change

src/Utils/hooks/useTracking.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }, [])
+
+  return {
+    tracking,
+    AnalyticsSchema,
```
ds300: I'm a bit concerned about passing the `AnalyticsSchema` as a prop here.  Might seem convenient, but it's only an import away, and the TS language server can do auto importing so it's even easier than destructuring the return value of `useTracking` if you lean into intellisense.

Also, importing `AnalyticsSchema` will still be required when wrapping things with `track` so having multiple ways to get at those values might create confusion. Especially when the destructuring approach implies dependency injection, which is not happening.

Personally I'd rather see a single named export for AnalyticsSchema (rather than the many options we have right now) which makes the auto import trivial, while also making it so there's one standard way to access analytics values 💆

src/Utils/hooks/useTracking.ts
ds300: I think what I'd suggest as an alternative to this would be to create our own version of `track` which injects a new context value (not kept in the system context). Then we also don't need to faff with the `useTracking(props.tracking)`

Something along these lines (just a rough sketch):

```ts
import { track as _track } from "react-tracking"

const track = trackingProps => Component =>
  _track(trackingProps)(props => (
    <ProvideTrackingContext trackingProps={props.tracking}>
      <Component {...props} />
    </ProvideTrackingContext>
  ))

```

And then useTracking would hook into the provided tracking context if present.

alloy:
I was under the impression that you were going to experiment with [a different type of pattern](https://artsy.slack.com/archives/CC56D129X/p1557430212058000) where analytics code would _not_ be inlined? I think we should focus on the design before implementation.

For instance; a pie in the sky design of something I imagined, based on the direction we were talking about, is:

```tsx
// ArtistHeader.tsx

const { track } from "Analytics"

export const handleFollowArtist = (event, artistID) => {
  // Perform API request...
}

// `useTracking` is injected as a prop here so that invocations of it can capture `props`
const ArtistHeader = track((({ useTracking, ...props }) => {
  const followArtist = useTracking(handleFollowArtist)

  return <div onClick={e => followArtist(e, props.artist.id)}>Follow Picasso</div>
})

export const ArtistHeaderContainer(
  ArtistHeader,
  {
    artist: graphql`
      fragment ArtistHeader_artist on Artist {
        id
        ... AnalyticsSchema_artist
      }
    `
  }
)
```

```tsx
// AnalyticsSchema.ts (or co-located)

import { registerTrack as track } from "Analytics"
import { handleFollowArtist } from "ArtistHeader"

// Types of function parameters are inferred by TS from the given function.
// `props` is the props passed to the component, except `artist` could be overwritten by the one defined by the fragment below.
track(
  handleFollowArtist,
  (props, _event, _artistID) => ({
    type: Schema.Type.Click,
    action: Schema.Actions.FollowArtist
    entity_id: props.artist.internalID,
  }),
  graphql`
    fragment AnalyticsSchema_artist on Artist {
      internalID
    }
  `
)
```

alloy:
I think I actually like the thought of co-locating in the same file more:

```tsx
const handleFollowArtist = (event: ClickEvent, artistID: string) => { … }

const ArtistHeader = track((({ useTracking, ...props }) => {
  const followArtist = useTracking(handleFollowArtist)
  return <div onClick={e => followArtist(e, props.artist.id)}>Follow Picasso</div>
})

export const ArtistHeaderContainer(
  ArtistHeader,
  {
    artist: graphql`
      fragment ArtistHeader_artist on Artist {
        id
        ... ArtistHeader_track_artist
      }
    `
  }
)

registerTrack(
  handleFollowArtist,
  (props, _event, _artistID) => ({
    type: Schema.Type.Click,
    action: Schema.Actions.FollowArtist
    entity_id: props.artist.internalID,
  }),
  graphql`
    fragment ArtistHeader_track_artist on Artist {
      internalID
    }
  `
)
```

src/Utils/hooks/useTracking.ts
pepopowitz: I like this suggestion. It would mean less steps to set up tracking in a component before we can actually call `track.trackEvent`, and as a consumer of `useTracking`, that makes me feel more confident that I'm not going to forget something.

damassi: @ds300 - I like it 👍  That would solve a few issues quite elegantly.

src/Artsy/SystemContext.tsx
damassi: These are the only dynamic values that could change and should trigger updates.

src/Utils/hooks/useTracking.ts
damassi: The main thought is convenience. There are multiple ways we handle tracking in Reaction, and this just makes things a little quicker without much possibility of causing harm. We're already returning `tracking` from `useTracking`, why not also pass along the schema. I think the chance of confusion is fairly small.

damassi:
@ds300, @pepopowitz - thoughts on https://github.com/artsy/reaction/pull/2405#issuecomment-491821837? 

More primarily, and for the sake of discussion, I'm curious also about y'alls opinions on the need for ^, vs something uncomplicated like 

```js
const { useTracking } from 'Utils/Hooks/useTracking'

const SomeChild = () => {
  const { tracking, AnalyticsSchema } = useTracking()

  const trackClick = () => {
    tracking.trackEvent({
      type: AnalyticsSchema.Type.Click
    })
  }

  return <div onClick={trackClick}>hi</div>
}
```

I always worry about over-engineering things while also admitting that there's a lot that's over my head in terms of understanding, and that I'm likely missing something. Having used our current tracking implementation quite extensively however, I haven't experienced much value other than the `Schema` file structure, which allows for intellisense. All of the type-safety in the current implementation comes down to `typeof Schema` in this implementation, which prevents folks from using unsafe values. (Again, I'm probably doing something wrong!) My personal experience is the inspiration for trying to keep things plain and simple. 

That said, I like @alloy's sketch, it makes sense and doesn't seem like too much overhead. It would be good to get more consensus around what we're trying to solve before I go further, though.

alloy:
(We have plenty of cases where we’re fetching additional data purely for analytics purposes, eg in some places we fetch a slug and a db ID)

damassi:
Oh right - what I mean is I have yet to see something that would necessitate anything more complex than adding an additional query prop to an existing component query, and then passing that into the tracking call. Meaning, 

```
// Before 
graphql`
  fragment Foo on Bar {
    name
    description
  }
`

// after 
graphql`
  fragment Foo on Bar {
    _id
    name
    description
  }
`
```

alloy:
Ahh, right! Well, if we co-locate inside the same file then an extra fragment should not be needed—that was more for if they would reside in different files.

pepopowitz:
@damassi I like the API in https://github.com/artsy/reaction/pull/2405#issuecomment-491934561 a lot. It's obvious to me what all the pieces are that I'm hooking up, and it makes sense why I need to hook them up.

We will `useTracking` many many times throughout the code, so for me it's most important that there are fewer things I have to hook up each time, and that I understand why I'm doing it. Otherwise I'll revert to cargo-culting and hope that I get it right (and then inevitably get it wrong).

ds300:
For me https://github.com/artsy/reaction/pull/2405#issuecomment-491934561 is preferable :+1:

I find it significantly easier to understand, and I suspect it would also be easier to use and read in real code.

src/Utils/hooks/useTracking.ts
ds300: I think your idea of convenience is different from my idea of convenience :)

src/Components/NavBar/NavBar.tsx
```diff
@@ -109,8 +97,8 @@ export const NavBar: React.FC<NavBarProps> = track<NavBarProps>({
                   Menu={NotificationsMenu}
                   Overlay={NotificationsBadge}
                   onClick={() => {
-                    tracking.trackEvent({
-                      subject: Schema.Subject.NotificationBell,
+                    tracking.trackEvent<{ new_notification_count: string }>({
```
damassi: For one-off tracking props that aren't a part of the AnalyticsSchema we can pass a generic

typings/react-tracking.d.ts
```diff
@@ -7,8 +7,8 @@
 declare module "react-tracking" {
   import * as React from "react"
 
-  export interface TrackingProp {
-    trackEvent({}): any
+  export interface TrackingProp<P = {}> {
+    trackEvent<T = {}>({  }: Partial<P> & T): any
```
damassi: This locks down what we can pass into `trackEvent`. The general flow (most of which the end-user won't touch, is like so: 

In the tracking context during overall app setup, pass our trackables into a (new) generic argument on `react-tracking`'s `TrackingProp`:
```
export interface TrackingContextProps {
  tracking?: TrackingProp<Trackables>
}
```
This constrains, in general, what we can pass into the function. 

Then later, in the rare occurrence when there's a one-off prop that doesn't belong in Trackables, the end-user can pass it in as a generic to `trackEvent`:

```
tracking.trackEvent<{ someRandomProp: string }>({
  ...
})
```

damassi:
I've updated this sketch based upon feedback, covering:

- New context provider
- Testing 
- Types

And updated the main description to cover the entire circle, per the above.

src/Artsy/Analytics/useTracking.ts
```diff
@@ -0,0 +1,12 @@
+import * as AnalyticsSchema from "Artsy/Analytics/Schema"
+import { TrackingContext } from "Artsy/Analytics/TrackingContext"
+import { useContext } from "react"
+
+export function useTracking() {
+  const { tracking } = useContext(TrackingContext)
```
ds300: What happens if there's no tracking context in the tree? What should happen? I'd vote for 'it should still work'.

src/Components/NavBar/__tests__/NavBar.test.tsx
```diff
@@ -11,24 +13,39 @@ jest.mock("Components/Search/SearchBar", () => {
...
+        <NavBar />
       </SystemContextProvider>
     )
   }
+
+  beforeEach(() => {
+    ;(useTracking as jest.Mock).mockImplementation(() => {
+      return {
+        tracking,
+        AnalyticsSchema,
```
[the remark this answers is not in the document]
damassi: TBH, last night it started to really annoy me too! Will remove and move to normal imports 👍

typings/react-tracking.d.ts
damassi: Was coding a bit too late last night 🤦‍♂and returning to this it feels somewhat loopy.... I realized here that a slightly simpler way to do this would be to inject `trackEvent` rather than the `tracking` object, and then type things right in Reaction. 

It seems a bit windy to pass a generic into `TrackingProp` only to pass it down to `trackEvent` _and then_ further narrow the function with a generic specific to the function, only later -- and only sometimes. Will play with this a bit more. 

(This would also allow devs to call `trackEvent`, rather than `tracking.trackEvent`, which will simplify addressing the "it should still work" comment below.)

artsyit:

Built with commit 22ae20d27c3df555b4b20ef59e72fdeb4a608e06

https://deploy-preview-2405--artsy-reaction.netlify.com

src/Artsy/Analytics/TrackingContext.tsx
```diff
@@ -1,22 +1,55 @@
 import { track as _track } from "Artsy/Analytics"
 import * as AnalyticsSchema from "Artsy/Analytics/Schema"
 import React from "react"
-import { TrackingProp } from "react-tracking"
 
 type Trackables = AnalyticsSchema.Trackables
 
 interface TrackingContextProps {
-  tracking?: TrackingProp<Trackables>
+  trackEvent?<P>(trackingProps: Partial<Trackables> & P): void
```
damassi: @ds300, @alloy - This approach simplifies things, and scopes to our needs (vs going up a level into react-tracking); however, I can't for the life of me figure out how to get this bit to work as I would like it to. In [the previous iteration](https://github.com/artsy/reaction/pull/2405/files#r283638766) it validated correctly and would force the user to expand the scope with a generic if something not a part of `Trackables` was passed. 

Meaning: 

```tsx
trackEvent({
  unknownProp: ...
}) 
```
should throw a type error, because we haven't passed `P`; e.g.,  

```tsx
trackEvent<{ unknownProp: string }>({
  unknownProp: ...
}) 
```

If I remove `P` from the interface above it correctly validates, and will check against `Partial<Trackables>`. 

I would like to to a) make sure the user isn't mistyping something from `Trackables`; while b) enforcing that they pass additional data for one-off props that don't belong in `Trackables`.

damassi:
Ready for another look y'all -- [this commit](https://github.com/artsy/reaction/pull/2405/commits/22ae20d27c3df555b4b20ef59e72fdeb4a608e06) encapsulates the updates. 

@ds300 - importing `AnalyticsSchema` was indeed much nicer 👍. I ended up adding an export so that VSCode auto-import would work so that we no longer have to use wildcards.

src/Artsy/Analytics/TrackingContext.tsx
ds300: Hmmm tough one, this is the best I could do:

https://www.typescriptlang.org/play/#src=interface%20Trackables%20%7B%0D%0A%20%20%20%20trackableProp%3A%20string%0D%0A%20%20%20%20otherTrackableProp%3A%20string%0D%0A%7D%0D%0A%0D%0Adeclare%20function%20trackEvent%3CP%20extends%20Trackables%3E(trackingProps%3A%20Partial%3CP%3E)%3A%20void%0D%0A%0D%0A%2F%2F%20doesn't%20expect%20all%20trackable%20props%0D%0AtrackEvent(%7B%20trackableProp%3A%20%22%22%20%7D)%0D%0A%2F%2F%20complains%20if%20you%20provide%20one%20that%20doesn't%20exist%0D%0AtrackEvent(%7B%20trackableProp%3A%20%22%22%2C%20newProp%3A%203%20%7D)%0D%0A%2F%2F%20doesn't%20complain%20if%20you%20explicitly%20add%20more%0D%0AtrackEvent%3CTrackables%20%26%20%7BnewProp%3A%20number%7D%3E(%7B%20trackableProp%3A%20%22%22%2C%20newProp%3A%203%20%7D)

alloy:
It’s unclear to me why we need our own tracking context. Can’t we piggy-back on [the existing context that `react-tracking`](https://github.com/nytimes/react-tracking/blob/d7cc4403733b5968878181af849cdebec41ab4ef/src/withTrackingComponentDecorator.js#L14) provides and have `useTracking` return a function (that’s typed with our schema) which uses that?

src/Components/NavBar/NavItem.tsx
```diff
@@ -23,15 +23,15 @@ export const NavItem: React.FC<NavItemProps> = ({
...
+  const { trackEvent } = useTracking()
   const [hover, toggleHover] = useState(active)
   const showMenu = Boolean(Menu && hover)
   const showOverlay = Boolean(Overlay)
   const hoverColor = hover ? "purple100" : "black80"
 
   const trackClick = () => {
     if (href && isString(children)) {
-      tracking.trackEvent({
+      trackEvent({
```
ds300: 💆 ✨

src/Components/NavBar/__tests__/NavBar.test.tsx
```diff
@@ -11,24 +12,40 @@ jest.mock("Components/Search/SearchBar", () => {
     SearchBarQueryRenderer: () => <div />,
   }
 })
+jest.mock("Artsy/Analytics/useTracking")
+jest.mock("Artsy/Analytics/TrackingContext", () => {
+  return {
+    provideTracking: () => Component => Component,
+  }
+})
```
ds300: So this all reminds me that in the order app we couldn't use the `mockTracking` helper at one point so we decided to leave `react-tracking` unmocked and we mocked out `postEvent` instead

https://github.com/artsy/reaction/blob/d562f3be73533115641130b299dc9395064d48b0/src/Apps/Order/Components/__tests__/PaymentPicker.test.tsx#L24-L28

And then in tests we were just like

https://github.com/artsy/reaction/blob/d562f3be73533115641130b299dc9395064d48b0/src/Apps/Order/Components/__tests__/PaymentPicker.test.tsx#L517-L522

Which I think is a lot simpler than mocking out `provideTracking` and `useTracking`. I wonder whether it makes sense to standardise on this?

damassi: Oh interesting, like this will skip a couple steps. My first reactionis that it's a little bit of indirection that will be a bit hard to keep track of over time, like devs will naturally want to mock out the thing that's closest (vs `postEvent` which being lower level, is unseen to most). But it does seem simpler... will think about this for a bit and give it a spin.

damassi:
> Can’t we piggy-back on the existing context that react-tracking provides

@alloy, good call, will investigate 👍

src/Artsy/Analytics/TrackingContext.tsx
damassi: This will work -- we could do something like 
```ts
export type TrackingEvent<P> = Trackables & P

... 

trackEvent<TrackingEvent<{newProp: number}>>({ trackableProp: "", newProp: 3 })
```

damassi:
@alloy / @ds300 / @pepopowitz  - Thoughts on merging this? Any other suggestions / feedback?

ds300:
I'd love to see some unit tests but otherwise happy to merge this :+1:

src/Artsy/Analytics/TrackingContext.tsx
```diff
@@ -0,0 +1,66 @@
+import { track as _track } from "Artsy/Analytics"
+import * as AnalyticsSchema from "Artsy/Analytics/Schema"
+import React from "react"
+
+type Trackables = AnalyticsSchema.Trackables
```
alloy: Perhaps replace this and the schema import with:

```ts
import { Trackables } from "Artsy/Analytics/Schema"
```

src/Artsy/Analytics/TrackingContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Used when tracking custom events, not in the schema:
```
alloy: Why would we have these?

src/Artsy/Analytics/useTracking.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (!tracking) {
+    console.error(
+      "[Artsy/Analytics/useTracking] Error: Attempting to call `useTracking` " +
+        "without a TrackingContext present. Did you forget to wrap the top of " +
+        "your component tree with `provideTracking`?"
+    )
+  }
+
+  // If missing a TrackingContext, return a noop for safety
+  const trackEvent = tracking ? tracking.trackEvent : x => x
```
alloy: Can we throw an error instead? Devs should catch this as they work on it.

src/Components/NavBar/Menus/MobileNavMenu.tsx
```diff
@@ -15,8 +12,12 @@ import {
   Serif,
 } from "@artsy/palette"
 
+import { SystemContext } from "Artsy"
+import { useTracking } from "Artsy/Analytics/useTracking"
+import * as auth from "Components/NavBar/Utils/auth"
```
alloy: Off-topic: This should be renamed to authentication, or at the very least authn (although I personally prefer actual words).

src/Artsy/Analytics/useTracking.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  return {
+    trackEvent,
+  }
```
alloy: Seeing as we only return 1 thing here, can we get rid of the object?

ds300: I like the idea of returning the object here for a couple of reasons:

- It guarantees consistent naming of `trackEvent`
- It leaves the api open for extension

alloy: > It guarantees consistent naming of trackEvent

What's the benefit of this?

> It leaves the api open for extension

Isn’t that very often the case and yet we don’t do that prematurely otherwise either? I also don't foresee any obvious future additions, but if there are known ones then please do share.

alloy:
@damassi _Because you created this PR from a fork I wasn’t able to easily create a PR to your branch (and I didn’t want to just push), so I created a branch with a PR you can view [here](https://github.com/artsy/reaction/pull/2433)._

The change there removes the need for another context and also improves the typing–when you cherry-pick that commit you'll see that there's still a bunch of type errors related to `trackEvent` invocations that do not provide enough data to be in accordance with the schema.

src/Artsy/Analytics/TrackingContext.tsx
alloy: Having just played around with the code and finding that the same additional type was passed in multiple places, I definitely don’t think we should do this, it will only introduce drift in the schema. So while I left [the generic in](https://github.com/artsy/reaction/pull/2433/files#diff-dbc9dc55a5bdbaf4e74b5a92ca77fd51R31) in my changes, I think we should not allow this.

src/Artsy/Analytics/useTracking.ts
ds300: > What's the benefit of this?

Consistent naming reduces cognitive load when reading.

> Isn’t that very often the case and yet we don’t do that prematurely otherwise either? I also don't foresee any obvious future additions, but if there are known ones then please do share.

It's just adding a couple of braces. I think it makes sense to not prematurely abstract things, but it would be academic to call this abstraction, or even indirection. It's another way to name a thing.

It also happens to be the idiomatic way to return things from hooks.

alloy: > Consistent naming reduces cognitive load when reading.

Good, this satisfies my question enough 👍 

> It also happens to be the idiomatic way to return things from hooks.

Is it? I skimmed the API docs before I asked this and only found array destructuring (e.g. state) or not destructuring (eg context).

ds300: Mostly referring to custom hooks e.g. https://github.com/beizhedenglong/react-hooks-lib

Here's a bunch more libraries: https://nikgraf.github.io/react-hooks/

It's not uncommon to see hooks that only return one unnamed value, but it tends to be the case that it's valid for those values to have different names depending on the context.

alloy: Thanks for digging those up! 🙏 

> It's not uncommon to see hooks that only return one unnamed value, but it tends to be the case that it's valid for those values to have different names depending on the context.

I see your point 👍

alloy:
Btw the changes re context can be done in a follow up PR imo. The only change I’d like to see here is to remove the ability to insert data that’s not in the schema, otherwise let’s go for it! 👍

src/Artsy/Analytics/useTracking.ts
damassi: Thanks for the good discussion y'all. Objects -> named exports better than default exports; lock the thing down if there should be consistency enforced across the codebase, as we'd expect here.

damassi:
@alloy - sounds good 👍 I'll merge this and then follow up today with adapting your changes in a follow up PR. @ds300 - I'll add some tests as well.

Thanks everyone for the review / feedback / discussion 👏 Mergin

artsyit:
:rocket: PR was released in v16.11.24 :rocket:

PR #2411, opened 2019-05-14
[Blocked] => Remove react-tracking.d.ts

damassi:
DefinitelyTyped PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35425

The above must be released before this can be merged. Looks like `Decorator` and `Options` aren't exported from the definition file.


artsyit:

Built with commit 778a3dafe148e19eeebcc89758b2442e6fa9d4f1

https://deploy-preview-2411--artsy-reaction.netlify.com

src/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -371,7 +371,8 @@ const PaymentPickerWithInnerRef: React.SFC<
   <PaymentPicker ref={innerRef} {...props as any} />
 )
 
-export const PaymentPickerFragmentContainer = createFragmentContainer(
+// FIXME: After deleting `react-tracking.d.ts, this needed to be cast due to HOC
+export const PaymentPickerFragmentContainer: any = createFragmentContainer(
```
damassi: @ds300 - After removing the local `d.ts` file this was needed. HOC's are tricky and I couldn't figure out how to fix.

ds300: Yeah doesn't look easy with how the `track` decorator is typed now.

We could fix that upstream, but for now would be good to turn this `any` into something less dangerous. Could do that by adding a cast inline after the `track()(...)` expression:

image

alloy:
DT PR is merged 👍

damassi:
Updated types to `@types/react-tracking@5.0.4` and addressed @ds30's comment (so _that's_ how you do it 😄)

artsyit:
:rocket: PR was released in v16.11.6 :rocket:

PR #2409, opened 2019-05-13
Changes adUnit and adDimension props from required to optional

ashleyjelks:
The work to migrate ads to Google Ad Manager (GAM) is still in progress, so for now the props used to populate ad components with GAM hosted data does not need to be required until the migration is complete.

PR #2413, opened 2019-05-14
Move new hosted ads to their own components in Standard Articles

ashleyjelks:
External testers were not seeing the new Google Ad Manager (GAM) hosted ads on 100% of articles because how often ads appear is contingent upon the Share of Voice (SOV) percentage.  One ad campaign has an SOV of 20%, this means that in order to see the new ads on 100% of articles we would have to activate 5 campaigns in the Staging Artsy Writer each day (as Staging Artsy Writer data is overwritten with Production data every night). Rather than doing this and having the logic of how new ads (GAM hosted ads) and current ads intertwined, this PR moves the new ads into their own component. Now the new ads will render on 100% of Standard Articles, on Staging, but they remain only viewable to Admin and/or Allowlisted users. 
Kapture 2019-05-14 at 15 11 00

### DIFF
--- a/src/Components/Publishing/Display/NewDisplayCanvas.tsx
+++ b/src/Components/Publishing/Display/NewDisplayCanvas.tsx
@@ -1,0 +1,46 @@
+import { Flex } from "@artsy/palette"
+import { ErrorBoundary } from "Components/ErrorBoundary"
+import { isHTLAdEnabled } from "Components/Publishing/Ads/EnabledAd"
+import { AdDimension, AdUnit } from "Components/Publishing/Typings"
+import React, { SFC } from "react"
+import styled from "styled-components"
+
+interface DisplayCanvasProps {
+  adUnit?: AdUnit
+  adDimension?: AdDimension
+}
+
+export const NewDisplayCanvas: SFC<DisplayCanvasProps> = props => {
+  const { adUnit, adDimension } = props
+
+  if (!isHTLAdEnabled()) {
+    return null
+  }
+
+  return (
+    <ErrorBoundary>
+      <DisplayCanvasContainer
+        flexDirection="column"
+        width="100%"
+        m={"0 auto"}
+        maxWidth="1250px"
+      >
+        <div
+          className="htl-ad"
+          data-unit={adUnit}
+          data-sizes={adDimension}
+          data-eager
+        />
+      </DisplayCanvasContainer>
+    </ErrorBoundary>
+  )
+}
+
+const DisplayCanvasContainer = styled(Flex)`
+  min-height: fit-content;
+  max-width: 1250px;
+  box-sizing: border-box;
+`
+
+// Set names for tests and DOM
+DisplayCanvasContainer.displayName = "DisplayCanvasContainer"
--- a/src/Components/Publishing/Display/NewDisplayPanel.tsx
+++ b/src/Components/Publishing/Display/NewDisplayPanel.tsx
@@ -1,0 +1,54 @@
+import { Box, color, Flex } from "@artsy/palette"
+import { isHTLAdEnabled } from "Components/Publishing/Ads/EnabledAd"
+import { AdDimension, AdUnit } from "Components/Publishing/Typings"
+import React, { SFC } from "react"
+import styled from "styled-components"
+import { ErrorBoundary } from "../../ErrorBoundary"
+
+export interface DisplayPanelProps extends React.HTMLProps<HTMLDivElement> {
+  adUnit?: AdUnit
+  adDimension?: AdDimension
+}
+
+export const NewDisplayPanel: SFC<DisplayPanelProps> = props => {
+  const { adDimension, adUnit } = props
+
+  if (!isHTLAdEnabled()) {
+    return null
+  }
+
+  return (
+    <ErrorBoundary>
+      <Wrapper m={["auto", "0 auto", "auto"]} color="black100">
+        <DisplayPanelContainer
+          flexDirection="column"
+          className="DisplayPanel__DisplayPanelContainer"
+          p={2}
+          m={["auto", "auto", "auto", "inherit"]}
+        >
+          <div
+            className="htl-ad"
+            data-unit={adUnit}
+            data-sizes={adDimension}
+            data-eager
+          />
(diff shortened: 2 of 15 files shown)


src/Components/Publishing/Display/NewDisplayCanvas.tsx
```diff
@@ -0,0 +1,55 @@
...
+              data-unit={adUnit}
+              data-sizes={adDimension}
+              data-eager
+            />
+          </DisplayContainer>
+        </ErrorBoundary>
+      )
+    )
+  }
+}
```
damassi: All new components should be functional components, not classes.

artsyit:

Built with commit 0a51aebe004cbc3e39c2968dca654102dc7933bc

https://deploy-preview-2413--artsy-reaction.netlify.com

src/Components/Publishing/Display/NewDisplayCanvas.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const Div: StyledFunction<DivProps> = styled.div
+
+export const DisplayContainer = Div`
```
damassi: No need for an intermediary interface or variable: 

```ts
export const DisplayContainer = styled.div<{ layout?: string }>`
  ...
` 
```

src/Components/Publishing/Display/NewDisplayPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { pMedia as breakpoint } from "../../Helpers"
+
+export interface DisplayPanelProps extends React.HTMLProps<HTMLDivElement> {
+  adUnit?: AdUnit
+  adDimension?: AdDimension
+}
+
+export class NewDisplayPanel extends Component<DisplayPanelProps> {
+  render() {
+    const { adDimension, adUnit } = this.props
```
damassi:
```js
if (!isHTLAdEnabled()) {
  return null
}
```

src/Components/Publishing/Display/NewDisplayPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  ${breakpoint.md`
+    margin: auto;
+  `};
+  ${breakpoint.sm`
+    margin: auto;
+  `};
+  ${breakpoint.xs`
+    margin: auto;
+  `};
```
damassi: Should use styled-system / Palette responsive props; `pMedia` is deprecated:

```js
<Flex margin={['auto', '... some other value ...']}>...</Flex>
```

src/Components/Publishing/Display/NewDisplayPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+`
+
+// Set names for tests and DOM
```
damassi: 👍

src/Components/Publishing/Display/__tests__/DisplayPanel.test.tsx
```diff
@@ -8,47 +8,22 @@ import { DisplayPanel } from "../DisplayPanel"
...
     const displayPanel = renderer
-      .create(
-        <DisplayPanel
-          unit={UnitPanelVideo}
-          campaign={Campaign}
-          adDimension={StandardArticleHostedAdPanel.adDimension}
-          adUnit={StandardArticleHostedAdPanel.adUnit}
-        />
-      )
+      .create(<DisplayPanel unit={UnitPanelVideo} campaign={Campaign} />)
```
damassi: Nice cleanup all around here 👍

src/Components/Publishing/Layouts/__tests__/StandardLayout.test.tsx
```diff
@@ -36,6 +38,49 @@ jest.mock("sharify", () => ({
   },
 }))
 
+// FIXME: Figure out how to implement a working mockImplementation to override the top level sharify mock values
```
damassi: Can do: 

```ts
const sd = require("sharify").data
jest.mock("sharify", () => ({ data: jest.fn() }))

...

afterEach(() => {
  jest.resetAllMocks()
})

// Then in your tests 
it('...', () => {
  ;(sd as jest.Mock).mockImplementation(() => {
    return stuff
  }
  ... 
})
```

(Something like this)

src/Components/Publishing/Display/NewDisplayCanvas.tsx
```diff
@@ -0,0 +1,44 @@
+import { Flex } from "@artsy/palette"
+import { ErrorBoundary } from "Components/ErrorBoundary"
+import { isHTLAdEnabled } from "Components/Publishing/Ads/EnabledAd"
+import { AdDimension, AdUnit } from "Components/Publishing/Typings"
+import React, { SFC } from "react"
+import styled from "styled-components"
+
+interface DisplayCanvasProps {
+  adUnit?: AdUnit
+  adDimension?: AdDimension
```
damassi: Are these props actually optional / will they ever not be used?

src/Components/Publishing/Display/NewDisplayCanvas.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const NewDisplayCanvas: SFC<DisplayCanvasProps> = props => {
+  const { adUnit, adDimension } = props
+
+  if (!isHTLAdEnabled()) {
+    return null
+  }
+
+  return (
+    <ErrorBoundary>
+      <DisplayCanvasContainer>
```
damassi: This should follow https://github.com/artsy/reaction/pull/2413/files#diff-9a095ec4c410c988e7febdfec25eefd5R22 in terms of inlining props

src/Components/Publishing/Display/NewDisplayCanvas.tsx
ashleyjelks: For now I'm leaving them option bc all of the new ad work hasn't been migrated to all article types, so it was causing some test failures that @eessex was blocked by. Here's the PR for reference: https://github.com/artsy/reaction/pull/2409

artsyit:
:rocket: PR was released in v16.11.10 :rocket:

PR #2414, opened 2019-05-14
DISCO 1029 - Consolidate url history management in state containers

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-1029.

During my work on pagination for the /collect, /collection/:id, and /artist routes, I learned more about the interactions between these routes and their corresponding state containers. This led me to a pattern in #2392 for managing browser history in the /artist route; I think it does a good job of consolidating "state" changes in one place. This PR applies that pattern to the /collect, /collection/:id, and /search routes. 

I'm also optimistic that these changes will make it easier to apply a fix for https://artsyproduct.atlassian.net/browse/DISCO-1032.

### DIFF
--- a/src/Apps/Artist/Routes/Overview/state.ts
+++ b/src/Apps/Artist/Routes/Overview/state.ts
@@ -159,10 +159,15 @@ export class FilterState extends Container<State> {
   }
 
   resetFilters = () => {
-    this.setState({
-      ...initialState,
-      showActionSheet: true,
-    })
+    this.setState(
+      {
+        ...initialState,
+        showActionSheet: true,
+      },
+      () => {
+        this.pushHistory()
+      }
+    )
   }
 
   unsetFilter(filter) {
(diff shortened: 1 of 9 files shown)


artsyit:

Built with commit 0a1389b7e422accf2450861858425b646405113e

https://deploy-preview-2414--artsy-reaction.netlify.com

ashfurrow:
@pepopowitz I forgot to explicitly say, when I assigned the PR back to you, that you should merge when ready 👍

pepopowitz:
> ...are there any remaining things we'd like to consolidate? Maybe not for this PR, but the `Search/FilterState.tsx` and `Collect/FilterState.tsx` are very similar; is there a case to unify them, or do we get some value from keeping them separate?

@ashfurrow I think there are probably pieces that could be unified. The duplication doesn't bother me too much yet, because (a) it's only a couple places, so I'm not totally sure we know the right abstraction yet, and (b) I do think that they are _similar_ components, so the code _should_ look similar in them. 

I'm going to merge as-is. My personal feeling is that if someone creates a new state container like these, that would be the better time to look for abstractions.

artsyit:
:rocket: PR was released in v16.11.14 :rocket:

PR #2420, opened 2019-05-15
Make FillwidthItem more forgiving of empty artwork images

pepopowitz:
This PR prevents the FillwidthItem from blowing up when there is no image associated. 

## Context

We had an issue today where a specific artwork was published with no image. We think that our system protects this from happening, but it happened somehow.

As a result, users were seeing a blank page when they visited https://artsy.net/artist/nadezda-nikolova-kratzer. (You can currently see this in staging - https://staging.artsy.net/artist/nadezda-nikolova-kratzer - at least, until data is sync'ed down from production, or this PR is deployed.)

image

Investigation into how this artwork was published with no images is happening separately.

### DIFF
--- a/src/Components/Artwork/FillwidthItem.tsx
+++ b/src/Components/Artwork/FillwidthItem.tsx
@@ -13,6 +13,10 @@ import { Mediator } from "Artsy"
 
 // @ts-ignore
 import styled, { StyledComponentClass } from "styled-components"
+import { get } from "Utils/get"
+import createLogger from "Utils/logger"
+
+const logger = createLogger("FillwidthItem.tsx")
 
 const IMAGE_QUALITY = 80
 
@@ -96,6 +100,13 @@ export class FillwidthItemContainer extends React.Component<
     let userSpread = {}
     if (user) {
       userSpread = { user }
+    }
+
+    const image = get(this.props, p => p.artwork.image)
+    if (!image) {
+      const href = get(this.props, p => p.artwork.href, "(unknown href)")
+      logger.error(new Error(`Artwork at ${href} does not have an image!`))
+      return null
     }
 
     return (
--- a/src/Components/Artwork/__tests__/FillwidthItem.test.tsx
+++ b/src/Components/Artwork/__tests__/FillwidthItem.test.tsx
@@ -1,0 +1,21 @@
+import { render } from "enzyme"
+import React from "react"
+import { FillwidthItem } from "../FillwidthItem"
+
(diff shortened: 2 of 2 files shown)


src/Components/Artwork/__tests__/FillwidthItem.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+  // This scenario _should not_ happen. But it did. Every artwork should have
+  //   an image, but somehow one snuck through in production and broke pages.
+  describe("No image associated with an artwork", () => {
+    it("Doesn't blow up when there is no image associated with an artwork", () => {
+      const artwork = {
+        " $fragmentRefs": null,
+        href: "my/artwork",
+      }
+
+      // @ts-ignore
```
pepopowitz: `ts-ignore` because our types prevent us from passing an artwork that doesn't have an image....in code. 😕

src/Components/Artwork/FillwidthItem.tsx
```diff
@@ -98,6 +102,13 @@ export class FillwidthItemContainer extends React.Component<
       userSpread = { user }
     }
 
+    const image = get(this.props, p => p.artwork.image)
+    if (!image) {
+      const href = get(this.props, p => p.artwork.href, "(unknown href)")
+      logger.error(new Error(`Artwork at ${href} does not have an image!`))
```
ashfurrow: Nice logging here 👍

artsyit:

Built with commit a616005764da1c3bdf72191b2b7b88aa4972674d

https://deploy-preview-2420--artsy-reaction.netlify.com

src/Components/Artwork/__tests__/FillwidthItem.test.tsx
ashfurrow: 😚👌

artsyit:
:rocket: PR was released in v16.11.12 :rocket:

pepopowitz:
For permanence: 

@damassi and I had a conversation about this PR, and I'm going to submit a follow-up with a couple changes. 

1. Call `logger.error` with the string directly, instead of a `new Error`. Logged `Error` objects end up getting sent to Sentry, and we don't know how often this error will happen. It's a data error, anyway, and Sentry isn't the right place to escalate those. They'll get annoying & quickly marked as "resolved". 

2. Remove `@ts-ignore` and pass `artist={artist as any}`. This is a less-extreme way to tell TS that the mismatch is intentional.

PR #2422, opened 2019-05-16
DISCO-1035 - Set the `major_periods` state property correctly when constructing the Artist, Collect, and Search apps.

pepopowitz:
This PR addresses https://artsyproduct.atlassian.net/browse/DISCO-1035.

At any of these URLs: /collect, /collection/:id, /artist/:id, or /search...
If you would select a time period filter, then click back, then click forward, the page would stop working or the results would stop updating. This was because in the constructors of the state containers for these Apps, we were initializing the `major_periods` state property improperly. 

We were setting it to an array of the value passed in through props. Unfortunately, the value passed in through props was _already_ an array, so it was effectively initializing the `major_periods` state property to `[ [ "1990" ] ]` (if I had selected 1990). This caused GraphQL to yell about the type - it was expecting an `Array<string>`, not an `Array<Array<string>>`.

### DIFF
--- a/src/Apps/Artist/Routes/Overview/state.ts
+++ b/src/Apps/Artist/Routes/Overview/state.ts
@@ -94,9 +94,7 @@ export class FilterState extends Container<State> {
     if (props) {
       Object.keys(this.state).forEach(filter => {
         if (props[filter]) {
-          if (filter === "major_periods") {
-            this.state[filter] = [props[filter]]
-          } else if (
+          if (
             [
               "for_sale",
               "acquireable",
--- a/src/Apps/Collect/FilterState.tsx
+++ b/src/Apps/Collect/FilterState.tsx
@@ -114,9 +114,6 @@ export class FilterState extends Container<State> {
         }
 
         switch (filter) {
-          case "major_periods":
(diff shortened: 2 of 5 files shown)


artsyit:

Built with commit 62dfef80d2a8eb4a37b54b134c342dd9396a1216

https://deploy-preview-2422--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/state.ts
```diff
@@ -94,9 +94,7 @@ export class FilterState extends Container<State> {
     if (props) {
       Object.keys(this.state).forEach(filter => {
         if (props[filter]) {
-          if (filter === "major_periods") {
```
ashfurrow: Gotcha, so these changes cause the `major_periods` case to fall through to the general `else` clause – nice!

pepopowitz: Oh, yes! I intended on pointing that out so you didn't have to figure it out, but I forgot.

artsyit:
:rocket: PR was released in v16.11.18 :rocket:

PR #2425, opened 2019-05-16
PURCHASE-1102 Remove handling of `null` from the `amount` helper in MP

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-1102

This is follow-up from artsy/metaphysics#1696

I checked all the fields, these are the only relevant cases in reaction.

Also, this is safe to merge, the related change in MP is now in production.

### DIFF
--- a/src/Apps/Artwork/Components/PricingContext.tsx
+++ b/src/Apps/Artwork/Components/PricingContext.tsx
@@ -72,7 +72,6 @@ export class PricingContext extends React.Component<PricingContextProps> {
     }
   }
 
-  // TODO: Investigate why metaphysics is returning null instead of zero for minPrice
   render() {
     const { artwork } = this.props
     if (!artwork.pricingContext) {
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit d0add139f74a5265c9f6d25f2351e9cb31acd069

https://deploy-preview-2425--artsy-reaction.netlify.com

src/Apps/Artwork/Components/PricingContext.tsx
```diff
@@ -119,10 +118,9 @@ export class PricingContext extends React.Component<PricingContextProps> {
             (bin, index): BarDescriptor => {
               const isFirstBin = index === 0
               const isLastBin = index === artwork.pricingContext.bins.length - 1
-              const binMinPrice = bin.minPrice != null ? bin.minPrice : "$0"
               const title = isLastBin
-                ? `${binMinPrice}+`
-                : `${isFirstBin ? "$0" : binMinPrice}–${bin.maxPrice}`
+                ? `${bin.minPrice}+`
+                : `${isFirstBin ? "$0" : bin.minPrice}–${bin.maxPrice}`
```
sweir27: Is there a way we can leverage metaphysics (or use a library here) to generate the `$` based on the currency of the artwork?

ds300: Hmm not sure. We get back the currency in the 3-char ISO format so need a way to convert that to a symbol. Seems like a task for another PR, when we make price-in-context multi-currency perhaps.

sweir27: Mind adding a TODO? We're trying to be more pro-active about handling currency 😅 but you're right that price-in-context has a ways to go.

ds300: done!

artsyit:
:rocket: PR was released in v16.11.19 :rocket:

PR #2430, opened 2019-05-16
DISCO-1032 Apply `medium` filter in path instead of querystring on /collect

pepopowitz:
~This is an incomplete PR. I wanted to put it out for people to look at while I am out tomorrow.~ This PR is green & ready for review/merge!

Addresses https://artsyproduct.atlassian.net/browse/DISCO-1032

## The Problem

In our `urlFragmentFromState` helpers in the Collect app's `FilterState`, we are tracking all state properties in the querystring. This is correct for all params *except* `medium`. 

We have special routes for the /collect page when filtered by `medium`: `/collect/:medium`. These URLs are indexed by bots, and we like them and want to use them. 

We have special logic to get the medium from the path instead of querystring, so going directly to the url `/collect?medium=photography` does _not_ filter by `medium=photography`. Going to `/collect/photography` _does_ filter by `medium=photography`, and we should direct our users there when they are interacting with our filters.

## A Wrench

To make the solution more difficult for this problem, the `FilterState` in the Collect app is shared by both `/collect` and `/collection/:id` routes. We need to apply these special `medium` rules to only the `/collect` route.

## This Solution

It's probably easiest to follow the PR one commit at a time; it's easier that way to see how I found a seam to allow different URL construction rules for the different routes. 

The idea is to inject a URL construction function into `FilterState`, so we can build the correct URL when we push history to the window.

### DIFF
--- a/src/Apps/Collect/urlBuilder.ts
+++ b/src/Apps/Collect/urlBuilder.ts
@@ -1,0 +1,9 @@
+import { FilterState } from "./FilterState"
+
+export const CollectUrlBuilder: (state: FilterState) => string = state => {
+  return "collect"
+}
+
+export const CollectionUrlBuilder: (state: FilterState) => string = state => {
+  return "collection"
+}
--- a/src/Apps/Collect/urlBuilder.ts
+++ b/src/Apps/Collect/urlBuilder.ts
@@ -1,26 +1,41 @@
 import qs from "qs"
 import { State } from "./FilterState"
 
-export const CollectUrlBuilder: (state: State) => string = state => {
-  return urlFragmentFromState(state)
+export const buildUrlForCollectionApp = (state: State): string => {
+  const params = removeDefaultValues(state)
+  const queryString = qs.stringify(params)
+
+  return queryString
+    ? `${window.location.pathname}?${queryString}`
+    : window.location.pathname
 }
 
-export const CollectionUrlBuilder: (state: State) => string = state => {
-  return urlFragmentFromState(state)
+export const buildUrlForCollectApp = (state: State): string => {
+  const fragment = buildCollectUrlFragmentFromState(state)
+
+  return fragment ? `/collect${fragment}` : "/collect"
 }
 
-// Returns a string representing the query part of a URL.
-// It removes default values.
-export const urlFragmentFromState = (state: State) => {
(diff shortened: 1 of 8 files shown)


artsyit:

Built with commit b7201574a52d940ddff8eba21556bdb71529ad7f

https://deploy-preview-2430--artsy-reaction.netlify.com

pepopowitz:
This PR is cleaned up & ready for review!

src/Apps/Collect/urlBuilder.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  return queryString
+    ? `${window.location.pathname}?${queryString}`
+    : window.location.pathname
+}
+
+export const buildUrlForCollectApp = (state: State): string => {
+  const fragment = buildCollectUrlFragmentFromState(state)
+
+  return fragment ? `/collect${fragment}` : "/collect"
```
damassi: This is just a matter of opinion but I've found that through time code that doesn't return logic is a little easier to debug and maintain. Meaning:

```js
export const buildUrlForCollectApp = () => {
   const fragment = ... 
   const url = fragment ? `/collect${fragment}` : "/collect"
   return url
}
```
It also assigns contextual meaning to the return result that, when combined with the function name, makes things easier to parse for devs who wern't originally responsible for writing the code.

pepopowitz: No argument from me, I'll make this change.

src/Apps/Collect/urlBuilder.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (isDefaultFilter(key, value)) {
+      return acc
+    } else {
+      return { ...acc, [key]: value }
+    }
+  }, {})
+}
+
+// This is used to remove default state params that clutter up URLs.
+const isDefaultFilter = (filter, value): boolean => {
```
pepopowitz: I'm going to add some tests to cover this stuff. I don't believe it's currently covered.

pepopowitz: Tests added. I also refactored the default value checks to be a `switch` instead of an `if`, as had been suggested several times 😄

artsyit:
:rocket: PR was released in v16.12.3 :rocket:

PR #2434, opened 2019-05-17
[Tracking] Remove unneeded custom tracking context

damassi:
Addresses follow-up feedback from 
- https://github.com/artsy/reaction/pull/2405 
- https://github.com/artsy/reaction/commit/cb131795a178b614cc2352e427892b0f10a2ee2c#diff-dd3251b7a5a60f2393c92482e5821cdbR35

We can piggy back on `react-tracking`'s context, rather than create our own. This simplifies things by allowing us to remove the `provideTracking` file and associated functionality. 

Closes https://github.com/artsy/reaction/pull/2433

Will follow up on shortening import once https://github.com/nytimes/react-tracking/pull/123 is merged.


src/Components/NavBar/Menus/MobileNavMenu.tsx
```diff
@@ -14,7 +14,7 @@ import {
 
 import { SystemContext } from "Artsy"
 import { useTracking } from "Artsy/Analytics/useTracking"
-import * as auth from "Components/NavBar/Utils/auth"
+import * as authentication from "Components/NavBar/Utils/authentication"
```
damassi: Updated per feedback, but note that the language in Force is `auth`, which this was designed to match. ("Auth" as a language choice is also very common and unambiguous.)

artsyit:

Built with commit 5865a99b5c2c33235df0848688c42d765896a73a

https://deploy-preview-2434--artsy-reaction.netlify.com

src/Components/NavBar/Menus/MobileNavMenu.tsx
alloy: It’s true that it is common usage, but not unambiguous. Depending on the reader it will mean either authentication or authorization, or both. As many people don’t realize the important differences between authentication and authorization, removing ambiguity at any opportunity is imo a minimal step towards education of these concepts.

A random example from the many search results for this topic https://stackoverflow.com/questions/6556522/authentication-versus-authorization

damassi: That makes sense, I didn't think about 'authorization' 👍

artsyit:
:rocket: PR was released in v16.12.0 :rocket:

PR #2439, opened 2019-05-20
Select default time period filter selection via RadioGroup instead of Radio

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-1046

We were setting the `Radio`.`selected` property on the time period that matched the param in the URL, but it had no effect because our `Radio`s are inside of a `RadioGroup`. It seems that `Radio`.`selected` is overridden by `RadioGroup`.`defaultValue` when the radios are located inside a group. 

This PR specifies the default via `RadioGroup`.`defaultValue`. I'll submit a second PR to palette, making a note for users that `Radio`.`selected` is overridden by `Radio`.`defaultValue`.

### DIFF
--- a/src/Apps/Collect/Components/Filters/TimePeriodFilter.tsx
+++ b/src/Apps/Collect/Components/Filters/TimePeriodFilter.tsx
@@ -2,6 +2,7 @@ import React from "react"
 import { FilterState } from "../../FilterState"
 
 import { Radio, RadioGroup } from "@artsy/palette"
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit c476ad32a38ea7f582c0f5c70b21492f362e39db

https://deploy-preview-2439--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v16.12.2 :rocket:

PR #2467, opened 2019-05-28
DISCO-1022 - Updates zero state copy

kierangillen:
- https://artsyproduct.atlassian.net/browse/DISCO-1022

- Stephen had requested that we retain state if the original search query had results before applying filters, however I decided it's better to just return the copy based on applied filters. If a user goes directly to a URL with filters applied we would need to do two queries to know if the original search term yielded results. I opted to just have a copy state for when filters are applied and one for only the search term.

### DIFF
--- a/src/Apps/Search/Components/ZeroState.tsx
+++ b/src/Apps/Search/Components/ZeroState.tsx
@@ -1,21 +1,15 @@
 import { Box, Flex, Serif } from "@artsy/palette"
 import { SendFeedback } from "Apps/Search/Components/SendFeedback"
-import qs from "qs"
 import React, { FC } from "react"
 
 interface Props {
   term: string
+  query: {}
 }
 
 export const ZeroState: FC<Props> = props => {
-  const { term } = props
-  let hasFilters = false
-  if (window.location.search.length) {
-    const params = Object.keys(
-      qs.parse(window.location.search, { ignoreQueryPrefix: true })
-    )
-    hasFilters = params.length > 1
-  }
+  const { term, query } = props
+  const hasFilters = Object.keys(query).length > 1
 
   return (
     <Flex
--- a/src/Apps/Search/Components/ZeroState.tsx
+++ b/src/Apps/Search/Components/ZeroState.tsx
@@ -1,14 +1,21 @@
 import { Box, Flex, Serif } from "@artsy/palette"
 import { SendFeedback } from "Apps/Search/Components/SendFeedback"
+import qs from "qs"
 import React, { FC } from "react"
 
 interface Props {
   term: string
-  entity: string
 }
 
 export const ZeroState: FC<Props> = props => {
-  const { term, entity } = props
+  const { term } = props
+  let hasFilters = false
+  if (window.location.search.length) {
+    const params = Object.keys(
+      qs.parse(window.location.search, { ignoreQueryPrefix: true })
+    )
+    hasFilters = params.length > 1
+  }
 
   return (
     <Flex
@@ -19,15 +26,14 @@ export const ZeroState: FC<Props> = props => {
     >
       <Box m={3} textAlign="center">
         <Serif size="6">
-          {`We couldn't find any ${entity} for 
-        "${term}"`}
+          {hasFilters ? "No results found." : `No results found for "${term}".`}
         </Serif>
-
         <Serif size="3">
-          Try checking your spelling or try another search term.
+          {hasFilters
+            ? "Try removing some filters or try another search term."
+            : "Try checking for spelling errors or try another search term."}
         </Serif>
       </Box>
-
       <Box width="100%">
         <SendFeedback />
       </Box>
--- a/src/Apps/Search/Routes/Artists/SearchResultsArtists.tsx
+++ b/src/Apps/Search/Routes/Artists/SearchResultsArtists.tsx
@@ -133,7 +133,7 @@ export class SearchResultsArtistsRoute extends React.Component<Props, State> {
       <LoadingArea isLoading={this.state.isLoading}>
         {artists.length === 0 ? (
           <Box mt={3}>
-            <ZeroState entity="artists" term={term} />
+            <ZeroState term={term} />
           </Box>
         ) : (
           this.renderArtists()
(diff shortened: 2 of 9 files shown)


ashfurrow:
Looks good! Merge on green.

src/Apps/Search/Components/ZeroState.tsx
```diff
@@ -1,14 +1,21 @@
...
 export const ZeroState: FC<Props> = props => {
-  const { term, entity } = props
+  const { term } = props
+  let hasFilters = false
+  if (window.location.search.length) {
+    const params = Object.keys(
+      qs.parse(window.location.search, { ignoreQueryPrefix: true })
+    )
+    hasFilters = params.length > 1
+  }
```
damassi: Can this be pulled from the router?

```
this.router.props.search
```

In general (with a few exceptions) all router / history state needs to go through [found](https://github.com/4Catalyzer/found) / [farce](https://github.com/4Catalyzer/farce) so that state is not out of sync.

damassi:
@ashfurrow / @kierangillen - I removed the merge on green label (sorry for the drive by) pending a quesion about pulling our current history out the router. Wanted to be sure we're using our framework before jumping right to the `window`, which can lead to sync issues.

kierangillen:
@damassi I posted this in #front-end too: 

I'm looking into the search filters' location state. It looks like when a filter is applied it's not updating the `location` prop, should filter params be stored in the `location.query` object when applied? This seems like a smell but also perhaps was done for a reason, if there's a page render on location pushes that we want to avoid for example.

The queries are there if the page is loaded using the params `/search?medium=sculpture&term=asdasd` for example.

src/Apps/Search/Components/ZeroState.tsx
pepopowitz: This is a case where we are manually pushing things to window.history, via [the FilterState state container](https://github.com/artsy/reaction/blob/76408fc817cc79aea707e9d75076c20531584889/src/Apps/Search/FilterState.tsx#L132); I don't think any of these URL updates are currently going through found/farce.

I _think_ the reason we are manually pushing things to the history instead of navigating via a found route is that we want the user to feel like they aren't leaving this page, they are just getting different results filtered. Pushing them to a new route would, I would expect, be more of a jarring experience than we want them to feel. I could be wrong about that; I don't know the history, and I'm mostly guessing, and I also don't know if found is capable of doing a transition between routes without re-rendering the entire route.

pepopowitz: There are a couple things that come to mind for modifying the code in this PR to not check `window.location`.

1. _Can_ we use found to manage the routes that are changed when you interact with the filters?
2. We could elevate the `FilterState` provider above this component in the tree. I _think_ the hierarchy currently looks like this:

```
<SearchApp>
  <ZeroState/>
  <SearchResultsArtworksRoute>
    <Provider inject={new FilterState(...)}>
      <SearchResultsFilterContainer>....actual filters and search results...</SearchResultsFilterContainer>
    </Provider>
  </SearchResultsArtworksRoute>
</SearchApp>
```

And I think we'd want to change that to be more like this:

```
<SearchApp>
  <Provider inject={new FilterState(...)}>
    <ZeroState/>
    <SearchResultsArtworksRoute>
      <SearchResultsFilterContainer>....actual filters and search results...</SearchResultsFilterContainer>
    </SearchResultsArtworksRoute>
  </Provider>
</SearchApp>
```

And then the `ZeroState` component would have access to the current filters in the `FilterState`.

kierangillen: Thanks @pepopowitz for looking into this, at this point if we need to refactor something just for this copy change its probably not worth it. The question is, I think, if we're capturing the location state correctly as is.

kierangillen:
@damassi What do you think after @pepopowitz comments? Should we merge this as is?

damassi:
@kierangillen, ah, apologies -- I could have sworn I responded to that comment. Since we're managing filter state with Unstated via the `FilterState` component, I agree with @pepopowitz that that should be the source or truth, not `window`. 

> at this point if we need to refactor something just for this copy change its probably not worth it

This isn't necessarily so. The idea for having centrally managed state is so that all parts of the system can stay in sync, regardless of scale. By having unstated manage filters, while window is checked for filter count, we're deferring to two engines, not one, and introducing side-effect-y code into the system.

The point of unstated is that it makes it trivially easy to grab state from anywhere in the React tree, without the needs to pass props. Steve's second example should only take a minute to implement, and we can check the filter length count there, and if zero show the zero state.

artsyit:

Built with commit 5fc62089ac198839164f823b3cd762c40e8aead8

https://deploy-preview-2467--artsy-reaction.netlify.com

artsyit:

Built with commit 1eee26b874ab563fdde7c887e6eae3fd246fe2bd

https://deploy-preview-2467--artsy-reaction.netlify.com

artsyit:

Built with commit 2a7cf2453d9aa12529c94aceb446e34c4fb3ce5a

https://deploy-preview-2467--artsy-reaction.netlify.com

src/Apps/Search/Routes/Artists/SearchResultsArtists.tsx
```diff
@@ -126,14 +126,14 @@ export class SearchResultsArtistsRoute extends React.Component<Props, State> {
...
+    const { query } = location
     const { term } = get(location, l => l.query)
-
     const artists = get(viewer, v => v.search.edges, []).map(e => e.node)
     return (
       <LoadingArea isLoading={this.state.isLoading}>
         {artists.length === 0 ? (
           <Box mt={3}>
-            <ZeroState entity="artists" term={term} />
+            <ZeroState term={term} query={query} />
```
damassi: Rather than passing stuff all around, could quickly do 
```tsx
<Subscribe to={[FilterState]}>
  {(filters: FilterState) => {
    const query = // parse filters state
    return <ZeroState term={term} query={query} />
   }}
</Subscribe>

damassi: State is populated at the get go, [via `location.query`](https://github.com/damassi/reaction/blob/e60c0c35cba024e557fe80f391f3faa584317b40/src/Apps/Search/Routes/Artworks/SearchResultsArtworks.tsx#L22-L29).

kierangillen: Ah, I see, got it! Will update.

PR #2468, opened 2019-05-28
Updates search results skeleton styling

kierangillen:
- Updates skeleton styling to more closely match the page layout
- Fixes some minor margin issues with filter layout

### DIFF
--- a/src/Apps/Search/Components/SearchResultsSkeleton/FilterSidebar.tsx
+++ b/src/Apps/Search/Components/SearchResultsSkeleton/FilterSidebar.tsx
@@ -1,34 +1,30 @@
-import { Box, color, Separator } from "@artsy/palette"
+import { Box, color, Flex, Separator } from "@artsy/palette"
 import React from "react"
 
 const FilterSidebarSection: React.SFC<any> = props => {
   return (
     <>
       <Box
-        width={100}
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
+        width={80}
+        height={14}
+        style={{ marginBottom: "15px", backgroundColor: color("black10") }}
       />
-      <Box
-        width={120}
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
-      />
-      <Box
-        width={120}
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
-      />
-      <Box
-        width={120}
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
-      />
-      <Box
-        width={120}
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
-      />
+      <Flex alignItems="center" mt={10}>
+        <Box width={20} height={20} bg={color("black10")} mr={1} />
+        <Box width={100} height={17} bg={color("black10")} />
+      </Flex>
+      <Flex alignItems="center" mt={10}>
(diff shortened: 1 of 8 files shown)


artsyit:

Built with commit ee62d0781aba4d7b3f667e0529a157575b9642a0

https://deploy-preview-2468--artsy-reaction.netlify.com

artsyit:
:rocket: PR was released in v16.12.18 :rocket:

src/Apps/Search/Components/SearchResultsSkeleton/FilterSidebar.tsx
```diff
@@ -1,34 +1,30 @@
...
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
-      />
-      <Box
-        width={120}
-        height={8}
-        style={{ marginTop: "20px", backgroundColor: color("black10") }}
+        width={80}
+        height={14}
+        style={{ marginBottom: "15px", backgroundColor: color("black10") }}
```
damassi: @kierangillen - `marginBottom` goes outside of our design system. All values need to be constrained within palette's [theme](https://github.com/artsy/palette/blob/master/packages/palette/src/Theme.tsx) file so that we don't sucumb to design drift. That means often rounding up or rounding down from the redlines.

damassi: Additionally, don't need to use `style` tags for props that can be pulled from `styled-system`'s API (meaning `background`): https://styled-system.com/table. In this case can use `bg`. 

If something isn't working via props, it's important to look directly at the source of truth -- [`Box`](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Box/Box.tsx#L60-L74) from within Palette.

ashfurrow: That makes sense – how can we tell whether a component _should_ use `marginBottom` or otherwise step outside our design system? Is there ever a legitimate case for this? The search results skeleton here is pretty unusual, for example, and I can see it needing more than our Palette theme can offer.

damassi: There's typically very few cases where we should step out of the design system, as it was built entirely on constraints. But I do see what you mean re the skeleton since that more graphical and overlays what will eventually be actual UI 👍

kierangillen: @damassi I would argue that this skeleton may fall in an area outside our normal design system, as I'm trying to achieve pixel perfect mirroring of the actual page. For this reason font sizes and margins may visually be different given the the line height of the fonts on the actual rendered page vs where they fall visually.

kierangillen: I can update the rest of the code to not use the `styled` attribute, that was code that I opted to not update where I wasn't touching or making changes. But best to take the time to update!

damassi: Side-note: with `marginBottom` we should still use `mb` as we do elsewhere, as a top-level prop; styled-system's lookup system will first check  the theme, and if it can't find the lookup key there it will convert the value to pixels, so all values are basically permitted.

damassi: > I would argue that this skeleton may fall in an area outside our normal design system

Yup agree.

Issue #2474, opened 2019-05-29
[RFC] Replace Moment and Moment-Timezone in Reaction and Palette


kierangillen:
Discover has an [open ticket](https://artsyproduct.atlassian.net/browse/DISCO-1002) to replace Moment and Moment-Timezone with another API and we've talked about switching to [Luxon](https://moment.github.io/luxon).

Currently 47.6% of Reaction's bundle size is attributed to Moment-Timezone.

The bundle size for [Luxon](https://bundlephobia.com/result?p=luxon@1.15.0) is significantly smaller than for [Moment](https://bundlephobia.com/result?p=moment@2.24.0) and [Moment-Timezone](https://bundlephobia.com/result?p=moment-timezone@0.5.25). 

So far, with some experimenting, there seems to be no reason Luxon should not work for our purposes in Reaction and Palette.

pepopowitz:
I'd be curious about edge cases. Have you found any unique/complex uses of timezones with moment-timezone in the existing codebase, or are they all pretty simple?

alloy:
Are there any other alternatives? And, if so, what are the pros/cons and are we perhaps already including one of the alternatives (or luxon) as a transitive dependency?

kierangillen:
@pepopowitz so far I haven't found any situation that can't be handled with Luxon, you can see my tests/work here https://github.com/artsy/reaction/pull/2476

@alloy I'm not an expert on these libraries and would be interested if others have thoughts, but from what I can tell it may come to personal preference as I can't find radical differences between alternative libraries. [Date-FNS](https://date-fns.org/) has a marginally smaller [bundle size](https://date-fns.org/) to Luxon's. I know @ds300 has used Luxon before, curious whether others have opinion or experiences.

kierangillen:
It should be noted that Luxon is maintained by the Moment team, so there may be slightly more support but this is conjecture.

anandaroop:
This caught my eye, from the author's own summary:

> 5. Abuse the Intl API horribly to provide time zone support. Only possible for modern browsers.

I wanted to see what the browser compatibility looked like, and I believe it's summarized in https://github.com/moment/luxon/blob/master/docs/matrix.md — so as an FYI this means we would either not support IE 11 and iOS Safari 9 — or polyfill for them.

damassi:
We currently don't support either of those browsers.

anandaroop:
Thanks @damassi I figured as much 👍 , but struggled to find an authoritative list.

[...]

PR #2476, opened 2019-05-29
Removes Moment-Timezone and introduces Luxon

kierangillen:
- This PR completely removes all references to `moment-timezone` and replaces it with `Luxon` as discussed in https://github.com/artsy/reaction/issues/2474
- This PR is the first step in completely removing both moment and moment-timezone but is isolated to moment-timezone for an easier transition. 
- Updates datetime formatting in accordance to our new date formatting guidelines where luxon formatting has been updated - https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794


src/Apps/Artwork/Components/OtherAuctions.tsx
```diff
@@ -1,6 +1,6 @@
 import { Box, Flex } from "@artsy/palette"
 import { OtherAuctions_sales } from "__generated__/OtherAuctions_sales.graphql"
-import { AuctionCardFragmentContainer as AuctionCard } from "Components/v2/AuctionCard"
+import { AuctionCardFragmentContainer as AuctionCard } from "Components/v2/AuctionCardLuxon"
```
damassi: Rather than rename, can we duplicate and then deprecate / delete old code?

src/Components/v2/AuctionCardLuxon.tsx
```diff
@@ -0,0 +1,191 @@
...
+      return "In progress"
+    } else if (isRegistered || isRegistrationClosed) {
+      return `Live in ${relativeTime(liveStartAt, now)}`
+    } else {
+      const dateTime = DateTime.fromISO(liveStartAt).setZone("America/New_York")
+      return `Register by ${dateTime.monthShort} ${dateTime.day}`
+    }
+  } else {
+    return `Ends in ${relativeTime(endAt, now)}`
+  }
```
damassi: One thing to note is a lot of similar behavior is already in MP: https://github.com/damassi/metaphysics/blob/d6b855cb928e51b8c8b298086540cd3b9734fa04/src/schema/sale/display.ts#L34

src/Components/v2/__stories__/AuctionTimerLuxon.story.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+
+import { SystemContextProvider } from "Artsy"
+import { AuctionTimerQueryRenderer } from "Components/v2/AuctionTimer"
+import { Section } from "Utils/Section"
+
+storiesOf("Styleguide/Components", module).add("AuctionTimerLuxon", () => {
+  return (
+    <React.Fragment>
```
damassi: `<>...</>`

damassi:
A less extensive 1st-pass moment-to-luxon refactor could also take place in Palette. A while back @dblandin ended up adding moment / moment-timezone and we always meant to follow up as it's increased the bundle-size quite a bit due to all of the timezone code.

src/Apps/Artwork/Components/OtherAuctions.tsx
dblandin: What do you think of updating the existing components to use Luxon vs creating largely duplicate Luxon variations? Might be easier to digest the diff.

Related to https://github.com/artsy/reaction/pull/2476#discussion_r288696089, it would be great to identify opportunities to delegate formatting / TZ manipulation to Metaphysics, where possible.

Also might be worthwhile to start with artsy/palette and work our way up through artsy/reaction to artsy/force.

kierangillen: I was thinking at first to make a copy with it's own story so its super clear that everything is displaying the same as the original. After that's verified I can replace the original files.

dblandin: Sounds reasonable but I hope we also have tests to verify that the formatted dates/times look correct.

artsyit:

Built with commit 49313cd7aba19dee3e3a4da4cf9c5aab24062803

https://deploy-preview-2476--artsy-reaction.netlify.com

src/Components/Publishing/Constants.ts
```diff
@@ -93,50 +92,45 @@ export const getAuthorByline = (authors, isEditoral = true) => {
...
+  }`
+  let hour
+  if (dateTime.hour > 12) {
+    hour = dateTime.hour - 12
+  } else if (dateTime.hour === 0) {
+    hour = 12
+  } else {
+    hour = dateTime.hour
+  }
+  const time = `${hour}:${minutes}${amPm}`
```
dblandin: I'll echo the same comment from  https://github.com/artsy/palette/pull/505/files/5c3707b56acd3e0f8ede9a53f9473c6d6041c4da#r292103479

This styling seems simple enough to defer to a format string, rather than constructing it ourselves.

kierangillen: Luxon's formatting strings don't provide the display formatting for our styling requirements.

kierangillen: Like lowercase am/pm for example.

damassi: (In the future these sorts of things should be called out in the description.)

dblandin: > Luxon's formatting strings don't provide the display formatting for our styling requirements.
>
> Like lowercase am/pm for example.

Ah, I didn't realize that! Thanks for clarifying :+1: 

I found a related issue and a good response to it here: https://github.com/moment/luxon/issues/224#issuecomment-400176269

Looks like we might be able to use `toLocaleParts` to customize the casing on the meridiem part. No need to block this PR though.

I'm also not sure where the decision was made to remove the space between `minutes` and `amPm`. Could you provide a link to that?

src/Components/Publishing/Byline/__tests__/Date.test.tsx
```diff
@@ -37,11 +37,11 @@ describe("Date", () => {
 describe("#getDate", () => {
   const timestamp = "2017-02-22T19:22:05.709Z"
   const expectedFormattedDates = {
-    monthYear: "February 2017",
+    monthYear: "Feb 2017",
```
zephraph: Is this actually going to be changing date formats to shorthand? Is that something we expect to see replicated in the UI?

src/Components/Publishing/Constants.ts
kierangillen: All of our time formatting is located in this notion doc https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794 @dblandin

src/Components/Publishing/Byline/__tests__/Date.test.tsx
kierangillen: Yes it's a recent change to our date formatting - https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794

zephraph: Ahh, I see. 👍

dblandin:
@kierangillen This PR seems to be accomplishing two things which is leading to some confusion:
1. Removing moment-timezone in favor of luxon
2. Updating date/time formatting to adhere to Artsy Brand Guidelines

What do you think of either:
1. Splitting these changes into two PRs, with each addressing one of the goals above?
2. Or updating the description to refer to both goals and linking to relevant Jira tickets / Notion documents?

kierangillen:
@dblandin I updated the description to mention there are changes to the date formatting. The date formatting changes are something that need to be done through all of our repos and are only done in this PR because the code producing the formatting is being touched. Therefore it's not one of the aims of this PR but a side effect.

src/Components/v2/AuctionTimer.tsx
```diff
@@ -41,7 +40,21 @@ export class AuctionTimer extends React.Component<Props> {
   }
 
   labelWithTimeRemaining() {
-    const display = moment(this.endDate).format("MMM D, h:mma")
```
zephraph: If there are many of these more complex formatting tasks it might be good to add a TODO to pull these out into a common datetime formatting file. 

I'm okay with that not happening in this PR though.

zephraph:
It's looking pretty good. I'm going to wait for the others to 👍 the review before merging.

dblandin:
> The date formatting changes are something that need to be done through all of our repos and are only done in this PR because the code producing the formatting is being touched. Therefore it's not one of the aims of this PR but a side effect.

Yep, that's clear to me now! Thanks for clarifying.

In the future, I'd suggest either splitting off drive-by changes to a following PR or making those considerations clear and up-front in the PR description to ease review :+1: 

Overall, this is great work! Thanks for tackling this! :tada:

src/Components/v2/AuctionTimer.tsx
kierangillen: Many of these formats are already in metaphysics which would be the most ideal. However removing moment and also updating all of the relay queries to pull in the correct formatting seamed like two separate tasks :)

zephraph:
I'm gonna go ahead and merge 👍

artsyit:
:rocket: PR was released in v16.16.0 :rocket:

PR #2480, opened 2019-05-29
Prevent right click on artwork images

pepopowitz:
**Update:** https://github.com/artsy/palette/pull/473 has been merged/released, and this branch rebased on it. This one's ready to go!

---

This addresses https://artsyproduct.atlassian.net/browse/DISCO-1062.

We used to block users from right-clicking on any artwork images, to discourage them from downloading them. This functionality disappeared from most of the site as pieces were moved from Force to Reaction.

In this PR, we add back the blocking of right-click on artwork images, unless the user is an admin.


src/Components/Artwork/GridItem.tsx
```diff
@@ -156,7 +160,7 @@ export const ArtworkGridItem = styled(ArtworkGridItemContainer)`
   }
 `
 
-export default createFragmentContainer(ArtworkGridItem, {
+export default createFragmentContainer(withSystemContext(ArtworkGridItem), {
```
pepopowitz: This is the one significant concern I have with this PR. This _seems_ like it should be just fine, but `GridItem` was previously relying on the `user` prop being passed into it, and `ArtworkGrid` (the only thing that renders a `GridItem`) was _also_ relying on the `user` prop being passed into _it_. That seems kind of strange to me, since `GridItem` could have just subscribed to the context to get the `user`. It makes me wonder if there is a reason that it's relying on it being passed in, and if this change will break something in a very unexpected way.

src/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -116,7 +116,6 @@ export class ArtworkGridContainer extends React.Component<
           <GridItem
             artwork={artwork}
             key={"artwork-" + j + "-" + artwork.__id}
-            user={this.props.user}
```
pepopowitz: This is removed because `GridItem` now subscribes to the context. If this is passed in, it will override what comes from the context (a surprise to me....but that is what I witnessed.)

That seems like it would be fine, because the user should always just be coming from the context...but there are several instances of `<ArtworkGrid>` that do not pass the user through. In those cases, the `user` coming from `withSystemContext` [here](https://github.com/artsy/reaction/pull/2480/files#diff-69470519ac81a7d74e2dbdb9d873891fR163) gets overwritten by `undefined`. 🤷‍♂

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -335,3 +348,5 @@ export class Lightbox extends React.Component<LightboxProps, LightboxState> {
     this.detectActivity()
   }
 }
+
+export const Lightbox = withSystemContext(LightboxComponent)
```
pepopowitz: This `withSystemContext` doesn't bother me, because `Lightbox` wasn't ever relying on `user` to be passed in.

src/Utils/getUser.ts
```diff
@@ -30,3 +30,7 @@ export function userHasLabFeature(user: User, featureName: string): boolean {
 
   return lab_features.includes(featureName)
 }
+
+export function userIsAdmin(user?: User): boolean {
```
pepopowitz: I _could_ add tests for this, but it seems pretty simple. If you feel like I should cover this with tests, let me know.

artsyit:

Built with commit 149a54584ed7f728d03a817dfd32e3c4346e67ae

https://deploy-preview-2480--artsy-reaction.netlify.com

src/Components/Artwork/GridItem.tsx
damassi: The grid stuff is like pre v1 of Reaction, right when we starting evaluating our new tech stack and were figuring things out. Tech debt.

src/Utils/getUser.ts
damassi: Test wouldn't hurt for utility things like this. Would just take a sec and prevent breakage in the future.

damassi: One other slight refactory-y thing (if you wouldn't mind :) would be to rename this file from `getUser` to just `user`, since it now contains multiple utilities. Originally it was mean to follow 1:1 with the exported file at the top, but has since expanded.

pepopowitz: Would you rather have it broken into individual files for each function, or do you prefer them all in that one file?

damassi: I think one file is fine for this

pepopowitz: All set on these changes! Let me know how they look.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -10,6 +10,7 @@ import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
 import { AuthModalIntent, openAuthModal } from "Utils/openAuthModal"
 import { Media } from "Utils/Responsive"
+import { userIsAdmin } from "Utils/user"
```
damassi: 👍

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -70,7 +74,7 @@ export interface LightboxState {
 }
 
 @track({ context_module: Schema.ContextModule.Zoom })
-export class Lightbox extends React.Component<LightboxProps, LightboxState> {
+class LightboxComponent extends React.Component<LightboxProps, LightboxState> {
```
damassi: Why change the name here? We should avoid encoding implementation details if possible.

damassi: Oh I see, it's for below 👍

pepopowitz: Is there a different naming convention you'd prefer?

damassi: No, not necessarily! Another way to do it could be 

```tsx
export const Lightbox = withSystemContext(class extends React.Component<...> {
  ...
})
```
But all in all this is totally fine as is.

artsyit:
:rocket: PR was released in v16.13.0 :rocket:

PR #2481, opened 2019-05-30
[NavBar] Fix notification count for zero notifications

damassi:
Fixes two issues caught in analytics QA testing:

- When a user has zero notifications, the badge is hidden, with its hover event. Adds a trackable hover event to the bell icon proper
- When there are zero notifications in the cookie, it comes back undefined, which is stripped out by analytics. If undefined, assume 0.

### DIFF
--- a/src/Components/NavBar/NavBar.tsx
+++ b/src/Components/NavBar/NavBar.tsx
@@ -50,6 +50,7 @@ export const NavBar: React.FC = track(
   const { xs, sm } = useMedia()
   const isMobile = xs || sm
   const isLoggedIn = Boolean(user)
+  const getNotificationCount = () => cookie.get("notification-count") || 0
 
   // Close mobile menu if dragging window from small size to desktop
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit baf2a4071da780dd894edcf435ff4c9464cb9977

https://deploy-preview-2481--artsy-reaction.netlify.com

damassi:
Merging per offline convo

artsyit:
:rocket: PR was released in v16.12.24 :rocket:

src/Components/NavBar/NavBar.tsx
```diff
@@ -50,6 +50,7 @@ export const NavBar: React.FC = track(
   const { xs, sm } = useMedia()
   const isMobile = xs || sm
   const isLoggedIn = Boolean(user)
+  const getNotificationCount = () => cookie.get("notification-count") || 0
```
mzikherman: Wondering why defer this?

damassi: For the client; `cookie` requires a window, so putting it behind a function serves as a guard.

mzikherman: Ah right, sorry- I've asked this before! Thanks, makes perfect sense.

PR #2487, opened 2019-05-31
PLATFORM-1440 - Removes React-Flickity-Component from Carousel

kierangillen:
- https://artsyproduct.atlassian.net/browse/PLATFORM-1440
- Removes `react-flickity-component` wrapper due to GPLv3 licensing issue.
- Removes versions of Carousels and consolidates to one Carousel file structure.

- Note: With this PR we lose some of the more edge case uses of Flickity's API, which `react-flickity-wrapper` provided, e.g. dynamically adding and removing slides. However, we do not currently have a need for these options, if we want to use these options we will need to use [portal](https://reactjs.org/docs/portals.html) to render the carousel children.


artsyit:

Built with commit f6ff48264ac003c5f1cb165565479c4878a7e02d

https://deploy-preview-2487--artsy-reaction.netlify.com

damassi:
@kierangillen - Can you reopen this PR as Carousel4? By replacing the main carousel (before we've had a chance to test the wrapper-less version) we're rolling things out to all consumers. Also difficult to see diff between V3 and V4.

PR #2489, opened 2019-05-31
[Collections] Related collections rail 

xtina-starr:
Addresses: [GROW-1228](https://artsyproduct.atlassian.net/browse/GROW-1228)

This adds related collection rails to the footer of the collection pages.
Screen Shot 2019-05-31 at 5 54 50 PM


src/Components/RelatedCollectionsRail/RelatedCollectionEntity.tsx
```diff
@@ -0,0 +1,157 @@
...
+import styled from "styled-components"
+
+export interface CollectionProps {
+  // collection: ArtistCollectionEntity_collection
+  collection: RelatedCollectionEntity_collection
+}
+
+@track()
+export class RelatedCollectionEntity extends React.Component<CollectionProps> {
+  @track<CollectionProps>(({ collection }) => ({
```
damassi: This work is already all done so not suggesting we refactor everything :) but we can now ditch all of the classes and use our new [`useTracking()` hook](https://github.com/artsy/reaction/blob/master/src/Components/NavBar/Menus/MoreNavMenu.tsx#L7).

src/Components/RelatedCollectionsRail/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+import React from "react"
+import { graphql, QueryRenderer } from "react-relay"
+import { RelatedCollectionsRailFragmentContainer as RelatedCollectionsRail } from "./RelatedCollectionsRail"
+
+interface Props {
+  category: string
+  showOnEditorial?: boolean
+}
+
+export const RelatedCollectionsRailContent: React.SFC<Props> = passedProps => {
```
damassi: `passedProps` -> `props`

src/Components/RelatedCollectionsRail/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <SystemContextConsumer>
```
damassi: We should be using hooks where we can now, which flattens things out quite nicely:

```tsx
import React, { useContext } from 'react'
import { SystemContext } from 'Artsy'

...

export const RelatedCollectionsRailContent: React.SFC<Props> = props => {
  
  return ( ... ) 
}
```

damassi: Or even better, rather than pulling from our context manually, we can create a new hook inside of `src/Artsy` called `useSystemContext` that we reference instead:

```tsx
import { useContext } from 'react'
import { SystemContext } from 'Artsy'

export const useSystemContext() {
  const systemContext = useContext(SystemContext) 
  return  systemContext
}
```
Then later
```tsx
import { useSystemContext } from 'Artsy'

...

const { relayEnvironment } = useSystemContext()
```

src/Components/RelatedCollectionsRail/RelatedCollectionEntity.tsx
xtina-starr: Sweet!! Good to know.

src/Components/RelatedCollectionsRail/index.tsx
xtina-starr: I opted to grab the `relayEnvironment` from the `useContext` api. I was thinking the later suggestion could be addresses in a separate PR.

src/Components/RelatedCollectionsRail/index.tsx
xtina-starr: I noticed I was really only using the `category` prop so I decided to just deconstruct that.

src/Components/RelatedCollectionsRail/index.tsx
damassi: @xtina-starr - Would you mind taking care of this in another PR and letting #front-end know about the new hook once it's merged?

src/Components/RelatedCollectionsRail/index.tsx
damassi: 👍 If ever wanting to pass the full props object we should still stick with just `props` as naming  since that's the typical react convention. Have noticed `passedProps` often used in tests but that's a different since we're passing props from the `createWrapper` helper function down _to_ the component, which is different from props being referenced inside of the component.

src/Components/RelatedCollectionsRail/index.tsx
xtina-starr: Will do!

artsyit:

Built with commit be1d746ed26c5c53576503881a691e1c3cc3a95a

https://deploy-preview-2489--artsy-reaction.netlify.com

xtina-starr:
I think this is ready to go. @damassi @l2succes let me know if you have any additional feedback.

artsyit:
:rocket: PR was released in v16.17.16 :rocket:

PR #2495, opened 2019-06-04
[NavBar] Fix search-on-enter, SSR query persistence, missing admin / CMS dropdown links

damassi:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-968
Addresses https://artsyproduct.atlassian.net/browse/DISCO-925

It was noticed today that search wasn't submitting on enter, due to mismatch described in 968 and overlooked during NavBar rebuild. Along the way I noticed 925 and fixed it, which involves search bar contents not being preserved from the server render to client-side mount pass, resulting in a UX flicker.

Companion PR: https://github.com/artsy/force/pull/4149


src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -127,7 +127,7 @@ export const ArtworkBannerFragmentContainer = createFragmentContainer(
           ... on ArtworkContextFair {
             name
             href
-            is_active
+            isActive
```
damassi: Fixes a graphql deprecation, picked up by apollo tooling

artsyit:

Built with commit a6baab3a6213abe5b8c054bb1ac517633d9b1b13

https://deploy-preview-2495--artsy-reaction.netlify.com

src/Artsy/SystemContext.tsx
```diff
@@ -24,6 +24,7 @@ export interface SystemContextProps {
    * FIXME: Ask alloy how to pass one-off props like this in from force
    */
   notificationCount?: number
+  searchQuery?: string
```
damassi: I need to follow up this week and get proper app-level interfaces in that extend from SystemContext, and can be imported into force as a contract.

src/Components/NavBar/Menus/UserMenu.tsx
```diff
@@ -30,8 +34,23 @@ export const UserMenu: React.FC = () => {
     })
   }
 
+  const isAdmin = user.type === "Admin"
```
pepopowitz: There's a [`userIsAdmin` helper](https://github.com/artsy/reaction/blob/master/src/Utils/user.ts#L34) that can be used to make this decision. I know it's only one line, but if some day we change the rules for what makes an admin, it'd be nice to have those decisions all made in one place.

src/Components/Search/SearchBar.tsx
```diff
@@ -368,12 +368,18 @@ export class SearchBar extends Component<Props, State> {
 
   render() {
     return (
-      <>
+      <form
```
pepopowitz: 👍 🙏 ❤️

src/Artsy/SystemContext.tsx
pepopowitz: How does `searchQuery` get updated? I see it defined here, and then read [in `SearchBar.tsx`](https://github.com/artsy/reaction/pull/2495/files#diff-195cf326e4088c5f9ff2c6d47cc4248cR424), but it's not obvious to me how that value ever changes in the `SystemContext`. I'm certain it's me not understanding - can you explain?

damassi: Yup! It's passed in from the server, in the companion PR [here](https://github.com/artsy/force/pull/4149/files#diff-3df7ffd25fef9729a61746506a33251bR39).

src/Components/NavBar/Menus/UserMenu.tsx
damassi: Oh yeah, that's much nicer; will update

artsyit:
:rocket: PR was released in v16.13.6 :rocket:

PR #2498, opened 2019-06-04
DISCO-1022 - Updates zero state copy

kierangillen:
- https://artsyproduct.atlassian.net/browse/DISCO-1022
- Adds `hasFilter` boolean to Search `FilterState`
- Moves `Provider` to pass `FilterState` to ZeroState component
- Updates copy to search results using a filter
- Refactored https://github.com/artsy/reaction/pull/2467

## Copy changes

58472539-73953c80-8114-11e9-8405-d61594cf198f

### DIFF
--- a/src/Apps/Search/FilterState.tsx
+++ b/src/Apps/Search/FilterState.tsx
@@ -71,7 +71,7 @@ const isDefaultFilter = (filter, value): boolean => {
   }
 
   if (filter === "medium") {
-    return value === "*"
+    return value === "*" || !value
   }
 
   if (filter === "price_range" || filter === "height" || filter === "width") {
--- a/src/Apps/Search/FilterState.tsx
+++ b/src/Apps/Search/FilterState.tsx
@@ -83,7 +83,6 @@ const isDefaultFilter = (filter, value): boolean => {
 
 export class FilterState extends Container<State> {
   state = cloneDeep(initialState)
-  hasFilters = false
 
   static MIN_PRICE = 50
   static MAX_PRICE = 50000
@@ -116,8 +115,6 @@ export class FilterState extends Container<State> {
           default:
             this.state[filter] = value
         }
-
-        this.checkForFilters()
       })
     }
   }
@@ -160,8 +157,8 @@ export class FilterState extends Container<State> {
     )
   }
 
-  checkForFilters() {
-    this.hasFilters = Object.keys(this.state).some(filter => {
+  get hasFilters() {
+    return Object.keys(this.state).some(filter => {
       const value = this.state[filter]
       if (
         value &&
@@ -205,7 +202,6 @@ export class FilterState extends Container<State> {
 
     this.setState(newPartialState, () => {
       this.pushHistory()
-      this.checkForFilters()
     })
   }
 
@@ -244,7 +240,6 @@ export class FilterState extends Container<State> {
 
     this.setState({ page: 1, ...newPartialState }, () => {
       this.pushHistory()
-      this.checkForFilters()
     })
   }
 
--- a/src/Apps/Search/FilterState.tsx
+++ b/src/Apps/Search/FilterState.tsx
@@ -203,10 +203,9 @@ export class FilterState extends Container<State> {
       newPartialState = { medium: "*" }
     }
 
-    this.checkForFilters()
-
     this.setState(newPartialState, () => {
       this.pushHistory()
+      this.checkForFilters()
     })
   }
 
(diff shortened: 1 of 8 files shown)


artsyit:

Built with commit 96bd287dccc4015fc2fc9a38c4f3d4866d857fce

https://deploy-preview-2498--artsy-reaction.netlify.com

artsyit:

Built with commit 2d473a3a62bb5b6106a87e2b70ae6b56cc3c0c8b

https://deploy-preview-2498--artsy-reaction.netlify.com

src/Apps/Search/FilterState.tsx
```diff
@@ -83,6 +83,7 @@ const isDefaultFilter = (filter, value): boolean => {
 
 export class FilterState extends Container<State> {
   state = cloneDeep(initialState)
+  hasFilters = false
```
damassi: Why not add this directly into state? Plucking instance variables out from inside the ZeroState (versus accessing via `state` directly) seems to circumvent the natural flow.

pepopowitz: I would prefer a different change: 

* Remove this line and any calls to `checkForFilters`.
* Modify `checkForFilters` to return the value that indicates there are filters, instead of setting the instance variable to that value
* Make `checkForFilters` a getter ([like `filteredState`](https://github.com/artsy/reaction/pull/2498/files#diff-c86a075efb459e31ca5d188d283748d3R125))
* Rename `checkForFilters` to `hasFilters`, so that we don't need to make any changes to `ZeroState`

The current code is maintaining state (as an instance variable) that is based entirely on other already-maintained state. When something can be determined by other state, I think it's best to look it up when you need it, instead of maintaining _another_ piece of state. [The React docs talk about doing this when identifying state to manage at the app-level](https://reactjs.org/docs/thinking-in-react.html#step-3-identify-the-minimal-but-complete-representation-of-ui-state), but this is also a good practice at the individual component level.

kierangillen:
@pepopowitz Hey just updated with requested changes!

src/Apps/Search/FilterState.tsx
```diff
@@ -157,6 +157,24 @@ export class FilterState extends Container<State> {
     )
   }
 
+  get hasFilters() {
+    return Object.keys(this.state).some(filter => {
```
pepopowitz: I think this currently covers all the filters logically, but I think this section could be more clear & less likely to change if we reused the `isDefaultFilter` helper from above.

For example, something like:

```
  get hasFilters() {
    return Object.entries(this.state)
      .some(([key, value]) => {
        return !isDefaultFilter(key, value)  
      })
  }
```

As the `hasFilters` function currently stands, if we add more filters/remove one, we need to make sure it's covered correctly both here and in `isDefaultFilter`. With the rewrite suggested, we would only have to make sure it's covered in `isDefaultFilter`.

Sorry I didn't catch this sooner, but looking at it made me wonder if all those clauses covered everything correctly. (Side note - that probably means it's worth writing tests for.)

damassi:
@kierangillen - would you mind updating the animated screenshot in the description? Something more real world, like items displayed then narrowed down to the ZeroState.

damassi:
@kierangillen - would you mind updating the animated screenshot in the description? Something more real world, like items displayed then narrowed down to the ZeroState.

src/Apps/Search/FilterState.tsx
```diff
@@ -70,6 +70,10 @@ const isDefaultFilter = (filter, value): boolean => {
     return value === "-decayed_merch"
   }
 
+  if (filter === "medium") {
```
pepopowitz: @kierangillen could you do me a favor and verify that the browser history & URLs do what you'd expect them to when you select/deselect a medium filter?

kierangillen: Made a slight update and it's good now, you can see it in the new gif in the description.

src/Apps/Search/__tests__/FilterState.test.ts
```diff
@@ -54,24 +74,37 @@ describe("FilterState", () => {
...
-  inquireable_only: null,
-  price_range: "*-*",
-  height: "*-*",
-  width: "*-*",
-  attribution_class: [],
-  artist_id: null,
-  color: null,
-  keyword: null,
-}
+  it("returns a height range tuple based on the state", () => {
```
pepopowitz: Thank you for adding these!

damassi:
(Peril shouldn't block a merge here)

artsyit:
:rocket: PR was released in v16.14.1 :rocket:

PR #2503, opened 2019-06-05
[Collections] do not render featured artist module when there are not featured artists

xtina-starr:
This addresses a featured artists module bug reported in [slack](https://artsy.slack.com/archives/C9SATFLUU/p1559758088059500). 

When there are no artworks for a collection there in turn are no featured artists. Currently we display "0" when there are no featured artists which is a bug. This PR strengthens conditional to only display when we have at least one featured articles.

**Before:**
Screen Shot 2019-06-05 at 2 42 09 PM

**After:**
Screen Shot 2019-06-05 at 2 52 37 PM

### DIFF
--- a/src/Apps/Collect/Components/Collection/Header/index.tsx
+++ b/src/Apps/Collect/Components/Collection/Header/index.tsx
@@ -238,7 +238,7 @@ export const CollectionHeader: FC<Props> = ({ artworks, collection }) => {
                       lgOffset={isColumnLayout ? null : 1}
                       xlOffset={isColumnLayout ? null : 1}
                     >
-                      {featuredArtists.length && (
+                      {featuredArtists.length > 0 && (
                         <Box pb={10}>
                           <Sans size="2" weight="medium" pb={15}>
                             {`Featured Artist${hasMultipleArtists ? "s" : ""}`}


artsyit:

Built with commit bc5b62c4de77db966cba333f30bfc69e2e2da9fa

https://deploy-preview-2503--artsy-reaction.netlify.com

src/Apps/Collect/Components/Collection/Header/index.tsx
```diff
@@ -238,7 +238,7 @@ export const CollectionHeader: FC<Props> = ({ artworks, collection }) => {
                       lgOffset={isColumnLayout ? null : 1}
                       xlOffset={isColumnLayout ? null : 1}
                     >
-                      {featuredArtists.length && (
+                      {featuredArtists.length > 0 && (
```
l2succes: I'm not sure if lodash's `take` guards against that but is there any chance featuredArtists could be `null`?

xtina-starr: `take` returns an empty array if passed `null`.

xtina-starr: But the more I think about it I think this is pointing to a bigger issue because in theory collections should have artworks and in the case of `/collection/andy-warhol-jean-michel-basquiat` there are works in staging but there aren't any on production.

artsyit:
:rocket: PR was released in v16.14.2 :rocket:

PR #2505, opened 2019-06-06
[NavBar] Address QA items

damassi:
More updates from todays NavBar QA session. Comments / images inline.

### DIFF
--- a/src/Components/Input.tsx
+++ b/src/Components/Input.tsx
@@ -35,6 +35,10 @@ export const Input: React.ExoticComponent<InputProps> = React.forwardRef(
 export const StyledInput = styled.input`
   ${borderedInput};
   ${block(24)};
+
+  /* Remove shadow and default border radius on iOS */
+  background-clip: padding-box;
+  border-radius: 0;
 `
 
 export const Title = styled.div`
--- a/src/Components/NavBar/Menus/MobileNavMenu.tsx
+++ b/src/Components/NavBar/Menus/MobileNavMenu.tsx
@@ -10,15 +10,15 @@ import {
   MenuIcon,
   Separator,
   Serif,
+  space,
 } from "@artsy/palette"
 
 import { AnalyticsSchema, SystemContext } from "Artsy"
 import { useTracking } from "Artsy/Analytics/useTracking"
-import * as authentication from "Components/NavBar/Utils/authentication"
 
 export const MobileNavMenu: React.FC = () => {
   const { trackEvent } = useTracking()
-  const { mediator, user } = useContext(SystemContext)
+  const { user } = useContext(SystemContext)
   const isLoggedIn = Boolean(user)
 
   const trackClick = event => {
@@ -34,13 +34,7 @@ export const MobileNavMenu: React.FC = () => {
   }
 
   return (
-    <MobileNavContainer
-      py={[1, 4]}
-      width="100%"
-      height="100vh"
-      flexDirection="column"
-      onClick={trackClick}
-    >
+    <MobileNavContainer py={[1, 4]} flexDirection="column" onClick={trackClick}>
       <MobileLink href="/">Home</MobileLink>
       <MobileLink href="/artists">Artists</MobileLink>
       <MobileLink href="/shows">Shows</MobileLink>
@@ -61,10 +55,14 @@ export const MobileNavMenu: React.FC = () => {
         </>
       ) : (
         <>
-          <MobileLink onClick={() => authentication.login(mediator)}>
+          <MobileLink
+            href={`/log_in?intent=signup&signupIntent=signup&trigger=click&contextModule=Header`}
+          >
             Login
           </MobileLink>
-          <MobileLink onClick={() => authentication.signup(mediator)}>
+          <MobileLink
+            href={`/sign_up?intent=signup&signupIntent=signup&trigger=click&contextModule=Header`}
+          >
             Sign up
           </MobileLink>
         </>
@@ -76,7 +74,7 @@ export const MobileNavMenu: React.FC = () => {
 interface MobileLinkProps {
   children: React.ReactNode
   href?: string
-  onClick?: () => void
+  onClick?: (event?: React.MouseEvent<HTMLElement>) => void
 }
 
 const MobileLink: React.FC<MobileLinkProps> = ({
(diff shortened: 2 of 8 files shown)


src/Components/Input.tsx
```diff
@@ -35,6 +35,10 @@ export const Input: React.ExoticComponent<InputProps> = React.forwardRef(
 export const StyledInput = styled.input`
   ${borderedInput};
   ${block(24)};
+
+  /* Remove shadow and default border radius on iOS */
+  background-clip: padding-box;
+  border-radius: 0;
```
damassi: Screen Shot 2019-06-05 at 6 34 55 PM

src/Components/NavBar/Menus/MobileNavMenu.tsx
```diff
@@ -61,10 +55,14 @@ export const MobileNavMenu: React.FC = () => {
...
-          <MobileLink onClick={() => authentication.login(mediator)}>
+          <MobileLink
+            href={`/log_in?intent=signup&signupIntent=signup&trigger=click&contextModule=Header`}
+          >
             Login
           </MobileLink>
-          <MobileLink onClick={() => authentication.signup(mediator)}>
+          <MobileLink
+            href={`/sign_up?intent=signup&signupIntent=signup&trigger=click&contextModule=Header`}
+          >
```
damassi: These are actually direct links. Grabbed urls old nav bar [here](https://github.com/damassi/force/blob/8b7bdfa6f9b0a7cc8c9b7bd7bd2b38e2c3a1ed7c/src/mobile/components/layout/templates/login-signup.jade#L1).

artsyit:

Built with commit 4c26f5610284626b4621ae51a9bc89ef8e47d125

https://deploy-preview-2505--artsy-reaction.netlify.com

src/Components/NavBar/Menus/MobileNavMenu.tsx
```diff
@@ -113,8 +111,12 @@ const MobileNavContainer = styled(Flex)`
   background-color: white;
   border-bottom: 1px solid ${color("black10")};
   position: relative;
-  z-index: 2;
   user-select: none;
+  overflow-y: scroll;
+  z-index: 2;
+  width: 100%;
+  height: calc(100vh - ${space(6)}px);
+  -webkit-overflow-scrolling: touch;
```
damassi: Menu scrolls when open: 

scroll2

src/Components/NavBar/Menus/MobileNavMenu.tsx
pepopowitz: Is there any learning from here that we should apply other places? Like...I know we use the mediator to do logins many places, why are these different? And is there a documented heuristic for when we should use the mediator to handle logins vs navigating to a specific URL?

src/Components/NavBar/NavBar.tsx
```diff
@@ -232,6 +231,8 @@ export const NavBar: React.FC = track(
   )
 })
 
+export const NavBarHeight = 60 - 1 // border offset
```
pepopowitz: 1px nudges like this (and the -1px margin nudge below) make me wonder if something is wrong somewhere else.

pepopowitz: Also - does `space(6) - 1` work, instead of `60 - 1`? That would feel more respectful* of the design system.

(Respectful is the wrong word, but I can't figure out the work I'm looking for. Basically, having a hard-coded 60 feels like we are abandoning the design system, instead of using it to help us define spacing. It probably doesn't matter much, but it feels like all spacing should be driven by the design-system, in case we ever change it.)

src/Components/NavBar/NotificationsBadge.tsx
```diff
@@ -42,6 +42,12 @@ export const NotificationsBadge: React.FC<{
 
         let count = totalUnread
 
+        // User has no notifications; clear the cookie
```
pepopowitz: It seems like this and the branch it was swapped with are mutually exclusive, and order shouldn't matter. Is this move because you prefer to `return null` early, or am I misunderstanding the code?

src/Components/Search/SearchBar.tsx
```diff
@@ -252,14 +252,11 @@ export class SearchBar extends Component<Props, State> {
...
-  renderSuggestionsContainer = (
-    { containerProps, children, query },
-    { xs }
-  ) => {
+  renderSuggestionsContainer = ({ containerProps, children, query }) => {
     const noResults = get(this.props, p => p.viewer.search.edges.length === 0)
     const { focused } = this.state
 
-    if (noResults || !focused || !query || (xs && !query)) {
+    if (noResults || !focused || !query) {
```
pepopowitz: 💪

src/Components/NavBar/Menus/MobileNavMenu.tsx
damassi: So how we currently have it is there's the `/desktop` app and then there's `/mobile` (m.artsy), each with subtly different flows. When we rebuilt the logins a little over a year ago we built in full-page route-based login screen vs a popup, due to the small screens. Before this change I had made the mistake of assuming everything was now totally responsive but when a mobile useragent is detected, for many screens it will do a hard redirect to the mobile app, where these routes then kick in.

src/Components/NavBar/NavBar.tsx
damassi: Ahh yes, I agree; will update 👍

src/Components/NavBar/NotificationsBadge.tsx
damassi: I originally had this branch further below and for clarity reasons moved it up to the top. No actual behavior has changed.

src/Components/NavBar/NavBar.tsx
damassi: Re the `- 1` stuff, it's ugly and annoying and feels hacky :( The only reason it's needed is because there's an outline on the nav bar, and when we hover over nav items we want the menu to appear perfectly flush, aligning _its_ borders with the menu border so things don't double up.

src/Components/NavBar/Menus/MobileNavMenu.tsx
pepopowitz: Thank you for the explanation! 🙏

artsyit:
:rocket: PR was released in v16.14.3 :rocket:

PR #2518, opened 2019-06-07
[CollectionFeatureArtists] only use aggregate artists when there is not explicit artist_ids

xtina-starr:
Address: [GROW-1262](https://artsyproduct.atlassian.net/browse/GROW-1262) 

**Work In Progress:**
This PR handles displaying featured artist based on `artist_ids` if they are explicitly provide and then falling back to `merchandisable_artists` when there is no value for `artist_ids`.

### Before:
Screen Shot 2019-06-07 at 7 39 22 PM

### After:
Screen Shot 2019-06-07 at 7 40 00 PM

### DIFF
--- a/src/Apps/Collect/Components/Collection/Header/__tests__/Header.test.ts
+++ b/src/Apps/Collect/Components/Collection/Header/__tests__/Header.test.ts
@@ -1,0 +1,97 @@
+import { featuredArtistsEntityCollection, getFeaturedArtists } from "../index"
+
+const mockedFeaturedArtistsEntityCollection = featuredArtistsEntityCollection as jest.Mock
+
+describe("getFeaturedArtists", () => {
+  const mockMediator = jest.fn()
+
+  const collection = {
+    title: "KAWS: Toys",
+    query: {
+      gene_id: null,
+      artist_id: null,
+      artist_ids: ["4e934002e340fa0001005336"],
+    },
+    artworks: {
+      merchandisable_artists: [
+        {
+          id: "kaws",
+          _id: "4e934002e340fa0001005336",
+          name: "KAWS",
+          imageUrl:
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit ade620516c2b67eb3bab63f8c0c04586199c5c69

https://deploy-preview-2518--artsy-reaction.netlify.com

src/Apps/Collect/Components/Collection/Header/__tests__/Header.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    getFeaturedArtists(
+      9,
+      collection,
+      true,
+      collection.artworks.merchandisable_artists,
+      mockMediator,
+      {}
+    )
+
+    expect(mockedFeaturedArtistsEntityCollection).toBeCalledWith(
```
xtina-starr: @l2succes I mock `featuredArtistsEntityCollection` function on line 3 but it looks like it's still using the actual function. Is there something obvious I'm missing?

l2succes: @xtina-starr Here's what you're looking for https://github.com/facebook/jest/issues/936#issuecomment-214556122

You're not actually mocking the function on line 3, you're only type casting it with Typescript. Here's what you actually want

```typescript
import all, { getFeaturedArtists } from "../index"
const mockedFeaturedArtistsEntityCollection = all.featuredArtistsEntityCollection = jest.fn()
```

xtina-starr: @l2succes ah that makes sense. The file I'm testing doesn't have a default export so could I change the above to something like this:

```TypeScript
const Header = require("../index")
jest.unmock("../index")
Header.featuredArtistsEntityCollection = jest.fn()
```

artsyit:
:rocket: PR was released in v16.17.7 :rocket:

PR #2521, opened 2019-06-10
[WIP] Make MP version configurable

zephraph:
This is the first step towards allowing a reaction app to be converted to use MP v2. 

Once this is merged, the next step will be to update force for the artist app to use v2 and to fix all the things that break in reaction. 

I suspect there might be tooling and other infra that needs to be updated too. We shall see 😬. 

## TODO:

- [ ] Sync v2 schema
- [ ] Separate v2 compilation into own relay step
- [ ] Ensure current relay/sync schema commands still trigger everything


artsyit:

Built with commit d76549d813c561ac5c7c637918beab45fcad3e00

https://deploy-preview-2521--artsy-reaction.netlify.com

alloy:
Great! Indeed, the network layer needs to be configurable; however, Force shouldn’t really be in charge of what version an app should use, that's an internal detail to the app. This is especially true as this is not a matter of preference for Force, the Reaction app needs to be compiled with either a v1 or v2 schema.

zephraph:
Ahh, hmm... fair point. I'll have to look into how to separate the compilation step.

damassi:
Something like https://www.npmjs.com/package/babel-plugin-global-define?

src/Artsy/Relay/createRelaySSREnvironment.ts
```diff
@@ -16,6 +16,7 @@ import {
   RelayNetworkLayer,
   urlMiddleware,
 } from "react-relay-network-modern/node8"
+import { MetaphysicsVersion } from "Utils/metaphysics"
```
damassi: Since the network layer we don't use `Utils/metaphysics` anymore. Maybe this file can be deleted?

zephraph: It's still used in two places (time.ts, helpers.ts).

zephraph:
Going to close this in favor of #2531

PR #2522, opened 2019-06-10
PURCHASE-1145 Clean up unused fields pt 1

ds300:
- Adds a script to find potential unused field selections
- Begins work to remove some field selections which were discovered by the script

The script is susceptible to false positives, which so far seem to happen for one of two reasons:

1. We need to check for the presence of a node, or to find out whether or not an array is empty, but the only way to do that is by selecting at least one child field. At @alloy's suggestion I've made the script ignore '__typename' and '__id' to enable this use case.
2. The fields are used _indirectly_.

To explain that second point: the script uses the TypeScript language server's 'find references' feature, which only finds _direct_ references to a named property. e.g.

```ts
interface MyInterface {
  someField: string
}
function directReference(obj: MyInterface) {
  console.log(obj.someField)
}
function indirectReference(obj: {someField: string}) {
  console.log(obj.someField)
}
declare const obj: MyInterface
directReference(obj)
indirectReference(obj)
```

typescript will only pick up the reference to the usage in the `directReference` function.

Sometimes we can help TS out by sprinkling stronger type annotations around. I've been doing this where possible. But other times the data is being passed to untyped APIs or being used in subtly polymorphic ways that I don't want to deal with right now. So as a temporary measure I've made it possible to ignore specific false positives by adding an entry to an object in the script file.


artsyit:

Built with commit 85b6cb5ea042d32705da17dfd6200c680e65b214

https://deploy-preview-2522--artsy-reaction.netlify.com

src/Apps/Artwork/Components/ArtworkImageBrowser/index.tsx
```diff
@@ -46,17 +46,10 @@ export const ArtworkImageBrowserFragmentContainer = createFragmentContainer<
...
     fragment ArtworkImageBrowser_artwork on Artwork {
-      title
       image_alt: to_s
-      image_title
-      href
-
       ...ArtworkActions_artwork
       image {
         id
-        url(version: "larger")
```
sweir27: I think that these iare actually used by force for the View in Room feature (see usage [here](https://github.com/artsy/force/blob/master/src/desktop/apps/artwork/client.tsx#L86), passed in [here](https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx#L106))

But actually, in general we'll need to make sure that we're not removing fields that are necessary for hooks within force/other parent apps.

ds300: > But actually, in general we'll need to make sure that we're not removing fields that are necessary for hooks within force/other parent apps.

Absolutely, I'm keeping that in mind. I think in this case we're safe. Unless I'm missing something, this image object doesn't get passed anywhere else, even transitively, and it's url, height, and width are not referenced anywhere.

sweir27: I believe they're referenced from force when they are passed in [via the `mediator`](https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx#L114) component.

sweir27: Oh oh, but I guess in this case they are included in the `ArtworkActions` fragment container (missed that!).

ds300: Right, but ArtworkActions has it's own fragment where it selects these fields https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx#L286

scripts/find-unused-fields.ts
```diff
@@ -0,0 +1,349 @@
...
+    console.log()
+    console.log("USAGE")
+    console.log()
+    console.log(
+      chalk.cyan.bold("  yarn find-unused-fields"),
+      chalk.cyan("[fragment name]")
+    )
+    console.log()
+    console.log("Add known exceptions in find-unused-fields.ts")
+    console.log()
```
damassi: Why not use template tag strings? 
```
console.log(`
  🔎${chalk.bold("find-unused-fields")} 🔍

  USAGE: 
    ... 
`)
```

scripts/find-unused-fields.ts
```diff
@@ -0,0 +1,349 @@
...
+      .filter(
+        f =>
+          f.fileName.match("__generated__") &&
+          !f.fileName.endsWith("Query.graphql.ts")
+      )
+      .sort((a, b) => compare(basename(a.fileName), basename(b.fileName)))
+  }
+}
+
+class CheckContext {
```
damassi: As there's a lot of complex behavior in this class, mind add a comment at the top and / or comments inline?

scripts/find-unused-fields.ts
```diff
@@ -0,0 +1,348 @@
...
+      return
+    }
+
+    for (const property of node.members) {
+      const propertyName = property.name.getText()
+      // ignore meta properties
+      if (
+        propertyName.startsWith('"') ||
+        propertyName === "__typename" ||
+        propertyName === "id"
```
alloy: In reaction this field is still called __id

scripts/find-unused-fields.ts
```diff
@@ -0,0 +1,348 @@
...
+import * as fs from "fs"
+import * as json5 from "json5"
+import * as klawSync from "klaw-sync"
+import { flatten } from "lodash"
+import { basename, relative } from "path"
+import * as ts from "typescript"
+
+function main() {
+  if (process.argv.includes("--help") || process.argv.includes("-h")) {
+    console.log(`
```
zephraph: Just a side note, but I recommend using [dedent](https://www.npmjs.com/package/dedent) so you don't have to worry so much about the weird spacing issues introduced via the multi-line strings.

scripts/find-unused-fields.ts
```diff
@@ -0,0 +1,348 @@
...
+          "Try a relay fragment name or the path of a generated relay typings file."
+        )
+        process.exit(1)
+      }
+    }
+    return [
+      languageService
+        .getProgram()
+        .getSourceFile(
+          "src/__generated__/ArtworkImageBrowser_artwork.graphql.ts"
```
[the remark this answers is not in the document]
ds300: Whoops, this should have been `arg`. Thanks!

src/Apps/Artist/Routes/CV/CVItem.tsx
```diff
@@ -229,7 +229,7 @@ export const CVPaginationContainer = createPaginationContainer(
     getFragmentVariables(prevVars, totalCount) {
       return { ...prevVars, count: totalCount }
     },
-    getVariables(props, { count, cursor }, fragmentVariables) {
+    getVariables(props: CVItemProps, { count, cursor }, fragmentVariables) {
```
alloy: This shouldn’t be necessary and may indicate an enhancement possibility in the react-relay typings.

scripts/find-unused-fields.ts
```diff
@@ -0,0 +1,360 @@
...
+      return
+    }
+
+    for (const property of node.members) {
+      const propertyName = property.name.getText()
+      // ignore meta properties
+      if (
+        propertyName.startsWith('"') ||
+        propertyName === "__typename" ||
+        propertyName === "__id"
```
alloy: I was slightly mistaken. The Global ID field will _always_ be selected, if available, the `__typename` field will only automatically be selected on union and interface types. If we can get away with only allowing the Global ID field, then that would probably be best.

ds300: Isn't the global ID only on Node types while __typename is on everything?

alloy: Not necessarily limited to types that explicitly implement the `Node` interface; but yes not all types might have a `__id` field, whereas `__typename` exists on all. I guess that does make it harder, we should just somehow tell people to prefer the ID field, when available. Perhaps that’s a GraphQL validation rule we can make (and also surface through vscode-apollo) 🤔

src/Apps/Artist/Routes/CV/CVItem.tsx
ds300: Yeah the types look weird. They have the props type available as a generic parameter but seems like they deliberatly decide to avoid it.

image

Could also have just been overlooked while they were migrating from non-generic code?

alloy: (They is probably me :D)

alloy: Now that I mention it, I recall this is actually something that’s on my TODO list https://github.com/artsy/emission/blob/7a06739211a4545cdd56eed77a5619006cf9c722/patches/%40types/react-relay%2B1.3.9.patch

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
```
alloy: Did you test how this may affect other imports at runtime?

ds300: No, but it won't because we use babel everywhere else.

    "esModuleInterop": true

is just

    "allowSyntheticDefaultImports": true

plus tsc emitting the import shim that babel emits by default.

alloy: Cool cool, sounds good 👍

alloy:
@ds300 FYI Just to be sure, I’m assuming this is still a WIP as you still have this self-assigned.

ds300:
No I just had a few people to ask to review specific changes, so didn't want to risk it being merged before that, but I've got those now so will assign it to Justin unless your approval means you've double-checked all the deletions?

zephraph:
You can assign it to me. I'm gonna go through and do some manual QA.

zephraph:
Okay, I did some manual validation. I mostly focused on verifying no adverse changes to the artist and artwork pages (but I also looked around collect, collections, orders). 

I think this is good to go!

Great job.

alloy:
Oh btw, will we be running this on CI?

ds300:
Eventually, I hope! Need to:

- Make the script happy by doing the rest of the removals.
- Make it return 1 if it finds any potential errors. 
- Optionally see if we can do anything else to mitigate false positives.
- educate folks about the script, why it exists, how to help it succeed, and so on.

artsyit:
:rocket: PR was released in v16.17.0 :rocket:

PR #2526, opened 2019-06-10
DISCO-1070 - Better handling of long artist names/origins

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/DISCO-1070.

This PR modifies the `ArtistCard` component to better handle long artist names/origins, in two ways:

1. It removes the static `margin-top` from the Follow button, replacing it with `margin-top: auto`. This allows the button to flow to the end of the parent container, instead of being placed relative to the ending of the artist text.

2. It applies `text-overflow:ellipsis` (and associated styles) to the artist origin, so that it doesn't flow to a second line. I did not use the `<Truncator>` component, for the reason described at the end of this PR.

These changes are only applied to the `LargeArtistCard`; mobile is not affected.

### What it looks like now

1070-wide-desktop

### What it looked like before

1070-wide-before

### What it looks like now

1070-narrow-desktop

### What it looked like before

1070-narrow-before

## Why I didn't use the `<Truncator>` component for truncation/ellipsis

Weirdly, I found that using the `<Truncator>` component to handle long text on the artist origin resulted in premature truncation. Even when I had plenty of width, and there was plenty of room horizontally for the entire origin to show, the `Truncator` would truncate the year and show an ellipsis, like below:

1070-truncator

This is an issue with `react-lines-ellipsis`, but I'm not sure if it's a bug or working as intended. It is a non-ideal experience, though, so I opted to just use plain-old CSS to handle the ellipsis, and included an explanatory comment.

### DIFF
--- a/src/Components/v2/ArtistCard.tsx
+++ b/src/Components/v2/ArtistCard.tsx
@@ -73,7 +73,7 @@ export const LargeArtistCard: SFC<Props> = props => (
       <Sans size="2">{props.artist.formatted_nationality_and_birthday}</Sans>
     </Flex>
 
-    <Flex flexDirection="column" alignItems="center" mt={3}>
+    <Flex flexDirection="column" alignItems="center" mt="auto">
       <FollowArtistButton
         artist={props.artist}
         user={props.user}
--- a/src/Components/v2/ArtistCard.tsx
+++ b/src/Components/v2/ArtistCard.tsx
@@ -21,6 +21,7 @@ import {
   space,
   Spacer,
 } from "@artsy/palette"
+import styled from "styled-components"
 
 interface Props {
   artist: ArtistCard_artist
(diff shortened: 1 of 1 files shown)


src/Components/v2/ArtistCard.tsx
```diff
@@ -54,6 +55,17 @@ export class ArtistCard extends React.Component<Props> {
   }
 }
 
+// Why not use <Truncator>? Because with only one line for the artist origin/
+//  birth year, the Truncator prematurely truncates, showing "American, b. 1938"
+//  as "American, b. ...".
```
damassi: This probably doesn't require a comment as we use `text-overflow` in quite a few places. The truncator is specific to large bodies of text + read more.

artsyit:

Built with commit aaa7545631da4a367e3d04f257c413cdc6b3e25f

https://deploy-preview-2526--artsy-reaction.netlify.com

src/Components/v2/ArtistCard.tsx
pepopowitz: What do you think about me adding this JSDoc comment to the `Truncator` component, instead?

```
/**
 * Truncate large bodies of text.
 * If you need to truncate a single line, consider using `text-overflow: ellipsis`.
 */
export const Truncator: React.SFC<Props> = ({
....
```

I would see this as useful guidance for someone, like me, who was uncertain about when to use the Truncator and when to do it with plain CSS.

damassi: That's better 👍

zephraph:
Thank you so much!

This drove me a little crazy 🙃

artsyit:
:rocket: PR was released in v16.14.11 :rocket:

PR #2527, opened 2019-06-10
Update logic to handle hiding/displaying ads

ashleyjelks:
This PR removes the DisplayAd component when ad content isn't served.

It also addresses 2 small bug fixes where the incorrect dimensions were being passed to interstitial (canvas) ads standard article mobile views + featured articles not showing ads when the article doesn't contain image collections. 

https://artsyproduct.atlassian.net/browse/GROW-1285
https://artsyproduct.atlassian.net/browse/GROW-1291
https://artsyproduct.atlassian.net/browse/GROW-1292

### DIFF
--- a/src/Components/Publishing/Display/DisplayAd.tsx
+++ b/src/Components/Publishing/Display/DisplayAd.tsx
@@ -18,6 +18,8 @@ export interface DisplayAdProps extends FlexProps {
 }
 
 export const DisplayAd: SFC<DisplayAdProps> = props => {
+  let isAdEmpty
+
   const {
     adDimension,
     adUnit,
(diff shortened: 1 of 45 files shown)


artsyit:

Built with commit cd4b56de3bdcb941e41f3a204a48a56a5045b2b7

https://deploy-preview-2527--artsy-reaction.netlify.com

src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -31,20 +33,28 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
...
+  const ad = (
+    <GPT
+      adUnitPath={`/21805539690/${props.adUnit}`}
+      targeting={props.targetingData}
+      slotSize={[width, height]}
+      onSlotRenderEnded={e => {
+        isAdEmpty = e.isEmpty
+      }}
+    />
+  )
```
damassi: could probably return 
```
if (isAdEmpty) {
  return null
}
```
further up, before all of this stuff initializes. Then can avoid potential confusion in return statement.

src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -31,20 +33,28 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
   }
 
   const [width, height] = adDimension.split("x").map(a => parseInt(a))
+  const ad = (
+    <GPT
+      adUnitPath={`/21805539690/${props.adUnit}`}
+      targeting={props.targetingData}
+      slotSize={[width, height]}
+      onSlotRenderEnded={e => {
+        isAdEmpty = e.isEmpty
```
damassi: Inside of an SFC this `isAdEmpty` will potentially get reset on each render cycle. If you'd like to keep some state around, best to use state:

```js
const [isAdEmpty, setAdEmpty] = useState(false)

...
onSlotRenderEnded={event => {
  setAdEmpty(event.isEmpty)
}
```

src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -26,20 +26,28 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
...
+  const [isAdEmpty, setAdEmpty] = useState(false)
+
+  const ad = (
+    <GPT
+      adUnitPath={`/21805539690/${adUnit}`}
+      targeting={targetingData}
+      slotSize={[width, height]}
+      onSlotRenderEnded={event => {
+        setAdEmpty(event.isEmpty)
+      }}
```
l2succes: Not sure if it's actually needed but we may want to look into adding the `collapseEmptyDiv` option here 
See docs: https://github.com/nfl/react-gpt/blob/master/docs/api/ReactGPT.md#props

ashleyjelks: I saw that option, the reason I didn't elect to go that route was that the ad's parent div has a fixed height so it will display regardless of the ad div is collapsed (or set to display:none), so we need a way to remove the parent container.

artsyit:
:rocket: PR was released in v16.17.8 :rocket:

PR #2530, opened 2019-06-11
[NavBar] Add animation to dropdown

damassi:
Adds some light animation to the dropdowns per design request. Note the below isn't quite capturing it  due to frame-rate -- it's a slide down with a very very light bounce: 

animate

### DIFF
--- a/package.json
+++ b/package.json
@@ -203,7 +203,7 @@
     "react-router": "^4.2.0",
     "react-sizeme": "^2.4.4",
     "react-slick": "^0.23.2",
-    "react-spring": "^5.7.2",
+    "react-spring": "^8.0.23",
     "react-stripe-elements": "^2.0.1",
     "react-tracking": "^7.0.1",
     "react-transition-group": "^2.3.0",
--- a/src/Components/NavBar/NavBar.tsx
+++ b/src/Components/NavBar/NavBar.tsx
@@ -59,9 +59,6 @@ export const NavBar: React.FC = track(
     }
   }, [isMobile])
 
-  // FIXME: Remove test
-  const [isWide, toggleMenuLayout] = useState(true)
-
   return (
     <>
       <NavBarContainer px={1}>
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -202,7 +202,7 @@
     "react-router": "^4.2.0",
     "react-sizeme": "^2.4.4",
     "react-slick": "^0.23.2",
-    "react-spring": "^5.7.2",
+    "react-spring": "^8.0.23",
```
damassi: We're not actually using this anywhere, so version bump is ok

artsyit:

Built with commit 96be8d69b757f7e17b224b30b0b1af1f921371e9

https://deploy-preview-2530--artsy-reaction.netlify.com

package.json
kierangillen: Do we want to keep it consistent with Palette so we're not using two versions on import https://github.com/artsy/palette/blob/12597a2ec97fc760e4902229d9156aa444a9a7ef/packages/palette/package.json#L124 ?

damassi: @sepans has a ticket to update Palette to latest react spring, so I think its ok if we have two versions for now, since we'll be migrating to the latest.

kierangillen:
@damassi Do we know why this PR is reintroducing `moment-timezone`, as a dependency of `moment`?

damassi:
Hmm, that looks like a weird false positive (or something?); looking at the `yarn.lock` diff there's no reference. I'll rebase from master just to be sure.

kierangillen:
Looks good! Feel free to merge when `moment-timezone` is verified as not a change.

artsyit:
:rocket: PR was released in v16.16.1 :rocket:

zephraph:
I suspect that something related to (or coincidentally triggered by) this change  is causing staging to fail. 

The errors coming from sentry:

```
TypeError: (0 , _reactSpring.useSpring) is not a function
```

kierangillen:
@damassi @zephraph Seeing some others having the same issue, possible fix - https://github.com/react-spring/react-spring/issues/354#issuecomment-461910298 ? Could do a quick PR and see if this fixes it.

zephraph:
Sure, sounds good. I've done a revert #2532 just to get staging back up. You can revert that and add this change to see if it fixes it.

PR #2535, opened 2019-06-12
Removes jsx and js from test regex

kierangillen:
- Removes `jsx` and `js` files from test regex as all files should be typescript

### DIFF
--- a/package.json
+++ b/package.json
@@ -240,7 +240,7 @@
       "^.+\\.(ts|tsx|js|jsx)$": "babel-jest",
       "\\.graphql$": "jest-raw-loader"
     },
-    "testRegex": "/__tests__/.+?\\.test\\.(ts|tsx|js|jsx)$",
+    "testRegex": "/__tests__/.+?\\.test\\.(ts|tsx)$",
     "testPathIgnorePatterns": [
       "<rootDir>/dist/",
       "<rootDir>/externals/"


artsyit:

Built with commit 4230c3c2a6e43544d08384917c0d317c0bd6f851

https://deploy-preview-2535--artsy-reaction.netlify.com

package.json
```diff
@@ -240,7 +240,7 @@
       "^.+\\.(ts|tsx|js|jsx)$": "babel-jest",
       "\\.graphql$": "jest-raw-loader"
     },
-    "testRegex": "/__tests__/.+?\\.test\\.(ts|tsx|js|jsx)$",
+    "testRegex": "/__tests__/.+?\\.test\\.(ts|tsx)$",
```
eessex: I think we could simplify this further by removing the folder designation -- the file extension should do it, `"testRegex": "(.test)\\.(ts|tsx)$"`

kierangillen: True, but I think keeping the folder designation will help ensure proper folder structure, so I'd lean on keeping it personally.

eessex: Fair enough, but it also means that if someone names the folder wrong, it doesn't get run -- we should find some way to check folder naming on pre-commit or something in the future... Force had a related issue recently where many directories were never being run because of typos like no plural "s" in "tests"

alloy: That could be a Peril rule.

artsyit:
:rocket: PR was released in v16.17.4 :rocket:

package.json
eessex: @alloy I'd be down to pair on this with you!

alloy: File a ticket and let’s do it (next week?)

eessex: https://artsyproduct.atlassian.net/browse/PLATFORM-1543

PR #2541, opened 2019-06-13
Scroll to top on all search results pagination

kierangillen:
- After clicking `next` or a different page number on a search results' tab the page should scroll to top.
- Had implemented a fix for this on Artworks without adding to all tabs

### DIFF
--- a/src/Apps/Artist/Routes/Articles/ArtistArticles.tsx
+++ b/src/Apps/Artist/Routes/Articles/ArtistArticles.tsx
@@ -66,9 +66,12 @@ export class ArtistArticles extends Component<
   }
 
   toggleLoading = isLoading => {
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit c23abea84394d2165a7716c25282247c5c652a40

https://deploy-preview-2541--artsy-reaction.netlify.com

kierangillen:
@ashfurrow  I agree there could be some refactoring but looking at adding a callback to each file seems to be comparable to the existing code. I think a serious refactor would be necessary to really cut down on the line count, might be worth it but I don't think for this PR, so going to merge.

artsyit:
:rocket: PR was released in v16.17.5 :rocket:

PR #2549, opened 2019-06-13
Extract action sheet management from Artist/state into Artist/ArtworkFilter

pepopowitz:
This PR is early stage preparation for unifying the 3 different-but-similar FilterState components. It moves the management of the mobile action sheet from Artist/...state.ts into Artist/...ArtworkFilter.ts, so that the three filter state components are even more similar. 

* Tests were added to guide the refactor. 
* Test fixture data was moved into the Fixtures folder.
* I did not modernize ArtworkFilter or state.ts with hooks, because I was more focused on moving the outliers out of state.ts. Hooks conversion will come in future work.

### DIFF
--- a/src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/__tests__/ArtworkFilter.test.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/__tests__/ArtworkFilter.test.tsx
@@ -1,14 +1,20 @@
 // FIXME: Uncomment tests, currently WIP
 
-import { ArtworkFilter_Test_Query } from "__generated__/ArtworkFilter_Test_Query.graphql"
+import { ArtworkFilterFixture } from "Apps/__tests__/Fixtures/Artist/Components/ArtworkFilter"
 import { ArtworkFilterFragmentContainer as ArtworkFilter } from "Apps/Artist/Routes/Overview/Components/ArtworkFilter"
 import { FilterState } from "Apps/Artist/Routes/Overview/state"
-import { MockBoot, MockRelayRenderer, renderUntil } from "DevTools"
+import { MockBoot, renderRelayTree } from "DevTools"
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit c51c0af60d35c856ccc0f9b566a99575d4fa60d4

https://deploy-preview-2549--artsy-reaction.netlify.com

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/__tests__/ArtworkFilter.test.tsx
```diff
@@ -17,154 +23,76 @@ describe("ArtworkFilter", () => {
...
+      await flushPromiseQueue()
+      wrapper.update()
+
+      expect(wrapper.find("Filters").length).toEqual(0)
+    })
+
+    function findButtonWithText(wrapper, buttonText) {
+      return wrapper
+        .find("Button")
+        .findWhere(button => {
```
damassi: Did not know about `findWhere` 👍

src/Apps/Artist/Routes/Overview/Components/ArtworkFilter/index.tsx
```diff
@@ -57,6 +65,16 @@ class Filter extends Component<Props> {
     return !this.existy.hasArtworks
   }
 
+  hideMobileActionSheet = () => {
+    this.setState({ showMobileActionSheet: false })
+    document.body.style.overflowY = "auto"
```
damassi: As an aside, this was sort of a get it out the door hack; I'm sure there are better ways of handling this if need be, via something like react-side-effect. Could also be fine for what it is!

damassi:
Looks great, and thanks for adding the additional tests. Things already feel simpler 👍

artsyit:
:rocket: PR was released in v16.17.9 :rocket:

PR #2550, opened 2019-06-13
Add more context to mock relay errors (the operation name)

pepopowitz:
While writing a test using `renderRelayTree`, I found that I failed to set up some mock data, but the error message didn't fully point me to _which_ mock data I needed to set up. Original slack conversation: https://artsy.slack.com/archives/C2TQ4PT8R/p1560442571197200

This PR adds a little more context to the error messages returned by the mock network layer - specifically, the operation name. In my specific situation, that information would have tipped me off that there was another query I wasn't considering in my test.

### DIFF
--- a/src/DevTools/createMockNetworkLayer/index.ts
+++ b/src/DevTools/createMockNetworkLayer/index.ts
@@ -8,6 +8,7 @@ import {
 import { IMocks } from "graphql-tools/dist/Interfaces"
 import getNetworkLayer from "relay-mock-network-layer"
 import { Network, RelayNetwork } from "relay-runtime"
+import { get } from "Utils/get"
 import uuid from "uuid"
 import schema from "../../../data/schema.graphql"
 import FormattedNumber from "./CustomScalars/formatted_number"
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 5d8c25ca6326ac1f1eaad8517f1202660f0bf4ad

https://deploy-preview-2550--artsy-reaction.netlify.com

src/DevTools/createMockNetworkLayer/index.ts
```diff
@@ -195,19 +197,24 @@ const inferUnionOrInterfaceType = (
...
+    type: string
+    path: string
+    operationName: string
+  }) => string
 ) {
   return new Error(
     renderMessage({
       path: responsePathAsArray(info.path).join("/"),
       type: info.returnType.inspect(),
+      operationName: get(info, i => i.operation.name.value, "(unknown)"),
```
pepopowitz: I don't have a test covering the "unknown" case. I don't really know what it would mean to not know the operation name. `info.operation.name` _is_ optional, though, so I took the safe route and used `get`.

alloy: I guess this is for when you don’t name an operation (which we shouldn’t be doing). I.e.

```graphql 
query {
  # ...
}
```

alloy: So perhaps it would be good to take that into account when printing the helpful messages.

ds300: @alloy I wonder if that should be a lint rule instead?

ds300: ah it already is

artsyit:
:rocket: PR was released in v16.17.10 :rocket:

PR #2565, opened 2019-06-18
GROW 1277 Draft

ashleyjelks:


src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -51,8 +51,32 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, State> {
     return isReadMoreExpanded || hasNoBio
   }
 
+  shouldShowCollectionsRail(collections, artistID): boolean {
+    const hasFeaturedArtistContent: boolean = collections.some(
+      collection => collection.is_featured_artist_content
+    )
+
+    const artistInCollectionQuery: boolean = collections.some(
```
damassi: One thing to note is that TS is smart enough to infer the type from the assignment -- meaning, since `array.some` can only return a boolean, the variable `artistInCollectionQuery` can only ever be a boolean. 

That said, these kinds of assignments are often good for readability.

ashleyjelks: > `array.some` can only return a boolean

Good point!

PR #2568, opened 2019-06-19
[WIP] Refactor search filterstate

pepopowitz:
This PR is..kind of a mess right now, but it shows where I'm headed with refactoring the FilterState components so that they can be better consolidated.

I've got `TODO` comments littered throughout the code that I need to clean up yet, and I'm tracking my work & thoughts in https://github.com/artsy/reaction/pull/2568/files?file-filters%5B%5D=#diff-2ad23fb76afe1c3b7015a7edda9e0662.

There are two problems I'm aware of with the FilterState stuff that I'm aiming to address: 

1. We aren't comfortable using `unstated` state containers anymore. This PR converts to a [React `context`](https://github.com/artsy/reaction/pull/2568/files?file-filters%5B%5D=#diff-b3a8cbc7267e0ba8d8866d60d5d11940), instead.

2. We've been bitten by the fact that there are multiple very-similar (and identically named) state containers in our app. You might think you're subscribing to the correct FilterState container, but find out that you're not when none of your actions have the state updates that you think they'd have. This PR **does not** address this yet, but the plan is for me to get this search/FilterContext into a position where it'd be easier to consolidate across multiple state containers. I'm finding that this might be pretty hard to do without making some significant abstractions, which might make it _more difficult_ to maintain. I have considered that maybe it makes sense _not_ to consolidate, but simply renaming the different state containers so that they can't be mistaken for each other would solve the problem.

This work is snowballing a bit. In addition, because I'm rewriting the entire state container for the search app, I'm introducing bugs. I'm tracking them as I find them and I've added a ton of testing around things (including some Cypress regression tests I have locally that are very brittle but have already caught some regressions). I'd love some feedback on whether my approach makes sense, whether the amount of snowballing is expected and I should just power through it, and if the 2 problems I described above are truly what we're trying to solve. Thanks for the 👀 !

### DIFF
--- a/src/Apps/Search/FilterContext.tsx
+++ b/src/Apps/Search/FilterContext.tsx
@@ -3,6 +3,7 @@ import { filterReducer } from "./filterReducer"
 
 export interface FilterContextValues {
   setFilter: (name: string, value: any) => void
+  unsetFilter: (name: string) => void
   hasFilters: boolean
   filters: Filters
 }
@@ -19,6 +20,9 @@ export const FilterContext = React.createContext<FilterContextValues>({
     page: 1,
   },
   setFilter() {
+    console.error("Shouldn't have gotten here.")
+  },
+  unsetFilter() {
     console.error("Shouldn't have gotten here.")
   },
   hasFilters: false,
@@ -54,6 +58,7 @@ export const FilterContextProvider: FC<FilterContextProps> = ({
     hasFilters: state.page !== 1,
     setFilter: (name, val) =>
       dispatch({ type: "set", payload: { name, value: val } }),
+    unsetFilter: name => dispatch({ type: "unset", payload: { name } }),
   }
   return (
     <FilterContext.Provider value={value}>{children}</FilterContext.Provider>
--- a/src/Apps/Search/FilterContext.tsx
+++ b/src/Apps/Search/FilterContext.tsx
@@ -1,4 +1,5 @@
 import React, { FC, ReactNode, useContext, useEffect, useReducer } from "react"
+import { useDidMount } from "Utils/Hooks/useDidMount"
 import { filterReducer } from "./filterReducer"
 
 export interface FilterContextValues {
@@ -47,10 +48,13 @@ export const FilterContextProvider: FC<FilterContextProps> = ({
   }
 
   const [state, dispatch] = useReducer(filterReducer, initialFilters)
+  const isMounted = useDidMount()
 
   useEffect(() => {
     // TODO - do this correctly.
-    console.log("page changed; push history here.")
+    if (isMounted) {
+      console.log("page changed; push history here.")
+    }
   }, [state.page])
 
   const value = {
--- a/src/Apps/Search/FilterContext.tsx
+++ b/src/Apps/Search/FilterContext.tsx
@@ -14,6 +14,8 @@ interface FilterContextProps {
   keyword?: string
   page?: number
   medium?: string
+  for_sale?: boolean
+  offerable?: boolean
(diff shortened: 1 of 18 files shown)


src/Apps/Search/FilterContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+  unsetFilter: (name: string) => void
+  hasFilters: boolean
+  filters: Filters
+  resetFilters: () => void
+  isDefaultValue: (name: string) => boolean
+  rangeToTuple: (name: string) => [number, number]
+}
+
+interface FilterContextProps {
+  children: ReactNode
```
pepopowitz: The repetition of this shape 5 times in this file irritates me, and is something I'd plan to address.

src/Apps/Search/FilterState.tsx
```diff
@@ -123,16 +124,13 @@ export class FilterState extends Container<State> {
...
   }
 
+  /* sjhsjhsjh done */
   get hasFilters() {
     return Object.entries(this.state).some(([key, value]) => {
       return !isDefaultFilter(key, value)
     })
   }
 
-  get filteredState() {
```
pepopowitz: This wasn't used anywhere.

src/Apps/Search/Routes/Artworks/Components/Filter/Filters/ColorFilter.tsx
```diff
@@ -58,134 +54,131 @@ const CheckmarkPositions = {
   },
 }
 
-export class ColorFilter extends React.Component<Props> {
-  toggleColor(color) {
-    const { filters } = this.props
-    if (filters.state.color === color) {
-      filters.unsetFilter("color")
+export const ColorFilter: FC = () => {
```
pepopowitz: All of these individual filter components were rewritten as FC's, using `useFilterContext` instead of having the filters passed in as props. This probably isn't necessary for the refactor, but it made me feel good.

src/Apps/Search/Routes/Artworks/Components/Filter/Filters/FilterContainer.tsx
```diff
@@ -140,28 +141,28 @@ export class FilterContainer extends React.Component<
 
                   <Spacer mb={2} />
 
-                  {this.props.children(filters)}
+                  {this.props.children(context)}
```
pepopowitz: It's possible that `context` doesn't need to be passed into `children` anymore. I haven't tracked that down yet.

damassi:
This is looking 💯 @pepopowitz 👍Just what I imagined. 

Mind updating the description with the game-plan discussed offline? As mentioned, I'm going to be offline for a few weeks but I can help pick up anything that may be needed when i get back (test writing, qa, features, whatever). It looks like we're pretty close here.

src/Apps/Search/FilterContext.tsx
damassi: Cool, yeah this looks like doing something as simple as 
```ts
FilterContextProps extends Filters {
  children: ReactNode
}
``` 
or
```ts
FC<Filters & { children: ReactNode }> { ... }
```
would be sufficient.

src/Apps/Search/Routes/Artworks/Components/Filter/Filters/ColorFilter.tsx
damassi: Lovely 👌

pepopowitz:
Closing, as https://github.com/artsy/reaction/pull/2572 replaces this.

PR #2572, opened 2019-06-21
[ArtworkFilter Refactor] Extract and consolidate

damassi:
This PR is a fork and further refactor of https://github.com/artsy/reaction/pull/2568. It does a few things:
- Extracts / restructures the filter from the Search implementation into `Components/v2/ArtworkFilter`, and removes all legacy code
- Creates a Storybook entry
- Collapses / simplifies the voluminous amount of graphql field / argument / component duplication from the currently live implementation into two components, a container which handles filters, and a grid which handles pagination. 

Once this is 🍏 #2568 can be closed.  

#### Todo
- [x] Fix types 
- [x] Re-add tracking via hooks
- [x] Test in Force 
- [ ] Follow up on `relay-operation-generics` linting error


src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -0,0 +1,168 @@
...
+  resetFilters: () => void
+  setFilter: (name: string, value: any) => void
+  unsetFilter: (name: string) => void
+}
+
+interface FilterContextProps extends Filters {
+  children: ReactNode
+}
+
+// TODO - consolidate these initial values....
```
damassi: Note that `TODO -` formatting like this isn't recognized by editor highlighting; need to write `TODO:`

artsyit:

Built with commit 76517dccf04a669dec55719487baee65d74755cf

https://deploy-preview-2572--artsy-reaction.netlify.com

src/Components/v2/ArtworkFilter/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+      current: filterContext.filters,
+      changed: {
+        [filterKey]: filterContext.filters[filterKey],
+      },
+    })
+
+    toggleFetching(true)
+
+    // FIXME: Re-enable fetch updates. Throws strange low-lever error ATM
+    return
```
damassi: @pepopowitz - Hit a bit of a blocker here when making the request to update the grid based on filter change; an error comes back about `aggregations`. I don't have any more time to look at this ATM but feel free to poke at it a bit if you want. I can give it a look when I get back.

damassi:
Ok @pepopowitz - I ended up refactoring this pretty extensively to basically pull it completely out of it's old form (no more renderProps, classes, etc). I also cleaned up some stuff here and there. I think it's in really good shape to merge and for us to pick up working on it when we can. I ran into one blocker as noted above (that's likely something silly), and obviously we need to test in force and knock out all of the kinks, but the over all shape is now where we want it to be.

src/Components/v2/ArtworkFilter/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+export const ArtworkFilterQueryRenderer = ({ term = "andy warhol" }) => {
+  const { relayEnvironment } = useSystemContext()
+
+  return (
+    <FilterContextProvider keyword={term} {...initialFilterState}>
+      <QueryRenderer<ArtworkFilterQueryType>
+        environment={relayEnvironment}
+        // FIXME: Passing a variable to `query` shouldn't error out in linter
+        /* tslint:disable:relay-operation-generics */
+        query={ArtworkQueryFilter}
```
damassi: @alloy - noticed that the linter doesn't like variables passed to `query`. Was there a specific reason for this?

alloy: Unsure what you mean by “passing a variable to `query`”. Can you provide an example and the error?

src/Components/v2/ArtworkFilter/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+        variables={{
+          term: "andy warhol",
+        }}
+        render={renderWithLoadProgress(ArtworkFilterFragmentContainer as any)}
```
alloy: This `as any` should not be necessary.

src/Components/v2/ArtworkFilter/ArtworkFilterArtworkGrid2.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { PaginationFragmentContainer as Pagination } from "../Pagination"
+import { useFilterContext } from "./ArtworkFilterContext"
+import { ArtworkFilterZeroState } from "./ArtworkFilterZeroState"
+
+interface ArtworkFilterArtworkGridProps {
+  columnCount: number[]
+  filtered_artworks: ArtworkFilterArtworkGrid2_filtered_artworks
+  isLoading?: boolean
+  relay: RelayRefetchProp
+  term: string
```
pepopowitz: `term` is specific to the Search app's usage of the artwork grid. We'll need to figure out a way to opt into this prop only for the apps that need it. 

Maybe that's just marking it as optional, and letting the apps decide what they pass in. When we get to the artist app, that one would pass in an optional `artist` or `artist_id`, too.

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -0,0 +1,184 @@
+import React, { useContext, useEffect, useReducer } from "react"
+import { useDidMount } from "Utils/Hooks/useDidMount"
+import { hasFilters } from "./Utils/hasFilters"
+import { isDefaultFilter } from "./Utils/isDefaultFilter"
+import { rangeToTuple } from "./Utils/rangeToTuple"
+import { urlFragmentFromState } from "./Utils/urlFragmentFromState"
+
+export interface Filters {
```
pepopowitz: Using this same shape for both the props passed into the provider and the filters available to consumers might be problematic. `major_periods`, for example, should not be required to be passed into the provider, but it should be always included for the consumers (because it's an array, and it's just easier to deal with it as an empty array).

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const isMounted = useDidMount()
+  // const updateOnChange = Object.keys(state).map(prop => state[prop])
+
+  useEffect(() => {
+    if (isMounted) {
+      const queryString = urlFragmentFromState(state)
+      window.history.pushState(
+        {},
+        null,
+        `${window.location.pathname}?${queryString}`
```
pepopowitz: Long-term this URL construction algorithm is another thing that will vary across apps. Most can use this string format, but the [Collect app's format](https://github.com/artsy/reaction/blob/9555125ef04e9af1f23a45990ec41cd893b94341/src/Apps/Collect/urlBuilder.ts#L15) is slightly different - because a selected medium factors in as part of the path, not the querystring.

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -0,0 +1,184 @@
...
+      }
+    }
+    case "UNSET": {
+      return {
+        ...state,
+        [action.payload.name]: initialFilterState[action.payload.name],
+      }
+    }
+    case "RESET": {
+      return initialFilterState
```
pepopowitz: I don't think this implementation of resetFilters will work. It will clobber the value of the `term` prop, which was passed in.

src/Components/v2/ArtworkFilter/ArtworkFilterZeroState.tsx
```diff
@@ -0,0 +1,40 @@
...
+    >
+      <Box m={3} textAlign="center">
+        <Serif size="6">
+          {hasFilters
+            ? "No results found."
+            : `No results found for "${props.term}".`}
+        </Serif>
+        <Serif size="3">
+          {hasFilters
+            ? "Try removing some filters or try another search term."
```
pepopowitz: This messaging is going to vary across apps, also. This current message makes sense for the Search app, but none of the others.

src/Components/v2/ArtworkFilter/Utils/urlFragmentFromState.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { isDefaultFilter } from "./isDefaultFilter"
+
+/**
+ * Returns a string representing the query part of a URL. It removes default
+ * values, and rewrites keyword -> term.
+ */
+export const urlFragmentFromState = (state: Filters) => {
+  const { keyword: term } = state
+  const filters = Object.entries(state).reduce((acc, [key, value]) => {
+    if (isDefaultFilter(key, value) || key === "keyword") {
```
pepopowitz: More `term`/`keyword` stuff that we'll have to make work with apps other than Search.

src/Components/v2/ArtworkFilter/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+          width: $width
+        )
+    }
+  }
+`
+
+interface ArtworkFilterProps {
+  viewer: ArtworkFilter_viewer
+  relay: RelayRefetchProp
+  term: string
```
pepopowitz: More `term`/Search-App-specific props

src/Components/v2/ArtworkFilter/Utils/isDefaultFilter.tsx
```diff
@@ -0,0 +1,19 @@
+export const isDefaultFilter: (name: string, value: any) => boolean = (
```
pepopowitz: I wonder if we can make this helper go away, based on the way you simplified the reducer `UNSET` action? It's accomplishing something very similar - trying to identify if something is set to its initial value or something different.

src/Components/v2/ArtworkFilter/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+  const tracking = useTracking()
+  const [isFetching, toggleFetching] = useState(false)
+  const [showMobileActionSheet, toggleMobileActionSheet] = useState(false)
+  const filterContext = useFilterContext()
+  const previousFilters = usePrevious(filterContext.filters)
+
+  /**
+   * Check to see if the mobile action sheet is present and prevent scrolling
+   */
+  useEffect(() => {
```
pepopowitz: 👌

src/Components/v2/ArtworkFilter/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+      setScrollable(true)
+    }
+  }, [showMobileActionSheet])
+
+  /**
+   * Check to see if the current filter is different from the previous filter
+   * and trigger a reload.
+   */
+  useEffect(() => {
+    Object.entries(filterContext.filters).forEach(
```
pepopowitz: This is neat. I like the hooks way of detecting if props have changed & determining if a refetch is necessary. I wonder if there's a custom hook in here about detecting changes to specific props.

src/Components/v2/ArtworkFilter/index.tsx
pepopowitz: He's referring to the variable named `ArtworkQueryFilter`. It's a variable because the same query is used twice, in both the fragmentContainer and queryRenderer. Instead of copy-pasting the same query and naming it different things, he's reusing the query definition in two places (here and https://github.com/artsy/reaction/pull/2572/files/9555125ef04e9af1f23a45990ec41cd893b94341#diff-822e8fa213e43cff5a5a98e1ce82eaeaR223).

The error without that `lint:disable` line is `expected a graphql`…` tagged-template tslint(relay-operation-generics)`.

pepopowitz:
> There is a small spelling error that I'm not sure is used. I'll correct that spelling and see how things go when this new subtree is used in Force before I merge this. Since it's a brand new app, I feel much better about merging it while in-progress.

- [x] I corrected the spelling error
- [x] I verified that I was able to connect force to the new component! I stashed that work locally, because it falls victim to the same problems I was having yesterday, where it's difficult to introduce a side-by-side path in the Search app based on a query-string param, and my work really just replaces the existing connection. 

I have a couple things I want to add to this - a README.md in this component's folder to give people an idea that it is WIP (and track the things we know we need to do yet), and pull back in the tests for the ArtworkFilterContext. Depending on time, I might skip that second one and merge sooner.

pepopowitz:
I'm merging this on green so that it doesn't sit for the next week+ while @damassi and I are out. It is currently not utilized by any active routes on the site, and is intentionally being committed as in-progress work.

artsyit:
:rocket: PR was released in v16.17.24 :rocket:

src/Components/v2/ArtworkFilter/index.tsx
damassi: Yup, the linter is looking for 

```
query={graphql` ... `}
```
or at least a "`".

src/Artsy/Analytics/Schema/index.ts
```diff
@@ -17,6 +17,14 @@ import { Label } from "./Label"
 import { Failure, Success } from "./Result"
 import { Type } from "./Type"
 
+interface Uncatagorized {
```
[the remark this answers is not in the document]
damassi: My spelling is so embarrassingly bad :/

damassi:
Thanks for the eagle eyes review @pepopowitz -- We'll def have to tighten some of the app-specific stuff up later. For things like the ZeroState I imagine we can make that generic, or pass it in as a prop; for some of the minor query differences we'll have to be a little more clever, since Relay queries are static. But these things are expected and we're in a good place now to figure this out. (I know @alloy has some ideas about how to leverage local schema extensions which we might be able to holistically work in now that our code is in a good place.)

src/Components/v2/ArtworkFilter/ArtworkFilterArtworkGrid2.tsx
```diff
@@ -0,0 +1,158 @@
...
+   * Refetch page of artworks based on cursor
+   */
+  function loadAfter(cursor, page) {
+    setLoading(true)
+
+    relay.refetch(
+      {
+        first: RefetchPageSize,
+        after: cursor,
+        filteredArtworksNodeID: this.props.filtered_artworks.__id,
```
damassi: Ahh -- stray `this`

src/Components/v2/ArtworkFilter/ArtworkFilterArtworkGrid2.tsx
```diff
@@ -0,0 +1,158 @@
...
+import React, { useEffect, useState } from "react"
+import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
+import { LoadingArea } from "../LoadingArea"
+import { PaginationFragmentContainer as Pagination } from "../Pagination"
+import { useFilterContext } from "./ArtworkFilterContext"
+import { ArtworkFilterZeroState } from "./ArtworkFilterZeroState"
+
+interface ArtworkFilterArtworkGridProps {
+  columnCount: number[]
+  filtered_artworks: ArtworkFilterArtworkGrid2_filtered_artworks
```
damassi: We need to fix a naming collision inside the `/artist` app by renaming that component to `ArtistArtworkFilterArtworkGrid`.

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
damassi: I think the way to go here is to tap into `farce` directly, and dispatch update events from there. This is a bit lower-level than the router (the router plugs into it) and it's more global. (I know we've gone down the "use the router" path before but I still have a hunch that we can make it work. If the component exists outside of a formal app (meaning, there's no router context available within the tree) then we can defer to the `window`. (Just brainstorming; this is a `TODO` for later.)

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -0,0 +1,184 @@
...
+      window.history.pushState(
+        {},
+        null,
+        `${window.location.pathname}?${queryString}`
+      )
+    }
+  })
+
+  /**
+   * TODO: use this to trigger resetting of fields when back/fwd buttons are hit
```
damassi: @pepopowitz - one thing I noticed is that the `TODO -` form doesn't trigger syntax highlighting in our editors; I updated these notes with the form that they expect (`TODO:` -- the ":" is what does it).

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -0,0 +1,184 @@
...
+      })
+    },
+  }
+
+  return (
+    <FilterContext.Provider value={value}>{children}</FilterContext.Provider>
+  )
+}
+
+const filterReducer = (state, action) => {
```
damassi: I removed some of the types around this because React's type definitions are very, very good and can infer all the way from the context, with autocomplete.

src/Components/v2/ArtworkFilter/README.md
```diff
@@ -0,0 +1,7 @@
+# v2/ArtworkFilter component
+
+The work in this folder is currently in progress. It is currently not connected to any routes in Force.
+
+It contains a large rewrite of artwork filters & grids, which are currently duplicated 3 times throughout the app (in the Collect app, Search app, and Artist app).
+
+This rewrite is currently being implemented for the Search app, with the intention of eventually be rolled out to the other apps.
```
damassi: 👍 tyty

src/Components/v2/ArtworkFilter/ArtworkFilters/SizeRangeFilters.tsx
```diff
@@ -0,0 +1,52 @@
...
+  MAX_HEIGHT,
+  MAX_WIDTH,
+  MIN_HEIGHT,
+  MIN_WIDTH,
+} from "../Utils/rangeToTuple"
+
+export const SizeRangeFilters: FC = () => {
+  const filterContext = useFilterContext()
+
+  const [initialMinHeight, initialMaxHeight] = filterContext.rangeToTuple(
```
damassi: I think we we can defer to the `rangeToTuple` util vs moving through the context

PR #2585, opened 2019-06-27
Return myLastOffer so we can correctly redirect

sweir27:
The bug is that currently, after countering, the user is not redirected to the `review` page.

This appears to be an error since we removed `moment`. (Specifically, [this change](https://github.com/artsy/reaction/commit/eca55f2cd40f634ebb2dfbca51235ee5427260ce#diff-99e28aa2a4cbde0ce6ae302e0fd18499R142)).

We've been failing to return `myLastOffer` from a counter response, but it was obscured because of this [odd moment behavior](https://github.com/moment/moment/issues/2697).

### DIFF
--- a/src/Apps/Order/Routes/Respond/index.tsx
+++ b/src/Apps/Order/Routes/Respond/index.tsx
@@ -366,6 +366,9 @@ export const RespondFragmentContainer = createFragmentContainer(
             id
             note
           }
+          myLastOffer {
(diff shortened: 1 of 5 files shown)


ashkan18:
merge on green

artsyit:
:rocket: PR was released in v16.17.32 :rocket:

PR #2588, opened 2019-06-27
Handle Flicker Affect When Ad Server Content is Empty

ashleyjelks:
This PR adds a boolean from the `react-gpt` API to collapse empty divs when no ad content is served.


artsyit:
:rocket: PR was released in v16.17.35 :rocket:

PR #2589, opened 2019-06-27
PURCHASE-1197 - Adds TabCarousel

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1197

- Adds solution for overflow tabs. On desktop when tabs exceed the viewport width an arrow will appear for the user to advance through the tabs in a carousel. On mobile the tabs are simply rendered in a `Flex` element which allows for swiping right and left.

- Depends on https://github.com/artsy/palette/pull/529

### DIFF
--- a/src/Components/v2/Carousel.tsx
+++ b/src/Components/v2/Carousel.tsx
@@ -1,4 +1,4 @@
-import { ChevronIcon, color, Flex, space } from "@artsy/palette"
+import { Box, ChevronIcon, color, Flex, space } from "@artsy/palette"
 import { Options as FlickityOptions } from "flickity"
 import React, { Fragment } from "react"
 import styled from "styled-components"
@@ -52,6 +52,11 @@ interface CarouselProps {
   options?: FlickityOptions
 
   /**
+   * Pass an optional position for left and right for the arrow wrapper element otherwise defaults to -40
+   */
+  arrowPosition?: number
+
+  /**
    * Show or hide arrows. Defaults to true
    */
   showArrows?: boolean
@@ -84,14 +89,14 @@ export class Carousel extends React.Component<CarouselProps> {
 
   render() {
     return (
-      <>
+      <Box width="100%">
         <Media greaterThan="xs">
           <LargeCarousel {...this.props} />
         </Media>
         <Media at="xs">
           <SmallCarousel {...this.props} />
         </Media>
-      </>
+      </Box>
     )
   }
 }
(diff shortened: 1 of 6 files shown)


src/Components/v2/Carousel.tsx
```diff
@@ -210,30 +215,39 @@ export class BaseCarousel extends React.Component<
...
 
     if (!showArrows) {
       return null
     }
 
+    const leftPosition =
+      this.props.arrowPosition !== null &&
+      this.props.arrowPosition !== undefined
+        ? this.props.arrowPosition
+        : -space(4)
```
zephraph: Why does this default to `-40`? I saw it mentioned in the comment earlier, but it I don't get a sense of why... is that to pull it off the screen?

kierangillen: Screen Shot 2019-07-02 at 1 57 19 PM

Yea the default is for the arrow to be 40px outside of the container and in space outside of the normal page container.

src/Components/v2/Carousel.tsx
```diff
@@ -150,7 +155,7 @@ export class BaseCarousel extends React.Component<
 > {
   state = {
     currentSlideIndex: 0,
-    lastItemVisible: false,
+    lastItemVisible: true,
```
sweir27: Will this affect other places this component is used?

kierangillen: Yes but this is better behavior. We should assume the carousel is completely visible by default and then only display the arrow if there's hidden content. This will prevent an arrow flashing.

artsyit:
:rocket: PR was released in v16.18.0 :rocket:

src/Components/v2/Carousel.tsx
```diff
@@ -265,18 +279,25 @@ export class BaseCarousel extends React.Component<
...
-    const { onArrowClick, renderRightArrow, showArrows } = this.props
+    const { onArrowClick, renderRightArrow, showArrows, height } = this.props
 
     if (!showArrows) {
       return null
     }
 
+    const rightPosition =
+      this.props.arrowPosition !== null &&
+      this.props.arrowPosition !== undefined
```
damassi: A shortcut to check null or undefined in one line is `this.props.arrowPosition != null` (single equals)

kierangillen: ah and this checks `undefined` too? interesting

damassi: lil js hack 😄

damassi:
Nicely done @kierangillen - dig the approach piggy-backing on the carousel.

PR #2590, opened 2019-07-01
[GROW-1314] refactor to be able to use image with authentication modal

xtina-starr:
Addresses: [GROW-1314](https://artsyproduct.atlassian.net/browse/GROW-1314)

This updates the ModalManager and ModalOptions interfaces to be able to accept an image url so that we can have signup modals that support images. There is a subsequent force PR that makes use of the  image property.


l2succes:
@xtina-starr you're adding back `metaphysics` and `elan` as submodules in this PR. can remove those before merging as well 🙏🏾

artsyit:
:rocket: PR was released in v16.18.4 :rocket:

PR #2592, opened 2019-07-02
Deprecate old ads from featured articles

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/GROW-1345.

1. Updates some tests that have gotten out of date, for FeaturedLayout.tsx.
2. Removes `<NewDisplayCanvas>` from Sections.tsx, in favor of rendering `<DisplayAd>` directly.
3. Refactors tests for Section.tsx to remove duplicated lines & improve readability.

### DIFF
--- a/src/Components/Publishing/Sections/Sections.tsx
+++ b/src/Components/Publishing/Sections/Sections.tsx
@@ -3,13 +3,13 @@ import {
   is300x50AdUnit,
   targetingData,
 } from "Components/Publishing/Display/DisplayTargeting"
-import { NewDisplayCanvas } from "Components/Publishing/Display/NewDisplayCanvas"
 import { AdDimension, AdUnit } from "Components/Publishing/Typings"
 import { clone, compact, findLastIndex, get, once } from "lodash"
 import React, { Component } from "react"
 import ReactDOM from "react-dom"
 import styled, { StyledFunction } from "styled-components"
 import { pMedia } from "../../Helpers"
+import { DisplayAd } from "../Display/DisplayAd"
 import { ArticleData } from "../Typings"
 import { Authors } from "./Authors"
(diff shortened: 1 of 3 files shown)


src/Components/Publishing/Sections/Sections.tsx
```diff
@@ -280,7 +280,7 @@ export class Sections extends Component<Props, State> {
 
         ad = (
           <AdWrapper mt={marginTop}>
-            <NewDisplayCanvas
+            <DisplayAd
               pt={is300x50AdUnit(adDimension) ? 2 : 4} // add 20px to mobile leaderboard ads until this component is converted to <DisplayAd />
```
pepopowitz: @ashleyjelks you mentioned that I should be able to remove this `pt=...` line once I converted to `DisplayAd`, but I am concerned with the visual results of that change.

The current ads in staging/production look like this - notice the top padding.

image

After removing the `pt=...` here, they looked like this - notice the smaller top padding.

image

In this component we're setting `pt` to either 2 or 4, and [in `DisplayAd` it's setting it to either 0 or 2](https://github.com/artsy/reaction/blob/master/src/Components/Publishing/Display/DisplayAd.tsx#L59). I don't know if this means that what is in `DisplayAd` is incorrect, or if we still need to apply the extra padding here. I left it as the latter until I could get your advice.

pepopowitz: @ashleyjelks I removed the `pt` specified in Sections.tsx now that we're handling it properly in DisplayAd.tsx.

ashleyjelks:
looks good!

merge on green

artsyit:
:rocket: PR was released in v16.18.13 :rocket:

PR #2597, opened 2019-07-03
Remove NewDisplayCanvas from news articles, in favor of DisplayAd

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/GROW-1347

* Replaces NewDisplayCanvas with DisplayAd in NewsLayout.tsx
* Updates tests to account for the replacement
* Refactors the tests a bit to remove `wrapper.find()` duplication


src/Components/Publishing/Layouts/NewsLayout.tsx
```diff
@@ -123,7 +123,7 @@ export class NewsLayout extends Component<Props, State> {
     return (
       <>
         {shouldAdRender && (
-          <NewDisplayCanvas
+          <DisplayAd
             pt={is300x50AdUnit(adDimension) ? 2 : 4} // add 20px to mobile leaderboard ads until this component is converted to <DisplayAd />
```
pepopowitz: I ran into the same issue [as my other similar PR](https://github.com/artsy/reaction/pull/2592#discussion_r300110145) when I removed this line, so I left it for now.

ashleyjelks: Were you able to confirm this behavior happens on a mobile device? I saw the screenshot you posted but, but I have noticed that the Chrome mobile simulator doesn't behave exactly the same as actual mobile devices.

src/Components/Publishing/Layouts/__tests__/NewsLayoutAdsEnabled.test.tsx
```diff
@@ -44,25 +43,19 @@ describe("News Layout with new ads enabled", () => {
...
+    expect(component.find(DisplayAd).length).toBe(1)
   })
 
   it("renders the news layout component with correct ad unit after the 3rd article on desktop", () => {
     props.articleSerial = 3
     const component = getWrapper()
 
-    expect(component.find(NewDisplayCanvas).length).toBe(1)
-    expect(component.find(DisplayCanvas).length).toBe(0)
+    const ads = component.find(DisplayAd)
```
ashleyjelks: This looks good, only now we can also delete this separate test file `NewsLayoutAdsEnabled.test.tsx` move these new tests to  `NewsLayoutAds.test.tsx`. Also in  `NewsLayoutAds.test.tsx` we should remove all assertions that test the `DisplayCanvas` or `DisplayPanel` components as well as remove any tests in `CanvasFooter` that are used for ads. Note: we used to use the `CanvasFooter` for 970x250 ads AND other components like the related articles rail, and the collections rails, so some `CanvasFooter` components need to remain.

src/Components/Publishing/Layouts/NewsLayout.tsx
```diff
@@ -123,7 +123,7 @@ export class NewsLayout extends Component<Props, State> {
     return (
       <>
         {shouldAdRender && (
-          <NewDisplayCanvas
+          <DisplayAd
```
ashleyjelks: In addition to converting this from a `NewDisplayCanvas` to a `DisplayAd` component, I would actually remove the ternary [on line 188](https://github.com/artsy/reaction/blob/94d93024a7874f4e6fff4b7ed78a8fe9e5b861de/src/Components/Publishing/Layouts/NewsLayout.tsx#L188) and return the result of the function call to [renderAd](https://github.com/artsy/reaction/blob/94d93024a7874f4e6fff4b7ed78a8fe9e5b861de/src/Components/Publishing/Layouts/NewsLayout.tsx#L125), but be sure to include the check for `areHostedAdsEnabled` since we only needed that duplicated `CanvasFooter` code to render NewsArticles both with and without ads.

src/Components/Publishing/Layouts/NewsLayout.tsx
ashleyjelks: If you haven't tried browsing sites served through your local dev server, lmk and we can go through how to get it up. I had to QA some mobile fixes on my phone via my machine running on localhost recently.

ashleyjelks:
Looks good to me!

artsyit:
:rocket: PR was released in v16.18.10 :rocket:

PR #2598, opened 2019-07-08
grow_1367 Featured artist module on Small viewport should be 3 per row

jiahaodavid:
address : [grow_1367](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1367)

- any changed on width     **->**     the numbers of artist module would be changed automatically

**less  than 768**

**between 768 - 899**

**between 900 - 1024**

**greater than 1024**


artsyit:
:rocket: PR was released in v16.18.6 :rocket:

PR #2599, opened 2019-07-08
Handles empty divs when ad blocker is enabled

ashleyjelks:
This PR removes the large empty div on articles when ad blockers are enabled.


artsyit:
:rocket: PR was released in v16.18.5 :rocket:

PR #2603, opened 2019-07-11
change sync correlator to true

ashleyjelks:
This PR sets the updateCorrelator option to true to prevent a bug that caused GAM to be unaware of which pages ads were being served on.

[For context on correlators in GAM visit this link](https://support.google.com/admanager/answer/183281?hl=en)


artsyit:
:rocket: PR was released in v16.18.10 :rocket:

PR #2605, opened 2019-07-11
Add initial auction registration flow

dblandin:
This implements one of our existing auction registration flows which
creates a payment method and registers the user to bid in the auction.

Zeplin spec :lock: :
https://app.zeplin.io/project/5c59f1624a2d813589f91e07/screen/5d123fbcd17e52705b13ae42
Jira ticket :lock: : https://artsyproduct.atlassian.net/browse/AUCT-462

Items to follow-up on:

- Show an error modal when a mutation against Metaphysics fails.
- Redirect routing if a user already has a payment method or if they're
  already registered to build
- Update Country select input to accept onBlur/onChange events and add
  to form
- Look into using `isCommittingMutation` helper instead of Formik's
  `isSubmitting` for loading indicator

Co-authored-by: @erikdstock 

screenshot 1564007281

### DIFF
--- a/src/Apps/Auction/AuctionApp.tsx
+++ b/src/Apps/Auction/AuctionApp.tsx
@@ -1,0 +1,95 @@
+import { Box, Separator, Serif } from "@artsy/palette"
+import { AuctionApp_sale } from "__generated__/AuctionApp_sale.graphql"
+import { AppContainer } from "Apps/Components/AppContainer"
+import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
+import { ErrorPage } from "Components/ErrorPage"
+import React, { Component } from "react"
+import { createFragmentContainer, graphql, RelayProp } from "react-relay"
+import {
+  Elements,
+  ReactStripeElements,
+  StripeProvider,
+} from "react-stripe-elements"
+import RegistrationForm from "./Components/RegistrationForm"
+
+declare global {
+  interface Window {
+    Stripe?: (key: string) => stripe.Stripe
+    sd: {
+      STRIPE_PUBLISHABLE_KEY: string
+    }
+  }
+}
+
+interface AuctionAppProps extends ReactStripeElements.InjectedStripeProps {
+  sale: AuctionApp_sale
+  relay?: RelayProp
+}
+
+interface AuctionAppState {
+  stripe: stripe.Stripe
+}
+
+export class AuctionApp extends Component<AuctionAppProps, AuctionAppState> {
+  state: AuctionAppState = { stripe: null }
+  CreditCardInput
+  componentDidMount() {
+    if (window.Stripe) {
+      this.setState({
+        stripe: window.Stripe(window.sd.STRIPE_PUBLISHABLE_KEY),
+      })
+    } else {
+      document.querySelector("#stripe-js").addEventListener("load", () => {
+        // Create Stripe instance once Stripe.js loads
+        this.setState({
+          stripe: window.Stripe(window.sd.STRIPE_PUBLISHABLE_KEY),
+        })
+      })
+    }
+  }
+
+  render() {
+    const { sale } = this.props
+
+    if (!sale) {
+      return <ErrorPage code={404} />
+    }
+
+    return (
+      <AppContainer>
+        <Box maxWidth={550} px={[2, 0]} mx="auto" my={[1, 0]}>
+          <Serif size="10">Register to Bid on Artsy</Serif>
+          <Separator mt={1} mb={2} />
+          <Serif size="4" color="black100">
+            Please enter your credit card information below. The name on your
+            Artsy account must match the name on the card, and a valid credit
+            card is required in order to bid.
+          </Serif>
+          <Serif size="4" mt={2} color="black100">
+            Registration is free. Artsy will never charge this card without your
+            permission, and you are not required to use this card to pay if you
+            win.
+          </Serif>
+          <StripeProvider stripe={this.state.stripe}>
+            <Elements>
+              <Box mt={2}>
+                <RegistrationForm sale={sale} relay={this.props.relay} />
+              </Box>
+            </Elements>
+          </StripeProvider>
+        </Box>
+      </AppContainer>
+    )
+  }
+}
+
+export const AuctionAppFragmentContainer = createFragmentContainer(
+  trackPageViewWrapper(AuctionApp),
+  {
+    sale: graphql`
+      fragment AuctionApp_sale on Sale {
+        id
+      }
+    `,
+  }
+)
--- a/src/Apps/Auction/AuctionApp.tsx
+++ b/src/Apps/Auction/AuctionApp.tsx
@@ -10,16 +10,8 @@ import {
   ReactStripeElements,
   StripeProvider,
 } from "react-stripe-elements"
-import RegistrationForm from "./Components/RegistrationForm"
-
-declare global {
-  interface Window {
-    Stripe?: (key: string) => stripe.Stripe
-    sd: {
-      STRIPE_PUBLISHABLE_KEY: string
-    }
-  }
-}
+import { data as sd } from "sharify"
+import { WrappedRegistrationForm } from "./Components/RegistrationForm"
 
 interface AuctionAppProps extends ReactStripeElements.InjectedStripeProps {
   sale: AuctionApp_sale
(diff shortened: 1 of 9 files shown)


src/Apps/Auction/routes.tsx
```diff
@@ -0,0 +1,39 @@
+// import { getRedirect } from "Apps/Order/getRedirect"
+// import { confirmRouteExit, redirects } from "Apps/Order/redirects"
+import { RouteConfig } from "found"
+import * as React from "react"
+import { graphql } from "react-relay"
+import { AuctionApp } from "./AuctionApp"
+
+// import { RegisterFragmentContainer as RegisterRoute } from "Apps/Auction/Routes/Register"
+
+// @ts-ignore
```
damassi: Are these `@ts-ignore`s still necessary? I recall these interfaces needed to be imported for automatic TS declaration output but recent TypeScripit versions seem to have fixed this.

src/Apps/Auction/routes.tsx
```diff
@@ -0,0 +1,39 @@
...
+// * `render` functions requires casting
+export const routes: RouteConfig[] = [
+  {
+    path: "/auction-registration/:saleID",
+    Component: AuctionApp,
+    render: ({ Component, props }) => {
+      if (Component && props) {
+        return <Component {...props} />
+      }
+    },
```
damassi: Best to avoid these `render` functions unless custom behavior is desired. In most cases `Component: AuctionApp` should be enough.

src/Apps/Auction/AuctionApp.tsx
```diff
@@ -0,0 +1,95 @@
...
+  ReactStripeElements,
+  StripeProvider,
+} from "react-stripe-elements"
+import RegistrationForm from "./Components/RegistrationForm"
+
+declare global {
+  interface Window {
+    Stripe?: (key: string) => stripe.Stripe
+    sd: {
+      STRIPE_PUBLISHABLE_KEY: string
```
damassi: Any reason `sharify.d.ts` doesn't [cover this](https://github.com/damassi/reaction/blob/8a0da9ad42b6cd5ef9986af68a63a804dc29a26e/typings/sharify.d.ts#L38)?

src/Apps/Auction/AuctionApp.tsx
```diff
@@ -0,0 +1,95 @@
...
+  componentDidMount() {
+    if (window.Stripe) {
+      this.setState({
+        stripe: window.Stripe(window.sd.STRIPE_PUBLISHABLE_KEY),
+      })
+    } else {
+      document.querySelector("#stripe-js").addEventListener("load", () => {
+        // Create Stripe instance once Stripe.js loads
+        this.setState({
+          stripe: window.Stripe(window.sd.STRIPE_PUBLISHABLE_KEY),
```
damassi: This is how we typically use sharify, as opposed to plucking directly from `window`:  
```js
import { data as sd } from 'sharify'
...
sd. STRIPE_PUBLISHABLE_KEY
```

src/Apps/Auction/AuctionApp.tsx
```diff
@@ -0,0 +1,95 @@
...
+export class AuctionApp extends Component<AuctionAppProps, AuctionAppState> {
+  state: AuctionAppState = { stripe: null }
+  CreditCardInput
+  componentDidMount() {
+    if (window.Stripe) {
+      this.setState({
+        stripe: window.Stripe(window.sd.STRIPE_PUBLISHABLE_KEY),
+      })
+    } else {
+      document.querySelector("#stripe-js").addEventListener("load", () => {
```
damassi: We should remove this listener on `componentWillUnmount`. Alternatively, if you wanted to use a functional component / hooks, could do something like this: 

```js
useEffect(() => {
  document.querySelector("#stripe-js").addEventListener("load", ...)

  return () => {
    document.querySelector("#stripe-js").removeEventListener("load", ...)
  }
}, [])
```
(No personal pref re classes / function components, just FYI)

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -0,0 +1,296 @@
...
+          })
+        }}
+        validationSchema={validationSchema}
+        render={InnerForm}
+      />
+    )
+  }
+}
+
+export default injectStripe(RegistrationForm)
```
damassi: We should be avoiding default exports as they're not statically analyzable

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -0,0 +1,296 @@
...
+            address_state: values.state,
+            address_zip: values.postal_code,
+          }
+
+          this.props.stripe.createToken(address).then(({ error, token }) => {
+            if (error) {
+              setFieldError("credit_card", error.message)
+              setSubmitting(false)
+            } else {
+              console.log(token)
```
damassi: stray log

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -0,0 +1,296 @@
...
+})
+
+export interface RegistrationFormProps
+  extends ReactStripeElements.InjectedStripeProps {
+  relay?: RelayProp
+  sale: AuctionApp_sale
+}
+
+class RegistrationForm extends Component<RegistrationFormProps> {
+  private createBidder(setSubmitting) {
```
damassi: Note that we could also use the React.FC form:

```tsx
const RegistrationForm: React.FC<RegistrationFormProps> = props => {
  function createBidder(setSubmitting) {
    ...
  }

  function createCreditCard() {
    ...
  }

  return (
    ...
  )
}
```
So as to avoid the use of `this`.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -0,0 +1,296 @@
...
+            input: { token },
+          },
+        }
+      )
+    })
+  }
+
+  render() {
+    return (
+      <Formik
```
damassi: 🙏 💯 for using Formik and not rolling this form behavior from scratch

src/Apps/__stories__/Apps.story.tsx
```diff
@@ -1,11 +1,25 @@
...
+  .add("Auction Registration", () => {
+    return (
+      <MockRouter
+        routes={auctionRoutes}
+        initialRoute="/auction-registration2/yuki-contemporary-art-july-22th"
+        context={{
+          mediator: {
+            trigger: x => x,
+          },
+        }}
```
damassi: If not using `mediator` no need to pass this context value in

src/Apps/Auction/AuctionApp.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
+import { ErrorPage } from "Components/ErrorPage"
+import React, { Component } from "react"
+import { createFragmentContainer, graphql, RelayProp } from "react-relay"
+import {
+  Elements,
+  ReactStripeElements,
+  StripeProvider,
+} from "react-stripe-elements"
+import RegistrationForm from "./Components/RegistrationForm"
```
damassi: We should be using named exports here

src/Apps/Auction/AuctionApp.tsx
dblandin: Looks cool! I'll look into using a functional component / hooks in the next iteration. :+1:

src/Apps/Auction/Components/ConditionsOfSaleCheckbox.tsx
```diff
@@ -0,0 +1,43 @@
...
+}) => {
+  return (
+    <Box mx="auto">
+      <StyledCheckbox {...{ checked: value, error, onChange, onBlur, name }}>
+        <Serif display="inline" color="black60" size="3t" ml={0.5}>
+          {"Agree to "}
+          <Serif display="inline" color="black100" size="3t">
+            <Link
+              color="black100"
+              href="https://www.artsy.net/conditions-of-sale"
```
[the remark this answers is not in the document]
dblandin: Good call! There's a pattern for this. We can use the `APP_URL` from sharify.

artsyit:
:rocket: PR was released in v16.19.0 :rocket:

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        props.stripe.createToken(address).then(({ error, token }) => {
+          if (error) {
+            setFieldError("credit_card", error.message)
+            setSubmitting(false)
+          } else {
+            createCreditCard(token.id).then(() => {
+              createBidder(setSubmitting)
+            })
```
erikdstock: This is so readable, very nice.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+          <Serif size="3t" mb={0.5}>
+            Credit card number*
+          </Serif>
+          <CreditCardInput
+            error={{ message: errors.credit_card } as stripe.Error}
+          />
+        </Box>
+      </Box>
+      <Box mt={4}>
+        <Serif weight="semibold" size="4t">
```
sweir27: @dblandin was there a reason to not re-use `AddressForm` here? (here's an [example](https://github.com/artsy/reaction/blob/master/src/Components/Payment/PaymentForm.tsx#L124) of where we're using both `AddressForm` and `CreditCardInput` in a custom wrapper).

PR #2611, opened 2019-07-13
Remove feature flag from display ads

ashleyjelks:
This PR removes the flag `areHostedAdsEnabled` which was a temporary placeholder used to render ads on staging, but not on production, while Grow was working on the new Ad Server project.

Now that ads are fully live on prod this feature no longer needs to be hidden behind this flag.

This PR removes the flags from ad components and updates corresponding tests.

[Links to GROW-1396](https://artsyproduct.atlassian.net/browse/GROW-1396)


src/Components/Publishing/Display/DisplayTargeting.tsx
```diff
@@ -1,7 +1,7 @@
-import { data as sd } from "sharify"
-
 export const targetingData = (id: string, pageType: string) => {
-  const isTesting = JSON.parse(sd.HASHTAG_LAB_ADS_ENABLED || "true")
+  // TODO: Create a sharify field to pass true for staging and false for prod env to the isTesting bool
```
pepopowitz: Does this need to be done before this can be merged?

src/Components/Publishing/Layouts/StandardLayout.tsx
```diff
@@ -67,49 +67,40 @@ export class StandardLayout extends React.Component<
...
-              ? AdUnit.Mobile_InContentMR1
-              : AdUnit.Desktop_TopLeaderboard
-          }
-          adDimension={adDimension}
-          displayNewAds={areHostedAdsEnabled}
-          targetingData={targetingData(article.id, "article")}
-        />
-      )
+      <DisplayAd
+        pt={is300x50AdUnit(adDimension) ? 2 : 4}
```
pepopowitz: Can we remove this `pt` value?

src/Components/Publishing/Layouts/__tests__/SeriesLayout.test.tsx
```diff
@@ -20,50 +20,43 @@ function mountComponent() {
...
   beforeEach(() => {
     props = {
-      areHostedAdsEnabled: true,
       article: SeriesArticleFixture,
       isSeries: true,
     }
   })
 
-  it("renders the series layout properly", () => {
+  it("renders the series layout with ads", () => {
```
pepopowitz: This is now a duplicate test of line 31 above. We should remove it.

src/Components/Publishing/Layouts/__tests__/StandardLayout.test.tsx
```diff
@@ -72,15 +71,8 @@ describe("Standard Article", () => {
...
-    props.article = extend(cloneDeep(StandardArticle), {
-      seriesArticle: SeriesArticle,
-    })
-    const article = getWrapper(props)
-    expect(article.find(DisplayAd).length).toBe(0)
-  })
-
-  it("Does not render display if article is super", () => {
+  // TODO: Find out why display ads are not expected to render in super articles
+  xit("Does not render display if article is super", () => {
```
eessex: I believe ads were left out of `superArticles` because they are not part of the infinite scroll feed, and use custom expandable and footers. I'm not sure that we need to exclude them from the article body, but would check in with Content Sales and Product on this. Examples of super articles can be found here: https://artsyproduct.atlassian.net/browse/GROW-1116

src/Components/Publishing/Layouts/__tests__/SeriesLayout.test.tsx
```diff
@@ -20,50 +20,43 @@ function mountComponent() {
...
-
-  it("does not render a DisplayAd", () => {
-    const component = mountComponent()
-
-    expect(component.find(DisplayAd).length).toBe(0)
-  })
 })
 
-describe("when ads are enabled", () => {
+describe("series layout with ads", () => {
```
pepopowitz: Since there is really no concept of "with ads" vs "without ads" anymore, I think we should consolidate this describe with the one above it.

src/Components/Publishing/Layouts/__tests__/StandardLayout.test.tsx
pepopowitz: This seems like it's worth figuring out before merging, since it's potentially a regression.

joeyAghion:
Since the point of staging is to host as equivalent an environment to production as possible, I would prefer a more specific configuration instead of `IS_STAGING`. Could it be something like `AD_TESTING` instead? Hopefully that's a small change here but prevents future developers from depending on staging vs. production conditionals.

src/Components/Publishing/Layouts/StandardLayout.tsx
ashleyjelks: No bc the padding still needs to be 20 px on mobile leaderboard ads and 40px on all others.The comment is misleading though. So I'll remove it.

ashleyjelks:
> Since the point of staging is to host as equivalent an environment to production as possible, I would prefer a more specific configuration instead of `IS_STAGING`. Could it be something like `AD_TESTING` instead? Hopefully that's a small change here but prevents future developers from depending on staging vs. production conditionals.

I opted to just expose the existing field `DEPLOY_ENV` client side to handle this case.

artsyit:
:rocket: PR was released in v16.18.20 :rocket:

PR #2613, opened 2019-07-15
[Trivial] Rename and export article card image container

eessex:
Renames so we can target styles in Positron.


artsyit:
:rocket: PR was released in v16.18.19 :rocket:

PR #2614, opened 2019-07-15
Test & refactor collection header

pepopowitz:
This PR addresses https://artsyproduct.atlassian.net/browse/GROW-1282.

It adds test coverage for the collection header, a few types, and some refactoring to remove concepts that are no longer needed.

### DIFF
--- a/src/Apps/Collect/Components/Collection/Header/__tests__/Header.test.tsx
+++ b/src/Apps/Collect/Components/Collection/Header/__tests__/Header.test.tsx
@@ -1,99 +1,207 @@
 import { EntityHeader } from "@artsy/palette"
-import { shallow } from "enzyme"
+import { MockBoot } from "DevTools/MockBoot"
+import { mount, shallow } from "enzyme"
 import React from "react"
-import { getFeaturedArtists } from "../index"
+import { CollectionHeader, getFeaturedArtists, Props } from "../index"
 
-describe("getFeaturedArtists", () => {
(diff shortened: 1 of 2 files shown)


xtina-starr:
@pepopowitz I pulled your branch down locally to continue working on this and I'm now wondering if [`collection.description`](https://github.com/artsy/reaction/blob/7d6b0e497720d379b49c22452943211ed93476d7/src/Apps/Collect/Components/Collection/Header/index.tsx#L35) getting typed as `JSX.Element` is even valid. It's actual type is `HTMLString` but [that isn't available as a type in typescript](https://github.com/microsoft/TypeScript/issues/4856). I'm wondering if it's cool to just type it as a string seeing as that's what it's represented as in the [kaws database](https://github.com/artsy/kaws/blob/0bfb0f3cbb905ef9391ed636169a3467d4db7e3a/src/Entities/Collection.ts#L25-L31)? @l2succes do you have any thoughts?

Also he is what the description looks like via the metaphysics query:

Screen Shot 2019-07-18 at 4 04 45 PM

l2succes:
`description: string` sounds good. Not sure why we had `JSX.Element | string` but if it breaks anything, the TS compiler will complain anyway so go for it.

ashleyjelks:
Awesome job!! 

merge on green!

artsyit:
:rocket: PR was released in v16.18.34 :rocket:

PR #2615, opened 2019-07-16
Remove ads from super articles

ashleyjelks:
Fixes a previously failing test where ads were being rendered on super articles.


ashleyjelks:
> Makes sense to me 👍 One thing to double check is if `isSuper` applies to sub-articles as well (they also render custom header/footer), in which case we would also want to pass `isSuper` to standard layout articles.

Ok, I'll look into that and follow up in a separate PR, thanks!

artsyit:
:rocket: PR was released in v16.18.22 :rocket:

PR #2623, opened 2019-07-22
PURCHASE-1317: Convert OtherWorks component to use contextGrids

sweir27:
This PR converts the `OtherWorks` component to consume the `contextGrids` API which was recently added to metaphysics.

The result is:
- We get to remove a whole bunch of code!
- The "Other works from the auction" grid is now server-side rendered
- "Other Auctions" rail is still client-side rendered, and I moved it into its own `QueryRenderer`

Visual changes:
- For the "Other works from this auction" grid, the title is now `Other works from <auction name` to match the updated logic we added for Emission
- When an artwork is in a fair, we _only_ show a grid of artworks from the same fair booth. We used to also show a grid of artworks from the gallery's most recent show, but these tended to be the same so it was duplicated.


src/Apps/Artwork/Components/OtherAuctions.tsx
```diff
@@ -36,3 +39,22 @@ export const OtherAuctionsFragmentContainer = createFragmentContainer(
     }
   `
 )
+
+export const OtherAuctionsQueryRenderer = () => {
```
damassi: So to confirm, we're not going to SSR render this?

src/Apps/Artwork/Components/OtherWorks/index.tsx
```diff
@@ -1,63 +1,138 @@
...
+/**
+ * Check to see if a connection's edges have a length; if false hide the grid.
+ */
+export function hideGrid(artworksConnection): boolean {
+  return Boolean(get(artworksConnection, p => !p.edges.length))
+}
+
+const populatedGrids = (grids: OtherWorks_artwork["contextGrids"]) => {
+  if (grids && grids.length > 0) {
+    return filter(grids, grid => {
```
damassi: Can we use the native array.filter form?

src/Apps/Artwork/Components/OtherAuctions.tsx
sweir27: Not now, that can be a follow up if we want it.

It's currently client-side rendered so this isn't any different 😄.

src/Apps/Artwork/Components/OtherWorks/index.tsx
```diff
@@ -1,63 +1,138 @@
...
+    }
+  }
 }
 
+export const OtherWorks: React.FC<{
+  artwork: OtherWorks_artwork
+}> = track()(props => {
+  const { context, contextGrids, sale } = props.artwork
+
+  const grids = contextGrids
```
damassi: #superminor no real need to create intermediate variable

src/Apps/Artwork/Components/__stories__/OtherAuctions.story.tsx
```diff
@@ -11,10 +11,10 @@ const OtherAuctions = ({ size }: { size?: number }) => {
   const { relayEnvironment } = useContext(SystemContext)
 
   return (
-    <QueryRenderer<OtherAuctionsQuery>
+    <QueryRenderer<OtherAuctionsTestQuery>
       environment={relayEnvironment}
       query={graphql`
-        query OtherAuctionsQuery($size: Int!) {
+        query OtherAuctionsTestQuery($size: Int!) {
```
damassi: Maybe `OtherAuctionsStoryQuery`?

src/Apps/Artwork/Components/OtherWorks/index.tsx
```diff
@@ -1,63 +1,134 @@
...
+    }
+    case "ShowArtworkGrid": {
+      return Schema.ContextModule.OtherWorksFromShow
+    }
+  }
 }
 
+export const OtherWorks: React.FC<{
+  artwork: OtherWorks_artwork
+}> = track()(props => {
```
ds300: this `props` parameter is typed as `any` because it's wrapped in `track` and typescript isn't clever enough to infer the type from the context.

try this instead:

```ts
export const OtherWorks = track()(
  (props: { artwork: OtherWorks_artwork } & SystemContextProps) => {
```

I noticed this because `props.mediator` was being used below but wasn't declared up here.

sweir27: Oooh, thank you!

artsyit:
:rocket: PR was released in v16.18.32 :rocket:

PR #2627, opened 2019-07-23
Fix Modal with image on smaller(xs) screens

l2succes:
Fixes some UI issue with the Modal component with image when rendered on a smaller screen. It should take the full with but only renders on half.

## Before
Screen Shot 2019-07-23 at 5 55 30 PM

## After
Screen Shot 2019-07-23 at 5 55 41 PM


l2succes:
merge on green

artsyit:
:rocket: PR was released in v16.18.31 :rocket:

PR #2634, opened 2019-07-25
[Collections]  Default Header

ashleyjelks:
This PR adds a default header to the collections page when a target header image has not been set.

[Links to GROW-1256](https://artsyproduct.atlassian.net/browse/GROW-1256)

Kapture 2019-07-26 at 15 02 38


src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+            )
+          })}
+        </HeaderArtworks>
+      </DefaultHeaderContainer>
+    </header>
+  )
+}
+
+const useViewportSize = () => {
+  const isClient = typeof window === "object"
```
damassi: `typeof window !== undefined`

src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
```diff
@@ -0,0 +1,170 @@
...
+              </a>
+            )
+          })}
+        </HeaderArtworks>
+      </DefaultHeaderContainer>
+    </header>
+  )
+}
+
+const handleViewportResize = () => {
```
pepopowitz: This feels like it would make a great custom hook! It could be named something like `useViewportSize` or `useWindowSize`, and it could go in our custom hooks folder (`src/utils/Hooks`).

src/Apps/Collect/Components/Collection/Header/__tests__/DefaultHeader.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      headerArtworks: defaultCollectionHeaderArtworks,
+      defaultHeaderImageHeight: 1000,
+    }
+  })
+
+  const getWrapper = headerProps => {
+    return shallow(<CollectionDefaultHeader {...headerProps} />)
+  }
+
+  it("renders a snapshot", () => {
```
pepopowitz: Is there any sense of safety we're getting out of this snapshot that we aren't getting out of our other tests? If not, I'd prefer to remove this in favor of the other more-specific tests. Snapshots of components usually result in me saying "I don't know why this is failing, so I'm just going to update it."

damassi: Very much agree with @pepopowitz

ashleyjelks: I find them helpful for catching unexpected UI changes personally. Before updating the snapshots I check to see why a change was triggered, and I've found its a useful tool for catching unintended changes.

src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
ashleyjelks: Ooh, good point! I'll add move this now.

src/Apps/Collect/Components/Collection/Header/__tests__/DefaultHeader.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  })
+
+  it("returns an array of artworks of equal length as the original array received when header can fit all works ", () => {
+    const artworks = defaultCollectionHeaderArtworks.hits
+
+    const result = getHeaderArtworks(artworks, 1308, false)
+    expect(artworks.length).toEqual(result.length)
+    expect(result).toHaveLength(10)
+  })
+})
```
pepopowitz: There's one more test case we could add that would be really valuable. `it("returns only the number of artworks necessary to fill the header at a small resolution")`. 

If a user is on a mobile device, they are going to only need a few images from the collection. Giving them all of the originals will result in them downloading a handful of images that they'll never be able to see.

I don't think the current code would pass this test, but I think adding this test & refactoring to support it might simplify the implementation. We would iterate through all images until it was wide enough to exceed the viewport width, and start over in the list of images when we reach the end.

l2succes: Great point @pepopowitz +1 to that

src/Utils/viewport.ts
```diff
@@ -0,0 +1,9 @@
+export const getViewportWidth = () => {
```
l2succes: I may make more sense to do

```typescript
const getViewport: () => { width: number, height: number } = () => {
 ...
} 
```

usage

```typescript
const { width } = getViewport()
```

that way you can also get the height of the viewport for another use-case

src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
```diff
@@ -0,0 +1,149 @@
...
+   * of the viewport then there's no need to duplicate/repeat the header's artworks
+   */
+  if (artworkWidthsInOriginalArray >= headerWidth) {
+    return originalArtworksArray
+  } else {
+    /**
+     * Otherwise, duplicate the original array of artworks and add a new artwork to the array
+     * until the summed widths of the artworks is greater than the viewport width
+     */
+    while (artworkWidthsInRepeatingArray <= headerWidth) {
```
l2succes: Nicely done 👏🏾 My only feedback is there's a lot of code duplication between this while block and those lines above. May make sense to extract into a local helper function

https://github.com/artsy/reaction/pull/2634/files#diff-224a40bc680d52b38baeff65c547243eR77-R84

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -125,6 +125,7 @@ export const CollectionAppFragmentContainer = createFragmentContainer(
         artworks(
           aggregations: $aggregations
           include_medium_filter_in_aggregation: true
+          sort: "-merchandisability"
```
sweir27: :ceiling_cat: Curious if this is intended or if we should be using `-decayed_merch` here (and in the other file). That's our standard sort on pages like `/collect`, and takes into account [other boosted parameters](https://github.com/artsy/gravity/blob/7b72b2766371890cc3702dbf5ed21feec0fa08ba/app/models/search/queries/filtered_query.rb#L25) aside from just merchandisability. (See [this slack convo](https://artsy.slack.com/archives/C9SATFLUU/p1562881100203900) for some extra context)

src/Apps/Collect/Components/Collection/Header/__tests__/DefaultHeader.test.tsx
ashleyjelks: Yep, decided to just loop through once, only adding enough artworks to fill the header regardless of the screen size.

src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const IMAGE_MARGIN_X = 10
+const LARGE_VIEWPORT_WIDTH = 880
+
+export const CollectionDefaultHeader: FC<Props> = ({
+  headerArtworks,
+  defaultHeaderImageHeight,
+}) => {
+  useWindowSize()
```
pepopowitz: I think you should be able to get the width from the return value here, and eliminate line 28 below (another call to `getViewportWidth`).

src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+export const getHeaderArtworks = (
+  artworksArray: any[],
+  headerWidth: number,
+  isSmallViewport: boolean
+) => {
+  let artworkWidths = 0
+  const headerArtworks = []
+
+  if (artworksArray.length < 1) {
```
pepopowitz: Nice catch! 👌

src/Apps/Collect/Components/Collection/Header/DefaultHeader.tsx
```diff
@@ -0,0 +1,131 @@
...
+    return []
+  }
+
+  /**
+   * Loop through the artworks array, appending a new artwork to the output array
+   * until the summed widths of the artworks are greater than the width of the viewport.
+   */
+  while (artworkWidths <= headerWidth) {
+    artworksArray.forEach((artwork, i) => {
```
pepopowitz: Looks great 👍

src/Apps/Collect/Components/Collection/Header/__tests__/DefaultHeader.test.tsx
pepopowitz: Do you mind adding a test case to capture this edge-case?

artsyit:
:rocket: PR was released in v16.19.7 :rocket:

PR #2637, opened 2019-07-30
Sync MP schema #trivial 

damassi:
Noticed MP was out of sync in Force


artsyit:
:rocket: PR was released in v16.19.3 :rocket:

PR #2641, opened 2019-07-30
Update VSCode launch configurations

dblandin:
This PR updates the VSCode launch configurations related to running Jest tests with a debugger attached. They're adapted from this resource from Microsoft's VSCode Recipes:

https://github.com/microsoft/vscode-recipes/tree/851693ce90da46a36a28edeb743e09b3da6b6e6f/debugging-jest-tests

The formerly defined "Run Tests With Debugger" launch configuration used an explicitly defined and deprecated `node debug` invocation and wouldn't attach properly without some adjustments.

I couldn't get the "Launch Server" or "Launch Chrome" tasks to work so I've removed those as well thinking that they're probably not used, but let me know if I'm mistaken.

cc/ @dleve123 because you were asking about testing with a debugger in artsy/reaction recently!

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -24,16 +24,30 @@
       "webRoot": "${workspaceRoot}/src"
     },
     {
-      "name": "Run Tests With Debugger (slower, use yarn for normal work)",
       "type": "node",
       "request": "launch",
-      "port": 5858,
-      "address": "localhost",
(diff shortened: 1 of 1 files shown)


damassi:
Thank you!

artsyit:
:rocket: PR was released in v16.19.6 :rocket:

PR #2642, opened 2019-07-30
Video Headers Won't Crash

mbilokonsky:
This PR is in support of [GROW-1444](https://artsyproduct.atlassian.net/browse/GROW-1444). The real change for that PR is in Positron, but _this PR should be merged first_.

As the JIRA ticket describes, this is a temporary fix so that videos can be added without causing a crash.

Diff Analysis: 
```
{
  "total_files_changed": 3,
  "test_files_changed": 2,
  "by_type": {
    "tsx": 2,
    "snap": 1
  }
}
```

### DIFF
--- a/src/Components/Publishing/Series/FixedBackground.tsx
+++ b/src/Components/Publishing/Series/FixedBackground.tsx
@@ -9,7 +9,7 @@ interface Props {
 export const FixedBackground: React.SFC<Props> = props => {
   const { backgroundColor, backgroundUrl } = props
 
-  if (backgroundUrl) {
+  if (backgroundUrl && !backgroundUrl.includes(".mp4")) {
     return (
       <FixedBackgroundContainer backgroundColor={backgroundColor}>
         <BackgroundImage backgroundUrl={backgroundUrl} />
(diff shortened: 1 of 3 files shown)


src/Components/Publishing/Series/FixedBackground.tsx
```diff
@@ -9,7 +9,7 @@ interface Props {
 export const FixedBackground: React.SFC<Props> = props => {
   const { backgroundColor, backgroundUrl } = props
 
-  if (backgroundUrl) {
+  if (backgroundUrl && backgroundUrl.indexOf(".mp4") === -1) {
```
damassi: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

src/Components/Publishing/Series/__tests__/FixedBackground.test.tsx
```diff
@@ -0,0 +1,34 @@
+import "jest-styled-components"
+import React from "react"
+import renderer from "react-test-renderer"
+import { FixedBackground } from "../FixedBackground"
+
+const image_url = "https://foo.com/image.png"
```
damassi: All JS should be camelCase; snake_case is for ruby

artsyit:
:rocket: PR was released in v16.19.8 :rocket:

PR #2584, opened 2019-06-26
fix GROW-1341(Carousel tweaks for artist series and related collections)

jiahaodavid:
addresses: [GROW-1341](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1341)

In this PR, we addresses the following:

1. No arrow when collections < 4 (desktop view)
Screen Shot 2019-06-26 at 1 44 42 PM
2. No infinite scroll,  just showing the remainder if next set is less than 4. 
    Also, reaching the end of the rail and starting from the beginning of the rail -> No arrow on the side.
    (desktop view)
Jun-26-2019 15-00-11

3. Can be swiped now if the artist only have 4 collections (mobile view)
Jun-26-2019 15-07-05

4. Swipe collection one by one. Not 4 collections each swipe. (mobile view)
Jun-26-2019 14-45-33

PR #2644, opened 2019-07-31
Grow 1354 Artist Series Module

jiahaodavid:
address: [Grow1354](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1354)

- arrow buttons is not showing in the screenshot because there are only 4 items. So, In the test, I set up a showing the arrow buttons test, and it is passed when there are more than 4 items.

Screen Shot 2019-07-31 at 09 03 20

### DIFF
--- a/src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
+++ b/src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
@@ -1,0 +1,140 @@
+import { Box, color, Flex, Link, Sans, Serif } from "@artsy/palette"
+import { ArtistSeriesEntity_member } from "__generated__/ArtistSeriesEntity_member.graphql"
+import currency from "currency.js"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { data as sd } from "sharify"
+import styled from "styled-components"
+import { get } from "Utils/get"
+
+export interface ArtistSeriesEntityProps {
+  member: ArtistSeriesEntity_member
+}
+
+export const ArtistSeriesEntity: React.FC<ArtistSeriesEntityProps> = ({
+  member,
+}) => {
+  const {
+    headerImage,
+    artworks: { hits },
+    price_guidance,
+    slug,
+    title,
+  } = member
+  const bgImages = hits.map(hit => hit.image.url)
+  const imageSize =
+    bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
+  return (
+    <Container p={0.5} m={0.5}>
+      <StyledLink href={`${sd.APP_URL}/collection/${slug}`}>
+        <ImgWrapper pb={1}>
+          {bgImages.length ? (
+            bgImages.map((url, i) => {
+              const artistName = get(hits[i].artist, a => a.name)
+              const alt = `${artistName ? artistName + ", " : ""}${
+                hits[i].title
+              }`
+              return (
+                <SingleImgContainer key={i}>
+                  <ImgOverlay width={imageSize} />
+                  <ArtworkImage key={i} src={url} width={imageSize} alt={alt} />
+                </SingleImgContainer>
+              )
+            })
+          ) : (
+            <ArtworkImage src={headerImage} width={265} />
+          )}
+        </ImgWrapper>
+        {<CollectionTitle size="3">{title}</CollectionTitle>}
+        {price_guidance && (
+          <Sans size="2" color="black60">
+            From $
+            {currency(price_guidance, {
+              separator: ",",
+              precision: 0,
+            }).format()}
+          </Sans>
+        )}
+      </StyledLink>
+    </Container>
+  )
+}
+
+export const ArtworkImage = styled.img<{ width: number }>`
+  width: ${({ width }) => width}px;
+  height: 125px;
+  background-color: ${color("black10")};
+  object-fit: cover;
+  object-position: center;
+  opacity: 0.9;
+`
+
+const ImgOverlay = styled(Box)<{ width: number }>`
+  height: 125px;
+  background-color: ${color("black30")};
+  opacity: 0.1;
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 7;
+`
+
+const Container = styled(Box)`
+  border: 1px solid ${color("black10")};
+`
+
+const SingleImgContainer = styled(Box)`
+  position: relative;
+  margin-right: 2px;
+
+  &:last-child {
+    margin-right: 0;
+  }
+`
+
+const CollectionTitle = styled(Serif)`
+  width: max-content;
+`
+
+export const ImgWrapper = styled(Flex)`
+  width: 265px;
+`
+
+export const StyledLink = styled(Link)`
+  text-decoration: none;
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+
+  &:hover {
+    text-decoration: none;
+
+    ${CollectionTitle} {
+      text-decoration: underline;
+    }
+  }
+`
+
+export const ArtistSeriesRailContainer = createFragmentContainer(
+  ArtistSeriesEntity,
+  {
+    member: graphql`
+      fragment ArtistSeriesEntity_member on MarketingCollection {
+        slug
+        headerImage
+        thumbnail
+        title
+        price_guidance
+        artworks(size: 3, sort: "-decayed_merch") {
+          hits {
+            artist {
+              name
+            }
+            title
+            image {
+              url(version: "small")
+            }
+          }
+        }
+      }
+    `,
+  }
+)
--- a/src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
+++ b/src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
@@ -1,0 +1,46 @@
+import { CollectionsHubLinkedCollections } from "Apps/__tests__/Fixtures/Collections"
+import { mount } from "enzyme"
+import React from "react"
+import { ArtistSeriesEntity, ArtworkImage } from "../ArtistSeriesEntity"
+jest.unmock("react-tracking")
+
+describe("ArtistSeriesEntity", () => {
+  let props
+
+  beforeEach(() => {
+    props = {
+      member: CollectionsHubLinkedCollections.linkedCollections[0].members[0],
+    }
+  })
+
+  it("Renders expected fields", () => {
+    const component = mount(<ArtistSeriesEntity {...props} />)
+
+    expect(component.text()).not.toMatch("Jasper Johns:")
+    expect(component.text()).toMatch("Flag")
+    expect(component.text()).toMatch("From $1,000")
+    expect(component.find(ArtworkImage).length).toBe(3)
+    const artworkImage = component
+      .find(ArtworkImage)
+      .at(0)
+      .getElement().props
+
+    expect(artworkImage.src).toBe(
+      "https://d32dm0rphc51dk.cloudfront.net/4izTOpDv-ew-g1RFXeREcQ/small.jpg"
+    )
+    expect(artworkImage.alt).toBe("Jasper Johns, Flag")
+    expect(artworkImage.width).toBe(85)
+  })
+
+  it("Returns proper image size if 2 artworks returned", () => {
+    props.member.artworks.hits.pop()
+    const component = mount(<ArtistSeriesEntity {...props} />)
+    const artworkImage = component
+      .find(ArtworkImage)
+      .at(0)
+      .getElement().props
+
+    expect(component.find(ArtworkImage).length).toBe(2)
+    expect(artworkImage.width).toBe(131)
+  })
+})
--- a/src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
+++ b/src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
@@ -18,81 +18,75 @@ describe("ArtistSeriesEntity", () => {
     expect(component.text()).toMatch("Flags unique collections")
     expect(component.text()).toMatch("From $1,000")
     expect(component.find(ArtworkImage).length).toBe(3)
-    const artworkImage = component
-      .find(ArtworkImage)
-      .at(0)
-      .getElement().props
-    expect(artworkImage.src).toBe(
+    expect(
+      component
+        .find(ArtworkImage)
+        .at(0)
+        .getElement().props.src
+    ).toBe(
       "https://d32dm0rphc51dk.cloudfront.net/4izTOpDv-ew-g1RFXeREcQ/small.jpg"
     )
   })
 
-  it("NOT rendering any artist name or the title of the artwork ", () => {
-    const component = mount(<ArtistSeriesEntity {...props} />)
-    expect(component.text()).not.toMatch("A great flag from Jasper")
-    expect(component.text()).not.toMatch("Jasper Johns")
-    expect(component.text()).not.toMatch("Back to 2046")
-    expect(component.text()).not.toMatch("Andy Warhol")
-    expect(component.text()).not.toMatch("An Apple")
-  })
-
   it("uses small image width when there are more than 2 hits", () => {
     const component = mount(<ArtistSeriesEntity {...props} />)
-    const artworkImage = component
-      .find(ArtworkImage)
-      .at(0)
-      .getElement().props
     expect(component.find(ArtworkImage).length).toBe(3)
-    expect(artworkImage.width).toBe(85)
+    expect(
+      component
+        .find(ArtworkImage)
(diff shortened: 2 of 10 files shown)


src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
```diff
@@ -0,0 +1,140 @@
...
+    bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
+  return (
+    <Container p={0.5} m={0.5}>
+      <StyledLink href={`${sd.APP_URL}/collection/${slug}`}>
+        <ImgWrapper pb={1}>
+          {bgImages.length ? (
+            bgImages.map((url, i) => {
+              const artistName = get(hits[i].artist, a => a.name)
+              const alt = `${artistName ? artistName + ", " : ""}${
+                hits[i].title
```
xtina-starr: I wonder if this should simply be the artwork title by the artist name if the artist name is present?

Something like: `${hits[i].title} by {artistName}`

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  let props
+
+  beforeEach(() => {
+    props = {
+      member: CollectionsHubLinkedCollections.linkedCollections[0].members[0],
+    }
+  })
+
+  it("Renders expected fields", () => {
+    const component = mount(<ArtistSeriesEntity {...props} />)
```
pepopowitz: It feels like there are a lot of unrelated assertions in this test. I think it would be beneficial to break them into individual tests, with names that tell us why the assertions are being made.

## Some examples of what I mean

- This test asserts that the text should _not_ match "Jasper Johns:". Why not? A separate test with a more specific name would tell me that. As this test stands, I don't know _why_ this is expected.

- We've rolled a test for the image width into this test, in addition to the [other test](https://github.com/artsy/reaction/pull/2644/files#diff-b458b45b0c5ac2ab639a1327179b482eR35). I'd rather see three individual tests to test the specific image widths - something like 
  1. `it("uses small image width when there are more than 2 hits")`
  2. `it("uses medium image width when there are 2 hits")`
  3. `it("uses large image width when there is 1 hit")`

- This test asserts that the `alt` text on the image equals `"Jasper Johns, Flag"`. I could see two separate tests for this:
  1. `it("uses the hit title for alt text if there is no artist")`, which would assert that the alt text was "Flag" if there was no associated artist
  2. `it("uses the artist name and hit title for alt text if there is an artist")`, which would assert what this test is asserting on line 31.

- There is another test case that is missing: `it("displays price guidance if it exists")`, which would verify that [this branch](https://github.com/artsy/reaction/pull/2644/files#diff-8a679cbe18e53c93fa1219c639cfbd82R49) is rendering when there is `price_guidance`.

## Why this matters

It's great to write tests for code coverage and to catch future regressions when we change the code, but it's important that they provide context to the person reading them when they break. That person is me, or someone else on the team, or even you a few months from now. They don't have the context that you have right now - adding smaller tests with more specific names preserves the context you have right now, and makes it easier for them to figure out _why_ a test broke.

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  })
+
+  it("Renders expected fields", () => {
+    const component = mount(<ArtistSeriesRail {...props} />)
+    expect(component.text()).toMatch("Trending Artist Series")
+    expect(component.text()).toMatch("Flags unique collectins")
+  })
+
+  it("No arrows when there are less than 5 collections", () => {
```
pepopowitz:
```suggestion
  it("Does not show arrows when there are less than 5 collections", () => {
```

A subtle difference, but I think it's important for readability for tests to read like a sentence.

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const component = getWrapper()
+    expect(component.find(ArrowButton).length).toBe(1)
```
pepopowitz: This assertion doesn't match the title of this test. Why do we assert that there is 1 arrow instead of 0?

src/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+          )
+        }}
+      />
+    </Content>
+  )
+}
+
+const Content = styled(Box)`
+  border: 1px solid ${color("black10")};
+  padding-left: 20px;
```
pepopowitz: With our design system [palette](https://github.com/artsy/palette), it's preferable to specify padding like this using the `px` (horizontal padding) and `py` (vertical padding) props on the `<Content>` declaration in the rendering of the component. 

This would mean removing lines 56-59 here, and modifying [line 19](https://github.com/artsy/reaction/pull/2644/files#diff-7146303d88d4a2a5cbd339d3eed6fcbdR19) to read like this: 

`<Content mb={2} px={2} py={3}>`.

We don't have great docs in Palette on what types of props are available for things like padding, but they come from a dependency named [`styled-system`](https://styled-system.com/api). Any of the props defined by `styled-system` are available on things like a `Box` or `Flex` component, and when we _can_ use them, we prefer that to specifying padding in a styled component like this.

One other trick to know is that when you _do_ specify something like `mb={2}` (margin-bottom), `px={2}` (horizontal padding), or `py={3}` (vertical padding), the values are generally changed by a factor of 10. So `px={2}` equates to `padding-left: 20px; padding-right: 20px`, and `py={3}` equates to `padding-top: 30px; padding-bottom: 30px`. These values are all based on our theme, [which defines all the different spacing values](https://github.com/artsy/palette/blob/master/packages/palette/src/Theme.tsx#L116) (amongst other things).

jiahaodavid:
> This is really great work @jiahaodavid! I have a few changes I've requested.
> 
> I tried to add as much context as I could. I hope that the context is helpful and assumes correctly your level of familiarity with our projects. Please let me know if I've provided too many details that you already know, so I can adjust my feedback in the future!

Thank you Steven. All the feedback you made is very useful. Quite honestly, I do not have too much experience on writing the test and working on the company project. Let me check the comments you made one by one.

jiahaodavid:
@pepopowitz 
I added more tests to the code. Basically, I

jiahaodavid:
@pepopowitz

jiahaodavid:
@pepopowitz 

- I added more test to the test file. Most of test is based on your feedback. I think I have more sense on how to write the test. 

- Also, you mentioned there is 1 arrow instead of 0 in the test "does not show arrows when there are less than 5 collections". The reason I set it to 1 is, if you look back to the index file, you can find out I set up a condition check for showing the right arrow button(line 41). Setting up the condition check is for when there are exactly 4 items in the carousel, it shows a right arrow button.(The arrow button problem is related to this PR https://github.com/artsy/reaction/pull/2584) In the line 36, if I did not set up the condition check for the left button, the test will keep indicating there is 1 arrow button when there are less than 5 collections, not 0. The reason is in the Carousel component, there is a boolean value (decide showing the arrow button or not )inside the styled-component. It seems we can not see arrows in the browser, but it is in the DOM. 
To avoid misleading others, I also set up the condition check for left. So, in the test, when there are less than 5 items, the test indicate 0 arrow button.

- The information about horizontal padding and vertical padding. I did not notice before. I made the changes.  

Thank you for the feedback. It is very helpful. 😄

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
```diff
@@ -13,34 +13,86 @@ describe("ArtistSeriesEntity", () => {
...
       .getElement().props
-
     expect(artworkImage.src).toBe(
       "https://d32dm0rphc51dk.cloudfront.net/4izTOpDv-ew-g1RFXeREcQ/small.jpg"
     )
-    expect(artworkImage.alt).toBe("Jasper Johns, Flag")
+  })
+
+  it("NOT rendering any artist name or the title of the artwork ", () => {
+    const component = mount(<ArtistSeriesEntity {...props} />)
```
pepopowitz: I'm still unsure why we are verifying that these aren't appearing. Is there value in not showing these? If so, then the test name should tell us _why_ it's import - like `it("doesn't render artist name or title because they XXXXX")`. 

If the test is verifying that we didn't render these fields to be certain that we didn't write code to render them..those types of assertions are typically avoided in tests. Writing tests that verify "we didn't do XXX" comprehensively would take a really long time, and it doesn't give enough value to justify them.

If you want, we can talk about this some more in person!

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
```diff
@@ -13,34 +13,86 @@ describe("ArtistSeriesEntity", () => {
...
+    expect(component.text()).not.toMatch("A great flag from Jasper")
+    expect(component.text()).not.toMatch("Jasper Johns")
+    expect(component.text()).not.toMatch("Back to 2046")
+    expect(component.text()).not.toMatch("Andy Warhol")
+    expect(component.text()).not.toMatch("An Apple")
+  })
+
+  it("uses small image width when there are more than 2 hits", () => {
+    const component = mount(<ArtistSeriesEntity {...props} />)
+    const artworkImage = component
```
pepopowitz: This is a very subtle change, but assertions should be written in order of specificity. The more general assertions should come first, and the more specific ones should come last. In this example, line 45 is the most general assertion - it should come before line 41. 

This is because tests fail on the first assertion that fails. It's helpful when looking at a failing test if the assertions are written in order of specificity, because it helps you narrow down exactly where your assertions went wrong. The way this code is currently written, if there were zero `ArtworkImage`'s, line 43 would fail because there would be no element `.at(0)`. This _might_ give us an error message that tells us that there are no `ArtworkImage` elements, but it also might not. The assertion in line 45 would _definitely_ give us that error message - so it should come first.

Let me know if this doesn't make sense!

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
```diff
@@ -13,34 +13,86 @@ describe("ArtistSeriesEntity", () => {
...
+    expect(artworkImage.width).toBe(265)
+  })
+
+  it("uses the hit title for alt text if there is no artist", () => {
+    const component = mount(<ArtistSeriesEntity {...props} />)
+    const artworkImage = component
+      .find(ArtworkImage)
+      .at(0)
+      .getElement().props
+    expect(component.find(artworkImage).length).toBe(3)
```
pepopowitz: I think it should be a capital A in `artworkImage`. We want to find the number of elements of a specific component type - which is `ArtworkImage`. `artworkImage` (lowercase a) is a specific element of that component type. I'm surprised this test passes - maybe Enzyme is looking for elements with the same name as the element passed in...regardless, we should correct it.

`expect(component.find(ArtworkImage).length).toBe(3)`

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
```diff
@@ -13,34 +13,86 @@ describe("ArtistSeriesEntity", () => {
...
+      .find(ArtworkImage)
+      .at(0)
+      .getElement().props
+    expect(component.find(ArtworkImage).length).toBe(1)
+    expect(artworkImage.width).toBe(265)
+  })
+
+  it("uses the hit title for alt text if there is no artist", () => {
+    const component = mount(<ArtistSeriesEntity {...props} />)
+    const artworkImage = component
```
pepopowitz: See my above comment about moving the more general assertion about the number of `ArtworkImage` elements above this line.

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
```diff
@@ -13,34 +13,86 @@ describe("ArtistSeriesEntity", () => {
...
+      .at(0)
+      .getElement().props
+    expect(component.find(artworkImage).length).toBe(3)
+    expect(artworkImage.alt).toMatch("A great flag from Jasper")
+  })
+
+  it("uses the artist name and title for alt text if there is an artist", () => {
+    props.member.artworks.hits[0].artist.name = "Jasper Johns"
+    const component = mount(<ArtistSeriesEntity {...props} />)
+    const artworkImage = component
```
pepopowitz: See my above comment about moving the more general assertion about the number of `ArtworkImage` elements above this line.

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
```diff
@@ -41,15 +41,28 @@ describe("ArtistSeriesRail", () => {
...
   })
 
-  it("No arrows when there are less than 5 collections", () => {
+  it("Does NOT show arrows when there are less than 5 collections", () => {
     const component = getWrapper()
-    expect(component.find(ArrowButton).length).toBe(1)
+    expect(component.find(ArrowButton).length).toBe(0)
+  })
+
+  it("Does NOT show arrows when there are exactly 4 collections", () => {
```
pepopowitz: This test and the one above it (`Does NOT show arrow when there are less than 5 collections`) seem redundant. I think we should remove one. 

Of the two, I prefer this test - I like seeing in the setup of the test that we're including 4 collections. But you're welcome to delete this one instead, if you prefer!

src/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx
```diff
@@ -10,13 +10,12 @@ import { ArtistSeriesRailContainer as ArtistSeriesEntity } from "./ArtistSeriesE
...
   collectionGroup: ArtistSeriesRail_collectionGroup
 }
-
 export const ArtistSeriesRail: React.FC<ArtistSeriesRailProps> = ({
   collectionGroup,
 }) => {
   const { members } = collectionGroup
   return (
-    <Content mb={2}>
+    <Content mb={2} px={2} py={3}>
```
pepopowitz: 👍🏽

src/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesEntity.test.tsx
jiahaodavid: Because there are two 'title' in the query, one is for whole group of the collection and one is for single artwork in the group. ( You can find at "src/Apps/__tests__/Fixtures/Collections.ts", one is under the "members", and one is under "hits" . ) I want to make sure the component is rendering the title from the whole group of the collection, not the title of the single artwork. 
The assertion seems it is also misleading people. So, for this one, I will remove this, you are right. It seems it is unnecessary since the test before this one just check the title and text.

jiahaodavid:
@pepopowitz 
I made some change based on the comment. I have a question, when we write the test, should we just test the basic functionality of the component? or we need to assume there are any missing props or any breaking points existed?
The reason I ask that is last time you mention the coverage of the code. It made me to be too sensitive when I was writing the test. Like, I make sure they are rendering the title staff from the right place, because there are 2 things called "title". I make 3 tests (**1**, great than 4; **2**, equal to 4; **3**, greater than 4) to check the arrow buttons will be showed or not, because one ticket I worked on is  when there are less than 5 items, arrows are not there, but the arrows display when there are exactly 4 items. 
 Again, thank you so much. For giving the feedback.

pepopowitz:
> @pepopowitz
> I made some change based on the comment. I have a question, when we write the test, should we just test the basic functionality of the component? or we need to assume there are any missing props or any breaking points existed?
> The reason I ask that is last time you mention the coverage of the code. It made me to be too sensitive when I was writing the test. Like, I make sure they are rendering the title staff from the right place, because there are 2 things called "title". I make 3 tests (**1**, great than 4; **2**, equal to 4; **3**, greater than 4) to check the arrow buttons will be showed or not, because one ticket I worked on is when there are less than 5 items, arrows are not there, but the arrows display when there are exactly 4 items.
> Again, thank you so much. For giving the feedback.

Following up on a conversation we had, so that we have it in writing - we talked about how we should write tests for all branches of code in the component, but generally don't need to worry about the absence of code. In the case of the test where we were verifying that we were using the _correct_ title, that test makes sense, but probably needed a name like `it("renders the correct title")` to give hints about its purpose.

artsyit:
:rocket: PR was released in v16.23.2 :rocket:

PR #2646, opened 2019-07-31
PURCHASE-1282: Updates offer views to handle non-USD currencies

sweir27:
This PR updates the buyer UI to correctly show prices in the order's currency (be it USD or otherwise).

The biggest changes were in [Exchange](https://github.com/artsy/exchange/pull/456) and [Metaphysics](https://github.com/artsy/metaphysics/pull/1858). Now all `amount`-wrapped fields on `Offer`s and `Order`s coming from metaphysics will display in the correct currency.

This just updates the few places where we display a currency symbol based on the `currencyCode` of the `Order` (previously this just used USD).

cc @mdole @iskounen in case this is relevant to the work on the CMS-side!

### Initial offer flow
Screen Shot 2019-07-31 at 11 28 42 AM

### Counteroffer flow
Screen Shot 2019-07-31 at 1 56 06 PM

Screen Shot 2019-07-31 at 1 56 15 PM

### DIFF
--- a/src/Apps/Order/Routes/Offer/index.tsx
+++ b/src/Apps/Order/Routes/Offer/index.tsx
@@ -168,6 +168,7 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
     }
 
     const listPriceCents = this.props.order.totalListPriceCents
+    const orderCurrency = this.props.order.currencyCode
 
     if (!lowSpeedBumpEncountered && offerValue * 100 < listPriceCents * 0.75) {
       this.showLowSpeedbump()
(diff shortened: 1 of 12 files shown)


sweir27:
I thought this would be WIP because it depends on metaphysics changes, but it doesn't actually consume anything from the newer schema (it just won't look correct in prod until everything is deployed).

ashkan18:
Nice and easy!! 🙌

artsyit:
:rocket: PR was released in v16.20.3 :rocket:

PR #2647, opened 2019-07-31
Collections Default Header Bug Fix

ashleyjelks:
This PR addresses 2 issues in the collections default header:

1) Fixes an infinite loop bug that caused collections pages with a default header to crash.
2) Limits the number of artworks in the standard collections header and default collections header to 12. 

[Links to GROW-1448](https://artsyproduct.atlassian.net/browse/GROW-1448)
Kapture 2019-07-31 at 15 00 27


src/Apps/Collect/Components/Collection/Header/__tests__/DefaultHeader.test.tsx
```diff
@@ -28,15 +28,15 @@ describe("artworks", () => {
...
   })
 
   it("returns only the number of artworks necessary to fill the header", () => {
     const artworks = defaultCollectionHeaderArtworks.hits
 
     const headerArtworks = getHeaderArtworks(artworks, 1375, false)
-    expect(artworks.length).toBeGreaterThan(headerArtworks.length)
-    expect(headerArtworks).toHaveLength(7)
+    expect(artworks.length).toBeGreaterThanOrEqual(headerArtworks.length)
+    expect(headerArtworks).toHaveLength(10)
```
pepopowitz: It's not obvious to me why this test would have gone from requiring 7 images to fill the header to requiring 10. Can you help me understand? I would have expected this not to change, since we're only placing as many images as it takes to fill the header.

src/Apps/Collect/Components/Collection/Header/__tests__/DefaultHeader.test.tsx
```diff
@@ -70,7 +70,7 @@ describe("default header component", () => {
   it("renders images in the header component ", () => {
     const wrapper = getWrapper(props)
 
-    expect(wrapper.find("Image").length).toBe(5)
+    expect(wrapper.find("Image").length).toBe(10)
```
pepopowitz: Same as above - it's not obvious to me why this test would have gone from requiring 5 images to fill the header to requiring 10. Can you help me understand? I would have expected this not to change, since we're only placing as many images as it takes to fill the header.

ashleyjelks: I'm returning more works than are in the set to handle window resizing.

pepopowitz: What will the effect be for a user on mobile who is viewing a collection with many header images? Will their device download only the number of images they can see, or will it download all (or 10) header images, only 3 of which they can see?

Also - I'm going to merge this as-is, since it fixes a crash. We can follow up with my question elsewhere.

ashleyjelks: Good question, the tests indicate that for a mobile device a user might have up to 4 images to process, and depending on wide their screen is they might see only 3, or 3 and part of 4 since the overflow is acceptable and the banner is intended to look like its a row of repeating images. I have a ticket in the backlog to trigger a component update on window re-size, since the hook isn't working as expected. But this should be merged pretty soon so we don't have collections pages timing out.

artsyit:
:rocket: PR was released in v16.20.2 :rocket:

PR #2650, opened 2019-07-31
[RecentlyViewed Carousel] Make consistent with recent patterns

damassi:
Noticed that our `RecentlyViewed` carousel hadn't been updated with some of our more recent patterns (show hide arrows, alignment, etc); this brings it in line. 

recently-viewed

### DIFF
--- a/src/Components/v2/RecentlyViewed.tsx
+++ b/src/Components/v2/RecentlyViewed.tsx
@@ -1,4 +1,4 @@
-import { Separator, Serif, Spacer } from "@artsy/palette"
+import { Box, Separator, Serif, Spacer } from "@artsy/palette"
 import { RecentlyViewed_me } from "__generated__/RecentlyViewed_me.graphql"
 import { RecentlyViewedQuery } from "__generated__/RecentlyViewedQuery.graphql"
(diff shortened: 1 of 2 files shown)


src/Components/v2/RecentlyViewed.tsx
```diff
@@ -67,6 +68,22 @@ export class RecentlyViewed extends React.Component<RecentlyViewedProps> {
...
+                    return (
+                      <ArrowContainer>
+                        <Arrow />
+                      </ArrowContainer>
+                    )
+                  }}
+                  renderRightArrow={({ Arrow }) => {
+                    return (
+                      <ArrowContainer>
+                        {me.recentlyViewedArtworks.edges.length > 4 && (
```
kierangillen: @damassi Ah we should expose the `checkLastItemVisible` so we don't need to do this type of thing. The carousel should decide to show the arrow based on what's still visible. Could scope for another PR.

damassi: Sounds good -- there are a few spots around Reaction that we could update all at once (in diff PR, as mentioned)

artsyit:
:rocket: PR was released in v16.20.1 :rocket:

PR #2640, opened 2019-07-30
Address more MP v2 deprecations

damassi:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-1641

Fixes a few more MP deprecations:
- `price` -> `listPrice`. List price can return `priceCents`, `minPriceCents` or `maxPriceCents`
- `short_description` -> `shortDescription`
- `to_s` -> `formattedMetadata`

The last remaining deprecations (to be fixed in https://github.com/artsy/reaction/pull/2552):

Screen Shot 2019-07-31 at 12 08 10 PM

PR #2655, opened 2019-08-01
Revert #2640 due to stitching edge case

zephraph:
cc @xtina-starr @ashleyjelks 

This temporarily reverts #2640 due to an edge case that was uncovered with schema stitching that causes Kaws collections to error out. 

Essentially the `ListPrice` union uses `ExactPrice` which is getting renamed to `MarketingExactPrice` when it shouldn't. Using `listPrice` in a collection query currently causes the query to fail. 

We'll restore these changes once we figure out a fix.


artsyit:
:rocket: PR was released in v16.21.1 :rocket:

alloy:
For a potential fix see https://github.com/artsy/metaphysics/pull/1878

PR #2656, opened 2019-08-01
[PURCHASE-1075, PURCHASE-1287]: Remove fake stitching, display SCA modal when action is required

lilyfromseattle:
__WIP till artsy/metaphysics#1863 is deployed to production__

# [PURCHASE-1075] Replace ecommerce with commerce for all mutations in reaction (get rid of fake stitching!)

Paired with @ds300 on this. It will need __A LOT__ of QA.

# [PURCHASE-1287] If user "requires_action" after submitting an order for buy now, they see a popup asking them to authenticate

The mutation returns a client secret which we pass to Stripe’s library and they handle the rest.

In the case of successful authentication, we re-submit on behalf of the user.

In the case of failed authentication, we won’t get notified, but presumably the user will close the popup and could re-try from there.

[PURCHASE-1075]: https://artsyproduct.atlassian.net/browse/PURCHASE-1075
[PURCHASE-1287]: https://artsyproduct.atlassian.net/browse/PURCHASE-1287


src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -41,10 +41,30 @@ export interface ReviewProps {
...
 const logger = createLogger("Order/Routes/Review/index.tsx")
 
 @track()
-export class ReviewRoute extends Component<ReviewProps> {
+export class ReviewRoute extends Component<ReviewProps, ReviewState> {
+  state = { stripe: null }
+  componentDidMount() {
+    if (window.Stripe) {
+      this.setState({
+        stripe: window.Stripe(window.sd.STRIPE_PUBLISHABLE_KEY),
```
ashkan18: @lilyfromseattle we talked about this, @ds300 @sweir27 is there a better way to get access to current stripe context and not get it from `window`?

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -63,39 +83,66 @@ export class ReviewRoute extends Component<ReviewProps> {
...
-          : (await this.submitOffer()).ecommerceSubmitOrderWithOffer
-              .orderOrError
+          ? (await this.submitBuyOrder()).commerceSubmitOrder.orderOrError
+          : (await this.submitOffer()).commerceSubmitOrderWithOffer.orderOrError
 
       if (orderOrError.error) {
         this.handleSubmitError(orderOrError.error)
         return
+      } else if (
+        this.isOrder(orderOrError) &&
```
ashkan18: this name is little confusing, are we trying to see if this order requires action? is it more like `this.orderRequiresAction`?

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -70,8 +70,10 @@ const logger = createLogger("Order/Routes/Shipping/index.tsx")
 export class ShippingRoute extends Component<ShippingProps, ShippingState> {
   state: ShippingState = {
     shippingOption: ((this.props.order.requestedFulfillment &&
-      this.props.order.requestedFulfillment.__typename.toUpperCase()) ||
-      "SHIP") as OrderFulfillmentType,
+      (this.props.order.requestedFulfillment.__typename === "CommerceShip"
```
ashkan18: is there a way we could make this little more readable? having hard time reading this line's logic

src/Apps/Order/Routes/Review/index.tsx
lilyfromseattle: This was to prevent type errors because offer responses will never have `actionData`. `isOrder` just checks to see if the order is an order and not an offer

src/Apps/Order/Routes/Review/index.tsx
sweir27: Usually we import `sd` variables through the `"sharify"` module. So add this:
```
import { data as sd } from "sharify"
```
to your imports and you can pull it directly off of `sd`. ([Example here](https://github.com/artsy/reaction/blob/master/src/Components/Payment/PaymentFormWrapper.tsx#L4))

src/Apps/Order/Routes/Shipping/index.tsx
lilyfromseattle: Just updated it!

src/Apps/Order/Routes/Review/index.tsx
ashkan18: cool, makes sense, i would just rename it to `isBuyNowOrder` then, since offer's have Order too.

ashkan18:
Metaphysics is deployed with these changes, we should be ok to not have this as WIP anymore.

ashkan18:
alright, merging! 🙌

artsyit:
:rocket: PR was released in v16.23.4 :rocket:

PR #2657, opened 2019-08-01
GROW-1453: this moves CollectionsHubRail into the CollectionApp

xtina-starr:
Addresses: [GROW-1453](https://artsyproduct.atlassian.net/browse/GROW-1453)

This brings in the CollectionHubsRail into the Collection page. This is not ready to launch yet so it's guarded by an admin flag.

Screen Shot 2019-08-01 at 5 36 33 PM


src/Apps/Collect/CollectionApp.tsx
```diff
@@ -83,7 +92,7 @@ export class CollectionApp extends Component<CollectionAppProps> {
 }
 
 export const CollectionAppFragmentContainer = createFragmentContainer(
-  CollectionApp,
+  withSystemContext(CollectionApp),
```
damassi: `withSystemContext` was crerated to address legacy code; we should be using the hook form

pepopowitz: It looks like this _is_ a legacy class component. Are you suggesting we convert it to a functional component as part of this work?

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -35,8 +37,10 @@ export class CollectionApp extends Component<CollectionAppProps> {
   }
 
   render() {
-    const { collection } = this.props
+    const { collection, user } = this.props
     const { title, slug, headerImage, description, artworks } = collection
+    const showCollectionHubs =
+      collection.linkedCollections.length > 0 && user && user.type === "Admin"
```
pepopowitz: There is a util named `userIsAdmin` that can be used to determine if the user is an admin - https://github.com/artsy/reaction/blob/master/src/Utils/user.ts#L34. I know it doesn't replace much code, but it has always felt better to me to use a common function instead of having `user && user.type === "Admin"` throughout the code. If we ever change the way we determine if the user is an admin, we'd have a _ton_ of places to change it.

src/Components/CollectionsHubRails/index.tsx
```diff
@@ -37,13 +35,12 @@ export const CollectionsHubRails = ({ marketingCollection }: Props) => {
...
   CollectionsHubRails,
   {
-    marketingCollection: graphql`
-      fragment CollectionsHubRails_marketingCollection on MarketingCollection {
-        linkedCollections {
-          groupType
-          ...FeaturedCollectionsRails_collectionGroup
-          ...OtherCollectionsRail_collectionGroup
-        }
+    linkedCollections: graphql`
```
pepopowitz: Can you give me some context as to why and how this fragment changed? I don't have any complaints or suggestions, I just don't really understand it.

src/Apps/Collect/CollectionApp.tsx
damassi: Ah, nope, since this is already a class all good 👍

ashleyjelks: Can you explain why we add this context here? Is it to pass props to CollectionsHubRails without having to pass through every component in the CollectionsApp tree?

src/Components/CollectionsHubRails/index.tsx
ashleyjelks: Same! Would love to understand this better!

ashleyjelks:
I had a couple of questions more about understanding how you're using system context in the hubs rail and Relay stuff, and I would recommend adding a test to check that this new component  is hidden behind an admin flag, but otherwise its a 👍 from me

src/Apps/Collect/CollectionApp.tsx
```diff
@@ -122,6 +131,9 @@ export const CollectionAppFragmentContainer = createFragmentContainer(
         relatedCollections {
           ...RelatedCollectionsRail_collections
         }
+        linkedCollections {
```
ashleyjelks: What's the difference between related collections and linked collections?

src/Apps/Collect/CollectionApp.tsx
pepopowitz: @ashleyjelks my understanding is that it gives access to the `user` prop, via `SystemContextProps`.

xtina-starr: Yeah, this is actually temporary. Sense we aren't ready to launch this yet grabbing the user from the context allows me to conditionally display this to only admins.

src/Apps/Collect/CollectionApp.tsx
xtina-starr: This makes sense!

src/Apps/Collect/CollectionApp.tsx
xtina-starr: [Related collections](https://github.com/artsy/kaws/blob/c45a8899867a15af35a5c4888e6d772c7e6c683b/src/Resolvers/Collections.ts#L93-L124) is based on either the `artist_ids` field in a collection query or the category. This is used to populate the related collection rail that's at the bottom of collection pages. Linked collections are hand curated collections developed by the marketing team. The presence of linked collection is essentially what makes a collection a hub. Not every collection will have linked collection but every collection will have related collections.

src/Components/CollectionsHubRails/index.tsx
xtina-starr: Previously we were passing in the entire collection (i.e. `marketingCollection`). For the hubs rail we only need the `linkedCollection` which is a property on the collection so I changed the fragment to reflect that.

Does that makes sense?

ashleyjelks: Totally, thanks for the clarification!

src/Apps/Collect/CollectionApp.tsx
ashleyjelks: Got it!

src/Apps/Collect/CollectionApp.tsx
ashleyjelks: Ohh, that makes sense. I think `curatedCollections` might be a better name in that case, just so its a little more clear how it's different from the `relatedCollections`. But I understand that renaming things while meeting Relay's naming conventions can be precarious.

xtina-starr:
@pepopowitz okay I've used the `userIsAdmin` util, added a test, and resolved the merge conflict. I'm going to go ahead and merge this

artsyit:
:rocket: PR was released in v16.22.4 :rocket:

PR #2666, opened 2019-08-05
Co-locate Vanguard assets and components

eessex:
Moves Vanguard assets and story to publishing directory, removes extra asset file.


artsyit:
:rocket: PR was released in v16.22.3 :rocket:

PR #2669, opened 2019-08-05
[EXPERIMENT] Refactor and consolidate /collect, /collection, /collections

damassi:
#### WIP

This PR drops in the new Artwork v2 filter into `/collect` while also refactoring / consolidating `/collect`, `/collection` and `/collections` to all exist under one app, which will substantially improve performance. 

#### Architecture 
- Currently there are multiple apps and route instances that render the `/collect` route ecosystem
- These apps / routers have all been consolidated under a single router, with subroutes broken out into `Apps/Collect/Routes/<RouteName>` pattern, following other apps
- By moving everything together, once the app renders on the client we switch to single page app mode and navigation between app routes all happens on the client, leading to instant feedback.


damassi:
Closing this until we drop the new ArtworkGrid in

PR #2672, opened 2019-08-06
Fix render implementation in AuctionApp routes

dblandin:
If provided, a `render` function in a [found][] router implementation will execute before and after a Relay response is available. To account for the early invocation, guard against the presence of `props`.

[found]: https://github.com/4Catalyzer/found

Co-authored-by: @dleve123

### DIFF
--- a/src/Apps/Auction/routes.tsx
+++ b/src/Apps/Auction/routes.tsx
@@ -49,20 +49,21 @@ export const routes: RouteConfig[] = [
     path: "/auction-registration2/:saleID",
     Component: AuctionAppFragmentContainer,
     render: ({ Component, props }) => {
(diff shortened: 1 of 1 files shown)


erikdstock:
I think this is save to #mergeongreen

mzikherman:
I think it's more of a Relay Query Renderer thing (vs. found), where it is rendered with `props` being undefined (allowing you to have a loading state), and then again with the actual data.

That's vaguely mentioned in the Relay docs: https://relay.dev/docs/en/query-renderer

A typical pattern might be:

```
if (props)
  return Component
else
  return Spinner
```

kinda thing.

artsyit:
:rocket: PR was released in v16.22.11 :rocket:

PR #2680, opened 2019-08-12
Checkbox onClick

kierangillen:
- Changes input to be on an `onClick` event rather than `onChange` to better handle mobile push event on the `CheckBox` component

### DIFF
--- a/src/Components/Checkbox.tsx
+++ b/src/Components/Checkbox.tsx
@@ -61,7 +61,7 @@ export class Checkbox extends Component<CheckboxProps, CheckboxState> {
             error={error}
           />
 
-          {checked && <PositionedCheckmark fill={fill} />}
+          {!disabled && checked && <PositionedCheckmark fill={fill} />}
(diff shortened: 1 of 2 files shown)


src/Components/Checkbox.tsx
```diff
@@ -20,9 +20,15 @@ export class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
-      this.props.onChange(event)
+  onClick = event => {
+    const { disabled } = this.props
+
+    if (disabled) {
+      return
+    }
+
+    if (this.props.onClick) {
+      this.props.onClick(event)
```
damassi: #minor, if already pulling out disabled from `props` might as well also pull out `onClick`

dleve123:
@dblandin and @erikdstock got this palette-ified. Thanks @kierangillen for showing a path forward here and getting the ball rolling. Closing in favor of https://github.com/artsy/reaction/pull/2682

PR #2681, opened 2019-08-12
[Vanguard] Article truncation

eessex:
Adds expand/close actions to artist stub articles on the vanguard page.  This PR updates the `Sections` component to accept an `isTruncatedAt` index, which crops the article sections array when present. This may be useful in the future as a way to consolidate our varied 'read more' strategies currently in place.

Text inversion has also been added to artist articles.

vanguard-expand


src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
```diff
@@ -28,37 +28,49 @@ export class VanguardArtistWrapper extends React.Component<
...
-          <Box position="absolute">
-            <Share
-              // TODO: We may need to use custom urls for in-page routing
-              url={getFullEditorialHref(layout, slug)}
-              title={title}
-            />
+      <Box background={isExpanded && "black"}>
+        <ArtistContainer pb={4} maxWidth={1000} px={4} mx="auto">
+          <Box textAlign="center">
+            <ArtistTitle size="8" color={textColor}>
```
ashleyjelks: Do we know how styled components will resolve an undefined color? Should we just default it to something?

src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
```diff
@@ -28,37 +28,49 @@ export class VanguardArtistWrapper extends React.Component<
     const { article } = this.props
     const { hero_section, layout, slug, title } = article
     const { isExpanded } = this.state
+    const textColor = isExpanded ? "white" : undefined
```
ashleyjelks: Use `color("white100")` here?

src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
```diff
@@ -28,37 +28,49 @@ export class VanguardArtistWrapper extends React.Component<
...
+
+          <ReadMoreButton
+            size="5"
+            weight="medium"
+            textAlign="center"
+            onClick={this.onExpand.bind(this)}
+            color={textColor}
+          >
+            {isExpanded ? "Close" : "Read More"}
+            <Sans size="8">{isExpanded ? "\u2191" : "\u2193"}</Sans>
```
ashleyjelks: I wonder if we should use the icons in Palette instead of unicode.  I know the icon is slight different than our up/down arrows, but I'm thinking about 1) browser font support or lack thereof and also if we should stay consistent with the icons in our design system.

src/Components/Publishing/Sections/ImageSetPreview/ImageSetLabel.tsx
```diff
@@ -14,20 +14,26 @@ export const ImageSetLabel = (props: ImageSetPreviewProps) => {
   const label = images.length === 1 ? "Image" : "Images"
   const imageCount = `${images.length} ${label}`
   const primaryTitle = title ? title : imageCount
+  const textColor = color !== "white" ? color : undefined
```
ashleyjelks: See previous comment

src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
eessex: They will default to black or the system color

src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
eessex: Heads up these had already been merged, I just reversed the order -- i'm open to using icons, going to check in with editorial to see how they feel about using our system icons. Are there known issues with unicode errors? They seem to be fine in most fonts and have been working without issue in the feature so far.

src/Components/Publishing/Sections/ImageSetPreview/ImageSetLabel.tsx
eessex: this is confirming the value of the input, not declaring it -- i believe `color("white100")` resolves to `"white"` so this should work as expected.

src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
ashleyjelks: I think it will be fine in most browsers, but from what I recall no single browser can decode every Unicode char (without a software install/update of some sort at least) so users running outdated or less popular browsers might see the text  `"\u2191"` instead of what it compiles to. But my suggestion to use Palette's icons was more about consistency throughout Artsy and using our design system moreso than browser issues.

artsyit:
:rocket: PR was released in v16.24.10 :rocket:

PR #2683, opened 2019-08-12
Extract tooltip provider from fullscreen logic

eessex:
Our `FullscreenProvider` was doing double duty, because it also contained logic related to fetching Tooltip data -- this PR separates the tooltip logic into its own component, `TooltipsDataProvider`. 

I ran into issues with the co-location when working on changes to slideshows in the Vanguard project, this should unblock that work.

### DIFF
--- a/src/Components/Publishing/Constants.ts
+++ b/src/Components/Publishing/Constants.ts
@@ -162,12 +162,14 @@ export const getArtsySlugsFromArticle = (
   article: ArticleData
 ): SlugsFromArticle => {
   const articleBody = article.sections
-    .map(section => {
-      if (section.type === "text") {
-        return section.body
-      }
-    })
-    .join()
+    ? article.sections
+        .map(section => {
+          if (section.type === "text") {
+            return section.body
+          }
+        })
+        .join()
+    : ""
 
   const artists = uniq(getArtsySlugsFromHTML(articleBody, "artist"))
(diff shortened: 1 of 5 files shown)


src/Components/Publishing/Constants.ts
```diff
@@ -162,12 +162,14 @@ export const getArtsySlugsFromArticle = (
...
-    })
-    .join()
+    ? article.sections
+        .map(section => {
+          if (section.type === "text") {
+            return section.body
+          }
+        })
+        .join()
+    : ""
```
eessex: In the old setup, we never wrapped `series` layouts, which have no `sections`. This is a minor change to make sure the provider doesn't fail on articles without `sections`.

src/Components/Publishing/ToolTip/TooltipsDataProvider.tsx
```diff
@@ -0,0 +1,19 @@
+import { ArticleProps } from "Components/Publishing/Article"
+import React from "react"
+import { TooltipsData } from "./TooltipsDataLoader"
+
+export class TooltipsDataProvider extends React.Component<ArticleProps> {
```
damassi: Minor, but can this be a `React.FC`?

artsyit:
:rocket: PR was released in v16.24.0 :rocket:

PR #2685, opened 2019-08-12
[Refactor] Consolidate /collect, /collections and /collection/id under a single App / Router

damassi:
Addresses: https://artsyproduct.atlassian.net/browse/PLATFORM-1692
Related: https://github.com/artsy/force/pull/4396

Continues work in https://github.com/artsy/reaction/pull/2669, but without the new `<ArtworkFilter>` component. 

This PR consolidates `/collect`, `/collection` and `/collections` to all exist under one app, which will substantially improve performance. 

#### Architecture 
- Currently there are multiple apps and route instances that render the `/collect` route ecosystem
- These apps / routers have all been consolidated under a single router, with subroutes broken out into `Apps/Collect/Routes/<RouteName>` pattern, following other apps
- By moving everything together, once the app renders on the client we switch to single page app mode and navigation between app routes all happens on the client, and navigating between pages, due to caching, is instant once the path has been visited once by the user.

#### Additional Changes
- Fixed performance issue in `/collection/id`, where we were cloningDeep on the react tree; refactored that bit of code into a new component
- Fixed performance issue in `/collections` due to thousands of items being rendered out on the dom. What happens now is on the server it will render out all of the items for (SEO), and then on the client it will truncate to 20, just below the fold. This will ensure that we have snappy transitions between `/collect`, `/collections` and `/collection/id` once the client mounts and the user is navigating back and forth between pages. 
- Added a `showFetchIndicator = true` prop to our route config to toggle off the spinner when transitioning between routes. When set to `false` this eliminates the hard jump in between SPA-style pages,  instead fetching the data in the background _and then_ transitioning.  

#### Screenshot: 

Note that perf will be much faster on prod, but this showcases page caching and so on: 

coll

#### Remaining Tasks

Once things look good in prod, will  rename `Apps/Collect2` to `Apps/Collect` in follow-up PR.


src/Apps/Collect2/Components/Collection/Header/FeaturedArtists.tsx
```diff
@@ -0,0 +1,80 @@
...
+  breakpointSize: Breakpoint
+}
+
+export const FeaturedArtists: React.FC<FeaturedArtistsProps> = props => {
+  const { featuredArtists, breakpointSize, hasMultipleArtists } = props
+  const artistCount = getArtistCountAtBreakpoint(breakpointSize)
+  const remainingCount = featuredArtists.length - artistCount
+  const truncatedArtists = featuredArtists.slice(0, artistCount)
+  const [showAll, setShowMore] = useState(false)
+  const headlineLabel = "Featured Artist" + (hasMultipleArtists ? "s" : "")
```
damassi: This code was refactored out of `Header/index`, fixing the performance bottleneck

src/Apps/Collect2/Components/Collection/Header/__tests__/Header.test.tsx
```diff
@@ -321,7 +324,7 @@ describe("collections header", () => {
 
       expect(entityHeaders.length).toEqual(4) // 4 = 3 artists + 1 "View More"
 
-      const viewMore = component.find(ViewMore)
+      const viewMore = component.find("ViewMore")
```
pepopowitz: Out of curiosity, is this a preference thing, or is there an objective argument to `component.find` by string instead of component reference?

src/Apps/Collect2/Components/Collection/Header/index.tsx
```diff
@@ -266,8 +226,8 @@ export const CollectionHeader: FC<Props> = ({ artworks, collection }) => {
                 )}
                 <MetaContainer mb={2} mt={[0, imageHeightSizes.sm + space(3)]}>
                   <BreadcrumbContainer size={["2", "3"]}>
-                    <a href="/collect">All works</a> /{" "}
-                    <a href={categoryTarget}>{collection.category}</a>
+                    <Link to="/collect">All works</Link> /{" "}
```
pepopowitz: 👍

src/Apps/Collect2/Components/Collection/Header/__tests__/Header.test.tsx
damassi: Less complexity IMO -- with an exported component, one 
a) has now made something that's effectively private, public
b) added an additional import

Most React components automatically have `displayName` attached, so its seamless to piggy back on that. Often, though, one needs to operate on a component (like `find(Foo).instance().callSomeMethod()` which will error out in TS if using string access. But this case is less common.

artsyit:
:rocket: PR was released in v17.0.0 :rocket:

PR #2686, opened 2019-08-13
Bump react to 16.8.4

eessex:
Bumps `react` version to match `react-dom` and versions used in force, in an attempt to mitigate Force testing issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "prettier": "1.16.4",
     "raf": "3.4.0",
     "raw-loader": "0.5.1",
-    "react": "16.8.2",
+    "react": "16.8.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -138,7 +138,7 @@
     "prettier": "1.16.4",
     "raf": "3.4.0",
     "raw-loader": "0.5.1",
-    "react": "16.8.2",
+    "react": "16.8.4",
```
damassi: Interesting that this fell out of sync with `react-dom` below; they should always be the same

damassi:
👍 

#mergeOnGreen

artsyit:
:rocket: PR was released in v16.24.2 :rocket:

PR #2689, opened 2019-08-13
Adds prop to hide ads from feature articles for vanguard development

ashleyjelks:
Currently, we are not expecting to render ads on the 2019 Artsy Vanguard article series.

This PR adds a boolean property we can use to render/hide display ads.


artsyit:
:rocket: PR was released in v16.24.5 :rocket:

PR #2692, opened 2019-08-14
Refactor svg blobs on artist headers

ashleyjelks:
Kapture 2019-08-14 at 0 48 15

### DIFF
--- a/src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
+++ b/src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
@@ -1,4 +1,4 @@
-import { Box, color, FlexProps, Sans, Serif } from "@artsy/palette"
+import { Box, color, Flex, FlexProps, Sans, Serif } from "@artsy/palette"
 import { Share } from "Components/Publishing/Byline/Share"
 import { getFullEditorialHref } from "Components/Publishing/Constants"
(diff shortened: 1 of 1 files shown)


src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/ArtistWrapper.tsx
```diff
@@ -112,11 +108,12 @@ const ArtistTitle = styled(Serif)`
   line-height: 1.35em;
   color: ${color("white100")};
 `
-
 const ReadMoreButton = styled(Sans)<{ onClick: () => void }>`
   text-transform: uppercase;
 `
 const ArtistContainer = styled(Box)`
+  position: relative;
+  bottom: 750px;
```
eessex: What is the `bottom` attr for?

ashleyjelks:
> One small q, otherwise looks good to me 👍

The SVG and article content are wrapped in a Flex container so that the height of each section adjusts properly without having text content overflowing into the next section. This creates a stacking effect where the SVG renders on top of the article content rather than beneath it like a background. To prevent this I added a `position: relative` attr to the article content wrapper so that when I apply the `bottom` attr it is relative to the parent (the `<ArtistWrapper>` component), and it pushes the content from the bottom of the container to the top, so that it renders on top of the SVG.

artsyit:
:rocket: PR was released in v16.24.6 :rocket:

PR #2693, opened 2019-08-14
[ Draft ] Grow-1371 Analytics tracking for "Artist Series" module

jiahaodavid:
@damassi 
Hi, I am working on the [Grow-1371](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1371).

I got stuck on tracking item number. The goal is when people click on one item in the carousel, the data collector will receive the index of the item in the carousel. I tried many ways but failed. It is in the ArtistSeriesEntity.tsx.
Do you have any idea on this one? 

Screen Shot 2019-08-14 at 11 12 52


src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
```diff
@@ -1,9 +1,10 @@
 import { Box, color, Flex, Link, Sans, Serif } from "@artsy/palette"
 import { ArtistSeriesEntity_member } from "__generated__/ArtistSeriesEntity_member.graphql"
+import * as Schema from "Artsy/Analytics/Schema"
```
damassi: `import { AnalyticsSchema } from 'Artsy/Analytics'`

src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
```diff
@@ -24,9 +25,24 @@ export const ArtistSeriesEntity: React.FC<ArtistSeriesEntityProps> = ({
   const bgImages = hits.map(hit => hit.image.url)
   const imageSize =
     bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
+  const { trackEvent } = useTracking()
+  const onClickLink = event => {
+    const link = event.target
+    const href = link.parentNode.parentNode.parentNode.getAttribute("href")
```
damassi: Can this `href` be added to the graphql request below and then plucked from props?

src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
```diff
@@ -24,9 +25,24 @@ export const ArtistSeriesEntity: React.FC<ArtistSeriesEntityProps> = ({
   const bgImages = hits.map(hit => hit.image.url)
   const imageSize =
     bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
+  const { trackEvent } = useTracking()
+  const onClickLink = event => {
```
damassi: So that this is a bit more readable, can a) a line break be added between 28 and 29 so that the function handler jumps out; and b) the variable renamed to `handleLinkClick`; `onLinkClick` is the form typically passed down from props e.g., 

```jsx
<Foo onClick={...}>
```
and then inside 
```jsx
const Foo = props => {
  if (props.onClick) { 
    ...
  }
}
```
Since we're "handling" an action, these functions are typically prefixed with "handle"; so :

```jsx
const handleLinkClick = event => { ... }
```

(Similarly, if it were a mouse-over, it would be `handleMouseOver`, etc.)

(cc @xtina-starr )

src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
```diff
@@ -24,9 +25,24 @@ export const ArtistSeriesEntity: React.FC<ArtistSeriesEntityProps> = ({
...
+      type: Schema.Type.Thumbnail,
+      destination_path: href,
+      // item_number: itemPlacement,
+    })
+  }
+
   return (
     <Container p={0.5} m={0.5}>
-      <StyledLink href={`${sd.APP_URL}/collection/${slug}`}>
+      <StyledLink href={`/collection/${slug}`} onClick={onClickLink}>
```
damassi: `onClick={handleClickLink}`

src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
jiahaodavid: href is not available from the graphql request. 

Screen Shot 2019-08-20 at 11 03 06

PR #2694, opened 2019-08-14
[TRIVIAL] Update fixture to use correct naming for related_article_ids

eessex:
Found an incorrect fixture while pairing with @mbilokonsky


artsyit:
:rocket: PR was released in v16.24.8 :rocket:

PR #2695, opened 2019-08-14
[Carousel] Pass index / props down to render items

damassi:
Updates the carousel to pass back props / index so that if needed, we can pass that data along to tracking (or whatever).

### DIFF
--- a/src/Components/v2/Carousel.tsx
+++ b/src/Components/v2/Carousel.tsx
@@ -39,12 +39,16 @@ interface CarouselProps {
   /**
    * Callback when forward / backward arrows are clicked
    */
-  onArrowClick?: () => void
+  onArrowClick?: ({
+    props: CarouselProps,
+    state: BaseCarouselState,
+    flickity: FlickityOptions,
+  }) => void
 
   /**
    * The render callback returns an image to display
    */
-  render: (slide) => React.ReactNode
+  render: (slide, slideIndex: number) => React.ReactNode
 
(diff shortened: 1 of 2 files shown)


src/Components/v2/Carousel.tsx
```diff
@@ -230,17 +234,21 @@ export class BaseCarousel extends React.Component<
...
     if (!showArrows) {
       return null
     }
 
     const leftPosition =
-      this.props.arrowPosition !== null &&
-      this.props.arrowPosition !== undefined
-        ? this.props.arrowPosition
-        : -space(4)
+      arrowPosition != null ? this.props.arrowPosition : -space(4)
```
jiahaodavid: I am curious about why did you make this change here? The reason I ask this is the undefined checking seems to make sense. 
Also, you already did the destructing above. When the condiction was true, is there any differences between this.props.arrowPosition and arrowPosition?

damassi: So, the change with `!== undefined` and `!== `null` to `!=` is a JS shortcut for checking both undefined _and_ null at the same time. Its a weird hack :) 

Re the last `this.props.arrowPosition`, oops! I need to update that spot too. Nice catch 👍

xtina-starr:
Thanks for adding this @damassi!! 🙏

artsyit:
:rocket: PR was released in v16.24.12 :rocket:

PR #2697, opened 2019-08-15
[TRIVIAL] Put back super article related_articles

eessex:
Changed the `SuperArticle` fixture incorrectly, it does indeed use `related_articles` and not the newer `related_article_ids` (https://github.com/artsy/reaction/pull/2694/files), breaking Force tests.


artsyit:
:rocket: PR was released in v16.24.11 :rocket:

PR #2698, opened 2019-08-15
GROW 1362 - Add hub entry points to collect page

pepopowitz:
In collaboration with @anandaroop. This PR addresses https://artsyproduct.atlassian.net/browse/GROW-1362. It adds a collection hubs rail to the Collect page, if the user is an admin.

From a design perspective, we want the tiles to be as large as possible to fill the width with 2 items per row at xs, 3 items per row at sm, and 6 items per row at md & above. You can see this effect below. Note that only one cell has a background image right now - it's the only one with a thumbnail defined. 

auto-scaled-hubs

The CSS to auto-scale the hub tiles & background images based on the available width is a bit tricky. We use CSS grid to lay out the rail, and for each tile, we render an image of specific size but hide it. This gives the container a certain width and height. The tile also includes a background image, which fills the entire space created by the hidden image.

## Notes

- The hub is currently displaying regular `marketingCollections` from metaphysics instead of `marketingHubCollections`, because there is no data in the `marketingHubCollections`. Once we get that data populated, we will need to follow up here.
- I've verified the auto-scaling solution in Chrome, Firefox, and Safari. I have yet to verify in Edge, but will once we (eventually) get a deploy preview.


src/Components/v2/ImageLink.tsx
```diff
@@ -0,0 +1,117 @@
...
+`
+
+// These images give the container the correct height/width/aspect ratio at each breakpoint.
+//   They are hidden by CSS, but they allow the background image to scale correctly
+//   based on the available width.
+const PlaceholderImagesForContainerHeightAndWidth = () => {
+  return (
+    <div>
+      <Media at="xs">
+        <ImagePlaceholder
```
pepopowitz: I'm not super happy that these are 3 specific images with specific sizes. It makes this component likely unusable by anyone with different needs. I'm going to look into using SVG's to render these placeholders instead.

src/Apps/Collect/CollectApp.tsx
```diff
@@ -9,12 +10,14 @@ import React, { Component } from "react"
 import { Link, Meta, Title } from "react-head"
 import { createFragmentContainer, graphql } from "react-relay"
 import { data as sd } from "sharify"
+import { CollectionsHubsNavFragmentContainer as CollectionsHubsNav } from "../../Components/v2/CollectionsHubsNav"
```
damassi:
```
import { CollectionsHubsNavFragmentContainer as CollectionsHubsNav } from "Components/v2/CollectionsHubsNav"
```

src/Apps/Collect/CollectApp.tsx
```diff
@@ -9,12 +10,14 @@ import React, { Component } from "react"
...
 import { getMetadataForMedium } from "./CollectMediumMetadata"
 import { CollectFilterFragmentContainer as ArtworkGrid } from "./Components/Base/CollectFilterContainer"
 import { SeoProductsForArtworks } from "./Components/Seo/SeoProductsForArtworks"
 
 export interface CollectAppProps {
   viewer?: CollectApp_viewer
+  marketingCollections?: CollectApp_marketingCollections
```
damassi: is `viewer` and `marketingCollections` optional?

src/Components/v2/CollectionsHubsNav.tsx
```diff
@@ -0,0 +1,58 @@
+import { CSSGrid } from "@artsy/palette"
```
damassi: One day we need to figure out the `v2` folder story :/ Some new stuff goes in it, some in the parent. I'll add it to the list...

src/Components/v2/ImageLink.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { space, SpaceProps } from "styled-system"
+import { Media } from "Utils/Responsive"
+
+export const ImageLink: React.SFC<ImageLinkProps> = ({
+  href,
+  imageUrl,
+  children,
+  fontSize,
+  ...rest
+}) => (
```
damassi: Implicit returns tend to always get unwound in the future, as additions (variables, debugger statements) get added.

src/Components/__stories__/CollectionsHubsNav.story.tsx
```diff
@@ -0,0 +1,37 @@
...
+storiesOf("Components/CollectionsHubsNav", module).add("default", () => (
+  <Theme>
+    <CollectionsHubsNavQueryRenderer />
+  </Theme>
+))
+
+const CollectionsHubsNavQueryRenderer = () => {
+  return (
+    <SystemContextConsumer>
+      {({ relayEnvironment }) => {
```
damassi: `useSystemContext`

src/Components/v2/CollectionsHubsNav.tsx
pepopowitz: I intentionally moved these components there after the Slack convo from a couple days ago. Should I not have?

src/Components/v2/ImageLink.tsx
pepopowitz: 👍 I probably did this in a rush and didn't notice it on commit. I'll fix.

src/Components/v2/CollectionsHubsNav.tsx
damassi: oh no its fine! I was just commenting on how there's no enforced consistency ATM. At least we can be 100% sure (at a glance) that components in v2 are new.

src/Components/v2/ImageLink.tsx
pepopowitz: I made SVG's work!

src/Components/v2/AutoScalingImageLink.tsx
```diff
@@ -0,0 +1,126 @@
...
+  href,
+  imageUrl,
+  children,
+  fontSize,
+  maxWidth,
+  maxHeight,
+  ...rest
+}) => {
+  return (
+    <LinkWithBackground href={href} imageUrl={imageUrl} {...rest}>
```
damassi: Does this link go to a `/collection/id` page?

src/Components/v2/CollectionsHubsNav.tsx
```diff
@@ -0,0 +1,60 @@
...
+      gridTemplateColumns={[
+        "repeat(2, 1fr)",
+        "repeat(3, 1fr)",
+        "repeat(6, 1fr)",
+      ]}
+      gridGap={20}
+    >
+      {props.marketingCollections.map(hub => (
+        <AutoScalingImageLink
+          href={`/collection/${hub.slug}`}
```
damassi: Same here; this will need to be updated with a `<LInk>` component, so that it supports the router.

src/Components/v2/__stories__/AutoScalingImageLink.story.tsx
```diff
@@ -0,0 +1,68 @@
...
+const imageSize = {
+  maxWidth: 346,
+  maxHeight: 216,
+}
+
+storiesOf("Components/AutoScalingImageLink", module)
+  .add("with specific dimensions", () => (
+    <Box width={300}>
+      <AutoScalingImageLink
+        href="http://example.com"
```
damassi: ditto

src/Components/v2/AutoScalingImageLink.tsx
pepopowitz: Yes! Good catch on the router stuff. I'll fix.

src/Apps/Collect2/Routes/Collect/index.tsx
```diff
@@ -1,21 +1,25 @@
...
+import { Collect_marketingCollections } from "__generated__/Collect_marketingCollections.graphql"
 import { Collect_viewer } from "__generated__/Collect_viewer.graphql"
 import { CollectFilterFragmentContainer as ArtworkGrid } from "Apps/Collect2/Components/Base/CollectFilterContainer"
 import { SeoProductsForArtworks } from "Apps/Collect2/Components/Seo/SeoProductsForArtworks"
 import { AppContainer } from "Apps/Components/AppContainer"
+import { withSystemContext } from "Artsy"
 import { track } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
 import { FrameWithRecentlyViewed } from "Components/FrameWithRecentlyViewed"
+import { CollectionsHubsNavFragmentContainer as CollectionsHubsNav } from "Components/v2/CollectionsHubsNav"
```
damassi: Is it appropriate to place this component inside of `Apps/Collect2/Collection/Components`? Like, ATM, will this *only* exist on collections-related pages? If so, thoughts on keeping it close until it becomes more general? Similar to the other CollectionsHub rails that were relocated. 

Also, just had another thought re the links -- if this component _isn't_ only on collect-related pages, then there won't be a global routing context, and the `<Link>` component won't work. 

(Which just gave me an idea to test for `router`, and as a failsafe fall back to a regular `a` tag 🤔).

src/Components/v2/CollectionsHubsNav.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// largest sized image at each breakpoint:
+// xs: 346x216 ~= 1.6 aspect ratio
+// sm: 254x159 ~= 1.6 aspect ratio
+// md: 138x86 ~= 1.6 aspect ratio
+// lg/xl: 168x105 = 1.6 aspect ratio
+export const CollectionsHubsNav: FC<CollectionsHubsNavProps> = props => {
+  return (
+    <CSSGrid
```
damassi: This worked really nicely 👍

src/Apps/Collect2/Routes/Collect/index.tsx
xtina-starr: These entry points will also live on the homepage.

damassi: Ok cool, then we'll have to implement this conditional Link thing. 

@pepopowitz, if you wouldn't mind rolling with the link changes mentioned below (and haven't already added them to the artsy.net homepage) I'll follow up with a PR that makes the link component a bit more safe.

src/Components/v2/AutoScalingImageLink.tsx
```diff
@@ -0,0 +1,126 @@
...
+const LinkText = styled(Serif)`
+  position: absolute;
+  vertical-align: middle;
+  text-align: center;
+`
+
+// This placeholder gives the container the correct height/width/aspect ratio at each breakpoint.
+//   It is hidden by CSS, but it allows the background image to scale correctly
+//   based on the available width.
+const PlaceholderImagesForContainerHeightAndWidth: React.FC<
```
anandaroop: #minor This is a singular item here, yeah? So maybe just `PlaceholderImageForContainerHeightAndWidth`?

src/Apps/Collect2/Routes/Collect/index.tsx
pepopowitz: Yes, I think this would be a logical move. It started to get in my way in `v2/components` as I started working on the next story, which is to put something similar but different on the home page.

src/Components/v2/AutoScalingImageLink.tsx
pepopowitz: Yeah, that makes sense. Historically, this thing was rendering one of multiple images - I should have updated it when it became only one. I'll fix!

src/Apps/Collect2/Routes/Collect/index.tsx
pepopowitz: Oh, my browser had a stale conversation, from before @xtina-starr 's comment. As of right now, the entry-points on home look pretty different than these, but there were some changed designs that I haven't looked at very closely. For now I'll leave it where it is, and when we get to the home page we can decide if this is a common component or unique to the collect page.

artsyit:
:rocket: PR was released in v17.2.2 :rocket:

PR #2703, opened 2019-08-15
Update Bidder Registration + Credit Card Form

dblandin:
- Add CountrySelect input to form
- Update Auctions-related Storybook entries
- Add Metaphytics API error handling
- Update register redirect rules

---

## Before

screenshot 1566331086

## After

screenshot 1566331097
screenshot 1566331146
screenshot 1566331160
screenshot 1566331222


src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -4,8 +4,10 @@ import { RegistrationFormCreateBidderMutation } from "__generated__/Registration
 import { RegistrationFormCreateCreditCardMutation } from "__generated__/RegistrationFormCreateCreditCardMutation.graphql"
 import { CreditCardInput } from "Apps/Order/Components/CreditCardInput"
 import { ConditionsOfSaleCheckbox } from "Components/Auction/ConditionsOfSaleCheckbox"
-import { Form, Formik, FormikProps } from "formik"
-import React from "react"
+import { ErrorModal } from "Components/Modal/ErrorModal"
```
dblandin: We should update this `ErrorModal` to use the [artsy/palette](https://palette.artsy.net/elements/dialogs/modal/) modal under the hood.

src/Components/v2/CountrySelect.tsx
```diff
@@ -13,6 +13,7 @@ export const CountrySelect = (props: CountrySelectProps) => {
 }
 
 const COUNTRY_SELECT_OPTIONS = [
+  { text: "Select a country", value: "" },
```
dblandin: Corresponding PR since this will affect other pages, including those in the OrderApp: https://github.com/artsy/reaction/pull/2708

src/Apps/Auction/Components/__stories__/RegistrationForm.story.tsx
```diff
@@ -0,0 +1,27 @@
...
+storiesOf("Apps/Auction/Components", module).add("RegistrationForm", () => {
+  const stripe = Stripe(sd.STRIPE_PUBLISHABLE_KEY)
+
+  return (
+    <Section>
+      <StripeProvider stripe={stripe}>
+        <Elements>
+          <WrappedRegistrationForm
+            sale={SaleAuction}
+            onSubmit={(token, actions) => {
```
dblandin: This overrides the default `onSubmit behavior` of:

1. Tokenize the credit card via Stripe
1. Create a credit card via Metaphysics
1. Create a bidder via Metaphysics

Instead, it continues to tokenize the card so we can see validation errors and raises a window alert.

src/Components/v2/CountrySelect.tsx
erikdstock: I rather like the approach @ds300 made in [this comment](https://github.com/artsy/reaction/pull/2708#issuecomment-522952081)!

src/Apps/__stories__/Apps.story.tsx
```diff
@@ -1,13 +1,8 @@
...
 
-storiesOf("Apps", module).add("Auction Registration", () => {
-  return (
-    <MockRouter
-      routes={auctionRoutes}
-      initialRoute="/auction-registration2/weekly-mocktion"
-    />
-  )
+storiesOf("Apps", module).add("Search Results", () => {
+  return <MockRouter routes={searchRoutes} initialRoute="/search?term=andy" />
```
erikdstock: Confused about this diff. Auction stories here changed to Search, and AuctionApp.story.tsx now holds this?

src/Apps/__tests__/Fixtures/Sale.ts
```diff
@@ -9,3 +10,8 @@ export const SaleAuction: routes_AuctionQueryResponse["sale"] = {
   is_preview: false,
   registrationStatus: null,
 }
+
+export const SaleAuction: AuctionApp_sale = {
+  " $refType": undefined,
+  id: "yuki-contemporary-art-july-22th",
```
erikdstock: Something tells me this sort of thing is a convention that i've learned before and forgotten but can you remind me?

src/Apps/Auction/Components/RegistrationForm.tsx
dleve123: Shall we ticket this follow up? Implement it now? What are your thoughts on next steps here?

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -260,41 +280,70 @@ const RegistrationForm: React.FC<RegistrationFormProps> = props => {
...
+
+          const { setFieldError, setSubmitting } = actions
+          const { sale } = props
+
+          props.stripe.createToken(address).then(({ error, token }) => {
+            if (error) {
+              setFieldError("credit_card", error.message)
+              setSubmitting(false)
+            } else {
+              if (props.onSubmit) {
```
dleve123: can you elaborate on this? When wouldn't `props.onSubmit` be true?

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -260,41 +280,70 @@ const RegistrationForm: React.FC<RegistrationFormProps> = props => {
...
+          street1: "",
+          country: "",
+          city: "",
+          credit_card: null,
+          state: "",
+          postal_code: "",
+          telephone: "",
+          agree_to_terms: false,
+        }}
+        onSubmit={(values: FormValues, actions) => {
```
dleve123: have you considered extracted a named function here? Maybe this is idiomatic TS and I'm not used to it, but seems like this file might be a bit easier to read by passing in and naming a function

src/Apps/__tests__/Fixtures/Sale.ts
dleve123: Is there a reason that this is tied to a staging sale? Should this be a more direct test flag-value, like `an-example-auction-sale`?

dblandin: This is just a fixture so we can name it however. I probably plucked this from another fixture. Agree, that we can rename this to be more example-focused!

src/Apps/__stories__/Apps.story.tsx
dblandin: Might have been a bad rebase here. I'll fix this up. Thanks!

src/Apps/Auction/Components/RegistrationForm.tsx
dblandin: This is intended to override the default onSubmit behavior for use in storybooks or tests. @erikdstock and I have been thinking of some component reorganization that would make this required on the form, moving some of the onSubmit behavior up a level.

src/Apps/Auction/Components/RegistrationForm.tsx
dblandin: Yep :+1:  As hinted to in https://github.com/artsy/reaction/pull/2703#discussion_r315741182, I think we're missed a component level here that should clean this up.

src/Apps/Auction/Components/RegistrationForm.tsx
dblandin: I'm discussing this with the Design Systems group. I think we can circle back to this.

artsyit:
:rocket: PR was released in v17.4.0 :rocket:

PR #2704, opened 2019-08-15
Improving styling on artist series on collection hub

jiahaodavid:
Screen Shot 2019-08-19 at 09 40 22

- Based on the QA feedback, make some change on the Artist Series on the collection hub. ref to [QA for 1334](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1354).

- For the smaller viewpoint, when there are only 4 items in the carousel, it will show the arrow button. See below GIF. 

small_viewpoint


src/Components/CollectionsHubRails/ArtistSeriesRail/ArtistSeriesEntity.tsx
```diff
@@ -23,11 +23,11 @@ export const ArtistSeriesEntity: React.FC<ArtistSeriesEntityProps> = ({
   } = member
   const bgImages = hits.map(hit => hit.image.url)
   const imageSize =
-    bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
+    bgImages.length === 1 ? 221 : bgImages.length === 2 ? 109 : 72
   return (
-    <Container p={0.5} m={0.5}>
+    <Container px={2} pt={2} pb={1.6} m={1}>
```
damassi: the only values that we should set inside of these spacing props are values [from our theme](https://github.com/artsy/palette/blob/master/packages/palette/src/Theme.tsx#L116-L138)

src/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx
```diff
@@ -14,40 +14,65 @@ export const ArtistSeriesRail: React.FC<ArtistSeriesRailProps> = ({
   collectionGroup,
 }) => {
   const { members } = collectionGroup
+  const broswerWidth = GetBroswerWidth()
```
damassi: Since `GetBrowserWidth` isn't a component or a class, it should be camelCase: `getBrowserWidth`

jiahaodavid:
Hold. I just found out. The whole collections hub got moved. I need to make some change.

src/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx
```diff
@@ -14,33 +15,49 @@ export const ArtistSeriesRail: React.FC<ArtistSeriesRailProps> = ({
...
       <Serif size="5" mb={1}>
         Trending Artist Series
       </Serif>
       <Carousel
-        height="200px"
-        width="500px"
+        height="250px"
         options={{
-          groupCells: sd.IS_MOBILE ? 1 : 4,
+          groupCells: sd.IS_MOBILE ? 1 : broswerWidth > 1024 ? 4 : 3,
```
damassi: Nice 👍

jiahaodavid: @damassi How about this ?
groupCells: (() => {
            if (sd.IS_MOBILE) {
              return 1
            } else {
              if (broswerWidth > 1024) {
                return 4
              } else {
                return 3
              }
            }
          })(),

damassi: Rather than writing a self-invoking function, it would be better to keep things minimal and assign to a variable just above, within the render method and before the `return`.

jiahaodavid: @damassi How about this time?
let groupCellsValue: number
  if(sd.IS_MOBILE){
    groupCellsValue = 1
  }else{
    if(broswerWidth > 1024 ){
      groupCellsValue =4
    }
    else
    groupCellsValue =3
  }
......
groupCells : groupCellsValue,

damassi: @jiahaodavid - Would you mind updating these comments to support proper github code snippet formatting?

damassi: `groupCellsValue` -> `groupCells`,  the same as the setting. That way things are 1:1, and we don't have to worry about variables diverging through time, and if we ever needed to update something globally, all relevant references would be returned. (Also, in general, no need to append "value" to a variable as by it's nature it's already a value.)

jiahaodavid: Yes. ```let groupCellsValue: number
if(sd.IS_MOBILE){
groupCellsValue = 1
}else{
if(broswerWidth > 1024 ){
groupCellsValue =4
}
else
groupCellsValue =3
}```

jiahaodavid: Yes. I made some change. How about this? If ok, I will apply to other condition checks like this one.
```
let groupCells: number
  if(sd.IS_MOBILE){
  groupCells = 1
  }
  else{
      if(broswerWidth > 1024 ){
      groupCells =4
      }
      else
      groupCells =3
  }

Comment (damassi) [reply src/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx]: Check out:  https://help.github.com/en/articles/creating-and-highlighting-code-blocks for better highlights.

But yeah, logic looks good!

Comment (jiahaodavid) [reply src/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx]: @damassi Hi, I make some change. You can see below. I use the better highlight this time. Last time I manually added the space to the code. 😭  
```typescript
let groupCells: number
  if (sd.IS_MOBILE) {
    groupCells = 1
  } else {
    if (broswerWidth > 1024) {
      groupCells = 4
    } else groupCells = 3
  }
```
..... 
```typescript
options={{
          groupCells,
          wrapAround: sd.IS_MOBILE ? true : false,
          cellAlign: "left",
          pageDots: false,
          draggable: sd.IS_MOBILE ? true : false,
          contain: true,
        }}
```
Also, Christina gave me some feedback. I made some change on other condition check. Thx @xtina-starr 
```typescript
const isSmallerViewpoint = broswerWidth < 1024
```
...
```typescript 
renderRightArrow={({ Arrow }) => {
          return (
            <ArrowContainer>
              {members.length > 4 ? (
                <Arrow />
              ) : (
                <Arrow showArrow={members.length === 4 && isSmallerViewpoint} />
              )}
            </ArrowContainer>
          )
        }
```

jiahaodavid: It seems I have to make the condition check before rendering the **Arrow** components. Otherwise, I will fail the test. Because the test only checks the length of the <ArrowButton> inside the <Arrow>. So, once the <Arrow> got rendered, the arrows maybe not show in the browser but the test still can detect 2 **Arrowbutton**. So, I am going to make some change like this. You can see below. And what do you think? @damassi @xtina-starr 
```typescript
renderRightArrow={({ Arrow }) => {
          const smallerViewpointAndFourItems =
            isSmallerViewpoint && members.length === 4
          return (
            <ArrowContainer>
              {members.length > 4 ? (
                <Arrow />
              ) : (
                smallerViewpointAndFourItems && <Arrow showArrow={true} />
              )}
            </ArrowContainer>
          )
        }
```

PR #2705, opened 2019-08-16
[Router] Add in  interstitial loading state

damassi:
Adds a new interstitial loading state for full-page routes: 

interstitial

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,6 +207,7 @@
     "react-sizeme": "^2.4.4",
     "react-slick": "^0.23.2",
     "react-spring": "^8.0.23",
+    "react-static-container": "^1.0.2",
     "react-stripe-elements": "^2.0.1",
     "react-tracking": "^7.0.1",
     "react-transition-group": "^2.3.0",
--- a/src/Apps/Collect2/Routes/Collect/CollectionApp.tsx
+++ b/src/Apps/Collect2/Routes/Collect/CollectionApp.tsx
@@ -35,7 +35,7 @@ export class CollectionApp extends Component<CollectionAppProps> {
     }
   }
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
     this.collectionNotFound(this.props.collection)
   }
 
--- a/src/Apps/Search/routes.tsx
+++ b/src/Apps/Search/routes.tsx
@@ -1,4 +1,3 @@
-import { renderSpinner } from "Artsy/Router/Utils/Route"
 import { RouteConfig } from "found"
 import React from "react"
 import { graphql } from "react-relay"
@@ -7,6 +6,7 @@ import { SearchResultsArtistsRouteFragmentContainer as SearchResultsArtistsRoute
 import { SearchResultsArtworksRouteFragmentContainer as SearchResultsArtworksRoute } from "Apps/Search/Routes/Artworks/SearchResultsArtworks"
 import { SearchResultsEntityRouteFragmentContainer as SearchResultsEntityRoute } from "Apps/Search/Routes/Entity/SearchResultsEntity"
 
+import { RouteSpinner } from "Artsy/Relay/renderWithLoadProgress"
 import { SearchAppFragmentContainer as SearchApp } from "./SearchApp"
 
 const prepareVariables = (_params, { location }) => {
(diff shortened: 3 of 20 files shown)


package.json
```diff
@@ -207,6 +207,7 @@
     "react-sizeme": "^2.4.4",
     "react-slick": "^0.23.2",
     "react-spring": "^8.0.23",
+    "react-static-container": "^1.0.2",
```
damassi: Got this tip from Taion

src/Apps/Collect2/Routes/Collect/CollectionApp.tsx
```diff
@@ -35,7 +35,7 @@ export class CollectionApp extends Component<CollectionAppProps> {
     }
   }
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
```
damassi: We're starting to get warnings in the console and so I ran the react codemod over the codebase.

package.json
zephraph: Taion?

damassi: Found router author

src/Artsy/Router/Utils/RenderStatus.tsx
```diff
@@ -0,0 +1,43 @@
...
+import ElementsRenderer from "found/lib/ElementsRenderer"
+
+export const RenderPending: React.FC = props => {
+  /**
+   * TODO: Add timeout here for when a request takes too long. Show generic error
+   * and notify Sentry.
+   */
+  return (
+    <Box>
+      <StaticContainer>{null}</StaticContainer>
```
zephraph: So this is just used to stop reconciliation? It seems like the relationship between `Box` and `StaticContainer` should always be the same between `RenderPending` and `RenderReady`. Is that the case?

src/Apps/Search/routes.tsx
```diff
@@ -29,8 +29,13 @@ const entityTabs = Object.entries(tabsToEntitiesMap).map(([key, entities]) => {
   return {
     path: key,
     Component: SearchResultsEntityRoute,
+
+    // FIXME: We shouldn't overwrite our route functionality, as that breaks
+    // global route configuration behavior.
```
zephraph: I'm trying to follow, but a little more context here would be good. So, providing a `render` method here technically overwrites the route functionality and can break things?

src/Artsy/Router/Utils/Route.tsx
```diff
@@ -1,79 +1,50 @@
...
       throw error
     }
 
-    if (prerender) {
-      prerender(renderArgs)
-    }
-
-    if (render) {
-      return render(renderArgs)
-    }
```
zephraph: Were these just never used?

zephraph:
Peril commented about a new dependency, but `react-static-container` is already being included as a dependency of found router. I hid the comment just to remove the clutter. 

The `yarn.lock` file wasn't changed, but it doesn't need to be because the correct version is already present.

src/Artsy/Router/Utils/Route.tsx
damassi: Yah, this was copy / paste from taion and were, I think, functions on his route defintion

src/Artsy/Router/Utils/RenderStatus.tsx
damassi: Can you elaborate?

src/Artsy/Router/Utils/Route.tsx
damassi: But we are using `render`, so let me revert this one line.

src/Artsy/Router/Utils/RenderStatus.tsx
zephraph: If someone came along and added an element between `<Box>` and `<StaticContainer>` in one of these (but not in the other) would it break?

src/Apps/Search/routes.tsx
damassi: Definiting a local `render` call on a route definition overwrites the handlers inside of Route.tsx, via a callback. We overwrite all over; the only reason I added this here was because soon I was going to merge this new app into the top level router and would like to return to this behavior.

src/Artsy/Router/Utils/RenderStatus.tsx
damassi: @zephraph - yup, impressive eye! i can move this into a shared component

artsyit:
:rocket: PR was released in v17.1.0 :rocket:

PR #2706, opened 2019-08-16
[Router] Fix zIndex #trivial

damassi:
Fixes zIndex of loader overlay so that it doesn't cover top nav. 

Also fixes issue on /collection/id where window would scroll to the top on pagination change and confuse focus, vs scrolling to the top of the grid where content updates. 

#trivial


artsyit:
:rocket: PR was released in v17.1.1 :rocket:

PR #2708, opened 2019-08-16
Add "Select a country" no value placeholder to CountrySelect

dblandin:
Instead of showing the first country in the list and no value selected internally, show a no value placeholder prompting a selection.

screenshot 1565979365

### DIFF
--- a/src/Components/v2/CountrySelect.tsx
+++ b/src/Components/v2/CountrySelect.tsx
@@ -13,6 +13,7 @@ export const CountrySelect = (props: CountrySelectProps) => {
 }
 
 const COUNTRY_SELECT_OPTIONS = [
+  { text: "Select a country", value: "" },
   { text: "Afghanistan", value: "AF" },
   { text: "Åland Islands", value: "AX" },
   { text: "Albania", value: "AL" },


dleve123:
Is it possible that we're leverage the _lack of_ a null state here to ensure that there's always a county included in a GraphQL mutation? Just want to make sure that the current system behavior isn't acting as an indirect validation. Can you confirm?

ds300:
Good question @dleve123, looks like we explicitly check for the empty string so this option wouldn't validate.

However, @dblandin I think rather than do it like this it would be a great idea to add a `placeholder` prop to the `LargeSelect` component in palette like so

image

And that way we get a nice greyed out option

image

erikdstock:
I'm working on a PR to artsy/palette following @ds300 's suggestion.

ashkan18:
thanks for doing this 💜 ! this looks good to me and I agree with @ds300 about a grayed out choice would be better. This already looks better than whats there so feel free to merge this or if you rather to wait till palette update.

dblandin:
Closing in favor of https://github.com/artsy/palette/pull/570

PR #2709, opened 2019-08-16
[Collections] Relocate CollectionsHub rails

damassi:
This relocates the CollectionsHub rails to be co-located with the Collections route. 

Additionally, updates internal links inside the rails to user our router's `<Link>`, thereby gaining the benefits of the router and avoiding the hard jump between pages.


src/Artsy/Router/RouterLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Wrapper component around found's <Link> component that can be easily styled
+ * via styled-components; e.g.,
+ *
+ * const StyledLink = styled(RouterLink)`
+ *   ...
+ * `
+ */
+export const RouterLink: React.FC<LinkProps> = ({
```
l2succes: So should we use this over `import { Link } from "@artsy/palette"` from now on?

damassi: If we need to navigate within the router framework to other `/collections` / `collection/id`, `/collect` pages, yup. Can also use `import { Link } from 'found'` if not needing to style fancy.

artsyit:
:rocket: PR was released in v17.1.4 :rocket:

PR #2712, opened 2019-08-16
add item_number to analytics tracking

xtina-starr:
This adds `item_number` to the analytics data for FeaturedCollections and OtherCollections.


src/Artsy/Analytics/Schema/index.ts
```diff
@@ -23,6 +23,7 @@ interface Uncategorized {
   item_type: any
   item_id: any
   query: any
+  item_number: any
```
l2succes: It’s safe to type this as `number`

PR #2719, opened 2019-08-19
[Router] Add ErrorPage handler

damassi:
Adds a redirect to an error page if an error is encountered from MP. Currently we've just been throwing, which we've needed to fix for a while.

Right now, because we don't have consistent error's that we can parse from MP, I'm passing a generic 404, which I think is less alarming than a 500. 

Note that this routing stuff is still something of a moving target and will likely change in minor ways over the next few weeks. Once it's stable will update test coverage where missing. 

error

### DIFF
--- a/src/Apps/Collect2/Routes/Collections/index.tsx
+++ b/src/Apps/Collect2/Routes/Collections/index.tsx
@@ -51,6 +51,7 @@ export class CollectionsApp extends Component<CollectionsAppProps> {
               <Box pb={0.3}>
                 <Sans size="3" weight="medium">
                   <Link to="/collect">View works</Link>
+                  <Link to="/collect/foo-error">View error</Link>
                 </Sans>
               </Box>
             </Flex>
--- a/src/Artsy/Router/Utils/RenderStatus.tsx
+++ b/src/Artsy/Router/Utils/RenderStatus.tsx
@@ -2,7 +2,13 @@ import React from "react"
 import StaticContainer from "react-static-container"
 
 import { Box } from "@artsy/palette"
+import { ErrorPage } from "Components/ErrorPage"
 import ElementsRenderer from "found/lib/ElementsRenderer"
+import { data as sd } from "sharify"
+import styled from "styled-components"
+import createLogger from "Utils/logger"
+
+const logger = createLogger("Artsy/Router/Utils/RenderStatus")
 
 export const RenderPending: React.FC = props => {
   /**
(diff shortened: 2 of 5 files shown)


zephraph:
Cool! This is pretty straightforward and I like how it gives us a little more granular control over the error code. I'm not certain about using a 404 vs a 500 here (@joey, thoughts?) but that seems relatively minor. 

Looks good!

zephraph:
Would be good to think about tests for this at some point too.

damassi:
> Would be good to think about tests for this at some point too.

Yes, this was mentioned in the PR description.

joeyAghion:
You mention "redirecting" to an error page but I don't see that in the code. I assume it's just rendered on the client and the URL remains the same? (I think that's better than redirecting somewhere else.)

I do *not* think we should default to rendering a 404 error. Our router can't punt on such a fundamental question. It's already pretty broken today how something as simple as an unpublished artwork will result in a 500 error, but it's a server-side error so we shouldn't pretend it's something else. I also worry about a transient 500 error causing Google to mark ranking pages as unavailable.

(I recall there being tickets about improving handling and communication of non-fatal errors in Metaphysics, but I can't find them. Our own blog post on the topic ranks pretty highly though, so I have faith that a solution is possible.)

damassi:
@mzikherman - Mind chiming in?

mzikherman:
Nice.

Re: the general comments here (propagating 'correct' error codes from downstream services, and be more resilient generally), these are all still open issues AFAIK.

We discussed this more in Slack [here](https://artsy.slack.com/archives/CA8SANW3W/p1566249724392600).

Since currently we don't have great error page/handling (I think any SSR errors encountered will serve our 500 error page, as currently), and errors occurring during client-side browsing...(result in a blank screen?).

artsyit:
:rocket: PR was released in v17.2.0 :rocket:

src/Apps/Collect2/Routes/Collections/index.tsx
```diff
@@ -51,6 +51,7 @@ export class CollectionsApp extends Component<CollectionsAppProps> {
               <Box pb={0.3}>
                 <Sans size="3" weight="medium">
                   <Link to="/collect">View works</Link>
+                  <Link to="/collect/foo-error">View error</Link>
```
alloy: Is this meant to stay in?

src/Artsy/Router/Utils/RenderStatus.tsx
```diff
@@ -39,6 +33,32 @@ export const RenderReady: React.FC<{
   )
 }
 
+export const RenderError: React.FC<{
+  error: { status?: number; data?: any }
+}> = props => {
+  logger.error(props.error.data)
```
alloy: Remind me, does this go to Sentry or paper-trail?

src/Apps/Collect2/Routes/Collections/index.tsx
damassi: This was removed subsequently

src/Artsy/Router/Utils/RenderStatus.tsx
damassi: Sentry. Papertrail mirror force's server logs

Issue #2675, opened 2019-08-08
Pretty text at the top got broke


orta:
Maybe prettier somehow?

![Screen Shot 2019-08-08 at 4 50 58 PM](https://user-images.githubusercontent.com/49038/62736984-c173ce00-b9fc-11e9-8fba-591eb38adea1.png)

dblandin:
@orta Thanks for the nudge on this!

Looks like a @fossabot script must have removed the leading indentation at the top of the file, causing the code block to break: https://github.com/artsy/reaction/pull/2449

All fixed now!

PR #2720, opened 2019-08-19
Bring back the cool ascii banner text in the README

dblandin:
This commit fixes the very cool ascii banner text in the README.

Wrapping the text in explicit code blocks should prevent it from breaking again in the future. It's likely that a script ran some sort of string strip command on the contents of the README, removing all leading and trailing whitespace, causing the implicit code block, marked by indentation, to break.

h/t to @orta for reporting :+1:

Addresses #2675

---

**Before**

screenshot 1566258894

**After**

screenshot 1566259079


damassi:
Thank you! this has been bothering me

artsyit:
:rocket: PR was released in v17.2.1 :rocket:

PR #2721, opened 2019-08-20
Handle <RouterLink> behavior better

damassi:
This PR updates our `<RouterLink>` behavior so that we can fallback to a regular `<a>` tag if outside of a router context. This allows us to use components anywhere on the site, with opt-in performance gains when the link being navigated to utilizes our routing framework.


src/Artsy/Router/RouterLink.tsx
```diff
@@ -9,15 +10,34 @@ import React from "react"
...
+    )
+  }
+}
+
+/**
+ * FIXME: Upgrade to Found v.4: https://github.com/4Catalyzer/found/releases/tag/v0.4.0,
+ * which supports React 14.3 context style.
+ */
+RouterLink.contextTypes = {
+  router: PropTypes.any,
```
damassi: Had to defer to old context behavior :/ Upgrading found will take some concerted attention.

src/Artsy/Router/RouterLink.tsx
```diff
@@ -1,23 +1,43 @@
...
+
+  if (isRouterContext) {
+    return (
+      <Link to={to} {...props}>
+        {children}
+      </Link>
+    )
+  } else {
+    return (
+      <a href={to as string} {...props}>
```
pepopowitz: It surprised me that `to` wasn't already a string. It looks like it's [of type `LocationDescriptor`](https://github.com/4Catalyzer/found/blob/master/types/index.d.ts#L345). I don't know that we've ever used it as anything but a string, so we should be fine blindly casting it as a string here....right?

src/Artsy/Router/__tests__/RouterLink.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+              return <RouterLink to="/foo">Foo</RouterLink>
+            },
+          },
+        ]}
+      />
+    ).renderUntil(enzyme => {
+      try {
+        return enzyme.find(Link).length > 0
+      } catch {
+        // Guard against p == null, which is the first render pass
```
pepopowitz: What does the `p` refer to in this comment? `props`?

src/Artsy/Router/RouterLink.tsx
```diff
@@ -1,23 +1,44 @@
...
+
+  if (isRouterContext) {
+    return (
+      <Link to={to} className={className} {...props}>
+        {children}
+      </Link>
+    )
+  } else {
+    return (
+      <a href={to as string} className={className} {...props}>
```
anandaroop: If `className` is just a pass-through in both branches here, does it need to be destructured out?

pepopowitz:
Heads up that I merged #2698 into master - it will require another removal of `<MockRouterProvider>` after you rebase.

src/Artsy/Router/RouterLink.tsx
damassi: That was my thinking here. I think if we ever need anything more advanced -- like passing around `LocationDescriptors` -- we'll be inside of a router context but we'll have to cross that when we get there.

src/Artsy/Router/__tests__/RouterLink.test.tsx
damassi: oo i forgot to update this var; will update

artsyit:
:rocket: PR was released in v17.2.3 :rocket:

PR #2723, opened 2019-08-20
[CollectionsHubsNav] Restyle collection hub nav according to new designs

pepopowitz:
This PR updates the styling of the collection hubs navigation to match updated designs.

_redesign


src/Components/v2/CollectionsHubsNavItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export const CollectionsHubsNavItem: React.FC<CollectionsHubsNavItemProps> = ({
+  href,
+  imageUrl,
+  children,
+}) => {
+  return (
+    <OuterLink to={href}>
+      <ImageContainer>
```
pepopowitz: This is a lot of layers, but they're needed. 

- The `ImageOverlay` renders a gray overlay on hover of the `OuterLink`. 
- The `ImageContainer` constrains the overlay to be the same size as the image.
- If we use the `OuterLink` to constrain the overlay and ditch the `ImageContainer`, we end up with the overlay covering both the image _and_ the title below it.

xtina-starr: This makes sense and what I would expect. Thanks for clarifying!

anandaroop: 👍

src/Components/v2/CollectionsHubsNavItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+    display: none;
+    color: #fff;
+    content: "";
+    background: rgba(0, 0, 0, 0.3);
+  }
+`
+
+const OuterLink = styled(RouterLink)`
+  text-decoration: none;
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```
anandaroop: 👀 ooh, never used this before. Specifically for mobile web?

src/Components/v2/CollectionsHubsNavItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const CollectionsHubsNavItem: React.FC<CollectionsHubsNavItemProps> = ({
+  href,
+  imageUrl,
+  children,
+}) => {
+  return (
+    <OuterLink to={href}>
+      <ImageContainer>
+        <ImageOverlay>
+          <HubImage src={imageUrl} width="100%" alt={"todo fill in"} />
```
anandaroop: Could the `alt` tag here just recapitulate the `hub.title`?

anandaroop:
If there's nothing to change re the alt tag handling for now, feel free to merge @pepopowitz

src/Components/v2/CollectionsHubsNavItem.tsx
pepopowitz: Oof, copy-pasta from an example I was looking at. I'll remove it.

src/Components/v2/CollectionsHubsNavItem.tsx
pepopowitz: Yes! Thanks for catching. I'll fix this. In retrospect I'm glad I put in a bad alt so that it would get caught, instead of just leaving it blank.

anandaroop: Cool cool, wasn't sure if you were deferring intentionally, since there's probably some complexity here (if we need to account for subtitles, or the title text having inline markup, etc.)

pepopowitz: I actually think I'm going to fix this as part of the next story (hubs nav on home page), because that one will force me to think about how to pass all the right things into each item.

artsyit:
:rocket: PR was released in v17.4.1 :rocket:

PR #2727, opened 2019-08-21
Add analytic event tracking to auction registration route

dblandin:
Jira: https://artsyproduct.atlassian.net/browse/AUCT-584

Event Registry: https://docs.google.com/spreadsheets/d/17oqFli15x1SmNBu_-fPZhFifMw2BNSPJe9VdOgs1aKs/edit#gid=0

## Local QA Gotcha

You'll need to `yarn add react-fast-compare@2.0.4` in Force for [reasons](https://artsy.slack.com/archives/C2TQ4PT8R/p1567089484035300)


dleve123:
We think that this PR fires the same number and quality of events that currently exists on production

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -165,7 +167,13 @@ const InnerForm: React.FC<FormikProps<FormValues>> = props => {
         )}
       </Flex>
 
-      <Button mt={1} size="large" width="100%" loading={isSubmitting}>
+      <Button
+        mt={1}
+        size="large"
+        width="100%"
+        loading={isSubmitting}
+        {...{ type: "submit" } as any}
```
erikdstock: We need `type="submit"` according to a console warning but it is not available on a palette button 🙃

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -186,17 +194,54 @@ const validationSchema = Yup.object().shape({
   ),
 })
 
+type TrackErrors = (errors: string[]) => void
```
erikdstock: We pass our form submission error handler in from above where tracking is set up. This is its proptype.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -186,17 +195,41 @@ const validationSchema = Yup.object().shape({
...
+    }
+  }
+  React.useEffect(effect, [formikProps.isSubmitting])
+
+  return null
+}
+
 export interface RegistrationFormProps
   extends ReactStripeElements.InjectedStripeProps {
+  tracking?: TrackingProp
```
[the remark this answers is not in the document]
erikdstock: 🔪

src/Apps/Auction/Components/__stories__/RegistrationForm.story.tsx
```diff
@@ -1,17 +1,23 @@
...
 import { storiesOf } from "storybook/storiesOf"
 import { Section } from "Utils/Section"
 import { StripeWrappedRegistrationForm } from "../RegistrationForm"
 
+const TrackedLayout = track()(Section)
+
 storiesOf("Apps/Auction/Components", module).add("RegistrationForm", () => {
   return (
-    <Section>
+    <TrackedLayout>
```
erikdstock: May not be needed any more- revisiting.

src/Apps/Auction/Routes/Register/index.tsx
```diff
@@ -13,27 +7,69 @@ import {
...
+      ...commonProperties,
+    })
+  }
+
   function createBidder() {
     return new Promise(async (resolve, reject) => {
       commitMutation<RegisterCreateBidderMutation>(relay.environment, {
-        onCompleted: (data, errors) => {
-          resolve()
+        onCompleted: data => {
```
erikdstock: Do we need to be handling the `(data, errors)` case rather than resolving unconditionally? Where does this second errors parameter come from rather than calling the `onError` parameter below?

src/Apps/Auction/Routes/Register/index.tsx
```diff
@@ -109,16 +148,28 @@ export const RegisterRoute: React.FC<RegisterProps> = props => {
...
       .catch(error => {
         logger.error(error)
+        let errorMessages: string[]
+        if (Array.isArray(error)) {
+          errorMessages = error.map(e => e.message)
+        } else if (typeof error === "string") {
+          errorMessages = [error]
+        } else if (error.message) {
+          errorMessages = [error.message]
+        }
```
erikdstock: In all cases we've tested the final `error.message` case applied here (though at least a `PayloadError[]` is a possible outcome), so this is just being pessimistic.

src/Apps/Auction/Routes/__tests__/RegisterRoute.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      }
+      me {
+        ...Register_me
+      }
+    }
+  `,
+})
+
+xdescribe("Auction Registration App", () => {
+  // const getWrapper = (props: any) => {
```
erikdstock: We haven't gotten _any_ of these tests working yet, so maybe we should not keep this file around.

src/Apps/Auction/Routes/__tests__/RegisterRouteAnalytics.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+    pending()
+  })
+
+  it("tracks form submission success", async () => {
+    pending()
+    // const wrapper = await renderRelayTree({
+    //   Component: (props: any) => (
+    //     <div>
+    //       <RegisterRouteFragmentContainer {...props} />
+    //     </div>
```
erikdstock: It's currently not easy to submit the form and mock out the entire success due to dependence on relay.

src/Apps/__tests__/Fixtures/Auction/Routes/Register/index.ts
```diff
@@ -1,23 +1,44 @@
...
+import { routes_RegisterQueryResponse } from "__generated__/routes_RegisterQuery.graphql"
 
-export interface RegisterQueryResponse {
-  sale: redirects_sale
-  me: redirects_me
+type Omit<T, K> = Pick<T, Exclude<keyof T, K>>
+type DeFraged<T> = Omit<T, " $refType" | " $fragmentRefs">
+
+export interface DeFragedRegisterQueryResponse {
+  sale: DeFraged<routes_RegisterQueryResponse["sale"] & Register_sale>
```
erikdstock: Our effort to remove confusing relay-specific attributes from these objects. If it's overkill, we will remove.

src/Apps/Auction/Components/__stories__/RegistrationForm.story.tsx
erikdstock: 🔪

src/Apps/Auction/Routes/__tests__/RegisterRouteAnalytics.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+  it("tracks form submission failure when create bidder mutation fails", () => {
+    pending()
+  })
+
+  it("tracks form submission failure when create credit card Stripe API call fails", () => {
+    pending()
+  })
+
+  it("tracks form submission success", async () => {
+    pending()
```
erikdstock: It's currently not easy to submit the form and mock out the entire success or strange cases with the stripe api due to dependence on relay.

src/Apps/Auction/Components/RegistrationForm.tsx
zephraph: PRs welcome 😉

dleve123:
I had myself as a reviewer to get Peril to stop it, assigned @zephraph as Erik and I were pairing on this.

src/Apps/Auction/Routes/Register/index.tsx
zephraph: `errors` here is different than `onError`. The `errors` parameter here are any potential errors that are returned from metaphysics. 

`onError` is an error that occurred with the request. I.E. the call to metaphysics likely failed.

src/Apps/Auction/Components/RegistrationForm.tsx
erikdstock: My understanding has been that we don't put html-specific attrs on palette components- what is the right way to do this?

zephraph: That's not a strict constraint. We just don't want to add HTML props that flow down to react native.

zephraph: Let's not worry with this right now. It's fine for this.

src/Apps/Auction/Routes/__tests__/RegisterRouteAnalytics.test.tsx
zephraph: What are these `pending()` function calls? I don't see a reference to that anywhere.

src/Apps/Auction/__tests__/redirects.test.ts
```diff
@@ -25,23 +25,23 @@ describe("Auction/redirects", () => {
...
-  ): RegisterQueryResponse => ({
+    data: DeFragedRegisterQueryResponse
+  ): DeFragedRegisterQueryResponse => ({
     sale: data.sale,
     me: data.me,
   })
 
   it("does not redirect if a sale is found", async () => {
-    const { redirect } = await render(
+    const result = await render(
```
zephraph: Why the change?

src/Apps/Auction/__tests__/redirects.test.ts
```diff
@@ -25,23 +25,23 @@ describe("Auction/redirects", () => {
...
       mockResolver(Fixture)
     )
-
+    const { redirect } = result
     expect(redirect).toBeUndefined
   })
 
   it("redirects to the auction registration modal if the user has a qualified credit card", async () => {
-    const { redirect } = await render(
+    const result = await render(
```
zephraph: This one too. It's not evident why the destructuring was removed. Was it for debugging purposes?

src/Apps/Auction/__tests__/redirects.test.ts
erikdstock: Artefact of debugging when we broke the relay found router.

src/Apps/Auction/__tests__/redirects.test.ts
erikdstock: correct.

src/Apps/__tests__/Fixtures/Auction/Routes/Register/index.ts
```diff
@@ -1,23 +1,44 @@
-import { redirects_me } from "__generated__/redirects_me.graphql"
-import { redirects_sale } from "__generated__/redirects_sale.graphql"
+import { Register_me } from "__generated__/Register_me.graphql"
+import { Register_sale } from "__generated__/Register_sale.graphql"
+import { routes_RegisterQueryResponse } from "__generated__/routes_RegisterQuery.graphql"
 
-export interface RegisterQueryResponse {
-  sale: redirects_sale
-  me: redirects_me
+type Omit<T, K> = Pick<T, Exclude<keyof T, K>>
```
zephraph: Omit is a first class member of TS now, you shouldn't need to redefine it. 

https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk

erikdstock: I'm pretty sure we do need it unless we bump the ts version; it was working fine because we hadn't set pinned it to the editor version in `code`, but then it failed in type-checking

src/Apps/Auction/__tests__/redirects.test.ts
erikdstock: (I can fix though)

src/Apps/__tests__/Fixtures/Auction/Routes/Register/index.ts
zephraph: All right, we can leave it for now. I'll look into bumping the version.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -186,17 +192,53 @@ const validationSchema = Yup.object().shape({
...
+    if (
+      formikProps.submitCount > 0 &&
+      !formikProps.isSubmitting &&
+      !formikProps.isValid
+    ) {
+      cb(Object.values(formikProps.errors))
+      formikProps.setSubmitting(false)
+    }
+  }
+  React.useEffect(effect, [formikProps.submitCount, formikProps.isSubmitting])
```
zephraph:
```suggestion
  React.useEffect(effect, [formikProps.submitCount, formikProps.isSubmitting, formikProps.isValid, formikProps.errors])
```

You need to track all props that are used within the effect.

src/Apps/Auction/Routes/__tests__/RegisterRouteAnalytics.test.tsx
zephraph: If these tests just aren't implemented yet, I recommend using `it.todo(...` instead.

src/Apps/Auction/Components/RegistrationForm.tsx
erikdstock: I think this would actually introduce a bug- i know the general way of thinking about it is to track all parts as dependencies, but the purpose of this effect is to run only after failed form submission. Validation runs every time, so by letting `isValid` or `errors` be dependencies we this will change the behavior of the effect. This approach was taken from the formik issue linked in the doc comment and pr comment above & I actually commented on that issue to clarify that my understanding was right (still waiting to hear back).

artsyit:
:rocket: PR was released in v17.9.0 :rocket:

PR #2728, opened 2019-08-22
[ArtworkFilter] Ensure page is reset to 1 on filter change #trivial

damassi:
Noticed a prod bug that we needed to fix in the new filter


artsyit:
:rocket: PR was released in v17.5.1 :rocket:

PR #2729, opened 2019-08-22
Add Reponsiveness to Primary and Subseries SVG Videos 

ashleyjelks:
Handles responsiveness and resizing issues on the Video in SVG Blobs.
[Links to GROW-1469](https://artsyproduct.atlassian.net/browse/GROW-1469)

Kapture 2019-08-23 at 14 31 43
Kapture 2019-08-23 at 14 26 40
Kapture 2019-08-23 at 14 28 28
Kapture 2019-08-23 at 14 30 09

### DIFF
--- a/src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/VanguardVideoBackground.tsx
+++ b/src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/VanguardVideoBackground.tsx
@@ -1,11 +1,6 @@
-import { Box, color } from "@artsy/palette"
+import { Box, color, Flex } from "@artsy/palette"
 import React from "react"
 import styled from "styled-components"
-
-// interface SVGVideoBackgroundProps {
-//   id?: string
-//   url?: string
-// }
 
 export const VanguardVideoPrimarySeries = props => (
   <VideoSVGWrapper>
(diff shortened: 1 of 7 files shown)


src/Components/Publishing/EditorialFeature/Components/Vanguard2019/Components/VanguardVideoBackground.tsx
```diff
@@ -1,84 +1,123 @@
...
+const PrimarySeriesWrapper = styled(Box)`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  overflow: hidden;
+`
+
+export const VanguardVideoSubSeries = props => {
+  const [ref, { height: svgHeight }] = useDimensions()
```
damassi: nice -- does this have SSR implications since there's no window during the server render pass? (not sure how y'all are handling that)

ashleyjelks: It doesn't seem to have an effect on the components its being used on `Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client.` There's a workaround posted here on how to handle it if the component using this lib renders server side https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85 but this isn't happening so we are good.

artsyit:
:rocket: PR was released in v17.6.1 :rocket:

PR #2735, opened 2019-08-24
Don't transpile to common js

zephraph:
Note that this is technically a breaking change for any consumers. 

This change (which is effectively just adding `"modules": false` to `@babel/env`'s config) stops reaction from transpiling modules to commonjs.

From [webpack's docs](https://webpack.js.org/guides/tree-shaking/) (emphasis mine):

> Tree shaking is a term commonly used in the JavaScript context for dead-code elimination. *It relies on the static structure of ES2015 module syntax*, i.e. import and export.

By having reaction compile everything to commonjs before passing it off to another consuming system, we limit the optimizations available to upstream consumers. Effectively it disables the ability to use tree shaking (removing dead code) and causes us to ship more than we need. 

I've labeled this as a breaking change because we _will_ need to ensure consuming systems can handle the update, but _all_ projects that use webpack should inherently handle this update fine.


zephraph:
I changed the `.babelrc` file to be `.babelrc.js`. This is because jest still needs tests compiled to commonjs (it won't do it on it's own). The babel config does have an `env` key to switch between environments, but because I'm changing a sub-option of one of the presets I couldn't really figure out the config merging stuff.

.babelrc.js
```diff
@@ -0,0 +1,50 @@
+module.exports = {
+  presets: [
+    "@babel/typescript",
+    "@babel/react",
+    [
+      "@babel/env",
+      {
+        useBuiltIns: "usage",
+        modules: process.env.NODE_ENV === "test" ? "auto" : false,
```
alloy: (FYI the babel config format has an `env` key that allows you to apply certain configuration only in some environments.)

alloy: Oh, my bad! 🙈

zephraph:
I have been unable to get force to successfully load with these changes. Webpack seems to be trying to wrap the module in a closure. 

I tried adding a `module` entry to reaction's package.json, ensuring that force doesn't transpile things in `node_modules`, and further playing with force's babel config. Nothing to this point has worked, so I'm gonna call this a failed experiment and move on.

alloy:
@zephraph let’s not delete such branches immediately, I might be interested in taking another stab at some point.

PR #2740, opened 2019-08-26
PURCHASE-1432: Improve error messaging for generic declines on the payment page

lilyfromseattle:
# PURCHASE-1432

# Problem
A new user received a generic error message when attempting to add a card on the payment page. We should be displaying the error message in the dialog, or a custom message based on the decline code.

https://artsy.slack.com/archives/CB110C6LE/p1566571791008000
https://exchange.artsy.net/admin/orders/db831680-b90c-46e2-96a2-09fe8d440025

# Solution
While I was unable to reproduce the error in the same as the customer received it, I figured out that that a blank error modal displays when the error detail is null. I was able to trigger a different error where detail was null and added a new error type "internal error" for these types of errors. The  error dialog message is the default error contact message and the title is 'An internal error occurred' so we can differentiate between these errors and other errors also using the default message. I also log these errors to sentry.

### DIFF
--- a/src/Apps/Order/Components/PaymentPicker.tsx
+++ b/src/Apps/Order/Components/PaymentPicker.tsx
@@ -108,6 +108,7 @@ export class PaymentPicker extends React.Component<
 
   getCreditCardId: () => Promise<
     | { type: "error"; error: string | undefined }
+    | { type: "internal_error"; error: string | undefined }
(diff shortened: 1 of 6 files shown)


ashkan18:
🙌 thanks!

artsyit:
:rocket: PR was released in v17.8.5 :rocket:

PR #2747, opened 2019-08-28
Add STRIPE_PUBISHABLE_KEY to .env.oss

dleve123:
Problem

Some app, particularly the new Auction app, requires a meaningful
STRIPE_PUBISHABLE_KEY for `yarn start` to start storybooks correctly.

It's kind of hard to know this though as there's no explicit call out
of this.

Solution:

- Add entry for `STRIPE_PUBISHABLE_KEY` to .env.oss
- Add instructions on how to get this from Force

### DIFF
--- a/.env.oss
+++ b/.env.oss
@@ -1,11 +1,14 @@
 NODE_ENV=development
 WEBPACK_DEVTOOL=cheap-module-eval-source-map
 
+# ENV variables to be sourced from artsy/force
+# Use `$ hokusai staging env get` in your local checkout of artsy/force.
 APP_URL=https://staging.artsy.net
 CMS_URL=https://cms-staging.artsy.net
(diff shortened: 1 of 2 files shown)


dleve123:
@erikdstock My rationale here is that it's not necessarily evident that one should get this secret from Force

.env.oss
```diff
@@ -14,3 +14,5 @@ METAPHYSICS_ENDPOINT=https://metaphysics-staging.artsy.net
 USER_ID=
 USER_ACCESS_TOKEN=
 
+# `cd path/to/force && hokusai staging env get | grep -i stripe`
+STRIPE_PUBISHABLE_KEY=
```
alloy: Perhaps move this up above the user specific vars to the rest of the env vars and make the comment generic so it applies to all those vars?

artsyit:
:rocket: PR was released in v17.8.2 :rocket:

PR #2748, opened 2019-08-28
Show correct edition set dimensions based on the selected edition set

ashkan18:
# Problem
When order has specific edition set, we still show the artwork's dimension on the `ItemReview` section instead of the being purchased edition sets dimensions.

# Solution (hacky one)
Ultimate solution would have been to fetch edition set information on that order's line item, but at the moment we are not getting edition set info back from MP.  Here we basically get all artwork's edition sets and then we also get order's edition set id and we try to find the specific edition set using the id from the list and show dimensions for that one.

### DIFF
--- a/src/Apps/Order/Routes/Status/index.tsx
+++ b/src/Apps/Order/Routes/Status/index.tsx
@@ -341,7 +341,6 @@ export const StatusFragmentContainer = createFragmentContainer(
         lineItems {
           edges {
             node {
-              ...ItemReview_lineItem
               fulfillments {
                 edges {
                   node {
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -341,6 +341,7 @@ export const StatusFragmentContainer = createFragmentContainer(
         lineItems {
           edges {
             node {
+              ...ItemReview_lineItem
```
ds300: looks like this fragment is not actually used and can be deleted anyway. Same goes for the `artwork` selection below.

ds300: just checked whether they were being used in this file (they were not) and then check whether any of the places we passed 'order' to were not fragment containers (they were all fragment containers).

PR #2751, opened 2019-08-28
PURCHASE-1428: Get failed payment flow working with case where payment intent of a already confirmed setupintent fails when seller accepts offer

lilyfromseattle:
# [PURCHASE-1428] Get failed payment flow working with case where payment intent of a already confirmed setupintent fails when seller accepts offer

current failed payment flow works when a seller accepts an offer and payment doesn't go through, we send email to buyer about the failed payment and they fix their issue.
we need to make sure this works also for the case were we confirmed setupintent and later when seller accepts offer we tried to process off-session payment intent but that fails

# Problem
The NewPayment page is not currently set up to handle SCA authentication, so payments requiring SCA will always fail.

# Solution
I updated the NewPayment page with similar logic to what already exists in the Review page so it will display the SCA modal if authentication is required.

[PURCHASE-1428]: https://artsyproduct.atlassian.net/browse/PURCHASE-1428

### DIFF
--- a/src/Apps/Order/Routes/NewPayment/index.tsx
+++ b/src/Apps/Order/Routes/NewPayment/index.tsx
@@ -110,19 +110,18 @@ export class NewPaymentRoute extends Component<
         orderOrError.actionData &&
         orderOrError.actionData.clientSecret
       ) {
-        this.state.stripe
-          .handleCardAction(orderOrError.actionData.clientSecret)
-          .then(scaResult => {
-            if (scaResult.error) {
-              this.props.dialog.showErrorDialog({
-                title: "An error occurred",
-                message: scaResult.error.message,
-              })
-              return
-            } else {
-              this.onContinue()
-            }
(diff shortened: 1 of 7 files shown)


src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -89,6 +106,25 @@ export class NewPaymentRoute extends Component<
       if (orderOrError.error) {
         this.handleFixFailedPaymentError(orderOrError.error.code)
         return
+      } else if (
+        orderOrError.actionData &&
+        orderOrError.actionData.clientSecret
+      ) {
+        this.state.stripe
+          .handleCardAction(orderOrError.actionData.clientSecret)
+          .then(scaResult => {
```
ds300: this promise chain should probably be changed to use `await` to be consistent with above and also to make use of the general error handling in this method.

artsyit:
:rocket: PR was released in v17.8.4 :rocket:

PR #2752, opened 2019-08-28
Update ImageLink to use a fixed aspect ratio and avoid jitter

anandaroop:
https://artsyproduct.atlassian.net/browse/GROW-1481

The hubs entrypoint nav for /collect was causing the layout to stutter as images were loaded.

Following @damassi's tip, the solution here is to use Palette's `ResponsiveImage`, which fills its containing element's available space, and takes an aspect ratio so that it can reserve a fixed-height space in the layout before the image is loaded.

---

## Before

deelay

---
## After

And here it is with fixed height images.

responsive


artsyit:
:rocket: PR was released in v17.7.7 :rocket:

PR #2749, opened 2019-08-28
Create Collection Hubs entrypoint nav component for homepage

anandaroop:
https://artsyproduct.atlassian.net/browse/GROW-1360
subtask: https://artsyproduct.atlassian.net/browse/GROW-1485

hp hubs

This creates the component that we will insert into the homepage, to allow direct navigation to our set of initial collection hubs.

Another Force PR will be required to stitch this into place, in the existing Jade template for the homepage.

PR #2755, opened 2019-08-29
Expose CollectionsHubsHomepageNav so that Force can use it

pepopowitz:
In #2749, we forgot to expose the CollectionsHubsHomepageNav component in the components manifest. This addresses that, so that we can stitch it into Force.


artsyit:
:rocket: PR was released in v17.8.3 :rocket:

PR #2758, opened 2019-08-30
Setup safer/more robust linking process for reaction/force

zephraph:
See related issue: https://github.com/artsy/force/pull/4475

We've had pretty continual issues with linking reaction inside of force. @alloy did an excellent job of summarizing the issues with linking [here](https://gist.github.com/alloy/4b9cef03ad197df37bda1efaca972acb). 

I started researching solutions and stumbled upon [this blog post]() by @mtfranchetto introducing [yalc](https://github.com/whitecolor/yalc). 

Conceptually yalc is a very simple tool. It _publishes_ our package (reaction in this case) locally to the fs and the upstream project (force) uses it like it would a normal dependency. That means after adding it to force, you have to go through the process of installing dependencies like normal. This really replicates what happens when we install a package directly from NPM, but it's local!

To further reduce the friction that people face when linking, I created a new yarn script to simplify the reaction/force development workflow. Now to test force in reaction you don't even need to open force. Just run the following command from reaction: 

```
yarn integrate force
```

There's a `quicklinks.sh` file inside of force that handles what linking _means_ to force. All the `integrate` workflow really does is call `yarn watch` and `./scripts/quicklink.sh` from the project you tell it to integrate with. Technically we could add a `quicklink.sh` script to volt, positron, etc and it would essentially do the same thing. 

Note that the force side now has a bit more changes so that should definitely have some dedicated eyes on too. 

---

This should go a long ways in simplifying our workflow and reducing friction we have when developing on reaction from inside of force.


zephraph:
This PR should be merged w/ https://github.com/artsy/force/pull/4475

docs/developing_in_force.md
```diff
@@ -0,0 +1,28 @@
+# Developing in force
+
+Often in development you'll to link the development version of reaction into force.
+Traditionally one would use `yarn link` in reaction and `yarn link @artsy/reaction` in a service like force
+to acheive this. Unfortunately, this mechanism is prone to a lot of issues. Below is a guide for a smoother
```
pepopowitz:
```suggestion
to achieve this. Unfortunately, this mechanism is prone to a lot of issues. Below is a guide for a smoother
```

docs/developing_in_force.md
```diff
@@ -0,0 +1,28 @@
+# Developing in force
+
+Often in development you'll to link the development version of reaction into force.
```
pepopowitz:
```suggestion
Often in development you'll want to link the development version of reaction into force.
```

docs/developing_in_force.md
```diff
@@ -0,0 +1,28 @@
...
+
+If you haven't already, install the dependencies for reaction.
+
+```
+yarn --check-files
+```
+
+## Development
+
+Use yarn to kick off the force development process
```
alloy:
```suggestion
Use yarn to kick off the force development process.
```

docs/developing_in_force.md
```diff
@@ -0,0 +1,28 @@
...
+
+```
+yarn dev-force
+```
+
+This command will build and compile reaction, publish it locally so it can be used
+in other places, link it into force for you, and then start reaction in watch mode.
+
+_Note:_ for this process to work force _must_ be in a sibling directory of reaction and
+be named `force`.
```
alloy: Why?

package.json
```diff
@@ -41,7 +41,8 @@
     "test:watch": "jest --watch --runInBand",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
     "type-check": "tsc --pretty --noEmit",
-    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'"
+    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch' 'node scripts/watch-link.js'",
+    "dev-force": "yarn compile && yarn yalc publish && ./scripts/link-force.sh && yarn watch"
```
alloy: Why not `link-force`?

zephraph: Just trying to make it representative of what you're trying to do, not how you're trying to do it.

scripts/watch-link.js
```diff
@@ -0,0 +1,38 @@
...
+      if (code === 0) {
+        resolve()
+      } else {
+        reject(new Error(`Exited with ${code}`))
+      }
+    })
+  })
+
+const updateDevPackage = () => {
+  run("yarn", "yalc publish --force --changed")
```
alloy: If you were to pass these as an array of `args` then you should be able to get rid of the need for an intermediate shell. (Won’t make much of a difference here, so more just a general tip for when it matters.)

package.json
```diff
@@ -41,7 +41,8 @@
     "test:watch": "jest --watch --runInBand",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
     "type-check": "tsc --pretty --noEmit",
-    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'"
+    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch' 'node scripts/watch-link.js'",
```
alloy: Noice! ✨

package.json
alloy: How about having `watch` just _always_ publish (or is it really costly?) and then add a package script _to force_ that runs `yalc link @artsy/reaction`? It’s 2 steps instead of 1, but I think it would keep things simpler to understand to the user.

zephraph: Watch does always publish. Ultimately I just want to make this whole process as seamless as possible. Ideally the dev doesn't even have to open force. The only reason I haven't grouped everything into a single command (including starting force) is because the terminal outputs would be mixed and I think that could cause further complications.

alloy: I hear you, but I’m wondering if having to go into Reaction to link into Force is a helpful level of indirection. The same is going to apply to Volt, Positron, etc.

I think it makes more sense to have the app be responsible for linking and would also allow for the command to perform the linking be the same regardless of what app using Reaction.

If it turns out that people have trouble doing so, we can always add more abstractions later on.

damassi:
@zephraph - Will this  change our existing workflows, which is quite vanilla? I think it would be good to ensure that this process is somewhat separate so that we're not wedded to additional pipelines _on top_ of npm, yarn, etc. -- e.g., something like `yarn dev` (in force and reaction) vs `yarn start`, while existing processes remain untouched and one can still use `yarn link` and so on.

zephraph:
I'm working through it. Trying to make it as transparent as possible.

zephraph:
Okay, I feel like I've largely worked through it. 

None of these changes break any current workflows. You can continue to use the same commands as before and `yarn link` like normal without having to touch this workflow. This does (or should) make it _substantially_ easier to develop with reaction linked to force. 

There's one command to "link" and start reaction + force up and the symlink linking issues won't be present. It's feeling pretty solid, but also doesn't _require_ folks to change how they currently work.

docs/developing_in_force.md
```diff
@@ -0,0 +1,28 @@
...
+```
+yarn --check-files
+```
+
+## Development
+
+Use yarn to kick off the force development process.
+
+```
+yarn dev-force
```
damassi: should this be `yarn integrate force`?

zephraph: Yep, I'm updating this atm

docs/developing_in_force.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+This command will build and compile reaction, publish it locally so it can be used
+in other places, link it into force for you, and then start reaction in watch mode.
+
+_Note:_ for this process to work force _must_ be in a sibling directory of reaction and
+be named `force`.
```
damassi: I believe this is now redundant

package.json
```diff
@@ -41,7 +41,8 @@
     "test:watch": "jest --watch --runInBand",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
     "type-check": "tsc --pretty --noEmit",
-    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'"
+    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch' 'node scripts/publish-on-change.js'",
+    "integrate": "./scripts/quicklink-to-project.sh"
```
damassi: Integrate feels a bit weird - maybe `yarn dev force`?

scripts/publish-on-change.js
```diff
@@ -0,0 +1,38 @@
+const chokidar = require("chokidar")
```
damassi:
```
// @ts-check
```

scripts/publish-on-change.js
```diff
@@ -0,0 +1,38 @@
...
+        reject(new Error(`Exited with ${code}`))
+      }
+    })
+  })
+
+const updateDevPackage = () => {
+  run("yarn", "yalc publish --force --changed --push")
+}
+
+console.log("Running yalc publish in watch mode")
```
damassi: Maybe `"[Reaction] Running yalc publish in watch mode"`

package.json
zephraph: That was what I had it as originally. Contextually that's a little weird because some projects us `yarn dev` like others use `yarn start`. It just sounds like starting force in dev mode, heh... which this does, amongst other things. 

Naming is hard.

docs/developing_in_force.md
```diff
@@ -0,0 +1,38 @@
...
+
+```
+PROJECT_PATH=path/to/force yarn integrate
+```
+
+If you need to attach force to a debugger, use the `ENABLE_DEBUGGER` command as follows
+
+```
+ENABLE_DEBUGGER=true yarn integrate force
+```
```
damassi: (Would be good to cross reference this in Force's README as well.)

scripts/quicklink-to-project.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Must supply a project name or set the PROJECT_PATH env"
+fi
+
+PROJECT=${PROJECT_PATH:-../$1}
+PROJECT_NAME="$(path_to_name $PROJECT)"
+
+CURRENT_NAME=$(path_to_name $(pwd))
+
+# Do initial compile just to make sure starting files are present
+# yarn compile
```
sweir27: Is this (and the `yarn yalc` command below) intended to be commented out?

docs/developing_in_force.md
```diff
@@ -0,0 +1,38 @@
+# Developing in force
+
+Often in development you'll want to link the development version of reaction into force.
+Traditionally one would use `yarn link` in reaction and `yarn link @artsy/reaction` in a service like force
+to achieve this. Unfortunately, this mechanism is prone to a lot of issues. Below is a guide for a smoother
```
sweir27: Doesn't have to be too verbose, but could you provide some color as to what the issues are?

docs/developing_in_force.md
zephraph: In force the debugger is enabled by default if you're running it in vscode. I've just essentially short circuited that for this process assuming that normally you _wouldn't_ want to run the debugger (given that it's so much slower).

alloy:
Alright, let’s give this one a whirl! ✨

artsyit:
:rocket: PR was released in v17.10.0 :rocket:

PR #2761, opened 2019-08-30
De-randomize subtitles in homepage hubs nav to eliminate hydration mismatches

pepopowitz:
I realize that this is only temporary data until we get real data, but until then, we're getting hydration mismatch errors in the console, when the server returns a different random subtitle than the client.

This approach chooses a subtitle based on the item index, instead of randomly.


artsyit:
:rocket: PR was released in v17.9.1 :rocket:

PR #2762, opened 2019-08-30
Rename "Credit card number*" to "Credit card" to fix scan cc on mobile

dleve123:
Problem:

On mobile safari the registration form was asking the user to scan their
credit card for _all_ form elements and was placing the credit card
number on the first billing info line for reasons unknown to human kind.

Solution:

Rename "Credit card number*" to "Credit card". This is more consistent
with our BN/MO flows as well.

Jira: https://artsyproduct.atlassian.net/browse/AUCT-596


erikdstock:
#mergeongreen

artsyit:
:rocket: PR was released in v17.9.1 :rocket:

PR #2765, opened 2019-09-03
Point collection hubs entry points at marketingHubCollections for real data

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/GROW-1495.

Collaboration with Roop.

We initially built the collections hubs entry points against `marketingCollections`, which was a temporary solution until we got real hubs data and could query `marketingHubCollections` instead. We currently have real hubs data in staging, and we have "empty" hubs data in production.

This PR points the hubs entry points at `marketingHubCollections` instead of `marketingCollections`. 

Once this is merged, I'll follow up with [the homepage entry point stitching PR in Reaction](https://github.com/artsy/force/pull/4476), to validate that it works end-to-end with these changes.


pepopowitz:
I think this is ready for us to merge!

* I verified that it will work against our "empty" collection hubs in production, by pointing my local instance at `metaphysics-production.artsy.net`. 
* Inspired by [this unrelated JIRA ticket](https://artsyproduct.atlassian.net/browse/GROW-1482), I noticed that we were rendering the images using the raw URL, which is via `http`. I snuck in a change to use the `resize` util against them, which means we'll pull them from a CDN via `https`.

src/Components/v2/CollectionsHubsNav.tsx
```diff
@@ -21,10 +22,10 @@ export const CollectionsHubsNav: FC<CollectionsHubsNavProps> = props => {
       ]}
       gridGap={20}
     >
-      {props.marketingCollections.map((hub, i) => (
+      {props.marketingHubCollections.slice(0, 6).map(hub => (
         <ImageLink
           to={`/collection/${hub.slug}`}
-          src={placeholderImages[i]}
+          src={resize(hub.thumbnail, { width: 357, height: 175 })}
```
anandaroop: Is it aprpopriate that these images and the ones in the other (homepage nav) component scale their images to the same size?

(Could very well be, I just can't remember off the top)

pepopowitz: The collect page hubs could probably be a little smaller, but I thought this was close enough that it wasn't worth trying to identify the max size possible. I'll take a look though, now that you've prompted me to do so :)

pepopowitz: These are the right dimensions for the collect tiles, but the homepage tiles should actually be a little bigger - 375x184 at their max.

Unfortunately, the images I received from Jeffrey are 357x175. They don't look bad slightly upscaled right now, but when he gets back from vacation we might want to follow up to get slightly larger ones.

artsyit:
:rocket: PR was released in v20.0.2 :rocket:

PR #2766, opened 2019-09-03
Ensure carousel arrows are the same size and long titles line-break

xtina-starr:
Addresses: [GROW-1488](https://artsyproduct.atlassian.net/browse/GROW-1488)

Addressing the following:
- Ensures the carousel arrows are the same size.
- Long titles line-break on smaller screens (we also truncate more of the description).

Screen Shot 2019-09-03 at 12 36 15 PM


xtina-starr:
@mbilokonsky I'm not the biggest fan of snapshot tests because I feel like most times we just run `--update` to get the spec to pass, but if you feel strongly about them I can add it.

mbilokonsky:
I won't block a release over it, but even when `--update` is run at least it creates an artifact that can be observed in a PR review - often with a small but relevant diff that makes it easy to review.

artsyit:
:rocket: PR was released in v17.9.7 :rocket:

PR #2767, opened 2019-09-03
Sync MP schema

yuki24:
We have updated the MP schema in https://github.com/artsy/metaphysics/pull/1923 and the schema file in Reaction needs to be updated.


erikdstock:
#mergeongreen

artsyit:
:rocket: PR was released in v17.9.5 :rocket:

PR #2772, opened 2019-09-04
Defensively access images' aspect ratio in Recently Viewed rail

anandaroop:
Fixes https://artsyproduct.atlassian.net/browse/GROW-1496

This fixes the white screen of death some users were [experiencing](https://cl.ly/10c5211b1b55) after they had viewed certain pathological artworks that were missing images altogether. (Missing whyyyy 😱  — follow-up ticket at https://artsyproduct.atlassian.net/browse/GALL-2132)

When those artworks were subsequently displayed in the Recently Viewed rail, scrolling down to that rail would trigger a lazy load, which in turn would result in trying to reference an `aspect_ratio` on a null image.

The solution is simply to use the well-established Force convention for safe access of nested properties, so that `aspect_ratio` will have a fallback value of `1` in the pathological case.

### DIFF
--- a/src/Components/v2/RecentlyViewed.tsx
+++ b/src/Components/v2/RecentlyViewed.tsx
@@ -11,6 +11,7 @@ import React, { useContext } from "react"
 import { QueryRenderer } from "react-relay"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
+import { get } from "Utils/get"
 
 export interface RecentlyViewedProps {
   me: RecentlyViewed_me
@@ -48,11 +49,11 @@ export class RecentlyViewed extends React.Component<RecentlyViewedProps> {
                 <Carousel
                   data={me.recentlyViewedArtworks.edges as object[]}
                   render={artwork => {
-                    const {
-                      node: {
-                        image: { aspect_ratio },
-                      },
-                    } = artwork
+                    const aspect_ratio = get(
+                      artwork,
+                      w => w.node.image.aspect_ratio,
+                      1
+                    )
 
                     return (
                       <FillwidthItem


src/Components/v2/RecentlyViewed.tsx
```diff
@@ -48,11 +49,11 @@ export class RecentlyViewed extends React.Component<RecentlyViewedProps> {
                 <Carousel
                   data={me.recentlyViewedArtworks.edges as object[]}
                   render={artwork => {
-                    const {
-                      node: {
-                        image: { aspect_ratio },
-                      },
-                    } = artwork
+                    const aspect_ratio = get(
```
mdole: @anandaroop small q on this: if I'm reading it correctly, the previous version sets the object containing aspect ration equal to `artwork`, while this version sets it to be equal to `aspect_ratio`. Should this be `artwork = get(...` or am I just not understanding the pattern here?

anandaroop: Ah yeah, this is a totally confusing ES6 concept at first. What's going on here is: [nested object destructuring](https://davidwalsh.name/nested-destructuring). 

The previous and current code assume that `artwork` is an object something like this:

```js
artwork = {
  node: {
    title: 'Nice painting',
    image: {
     aspect_ratio: 1.414,
     url: 'http://somewhere.jpg',
    }
  }
}
```

The pre-ES6 way of accessing these deeply nested `image` properties would be simply:

```js
const aspect_ratio = artwork.node.image.aspect_ratio
const url = artwork.node.image.url
```

ES6 object destructuring gives us the option to write this more concisely as:

```js
const { node: { image: { aspect_ratio, url } } } = artwork
```

The above two snippets are exactly equivalent (`aspect_ratio` and `url` still constitute the left-hand side of the assignment), but the latter scales more concisely as you pull out more and more of those nested properties.

However _concise_ is not always the same thing as _clear_ 😅 

Note that both versions would blow up in the case where `image` was null or undefined — exactly the cause of the present bug.

mdole: ah that makes sense - thank you for the thorough explanation! Now I understand a little more ES6 than I did this morning 😄

artsyit:
:rocket: PR was released in v17.11.1 :rocket:

PR #2773, opened 2019-09-05
PURCHASE-1470: Reuse SetupIntent for make offer

lilyfromseattle:
[PURCHASE-1470]

Reaction should pass confirmedSetupIntentId so that the same confirmed setup intent is used when the order is submitted a second time

[PURCHASE-1470]: https://artsyproduct.atlassian.net/browse/PURCHASE-1470

### DIFF
--- a/src/Apps/Order/Routes/Review/index.tsx
+++ b/src/Apps/Order/Routes/Review/index.tsx
@@ -80,12 +80,13 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
       },
     ],
   }))
-  async onSubmit() {
+  async onSubmit(setupIntentId: null) {
     try {
       const orderOrError =
         this.props.order.mode === "BUY"
           ? (await this.submitBuyOrder()).commerceSubmitOrder.orderOrError
-          : (await this.submitOffer()).commerceSubmitOrderWithOffer.orderOrError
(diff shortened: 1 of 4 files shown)


src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -80,12 +80,13 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
       },
     ],
   }))
-  async onSubmit() {
+  async onSubmit(setupIntentId?: null) {
```
ashkan18: is the `?` new Typescript thing? never seen it before!

ds300: no it's been around forever. Just means the argument is optional.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -170,11 +171,12 @@ export class ReviewRoute extends Component<ReviewProps, ReviewState> {
     })
   }
 
-  submitOffer() {
+  submitOffer(setupIntentId) {
```
ds300: this should have a type annotation. Do we always setupIntentId for an offer order now?

artsyit:
:rocket: PR was released in v17.11.10 :rocket:

PR #2776, opened 2019-09-06
Allow images without zoom

kierangillen:
- We had been returning `null` if an image had zoom disabled, which would remove the image from the render. This PR allows for images to be returned that are not zoomable.

### DIFF
--- a/src/Components/v2/Lightbox/Lightbox.tsx
+++ b/src/Components/v2/Lightbox/Lightbox.tsx
@@ -26,7 +26,6 @@ const DeepZoomContainer = styled.div`
 `
 
 const StyledImage = styled(Image)`
-  cursor: zoom-in;
   max-width: 100%;
   max-height: 100%;
 `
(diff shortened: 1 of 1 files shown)


src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -299,31 +298,35 @@ class LightboxComponent extends React.Component<LightboxProps, LightboxState> {
     if (!this.state.element) {
       return (
         <Flex justifyContent="center" height={height} alignItems="center">
-          <StyledImage alt={imageAlt} src={src} preventRightClick />
+          <StyledImage
+            style={{ cursor: enabled ? "zoom-in" : "auto" }}
```
damassi: 👍 nice

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -299,31 +298,35 @@ class LightboxComponent extends React.Component<LightboxProps, LightboxState> {
...
-              preventRightClick={!isAdmin}
-            />
-          </Flex>
-        )}
-      </React.Fragment>
+        <Flex
+          justifyContent="center"
+          alignItems="center"
+          height={height}
+          onClick={enabled ? this.show.bind(this) : null}
```
kierangillen: Moved the check for `enabled` to the onClick call instead of the render

artsyit:
:rocket: PR was released in v17.11.2 :rocket:

PR #2783, opened 2019-09-09
Add intercom to order app

ds300:
Following artsy/force#4515 we're now invoking the setup logic for Intercom.

Didn't have a chance to test this on eigen because of the MP issues that prevented our QA session today, so give it a go before merging if you have the time. (note you'll need to grab the `INTERCOM_*` env vars from force staging then run force locally with this reaction branch linked. And then you need to configure eigen to use local force (think this can be done from the dev menu, but I do it by editing `ARDefaults.m`)

### DIFF
--- a/src/Apps/Order/redirects.tsx
+++ b/src/Apps/Order/redirects.tsx
@@ -1,4 +1,3 @@
-import { routes_OrderQueryResponse } from "__generated__/routes_OrderQuery.graphql"
 import { Location, RouteConfig, Router } from "found"
 import { DateTime } from "luxon"
 import { graphql } from "react-relay"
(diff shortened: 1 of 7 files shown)


src/Apps/Order/redirects.tsx
```diff
@@ -41,7 +40,7 @@ export const confirmRouteExit = (
 }
 
 const goToStatusIf = (
-  pred: (order: routes_OrderQueryResponse["order"]) => boolean,
+  pred: (order: redirects_order) => boolean,
```
sweir27: Was this just leftover from before?

ds300: Yep

ds300:
I checked on eigen and it works beautifully. Intercom is very impressive!

artsyit:
:rocket: PR was released in v17.11.6 :rocket:

PR #2786, opened 2019-09-09
Vanguard QA

eessex:
- Prevent flashing of inverted body text
- Fix text inversion on desktop safari
- Remove roman numerals from video section headers
- Remove black background from inverted text over video


artsyit:
:rocket: PR was released in v17.11.5 :rocket:

PR #2714, opened 2019-08-17
[Refactor] Replace <ArtworkFilter> in /collect 

damassi:
This PR swaps in the new generic ArtworkFilter component into `/collect`, and rounds out the new `ArtworkFilter` API.

#### Todo: 
- [x] Tests 
- [x] Add toLowerCase on filter urls -- `works-on-Paper` errors out in MP
- [x] Fix time period filter; state doesn't update and padding is off (major_periods array)

#### Example

```tsx
<ArtworkFilter
  viewer={viewer}
  filters={{ foo, bar, baz }}
  sortOptions={[
    { value: "-decayed_merch", text: "Default" },
    { value: "-partner_updated_at", text: "Recently updated" },
    { value: "-published_at", text: "Recently added" },
    { value: "-year", text: "Artwork year (desc.)" },
    { value: "year", text: "Artwork year (asc.)" },
  ]}
  onChange={filters => {
    const url = buildUrlForCollectApp(filters)

    if (typeof window !== "undefined") {
      window.history.pushState({}, null, url)
    }
  }}
  onFilterClick={(key, value, filterState) => {
    trackEvent({
      action_type: Schema.ActionType.CommercialFilterParamsChanged,
      changed: { [key]: value },
      current: filterState,
    })
  }}
  onBrickClick={props => doStuff}
  ZeroState={SomeCustomZeroState} 
/>
```

filter2

PR #2789, opened 2019-09-10
[Refactor] Replace <ArtworkFilter> in /collect -> Take 2 

damassi:
Reopening https://github.com/artsy/reaction/pull/2714 with a compromise fix for back buttons:

This uses `history.replaceState` to replace URL bar on changes to artwork filter. At any point the user is able to deep link in / share config, but back / forward buttons don’t navigate back / forward through interactions with artwork filter. Since the back / forward buttons are already pretty broken with existing filter this isn’t so bad. Using them will take user back to previous page, just before replaceState kicked in.


damassi:
@mzikherman - merging per offline convo

artsyit:
:rocket: PR was released in v18.0.0 :rocket:

PR #2793, opened 2019-09-10
PURCHASE-1476: If work can only ship domestically, the shipping form only displays the artwork's origin country

lilyfromseattle:
# PURCHASE-1476

If I'm buying a work that can only be shipped domestically, when I enter my shipping information, the shipping form has the artwork's origin country selected, and I cannot select any other country.

Previously, this only happened for US works and the copy said "continental US" instead of "domestic".

Screen Shot 2019-09-11 at 11 10 14 AM

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -1,3 +1,5 @@
+directive @principalField on FIELD
+
 input AddAssetToConsignmentSubmissionInput {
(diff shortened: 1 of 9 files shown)


artsyit:
:rocket: PR was released in v18.0.10 :rocket:

PR #2800, opened 2019-09-11
Switch to using metaphysics field for tax information

sweir27:
Will need to wait on the merge until Metaphysics is deployable again.

Per this change: https://github.com/artsy/metaphysics/pull/1940 we've moved the logic that determines this string to Metaphysics and can remove it from the view logic here.

image


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -376,9 +377,9 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
             {artwork.shippingInfo}
           </Sans>
         )}
-        {artwork.is_for_sale && artwork.priceIncludesTax && (
+        {artworkEcommerceAvailable && artwork.priceIncludesTaxDisplay && (
```
sweir27: This felt like the right logic (based on `shippingInfo`, etc. above). The logic for `is_acquireable` and `is_offerable` includes a check for "For sale", and I don't think we want this to show for works that are inquireable or at auction.

kierangillen:
Merge when ready

artsyit:
:rocket: PR was released in v17.11.18 :rocket:

PR #2809, opened 2019-09-12
PURCHASE-1416 - Adds AuctionFAQ 

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1416
- Adds a responsive view for the Auction FAQ which we can use for the iOS artwork view at `/auction-faq`
- Force PR for the route is here: https://github.com/artsy/force/pull/4553

Screen Shot 2019-09-13 at 11 35 05 AM

Screen Shot 2019-09-13 at 11 29 26 AM


src/Components/v2/AuctionFAQ.tsx
```diff
@@ -0,0 +1,184 @@
...
+            $5,000 - $9,999: $500
+            <br />
+            $10,000 - $19,999: $1,000
+            <br />
+            $20,000 - $49,999: $2,000
+            <br />
+            $50,000 - $99,999: $5,000
+            <br />
+            <br />
+            At or above $100,000: $10,000 Note: Our usual bidding increments are
```
ds300: should there be some line breakage before the note here?

ds300: image
might want to keep it to just one br between the $50,000 line and the 'at or above' line too ?

kierangillen: Oh thanks! I misplaced the last break. I'm going to keep 2 breaks for now which is consistent with the previous design, can update later if necessary.

src/Components/v2/AuctionFAQ.tsx
```diff
@@ -0,0 +1,186 @@
...
+export const AuctionFAQ: React.FC = () => {
+  return (
+    <Theme>
+      <AuctionWrapper>
+        <Serif size="8" color="black100">
+          Auction FAQs
+        </Serif>
+        <Spacer mb={2} />
+        <Serif size="3" color="black100">
+          How can we help you? Below are answers to some of the most common
```
sweir27: Currently, we serve this content via the `Page` model in gravity which is easily edited by admins and also powers the other `multi_page` instances.

Given we also want to follow up with the collector FAQ which follows a similar pattern, it would probably make sense to start building it out that way!

If you're worried about time/blocking a release then we can merge this and follow up when we do the collector FAQ (but perhaps add a `TODO` somewhere in here).

src/Components/v2/AuctionFAQ.tsx
```diff
@@ -0,0 +1,186 @@
+import { Box, media, Serif, space, Spacer, Theme, Toggle } from "@artsy/palette"
```
dleve123: Can / should this be in [`src/App/Auction/Components`](https://github.com/artsy/reaction/tree/master/src/Apps/Auction/Components) – the `RegistrationForm` is located in that directory, so would probably be the most logical to co-locate all components for the same Auction application

dleve123: Could also be in `src/App/Auctions/Routes/FAQ` as there is an FAQ route

src/Components/v2/AuctionFAQ.tsx
kierangillen: Made this update, now it's being served from metaphysics

src/Components/v2/AuctionFAQ.tsx
kierangillen: Moved to the components folder.

damassi: The pattern we follow in other apps is if the component relates to a specific route it is typically co-located in a
```
src/App/Auctions/Routes/FAQ/Components
```
folder under the route. If it's more general or free floating (like maybe the FAQModal indeed is) then `src/App/Auctions/Components`.

src/Apps/Auction/Components/AuctionFAQ.tsx
```diff
@@ -0,0 +1,158 @@
...
+import React from "react"
+import Markdown from "react-markdown"
+import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import styled from "styled-components"
+
+interface AuctionFAQProps {
+  viewer: AuctionFAQ_viewer
+}
+
+export const AuctionFAQ: React.FC<AuctionFAQProps> = ({ viewer }) => {
```
sweir27: Our old approach for `multi-page` components like this FAQ was to have a [config file](https://github.com/artsy/force/blob/master/src/desktop/components/multi_page/config.coffee) that specified all of the necessary metadata for rendering these.

It seems like we could build something similar here instead of making this specific to auctions. I'm curious if you think going down that path (a single place to define the necessary data, with a generic component that could work for any of these) would be worthwhile?

We'll be building out the Collector FAQs next, so it would be good to make sure there aren't any limitations to that approach at this point, even if we do the refactor as part of that effort.

src/Apps/Auction/Components/AuctionFAQ.tsx
```diff
@@ -0,0 +1,158 @@
...
+        id: "how-auctions-work-payments-and-shipping"
+      ) {
+        content
+        name
+      }
+      emailsAndAlerts: staticContent(
+        id: "how-auctions-work-emails-and-alerts"
+      ) {
+        content
+        name
```
sweir27: It doesn't look like we use these `name`s anywhere.

If those aren't intended to be displayable values and we can't use them, then we could remove them from the fetches here I think.

src/Apps/Auction/Components/AuctionFAQ.tsx
```diff
@@ -0,0 +1,158 @@
...
+          <StyledMarkdown
+            source={
+              viewer.buyersPremiumTaxesAndFees &&
+              viewer.buyersPremiumTaxesAndFees.content
+            }
+            containerTagName="div"
+            unwrapDisallowed
+          />
+        </Toggle>
+        <Toggle label="Payments and Shipping">
```
sweir27: If there's no `content` for these section, we probably shouldn't be showing the section headers either.

src/Apps/Auction/__tests__/routes.test.ts
```diff
@@ -31,6 +31,12 @@ describe("Auction/routes", () => {
     me: data.me,
   })
 
+  it("renders the Auction FAQ view", async () => {
+    const { status } = await render("/auction-faq", mockResolver(Fixture))
```
sweir27: Can we add a couple more tests here?

Would be nice to verify toggle behavior, and also that entire sections don't show when content is missing.

src/Apps/Auction/Components/AuctionFAQ.tsx
kierangillen: This is such a simple component I think even if we had to refactor it would be trivial. I'd rather not add anymore complexity to this for now.

src/Apps/Auction/__tests__/routes.test.ts
kierangillen: Added a test file with these tests.

artsyit:
:rocket: PR was released in v20.0.11 :rocket:

PR #2811, opened 2019-09-12
Grow 1498 fix the broken scrolling in small size

jiahaodavid:
Address [Grow 1498](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1498)

smallSizeS

To fix the broken scrolling in small size. This PR is going to remove the condition check on "draggable". Then it going to use its default value, which is true.
The condition check should be work, but it doesn't. There will be another ticket to investigate.


artsyit:
:rocket: PR was released in v18.0.11 :rocket:

PR #2801, opened 2019-09-12
[WIP] Wire up new <ArtworkFilter> into /collection/id 

damassi:
#### WIP 

Follow-up to https://github.com/artsy/reaction/pull/2789.

Wires in new ArtworkFilter into /collect. 

#### Todo: 
- [ ] Cleanup
- [ ] Fix reset filter

PR #2816, opened 2019-09-13
[Search] Replace old artwork filter with new generic implementation

damassi:
Follow up to 
- `/collection/id` swap https://github.com/artsy/reaction/pull/2801
- `/collect` swap https://github.com/artsy/reaction/pull/2789

Replaces / deletes old artwork filter and replaces it with generic implementation.


artsyit:
:rocket: PR was released in v18.1.2 :rocket:

mzikherman:
I'm loving all these code deletions!!

damassi:
it feels so good 😄

PR #2820, opened 2019-09-13
[Refactor] Add new ArtworkFilter to /artist/id

damassi:
Updates `/artist/id` with new artwork filter. 

Related: 
- `/collection/id` swap https://github.com/artsy/reaction/pull/2801
- `/collect` swap https://github.com/artsy/reaction/pull/2789
- `/search` swap https://github.com/artsy/reaction/pull/2816

#### Todo: 
- [x] Add additional artist-only aggregation filters
- [x] Add missing color / size filters
- [x] Fix aggregation counts on http://localhost:5000/collection/minimalist-prints | see https://www.artsy.net/collection/minimalist-prints
- [x] Add disabled state to filters when zeroState is showing
- [x] Delete code 
- [x] Cleanup


src/Apps/Artist/Routes/Shows/index.tsx
```diff
@@ -61,15 +61,15 @@ export const ShowsRouteFragmentContainer = createFragmentContainer(ShowsRoute, {
         pastShowsStatus: { type: "String", defaultValue: "closed" }
         pastShowsSort: { type: "PartnerShowSorts", defaultValue: END_AT_DESC }
       ) {
-      artist_currentShows: artist(id: $artistID) {
+      artist_currentShows: artist(id: $artist_id) {
```
damassi: Had to update this form because this variable name is passed verbatim directly to relay and to mp via found, etc. The filter needed it to match what our filter_artworks query params were expecting.

mzikherman:
Taking this for a walk in Storybooks, and we're over-fetching to the filter endpoint.

Screen Shot 2019-09-23 at 10 27 34 AM

There's two API calls to `/filter/artworks` made to initially render, as well as two that happen when you paginate, or interact with the filter.

It looks like one fetches aggregations with `size: 0` (and fetches them on every pagination, which is unneeded), and the other call fetches the actual data (with all aggregation params included). The version currently on `master` makes 1 call for these. Also interesting cause some optimizations in Metaphysics should have helped make sure this doesn't happen! Looking more carefully and the params differ, preventing the consolidation.

Also, when you interact with the filter, it looks like it re-triggers the `OverviewQuery`? As in, it re-queries for data used elsewhere on the main tab of the artist page. Here's that output:

Screen Shot 2019-09-23 at 10 32 57 AM

That's what was printed out when clicking 'Buy Now' ^. It makes a bunch of calls (layers, the related genes, etc.) that don't seem to be related to the grid or artist. The version on `master` makes just 2 calls on filter interactions: the artist fetch and the filter call.

damassi:
@mzikherman - will investigate; might need to pair on this tho.

mzikherman:
Definitely!

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -10,12 +10,12 @@ import { rangeToTuple } from "./Utils/rangeToTuple"
  */
 export const initialArtworkFilterState = {
   attribution_class: [],
-  height: "*-*",
+  // height: "*-*",
```
mzikherman: Not 100% sure, but these seemed to be defaults in one request but not another. This, combined with that additional `artworks_connection` load means the initial render of an artist page (unfiltered), takes one request.

mzikherman:
Pushed a commit that consolidates the initial request into one. Sidebar aggregations and the main grid are all fetched in one go.

There's still the 2 API requests when you interact with the filter or paginate: it's re-requesting the sidebar aggregations with `size: 0` and no filter, _and_ the filtered content. The sidebar aggregations don't need to be requested ever once the page is loaded. Since those are included with the grid in a refetch container, that's why we're seeing that re-request, so the component may need to be refactored? Or we might be clever-er, and similar to the `hasFilter` variable to include a snippet, we can add a `isRefetch`/`isAlreadyFetched` to leave out the sidebar.

However, the component will then re-render, and the sidebar will go away/crash. I think (since the whole thing is a refetch container, and we don't want to refetch after the initial load). So that's why I'm thinking a refactor there might be in order, vs some other solution (read in the aggregations into local state and rendering those, etc.)

src/Apps/Artist/Routes/Overview/Components/ArtistArtworkFilter.tsx
```diff
@@ -0,0 +1,206 @@
...
+      onFilterClick={(key, value, filterState) => {
+        tracking.trackEvent({
+          action_type: Schema.ActionType.CommercialFilterParamsChanged,
+          changed: { [key]: value },
+          current: filterState,
+        })
+      }}
+    >
+      <BaseArtworkFilter
+        relay={relay}
```
alloy: Why is there a need to pass the relay prop around?

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -1,37 +1,76 @@
...
+  page?: number
   partner_id?: string
-  price_range: string
-  sort: string
+  price_range?: string
+  sort?: string
   term?: string
-  width: string
+  width?: string
+}
```
alloy: FYI If all props in an object are to be optional, you can also use the `Partial<...>` type to do so.

src/Apps/Artist/Routes/Overview/Components/ArtistArtworkFilter.tsx
damassi: The artwork filter is an abstraction that fits into various app's existing queries. This is only needed if the schema deviates significantly from the default. (See `/collection/id` communicating with Kaws, for example. )

src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx
damassi: This might have an impact on other pages, let me see...

mzikherman:
Let's do this!

artsyit:
:rocket: PR was released in v19.0.0 :rocket:

PR #2821, opened 2019-09-16
[ArtworkFilter] Fix aggregations

damassi:
Fixes aggregations so they narrow down options on `/collection/id` while remaining generic on `/collect` and `/search`: 

aggregations


artsyit:
:rocket: PR was released in v18.1.4 :rocket:

PR #2822, opened 2019-09-16
Fix Collection rail on Artist page

sweir27:
As [reported in slack](https://artsy.slack.com/archives/C9SATFLUU/p1568654554019600), there's a bug right now where collection rails don't show up if an artwork's image URL is missing.

This updates reaction to be defensive to this case. I believe there's a follow-up to fix the underlying issue (where metaphysics is returning artworks that have been deleted).

image


artsyit:
:rocket: PR was released in v18.1.5 :rocket:

PR #2823, opened 2019-09-16
Vanguard sharing

eessex:
Updates sharing components to use meta from sub articles.


artsyit:
:rocket: PR was released in v18.1.6 :rocket:

PR #2824, opened 2019-09-17
Handle case for payment_method_confirmation_failed

ashkan18:
# Change
This is a follow up on https://github.com/artsy/exchange/pull/508 where we introduced new `payment_method_confirmation_failed` processing error which is the case where payment method confirmation fails on Stripe end and show proper new message.

more discussions about copy: https://artsy.slack.com/archives/C9YNS4X32/p1568669811033800


artsyit:
:rocket: PR was released in v18.1.7 :rocket:

PR #2829, opened 2019-09-18
Don't tie `yalc` publishing to `yarn watch`

zephraph:
The publishing mechanism for `yalc` was tied directly into `yarn watch` for simplicity when designing that workflow. That said, `yarn link` or just local development in reaction is slowed down by having that publishing step backed into `yarn watch`, so this pulls them apart. 

Thanks @damassi for the recommendation!

### DIFF
--- a/scripts/quicklink-to-project.sh
+++ b/scripts/quicklink-to-project.sh
@@ -43,4 +43,4 @@ fi
 
 echo ""
 echo "Beginning integration..."
-yarn concurrently --kill-others --names "$CURRENT_NAME,$PROJECT_NAME" "yarn watch" "cd $PROJECT && ./scripts/quicklink.sh @artsy/reaction"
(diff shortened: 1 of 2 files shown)


scripts/quicklink-to-project.sh
```diff
@@ -43,4 +43,4 @@ fi
 
 echo ""
 echo "Beginning integration..."
-yarn concurrently --kill-others --names "$CURRENT_NAME,$PROJECT_NAME" "yarn watch" "cd $PROJECT && ./scripts/quicklink.sh @artsy/reaction"
\ No newline at end of file
+yarn concurrently --kill-others --names "$CURRENT_NAME,$CURRENT_NAME,$PROJECT_NAME" "yarn watch" "node scripts/publish-on-change.js" "cd $PROJECT && ./scripts/quicklink.sh @artsy/reaction"
```
zephraph: I still wanted the `yalc publish` step to be viewed as coming from reaction, so I just added the same name twice for concurrently.

damassi: 👍

damassi:
Thank you! this keeps things a bit more vanilla 👍

artsyit:
:rocket: PR was released in v18.1.10 :rocket:

PR #2830, opened 2019-09-18
Collect2 app passes strict null checks

eessex:
An experiment related to https://artsyproduct.atlassian.net/browse/PLATFORM-1780

Turned on strict null checking for typescript, and converted the `Apps/Collect2` directory to pass all checks. I've disabled strict null checking for the moment so there is a chance these changes could be accidentally removed. 

Next steps: 
- implement a strategy to turn on type-checking selectively by path.
- Develop list of apps that have greatest need for checking so we can prioritize these changes.

### DIFF
--- a/src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
+++ b/src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
@@ -29,108 +29,83 @@ export class SeoProducts extends React.Component<SeoProductsProps> {
     // here the filtering is necessary so we can re-use the artwork list shown in the page (could include
     // non-acquireable artworks) without making an extra request. Also, seller image is a required field
     // so excluding those that don't have `partner.profile.icon.url`.
-
-    const edges = artworks_connection && artworks_connection.edges
-    const artworksForSeoProduct =
-      edges &&
-      edges.filter(edge => {
-        return get(edge, e => {
-          const node = e && e.node !== null && e.node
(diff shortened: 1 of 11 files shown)


src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
```diff
@@ -29,72 +29,108 @@ export class SeoProducts extends React.Component<SeoProductsProps> {
...
+              partner &&
+              partner.profile &&
+              partner.profile.icon &&
+              partner.profile.icon.url
+            return is_acquireable && iconUrl
+          } else {
+            return false
+          }
+        })
+      })
```
damassi: Why not use our `get` utility throughout? The verbosity can be reduced

alloy: Isn’t this all already wrapped inside a `get` block, in which case you could perhaps indeed reduce the verbosity by using the `!` operator to remove `null` from these intermediate values. i.e. `partner!.profile!.icon!.url!`

alloy: FWIW I think it’s fine to use the verbose version in places where you’re not already using `get` to wrap a larger block, no need to imo to replace it everywhere.

artsyit:
:rocket: PR was released in v18.1.10 :rocket:

PR #2834, opened 2019-09-19
Update fake hubs data with more accurate fake images

pepopowitz:
This PR updates the images for the fake hubs tiles to something less jarring, and hopefully less likely to prompt inquiries into whether the rail is broken.

The tiles look like this: 

image
 
It's still all temporary/fake data...until we can figure out how to get genes fully automated for the 6 hubs.


damassi:
👍

pepopowitz:
I'm going to self-merge to try to get this out today.

artsyit:
:rocket: PR was released in v18.1.12 :rocket:

PR #2837, opened 2019-09-19
Revert "Add error log"

damassi:
Reverts artsy/reaction#2835

@joeyAghion reverting this as we can reproduce locally.


artsyit:
:rocket: PR was released in v18.1.14 :rocket:

PR #2839, opened 2019-09-20
PURCHASE-1091: Buyer receives a more friendly, informative error when their card is declined in checkout

lilyfromseattle:
[PURCHASE-1091]
Desired Behavior:
Heading: Your card was declined
Body: Please enter another payment method or contact your bank for more information.
CTA: Continue

__Before:__
Screen Shot 2019-09-20 at 1 42 13 PM

__After:__
Screen Shot 2019-09-20 at 1 39 37 PM

[PURCHASE-1091]: https://artsyproduct.atlassian.net/browse/PURCHASE-1091

### DIFF
--- a/src/Apps/Order/Routes/NewPayment/index.tsx
+++ b/src/Apps/Order/Routes/NewPayment/index.tsx
@@ -91,7 +91,9 @@ export class NewPaymentRoute extends Component<
 
       if (result.type === "error") {
         this.props.dialog.showErrorDialog({
-          message: result.error,
+          title: "Your card was declined",
+          message:
+            "Please enter another payment method or contact your bank for more information.",
         })
         return
       }
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -74,7 +74,9 @@ export class PaymentRoute extends Component<PaymentProps, PaymentState> {
 
       if (result.type === "error") {
         this.props.dialog.showErrorDialog({
-          message: result.error,
+          title: "Your card was declined",
+          message:
+            "Please enter another payment method or contact your bank for more information.",
         })
         return
       }
--- a/src/Apps/Order/Routes/__tests__/NewPayment.test.tsx
+++ b/src/Apps/Order/Routes/__tests__/NewPayment.test.tsx
@@ -175,8 +175,8 @@ describe("Payment", () => {
     const page = await buildPage()
     await page.clickSubmit()
     await page.expectAndDismissErrorDialogMatching(
-      "An error occurred",
-      "This is the description of an error."
+      "Your card was declined",
+      "Please enter another payment method or contact your bank for more information."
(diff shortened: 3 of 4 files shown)


ds300:
My knowledge might be completely out of date, but looks like this error message will only show as a result of an error on the `createCreditCard` mutation, which is not where a card will be declined. That will happen on the 'sumbitOrder' or 'fixFailedPayment' mutations, right?

lilyfromseattle:
@ds300 This story is for an error message when the card is declined when attaching it to a customer (on the payment page). In the screenshot attached to the jira ticket you can see the payment page behind the error modal.

artsyit:
:rocket: PR was released in v18.1.17 :rocket:

sweir27:
@lilyfromseattle I'm not sure this is doing exactly what we want.

It looks like now for _any_ card error, we'll show this message, when it should only show if we're getting back a `card_declined` code from Stripe without any other information (which is the situation that occurs to bring up the message in the screenshot).

With this change, we're covering up other useful errors from Stripe that we used to show, such as  `"Your card has insufficient funds."` or `"Your card's security code is incorrect."`.

Unless it was intentional to not show those ☝️ errors, I think we should revisit this. I recall looking into it briefly with @ashkan18 and we may need to update gravity to return a more informative error code to easily key off of.

ashkan18:
I agree with @sweir27 , the ideal fix for this might be:
- separate `Stripe::CardError` and `PaymentError` here https://github.com/artsy/gravity/blob/0f07584accaa0eab32ab1a9392eaec8d27c94c41/app/api/util/error_handlers.rb#L56
- for `Stripe::CardError` maybe add `decline_code` and `code` to `detail`. (sadly i don't think we can change current `type` from `payment_error` to anything else since its most likely used by other clients.
- make sure MP has access to these fields and expose them properly.
- here based on the detail `decline_code` or `code` decide what to show.

Knowing ☝️ is more involved, another alternative could be use a more generic message in here since in all of the known cases what actually happens is, the card gets declined and user need to use a different payment method. So maybe this or a more generic error is also ok in this case? 🤷‍♂

sweir27:
@ashkan18 That makes sense.

I think a generic error message would be OK, but it shouldn't be so generic that we drown out useful errors (which is what we're doing in this case).

Another option is to change this so we use the generic message, but change the title to be the "specific" message we're getting back from Stripe.

So in these cases we'd have:
```
Heading: Your card has insufficient funds
Body: Please enter another payment method or contact your bank for more information.
```

```
Heading: Your card's security code is incorrect
Body: Please enter another payment method or contact your bank for more information.
```

```
Heading: Your card was declined
Body: Please enter another payment method or contact your bank for more information.
```

ashkan18:
this makes sense! it may actually make it easier in a way too since we just show the message we get back from Stripe and the detail stays the same so we may not need that extra `decline_code` and `code` for now (although would be nice to add it)

PR #2838, opened 2019-09-20
Handle collections header flicker

ashleyjelks:
This PR attempts to address the flickering effect when navigating between various collection hubs (the header switches on initial page load). 

[Links to GROW-1506](https://artsyproduct.atlassian.net/browse/GROW-1506)

Kapture 2019-09-20 at 12 46 38

PR #2840, opened 2019-09-20
[GROW-1506] use unique key

xtina-starr:
The PR is related to https://github.com/artsy/reaction/pull/2838. After this was merge I was still seeing the delay on staging. I thought it might be because the key needs to be unique. I made the minor adjustment.


artsyit:
:rocket: PR was released in v18.1.16 :rocket:

PR #2842, opened 2019-09-23
Persist telephone from bidder registration form to user profile

dblandin:
This PR updates the first GraphQL mutation to also update the user's profile with the provided telephone number.

This PR also updates the form validation to disallow blank input for all fields, which is in parity with validations on the former bidder registration form.

jira ticket :lock:: https://artsyproduct.atlassian.net/browse/AUCT-629

<a href="https://www.loom.com/share/9500245bf1394d6aa4523112768e517f"> <p>Storybook - Watch Video</p>  </a>


src/Apps/Auction/Components/__tests__/RegistrationForm.test.tsx
```diff
@@ -26,7 +26,7 @@ describe("RegistrationForm", () => {
     trackSubmissionErrors: jest.fn(),
   }
 
-  beforeAll(() => {
+  beforeEach(() => {
```
dleve123: was there a bug latent within these tests? what motivated this change?

src/Apps/Auction/Routes/Register/index.tsx
```diff
@@ -104,8 +107,17 @@ export const RegisterRoute: React.FC<RegisterProps> = props => {
         },
         onError: reject,
         mutation: graphql`
-          mutation RegisterCreateCreditCardMutation($input: CreditCardInput!) {
-            createCreditCard(input: $input) {
+          mutation RegisterPreCreateBidderMutation(
```
dleve123: So is this name intended to name the sequence of mutations: { `updateMyUserProfile`, `createCreditCard` }? Does the name of  `RegisterPreCreateBidderMutation` follow some existing pattern?

Perhaps that this is just me, but the name of `RegisterPreCreateBidderMutation` really doesn't mean much. In particular, the "pre" is confusing.

Thoughts on a name like `BidderRegistrationMutation`? Maybe the function could also be named `createBidderRegistration` as well for parity? Right now there are lot of terms flowing around: requirements, pre-create bidder, etc – and I think for the sake of clarity it would be useful to reduce them.

src/Apps/Auction/Routes/__fixtures__/MutationResults/persistBidderRequirements.ts
```diff
@@ -0,0 +1,35 @@
...
+        node: {
+          last_digits: "4242",
+        },
+      },
+    },
+  },
+}
+
+export const persistBidderRequirementsFailed: RegisterPreCreateBidderMutationResponse = {
+  updateMyUserProfile: {
```
dleve123: is it possible/sensible for the `updateMyUserProfile` to return an error? It doesn't seem like it's modeled in the MP API, so I presume not, meaning that, I _believe_ [this `catch`](https://github.com/artsy/reaction/blob/53e90c8ee71ba37177c2d30b49ef2d908e3dd55a/src/Apps/Auction/Routes/Register/index.tsx#L181) would rescue an exception.

src/Apps/Auction/Components/__tests__/RegistrationForm.test.tsx
dblandin: We started adding a test here and noticed some leakage caused by this not being called before each spec. I left it in because it does fix that in case someone ends up adding more specs here.

src/Apps/Auction/Routes/__fixtures__/MutationResults/persistBidderRequirements.ts
dblandin: > is it possible/sensible for the updateMyUserProfile to return an error?

Possible, but that seems out of scope for this PR. At the moment, the mutation doesn't model an error. Any error thrown by metaphysics should be caught by the top-level `errors` handling or the catch you linked to.

src/Apps/Auction/Routes/Register/index.tsx
dblandin: Before this change we had two mutations:

1. createCreditCard
2. createBidder

We're now updating user profile information in the first mutation. Instead of changing that to something like `createCreditCardAndUpdatePhoneNumber`, we went with the "pre" name which comes before bidder creation.

We already have a `RegisterCreateBidderMutation` defined which executed the `createBidder` behavior. This other mutation is more of a "pre" step, hence the name.

Open to other suggestions though.

dleve123: @dblandin and I chatted briefly IRL and decided that `createCreditCardAndUpdatePhoneNumber` was the least mysterious name for now while requiring minimal further refactorings

dleve123:
Approving presuming that we'll update the name of the mutation!

dblandin:
#mergeongreen

artsyit:
:rocket: PR was released in v18.1.19 :rocket:

PR #2845, opened 2019-09-25
Fix ArtistCollectionsRail's fetching of collections' artworks

anandaroop:
Addresses https://artsyproduct.atlassian.net/browse/GROW-1533
Addresses  https://artsyproduct.atlassian.net/browse/GROW-1541

### Problem

The `ArtistCollectionsRail` that  appears on the artist page — the one that shows a few triplets of thumbnails  — was breaking in two ways:

- erroring as it tried to display (non-existent) thumbnails for deleted works (GROW-1533)
- showing incorrect thumbnails unrelated to the given artist (GROW-1541)

### Background

There are currently two ways of accessing a collection's artworks via MP:

1. `marketingCollection.artworks.hits`
2. `marketingCollection.artworks.artworks_collection`

My investigation showed that method 1 was broken and ignoring input parameters in a way that explained both 1533 and 1541.

This turned out to be due to underlying bug in Gravity that @mzikherman diagnosed and fixed (https://github.com/artsy/gravity/pull/12588 and https://github.com/artsy/gravity/pull/12590)

That actually fixed the immediate issue and took the heat off of these tickets. _But…_

- we decided to convert `ArtistCollectionsRail` from method 1 to method 2 in any case, since using a connection is a better pattern for this use case in general, _and_ this gets us closer to the kind of usage we will want to have in MPv2

- doing so also entailed fixing a couple of other rails, since they relied on shared fixture data

### Solution

This simply updates three collection-related rails to use the `artwork_connection` field instead of the `hits` field — 

- `ArtistCollectionsRail`, where the problems were initially noticed a923a4ebb
- `ArtistSeriesRail`, a new rail that appears on hub pages 2eef5b331
- `RelatedCollectionsRail`, another rail that appears on collection pages 0f959f82f

### DIFF
--- a/src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
+++ b/src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
@@ -30,14 +30,15 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
 
   render() {
     const {
-      artworks: { hits },
       headerImage,
       price_guidance,
       slug,
       title,
+      artworks: { artworks_connection },
     } = this.props.collection
+    const artworks = artworks_connection.edges.map(({ node }) => node)
     const formattedTitle = (title && title.split(": ")[1]) || title
-    const bgImages = compact(hits.map(hit => hit.image && hit.image.url))
+    const bgImages = compact(artworks.map(({ image }) => image && image.url))
     const imageSize =
       bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
 
@@ -50,9 +51,9 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
           <ImgWrapper pb={1}>
             {bgImages.length ? (
               bgImages.map((url, i) => {
-                const artistName = get(hits[i].artist, a => a.name)
+                const artistName = get(artworks[i].artist, a => a.name)
                 const alt = `${artistName ? artistName + ", " : ""}${
-                  hits[i].title
+                  artworks[i].title
                 }`
                 return (
                   <SingleImgContainer key={i}>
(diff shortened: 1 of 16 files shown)


src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
```diff
@@ -150,14 +151,18 @@ export const ArtistCollectionEntityFragmentContainer = createFragmentContainer(
...
+            edges {
+              node {
+                artist {
+                  name
+                }
+                title
+                image {
+                  url(version: "small")
+                }
+              }
```
anandaroop: This is the heart of the this change, and is modeled after a known-to-be-working collection → artworks query —namely the artwork grid on the /collection/:foo page, as deduced from the Relay debugger dev tool.

(Just about everything else in this PR is relay compiler / test fixture fallout)

src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
```diff
@@ -30,14 +30,15 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
...
       headerImage,
       price_guidance,
       slug,
       title,
+      artworks: { artworks_connection },
     } = this.props.collection
+    const artworks = artworks_connection.edges.map(({ node }) => node)
     const formattedTitle = (title && title.split(": ")[1]) || title
-    const bgImages = compact(hits.map(hit => hit.image && hit.image.url))
+    const bgImages = compact(artworks.map(({ image }) => image && image.url))
```
pepopowitz: I appreciate the caution that an artwork might not have an image 👍

src/Components/RelatedCollectionsRail/RelatedCollectionEntity.tsx
```diff
@@ -45,12 +46,13 @@ export class RelatedCollectionEntity extends React.Component<CollectionProps> {
           href={`${sd.APP_URL}/collection/${slug}`}
           onClick={this.onLinkClick.bind(this)}
         >
+          Hey!
```
pepopowitz: This looks like a test message that snuck into the commit.

anandaroop: New brand guidelines. Must be excited! in! every! component!

heyy

(Lol ok this is fixed now 🙈 )

mzikherman:
LGTM! `hits` is deprecated and a connection is preferred (altho https://github.com/artsy/gravity/pull/12588 should have fixed that actual bug).

anandaroop:
Yes thanks again for the upstream fix 🙌 

Meanwhile should that `hits` field be like _actually_ deprecated in the schema?

We seem to have uncovered a lot of usage of it recently, so might be good to nip it in the bud. I know MP v1's days are numbered, but if it's a small change, might be worthwhile 🤷‍♂.

artsyit:
:rocket: PR was released in v21.0.3 :rocket:

PR #2846, opened 2019-09-25
PURCHASE-1433: Refactor carousel to use explicit height instead of 60vh

lilyfromseattle:
[PURCHASE-1433]

https://artsyproduct.atlassian.net/browse/PURCHASE-1420 brought up the issue that Artwork images are being incorrectly rendered by google. The image carousel is currently using `60vh` to render the height of the image. We need to adjust the carousel height to use a height without using `vh`.

I chose to keep the view-height heights and just add a max height as well. This way the images can still resize dynamically, but only until they hit a certain size. I believe this will solve the problem, but we will need to wait for this to be released before we can verify in google search console.

[PURCHASE-1433]: https://artsyproduct.atlassian.net/browse/PURCHASE-1433

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
@@ -54,7 +54,7 @@ export class LargeArtworkImageBrowser extends React.Component<
           showArrows={hasMultipleImages}
           options={options}
           oneSlideVisible
-          height="60vh"
+          height="auto"
           setCarouselRef={setCarouselRef}
           data={carouselImages}
           renderLeftArrow={({ flickity }) => (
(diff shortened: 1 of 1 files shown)


damassi:
This will get really weird on large monitors, leading to a lot of space as well as behavior that doesn't match the old page. We've discussed this a lot in the past (cc @kierangillen and Jeffry).

sweir27:
@lilyfromseattle I'm going to assign this to @kierangillen now that he's back in the office and originally identified this issue.

kierangillen:
@lilyfromseattle Could we keep the height `60vh` and add a maxheight of `3000px` to the image and it's wrapper? That should cover us on all fronts.

sweir27:
@lilyfromseattle it looks like these changes are also included in this PR: https://github.com/artsy/reaction/pull/2879

Would you mind closing one of them if it is no longer relevant?

PR #2847, opened 2019-09-25
Consolidate payment form components

dblandin:
Co-authored-by: @sweir27 

jira ticket :lock: : https://artsyproduct.atlassian.net/browse/DSG-2

## Before

Screenshot_2019-10-07_17-22-18

## After

Screenshot_2019-10-07_17-23-53

## Mobile Screencast

screencast


src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -1,8 +1,8 @@
-import { Box, Button, Flex, Input, Sans, Serif } from "@artsy/palette"
+import { Box, Button, Flex, Sans, Serif } from "@artsy/palette"
+import { Address, AddressForm } from "Apps/Order/Components/AddressForm"
```
dblandin: The `AddressForm` should move out of `src/Apps/Order` and into `src/Components`. I can tackle that in a separate PR.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -78,77 +55,16 @@ const InnerForm: React.FC<FormikProps<FormValues>> = props => {
...
-          </Serif>
-          <CountrySelect
-            selected={values.country}
-            onSelect={value => {
-              setFieldValue("country", value)
-              setFieldTouched("country")
+          <AddressForm
+            value={values.address}
+            onChange={(address, _key) => {
+              setFieldValue("address", address)
```
dblandin: We're dropping down to Formik's `setFieldValue` function to set the entire address as provided by the `AddressForm` component. We don't really care what changed specifically. We just want the current state.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -78,77 +55,16 @@ const InnerForm: React.FC<FormikProps<FormValues>> = props => {
...
-            onBlur={handleBlur}
-            onChange={handleChange}
-            type="tel"
-            error={touched.telephone && errors.telephone}
-            required
-            title="Telephone"
-            name="telephone"
+            errors={errors.address}
+            touched={touched.address}
+            billing
```
dblandin: Hides the "Required for shipping logistics" label for billing address.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -193,13 +109,15 @@ Yup.addMethod(Yup.string, "present", function(message) {
...
-  postalCode: Yup.string().present("Postal code is required"),
-  telephone: Yup.string().present("Telephone is required"),
+  address: Yup.object({
+    name: Yup.string().present("Name is required"),
+    addressLine1: Yup.string().present("Address is required"),
+    country: Yup.string().present("Country is required"),
+    city: Yup.string().present("City is required"),
+    region: Yup.string().present("State is required"),
+    postalCode: Yup.string().present("Postal code is required"),
+    phoneNumber: Yup.string().present("Telephone is required"),
```
dblandin: We're restructuring the validation rules here to account for the new values coming from the `AddressForm` component.

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -232,7 +150,13 @@ const OnSubmitValidationError: React.FC<{
       !formikProps.isSubmitting &&
       !formikProps.isValid
     ) {
-      cb(Object.values(formikProps.errors))
+      const clonedErrors = Object.assign({}, formikProps.errors)
+      const addressErrors = clonedErrors.address
+      delete clonedErrors.address
+
+      const errors = Object.assign({}, clonedErrors, addressErrors)
```
dblandin: We're flattening the errors here so we end up with a flat array of error messages to send via our analytics event.

erikdstock: If both are arrays you can use the same `Object.assign` spread syntax as with objects: `[...clonedErrors, ...addressErrors]` #minor

dblandin: > I guess this could be a spread instead?

Nope. TypeScript doesn't seem to like that /shrug

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -83,6 +84,12 @@ export class AddressForm extends React.Component<
     )
   }
 
+  getPhoneNumberInputDescription = () => {
+    if (!(this.props.billing && this.props.showPhoneNumberInput)) {
+      return "Required for shipping logistics"
+    }
+  }
```
erikdstock: This logic is a bit confusing: I'm reading the body of the function as 'if NOT (`billing` AND `showPhoneNumberInput`) return 'required....', else return nothing.

Other ways we could state this:
- `'if NOT `billing` OR NOT `showPhoneNumberInput` return 'required...' else return nothing
- `if `billing AND showPhoneNumberInput` return nothing, else return 'required'

Assuming this logic is correct I prefer the final case personally.

which also feels confusing to me.

dblandin: Updated this to hopefully read a bit clearer. It's still a bit awkward though. I wish this could be more like the following but trying to avoid other changes to this components API for now.

```typescript
  phoneNumberInputDescription = (): string | null => {
    if (this.props.shipping) {
      return "Required for shipping logistics"
    } else {
      return null
    }
  }
```

dblandin:
@erikdstock Added that mobile screencast!

Let me know if you have any other suggestions. Thanks!

artsyit:
:rocket: PR was released in v21.4.0 :rocket:

PR #2848, opened 2019-09-25
[GROW-1547] increment correctly at smaller breakpoints

xtina-starr:
Addresses: [GROW-1547](https://artsyproduct.atlassian.net/browse/GROW-1547)

This PR ensures that the collection rails increment by one at smaller breakpoints.

collection hubs rail

### DIFF
--- a/src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
+++ b/src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
@@ -1,10 +1,8 @@
 import { CollectionsHubLinkedCollections } from "Apps/__tests__/Fixtures/Collections"
 import { useTracking } from "Artsy/Analytics/useTracking"
 import { ArrowButton } from "Components/v2/Carousel"
-import { MockBoot } from "DevTools/MockBoot"
 import { mount } from "enzyme"
 import "jest-styled-components"
-import { clone } from "lodash"
 import React from "react"
 import { ArtistSeriesRail } from "../index"
 
--- a/src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
+++ b/src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
@@ -1,6 +1,7 @@
 import { CollectionsHubLinkedCollections } from "Apps/__tests__/Fixtures/Collections"
 import { useTracking } from "Artsy/Analytics/useTracking"
 import { ArrowButton } from "Components/v2/Carousel"
+import { MockBoot } from "DevTools/MockBoot"
 import { mount } from "enzyme"
 import "jest-styled-components"
 import { clone } from "lodash"
@@ -8,12 +9,6 @@ import React from "react"
 import { ArtistSeriesRail } from "../index"
 
 jest.mock("Artsy/Analytics/useTracking")
-jest.mock("Utils/Hooks/useMedia", () => ({
-  useMedia: () => ({
-    xl: true,
-  }),
-}))
-
 jest.mock("found", () => ({
   Link: props => <div>{props.children}</div>,
 }))
@@ -69,7 +64,11 @@ describe("ArtistSeriesRail", () => {
       singleData(),
       singleData(),
     ]
-    const Component = mount(<ArtistSeriesRail {...newprops} />)
+    const Component = mount(
+      <MockBoot>
+        <ArtistSeriesRail {...newprops} />
+      </MockBoot>
+    )
     expect(Component.find(ArrowButton).length).toBe(0)
   })
 
@@ -84,7 +83,11 @@ describe("ArtistSeriesRail", () => {
       singleData(),
       singleData(),
     ]
-    const Component = mount(<ArtistSeriesRail {...newprops} />)
+    const Component = mount(
+      <MockBoot>
+        <ArtistSeriesRail {...newprops} />
+      </MockBoot>
+    )
     expect(Component.find(ArrowButton).length).toBe(2)
   })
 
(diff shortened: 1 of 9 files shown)


jiahaodavid:
Wow!! 👍

src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx
```diff
@@ -95,7 +88,21 @@ export const ArtistSeriesRail: React.FC<ArtistSeriesRailProps> = ({
...
+    )
+  }
+
+  return (
+    <Content mt={2} py={3}>
+      <Serif size="5" mb={1}>
+        Trending Artist Series
+      </Serif>
+      <Media lessThan="lg">
+        {renderCarousel({ xs: true, sm: true, md: true })}
```
pepopowitz: This threw me off at first; the line reads like we are saying that the current breakpoint is _always_ `xs`, `sm`, _and_ `md`. 

I think I understand how `renderCarousel` is using these values, but I wonder what it would look like if we passed the breakpoint-based computed values _into_ `renderCarousel` instead of passing the breakpoints themselves? Something like this:

```
<Media lessThan="lg">
  {renderCarousel({
    groupCells: 1, 
    showArrows: members.length <= 4 && !sd.IS_MOBILE
  })
</Media>
<Media at="lg">
  {renderCarousel({
    groupCells: 4, 
    showArrows: members.length <= 4 && !sd.IS_MOBILE
  })
</Media>
<Media at="lg">
  {renderCarousel({
    groupCells: 4, 
    showArrows: false
  })
</Media>
```

And then `renderCarousel` itself would become a little simpler, and just display based on the computed values passed in.

This, to me, feels like I can tell more quickly what the differences are between the different breakpoints, because all the decisions are in one place.

pepopowitz: I also think the names `showArrows` in my example and `showArrowChecker` in the actual code are slightly misleading. They imply that this flag decides 100% if the arrows are showing...but it seems to be more complicated than that. I'm not sure what it would be, but is there a more direct name?

src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/FeaturedCollectionsRails/__tests__/FeaturedCollectionsRails.test.tsx
```diff
@@ -60,7 +54,11 @@ describe("FeaturedCollectionsRails", () => {
   })
 
   it("Renders no arrows when there are less than 5 collections", () => {
-    const component = mount(<FeaturedCollectionsRails {...props} />)
+    const component = mount(
```
pepopowitz: Now that we're using `MockBoot`, could we add some tests to target different breakpoints, and make sure our logic is correct at each of them?

src/Components/v2/Carousel.tsx
```diff
@@ -143,7 +144,6 @@ export const SmallCarousel: React.FC<CarouselProps> = props => {
         wrapAround: false,
         ...props.options,
       }}
-      {...props}
```
xtina-starr: Here we were overriding the option property. By fixing this, we able to use the Carousel component as is (which is responsive) and make the logic in the collection hub rails much simpler.

src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.test.tsx
```diff
@@ -56,41 +56,6 @@ describe("ArtistSeriesRail", () => {
...
-      singleData(),
-    ]
-    const Component = mount(
-      <MockBoot>
-        <ArtistSeriesRail {...newprops} />
-      </MockBoot>
-    )
-    expect(Component.find(ArrowButton).length).toBe(2)
-  })
-
```
xtina-starr: Because we are now relying on the default functionality of the carousel component, I removed these arrow tests.

src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/FeaturedCollectionsRails/index.tsx
```diff
@@ -23,6 +23,29 @@ interface Props {
   collectionGroup: FeaturedCollectionsRails_collectionGroup
 }
 
+export const renderCarousel = (
```
xtina-starr: I pulled out the `renderCarousel` func to be able to test that it gets called with the correct value.

src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/FeaturedCollectionsRails/__tests__/FeaturedCollectionsRails.test.tsx
```diff
@@ -53,22 +54,18 @@ describe("FeaturedCollectionsRails", () => {
...
-        <FeaturedCollectionsRails {...props} />
-      </MockBoot>
-    )
-    expect(component.find(ArrowButton).length).toBe(2)
-  })
+  //   console.log("comp", component.props())
+  //   console.log("key", component.key())
+
+  //   expect(renderCarousel).toBeCalledWith("430px")
+  // })
```
xtina-starr: This test is just about there. I need to properly mock `renderCarousel` and update the assertion to include the members prop and tracking function that gets passed to renderCarousel.

I forgot how to mock renderCarousel so that it's in scope with the render/mount of the component on line 58. Any ideas @pepopowitz?

pepopowitz: Is `jest.spyOn` what you're looking for? https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname

pepopowitz: Also, for the record, I don't feel like it's a requirement that this test exists, if it's too much of a challenge to get working.

artsyit:
:rocket: PR was released in v21.0.4 :rocket:

PR #2849, opened 2019-09-26
Updates jscodeshift to 0.6.4 (maybe)

zephraph:
For the id codemod I need to be able to use promises in the transform. I've updated the version of jscodeshift in reaction with this patch.

This _isn't_ theoretically required. We could always just remove `jscodeshift` altogether from reaction and run the codemods like...

```../codemods/node_modules/.bin/jscodeshift ...```

That'd reduce a dep in reaction and centralize the dependency........ hmmm

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,6 +132,7 @@
     "jest-junit": "6.4.0",
     "jest-raw-loader": "1.0.1",
     "jest-styled-components": "7.0.0-2",
+    "jscodeshift": "^0.6.4",
     "json5": "2.1.0",
     "klaw-sync": "6.0.0",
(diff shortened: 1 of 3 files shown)


zephraph:
Yep, maybe let's do that thing.

alloy:
Yup, that’s what I’ve been doing, which means it can be removed from Reaction altogether.

alloy:
Ah it wasn’t in Reaction, I got confused by the mention of “update”.

zephraph:
I was confused too, no worries lol

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2850, opened 2019-09-26
Add codemod command to simplify running codemods in reaction

zephraph:
This is a really small change just to make running codemods easier in reaction. 

I'll likely pull this out into `artsy/cli` later, but for the MPv2 efforts it'll be a time saver. 

Just run it like this:

```
yarn codemod <codemod-file-name>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "styled-components": "^4"
   },
   "scripts": {
+    "codemod": "./scripts/codemod.sh",
     "clean": "rm -rf .cache && rm -rf dist",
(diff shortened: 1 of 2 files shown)


mzikherman:
Idk why the license bot showed up, but LGTM!

artsyit:
:rocket: PR was released in v20.0.1 :rocket:

alloy:
I don’t know if this really simplifies things. Depending on what codemod we’re running we need to pass different arguments to `jscodeshift` (eg `--ignore-pattern`) and sometimes even to the codemod script itself (see the ‘update import’ codemod).

Having this convenience wrapper exist will only make it seem to others as if this is more complex than it really is. I think abstracting away tooling that is worthwhile to learn how to use for when you need it to, should really _heavily_ outweigh any downsides of using the tool directly.

PR #2863, opened 2019-10-01
[Refactor] Replace router code with @artsy/arc

damassi:
As our routing code has been gradually growing more tangled within reaction for a while I've  cleaned things up and pulled it into a lightweight Relay framework called [`@artsy/arc`](https://github.com/artsy/arc). The gist of it is https://github.com/artsy/arc#example, however since we've got a lot of custom behavior I've created a factory function that wraps some of our more custom bits; e.g., 

```js
import { buildServerApp as makeApp }  from "@artsy/arc"

export const buildServerApp(config) {
  return makeApp({
    ...config,

    routes: createRoutes(config.routes)
    ...
  })
}
```

Nothing needs to change over in force since the  paths and whatnot have remained the same. 

arc


src/Artsy/Router/Utils/RenderStatus.tsx
```diff
@@ -11,47 +10,29 @@ import createLogger from "Utils/logger"
 const logger = createLogger("Artsy/Router/Utils/RenderStatus")
 
 export const RenderPending: React.FC = props => {
-  const { isFetching, setIsFetching } = useSystemContext()
-
-  useEffect(() => {
-    if (!isFetching) {
-      setIsFetching(true)
-    }
-  })
```
damassi: This was a bit of unnecessary indirection, removing for now

mzikherman:
Merge (and QA on staging, deploy block if needed, etc.) at will! Took this for a 🚶 locally linked with Force and all seemed good!

artsyit:
:rocket: PR was released in v21.0.0 :rocket:

PR #2864, opened 2019-10-01
WIP: Add tooling to enable incremental migration to strict type checking in typescript

zephraph:
TL;DR This PR will add a new tool which will run strict type checking in CI and _disallow_ the introduction of any new strict type checking failures. Any failures that we currently have will be okay (until we fix them). 

## What tool will help with this migration?

[Betterer](https://github.com/phenomnomnominal/betterer) is a tool designed to assist teams in making large, incremental changes to a codebase. It does this by applying some change and snapshotting the errors that result from that change. That snapshot becomes the baseline. Any _additions_ result in a failure and any removals result in a new baseline. 

## How will the migration work?

For our typescript strict checking migration, betterer will generate a snapshot of all strict type checking failures that exist in reaction today. That'll be our baseline. New PRs will _fail CI_ if additional strict checking failures are added. As existing strict type checking failures are fixed the baseline will continually get smaller until we're eventually fully migrated over to strict type checking. 

## What's remaining?

For this tool to be successful, it has to set our team up for success. To do so it needs to provide information both at development time and via CI. At development time we need hints about what will cause CI to fail... ideally that should be fast and run as a part of our normal process. At CI time we need the proper failures (which betterer will provide) _and_ clear direction of what to fix to get those errors to go away. 

So, two things:
1. Dev time feedback
2. Clear CI time direction on what to fix


alloy:
Ooohh, and??

zephraph:
It's pretty promising. My only concern about turning it on right now, is that the information it records isn't very granular. So it captures the total number of compiler errors and fails if it raises. It doesn't really give you a sense of what new errors you added... you'd have to look through the whole list. 

I'm tossing ideas around with @phenomnomnominal here: https://github.com/phenomnomnominal/betterer/issues/4

alloy:
I see. Perhaps also worth looking at these scripts I added to emission, which give pretty clear insights into how many TS errors occurred and of what type:

https://github.com/artsy/emission/blob/05ce664e09d96770060935ee78121cde26d31815/package.json#L17-L18

zephraph:
Okay @alloy, this is looking much... betterer, ha. 

Close now. Check out the actual `.betterer.results` file.  It's looking mighty nice thanks to @phenomnomnominal's recent improvements 🙏. 

The last thing I want to do before merging this is hook up some github checks and wire it to CI.

damassi:
@zephraph - before merging, can you provide thorough explanation of what betterer is, how it fits into reaction, and how we can use it going forward? Basically just fill in the details? Everything about this is pretty opaque and would benefit from a description that others can refer to (other than the betterer docs, which are confusing).

package.json
```diff
@@ -43,7 +43,8 @@
     "test:watch": "jest --watch --runInBand",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
     "type-check": "tsc --pretty --noEmit",
-    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'"
+    "watch": "concurrently --raw --kill-others 'yarn relay --watch' 'yarn compile -w' 'yarn emit-types --watch'",
+    "betterer": "betterer"
```
damassi: alpha pls

zephraph:
Yep. I'll do that when I get back from vacation.

PR #2866, opened 2019-10-01
[Lib] Remove unused router packages

damassi:
Remove packages that are no longer needed due to https://github.com/artsy/reaction/pull/2863 and https://github.com/artsy/arc.


artsyit:
:rocket: PR was released in v21.0.1 :rocket:

PR #2870, opened 2019-10-03
[Router] Make user path a bit safer

damassi:
Fixes an object path to make safer. #trivial


artsyit:
:rocket: PR was released in v21.0.5 :rocket:

PR #2875, opened 2019-10-03
Confirm Bid Page Layout

erikdstock:
This PR sets up a basic layout, route and an initial querying pattern for the Confirm Bid page in reaction.

🔒 Jira ticket: https://artsyproduct.atlassian.net/browse/AUCT-654

The ticket above is broken off from a larger ticket, so followup work is tracked there.

Current state with live artwork data and submittable form:

image


src/Apps/Auction/Routes/Bid/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  )
+}
+
+const TrackingWrappedBidRoute: React.FC<BidProps> = props => {
+  const Component = track({
+    context_page: Schema.PageName.AuctionBidPage,
+  })(BidRoute)
+
+  return <Component {...props} />
+}
```
damassi: why not 
```tsx
const TrackingWrappedBidRoute = track({
  context_page: Schema.PageName.AuctionBidPage,
})((props: BidProps) => {
  ...
}

File: src/Apps/Auction/Routes/Bid/index.tsx [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+  {
+    artwork: graphql`
+      fragment Bid_artwork on Artwork {
+        _id
+        title
+        imageUrl
+      }
+    `,
+
+    sale: graphql`
```
mzikherman: I haven't seen this used much, multiple props defined like this. I would just start with one incoming prop, most likely `artwork`? All the relevant fields can just be listed inline, w/o fragments. Then, as you build up this component and it starts to have sub components you can split out fragments. I don't know how much use/good it is currently, to have a `Bid_sale` fragment for instance, without a proper component for it (which would clean up some of this config).

src/Apps/Auction/routes.tsx
```diff
@@ -22,6 +23,51 @@ export const routes: RouteConfig[] = [
...
+        const redirect = findRedirectBid(artwork.sale_artwork.sale, me)
+        handleRedirect(redirect, location)
+
+        return <Component {...props} />
+      }
+    },
+    query: graphql`
+      query routes_BidQuery($saleID: String!, $artworkID: String!) {
+        artwork(id: $artworkID) {
+          ...Bid_artwork
```
mzikherman: I would list all of these in-line w/o fragments, since currently the fragments aren't being split into their own components.

src/Apps/Auction/routes.tsx
```diff
@@ -22,6 +23,50 @@ export const routes: RouteConfig[] = [
     path: "/auction-faq",
     Component: AuctionFAQ,
   },
+  {
+    path: "auction/:saleID/bid2/:artworkID",
```
dblandin: I think we should match this against bid and bid2, as done below. Removes one more change to make later on. Force will ensure that the right thing is rendered. Not much in harm in having reaction ready to handle both routes now.

src/Apps/Auction/Routes/Bid/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { TrackingProp } from "react-tracking"
+import createLogger from "Utils/logger"
+
+const logger = createLogger("Apps/Auction/Routes/Bid")
+
+interface BidProps {
+  artwork: any
+  me: any
+  relay: RelayProp
+  tracking: TrackingProp
```
dblandin: We can probably remove everything related to tracking for now, including the schema change. We can circle back to that when we're ready to add all of the desired events.

src/Apps/Auction/Components/LotInfo.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { createFragmentContainer, graphql, RelayProp } from "react-relay"
+
+interface Props {
+  artwork: LotInfo_artwork
+  saleArtwork: LotInfo_saleArtwork
+  relay: RelayProp
+}
+
+export const LotInfo: React.FC<Props> = ({ artwork, saleArtwork }) => {
+  console.log(artwork, saleArtwork)
```
damassi: stray log

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Box maxWidth={550} px={[2, 0]} mx="auto" mt={[1, 0]} mb={[1, 100]}>
+        <Serif size="8">Confirm your bid</Serif>
+        <Separator />
+        <LotInfo artwork={artwork} saleArtwork={artwork.saleArtwork} />
+        <Separator />
+        <BidForm
+          saleArtwork={saleArtwork}
+          onSubmit={(values, actions) => {
+            alert(Object.keys(values))
+            alert(Object.keys(actions))
```
damassi: Intentional?

src/Apps/Auction/routes.tsx
```diff
@@ -22,6 +23,50 @@ export const routes: RouteConfig[] = [
     path: "/auction-faq",
     Component: AuctionFAQ,
   },
+  {
+    path: "auction/:saleID/bid(2)?/:artworkID",
+    Component: ConfirmBidRoute,
+    render: ({ Component, props }) => {
+      if (Component && props) {
```
damassi: When using `render` default router behavior is overwritten. This stuff should be checked inside of `ConfirmBidRoute` component.

src/Apps/Auction/routes.tsx
```diff
@@ -22,6 +23,50 @@ export const routes: RouteConfig[] = [
...
+            ...BidForm_saleArtwork
+            _id
+            id
+            sale {
+              _id
+              id
+              name
+            }
+          }
+        }
```
damassi: Can this stuff be moved into a `ConfirmBidRoute` `createFragmentContainer`? The main route `query` key (ideally, but no strictly) only contains references to fragment spreads related to components in the subtree.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
erikdstock: yes

src/Apps/Auction/routes.tsx
erikdstock: Yes, for the purposes of this PR the immediate goal was to just get a skeleton with some presentation set up while some other work could not be parallelized. I only began breaking things out into fragment containers as a final step up above. But yes, this will eventually be moved in as much as possible.

src/Apps/Auction/routes.tsx
erikdstock: Interesting. I think that might be a good optimization to consider later for both this and the Register route cc @dblandin 

Thanks!

dblandin: @damassi Cool! Yeah, that's a great idea.

artsyit:
:rocket: PR was released in v21.5.0 :rocket:

PR #2879, opened 2019-10-04
PURCHASE-1433, PURCHASE-1534: Informative error messages and maxHeight for image carousel

lilyfromseattle:
# [PURCHASE-1433]: Refactor carousel to use explicit height instead of 60vh

https://artsyproduct.atlassian.net/browse/PURCHASE-1420 brought up the issue that Artwork images are being incorrectly rendered by google. The image carousel is currently using `60vh` to render the height of the image. We need to adjust the carousel height to use a height without using `vh`.

I ended up adding a maxHeight of 2000px, which shouldn't affect behavior on large screens.

__On a large screen:__

__Before:__
Screen Shot 2019-10-09 at 6 23 05 PM

__After:__
Screen Shot 2019-10-09 at 6 22 35 PM

# [PURCHASE-1534]: User sees informative error message during checkout flow if card is declined

This is a follow-up from https://artsyproduct.atlassian.net/browse/PURCHASE-1091

(See conversation in https://github.com/artsy/reaction/pull/2839 for context)

As a user
When I add a credit card that is declined (for various reasons)
I see the reason the card was declined, along with a message telling me to call my bank or use a different card.

I changed the error modal text to be a hardcoded string and made the title the error message detail. It looks like this:

Screen Shot 2019-10-04 at 12 45 10 PM

Screen Shot 2019-10-04 at 12 46 18 PM

[PURCHASE-1433]: https://artsyproduct.atlassian.net/browse/PURCHASE-1433
[PURCHASE-1534]: https://artsyproduct.atlassian.net/browse/PURCHASE-1534


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -81,6 +81,7 @@ export class LargeArtworkImageBrowser extends React.Component<
                 width="100%"
                 px={hasMultipleImages ? [2, 2, 0] : 0}
                 height="60vh"
+                maxHeight={image.height}
```
damassi: @lilyfromseattle - this keeps coming up and I'm sorry for constantly flagging! This is the wrong approach as the image is supposed to scale. We can't lock it at a specific height because then it is not responsive. It's supposed to stretch at larger sizes. Totally down to jump on a call to explain this more in person and show exactly why.

kierangillen: For now lets leave this out of this PR and handle it in its own instance.

damassi: (will need to remove each of the other `maxHeights`)

@kierangillen - can you update the JIRA ticket with why, or close / retitle so that this isn't picked up again via this approach? I can also do this. Ideally, though, I can jump on a call with both of y'all and screen share show whats going on and why these various height solutions don't work for responsivity, independent of Google Bot.

lilyfromseattle: That height is the size of the image we're getting from the query. You're saying that the image is supposed to stretch beyond its actual dimensions?

lilyfromseattle: It seems like the image should only be scalable up to its actual dimensions, otherwise wont it pixelate?

damassi: It will, and that's ok. For larger screens this is important, and depending on the aspect ratio it will stop at the edges.

damassi: The way the page is laid out with `60vh` is deliberate, not a bug, and was carried over from the old Artwork page and discussed at length in the new artwork page rebuild. 

@kierangillen mentioned earlier that to fix the google bot screenshot we need to set an arbitrarily large maxHeight (say 2000px). When a bot visits that doesn't possess normal screen dimensions, it will cap there and be passible. There are possibly other solutions to this (using a media query for example), but capping a height won't work for our users. Happy to brainstorm (not sure how much priority this has) and can happy to go into more depths about this on a call.

lilyfromseattle: `image.height` is going to be around 2000px since we're fetching a large image. Is it better to hard code an arbitrary number?

damassi: Is this always the case though? It depends on what has been entered into CMS. But if in general, based upon testing a bunch of different pages on a large apple monitor (not a laptop) things look as they currently do in prod, then 👍on this change.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -138,7 +138,7 @@ export class SmallArtworkImageBrowser extends React.Component<
                   enabled={image.is_zoomable}
                   isDefault={image.is_default}
                   src={image.uri}
-                  maxHeight={image.height}
+                  maxHeight="2000px"
```
damassi: Why 2000 here, versus 3000 up above?

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -81,7 +81,7 @@ export class LargeArtworkImageBrowser extends React.Component<
                 width="100%"
                 px={hasMultipleImages ? [2, 2, 0] : 0}
                 height="60vh"
-                maxHeight={image.height}
+                maxHeight="3000px"
```
damassi: Can we add a comment just above this line about the reason this is needed, and perhaps link to JIRA ticket?

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
lilyfromseattle: This ArtworkImageBrowser is for mobile, so extra large screens aren't really a concern.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
damassi: @lilyfromseattle - thoughts?

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
damassi: ArtworkImageBrowser is also for desktop

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
lilyfromseattle: I changed it to a hardcoded maxHeight

damassi: I'm wondering about 

> But if in general, based upon testing a bunch of different pages on a large apple monitor (not a laptop) things look as they currently do in prod

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
lilyfromseattle: There are 2 in the component, `SmallArtworkImageBrowser` and `SmallArtworkImageBrowser`. But I've decided to make it 2000px for both.

oxaudo:
This is not something that was touched in this PR but wanted to put on your radar that I've moved `ModalDialog` to palette where it's called `Dialog`. Probably worthwhile to use that instead if Reaction one so we can deprecate that and keep things consistent.

sweir27:
@oxaudo I'll add a ticket to our backlog to consolidate around the palette version. Thanks for building that out!

zephraph:
@lilyfromseattle are there still pending changes to the max height values or is this good to go?

damassi:
Also @lilyfromseattle - I've still got a couple unresolved questions in the comments up above.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
lilyfromseattle: I tested on large external monitor, which is high resolution and the monitor apple recommends. It looks the same, as you can see in the screenshots above.

damassi: Sweet, tyty 👍

PR #2880, opened 2019-10-04
Grow-1482 collection rail images are insecure and oversized

jiahaodavid:
Address: [Grow-1482](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1482)

This PR is going to do 3 things. You can see below.

1. For the header image, the resized will transfers the format from PNG to JPG on the collection page. You can see the image below, see the size is much smaller.
Screen Shot 2019-10-04 at 12 40 51

2. For the images in the collection rail and related collection rail, it will get a secure URL when it does the query.

3. Based on the Zeplin, [Zeplin Collection Page](https://app.zeplin.io/project/5aabc5e0786bbb29b6e3dc7f/screen/5cd1af05e05b3067d84963cd), [Zeplin Artist Page](https://app.zeplin.io/project/5aabc5e0786bbb29b6e3dc7f/screen/5c7403e192623b407fcec158),
In order to fit the requirement in the Zeplin, this PR changes the size a little bit.

### DIFF
--- a/src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
+++ b/src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
@@ -38,9 +38,11 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
     } = this.props.collection
     const artworks = artworks_connection.edges.map(({ node }) => node)
     const formattedTitle = (title && title.split(": ")[1]) || title
-    const bgImages = compact(artworks.map(({ image }) => image && image.url))
+    const bgImages = compact(
+      artworks.map(({ image }) => image && image.resized.url)
+    )
     const imageSize =
-      bgImages.length === 1 ? 265 : bgImages.length === 2 ? 131 : 85
+      bgImages.length === 1 ? 262 : bgImages.length === 2 ? 130 : 86
 
     return (
       <Box pr={2}>
@@ -73,7 +75,7 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
               <ArtworkImage
                 src={headerImage}
                 lazyLoad={this.props.lazyLoad}
-                width={265}
+                width={262}
                 style={{ objectFit: "cover", objectPosition: "center" }}
               />
             )}
@@ -139,7 +141,7 @@ export const ArtworkImage = styled(Image)<{ width: number }>`
 `
 
 const ImgWrapper = styled(Flex)`
-  width: 265px;
+  width: 262px;
 `
 
 export const ArtistCollectionEntityFragmentContainer = createFragmentContainer(
@@ -160,7 +162,9 @@ export const ArtistCollectionEntityFragmentContainer = createFragmentContainer(
                 }
                 title
                 image {
-                  url(version: "small")
+                  resized(width: 262) {
+                    url
+                  }
                 }
               }
             }
--- a/src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
+++ b/src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
@@ -39,7 +39,7 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
     const artworks = artworks_connection.edges.map(({ node }) => node)
     const formattedTitle = (title && title.split(": ")[1]) || title
     const bgImages = compact(
-      artworks.map(({ image }) => image && image.resized.url)
+      artworks.map(({ image }) => image && image.resized && image.resized.url)
     )
     const imageSize =
       bgImages.length === 1 ? 262 : bgImages.length === 2 ? 130 : 86
--- a/src/Apps/Collect2/Routes/Collection/Components/Header/index.tsx
+++ b/src/Apps/Collect2/Routes/Collection/Components/Header/index.tsx
@@ -198,6 +198,7 @@ export const CollectionHeader: FC<Props> = ({ artworks, collection }) => {
             width: imageWidth * (xs ? 2 : 1),
             height: imageHeight * (xs ? 2 : 1),
             quality: 80,
+            convert_to: "jpg",
           })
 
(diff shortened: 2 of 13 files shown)


src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
```diff
@@ -38,9 +38,11 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
     } = this.props.collection
     const artworks = artworks_connection.edges.map(({ node }) => node)
     const formattedTitle = (title && title.split(": ")[1]) || title
-    const bgImages = compact(artworks.map(({ image }) => image && image.url))
+    const bgImages = compact(
+      artworks.map(({ image }) => image && image.resized.url)
```
pepopowitz: I don't know if our back-end services prevent us from having an undefined `image.resized` from this query, but we should be more cautious just in case.

Two options come to mind.

1. Chain another `&&`:

```
const bgImages = compact(
      artworks.map(({ image }) => image && image.resized && image.resized.url)
```

2. Use the `get` helper, which takes the top level object (`image`), and safely traverses a path to a property.

import the helper:

```
import { get } from "Utils/get"
```

and use it like this: 

```
const bgImages = compact(
      artworks.map(({ image }) => get(image, i => i.resized.url))
```

[Example](https://github.com/artsy/reaction/blob/0d6255c76828a43621067bdfa0c62c7c9b503d49/src/Apps/Artist/Routes/Overview/index.tsx#L79)

src/Apps/Collect2/Routes/Collection/Components/Header/index.tsx
```diff
@@ -198,6 +198,7 @@ export const CollectionHeader: FC<Props> = ({ artworks, collection }) => {
             width: imageWidth * (xs ? 2 : 1),
             height: imageHeight * (xs ? 2 : 1),
             quality: 80,
+            convert_to: "jpg",
```
pepopowitz: 👍

src/Utils/resizer.ts
```diff
@@ -48,9 +48,10 @@ export const resize = (
     width?: number
     height?: number
     quality?: number
+    convert_to?: string
```
pepopowitz: Does Gemini have a specific set of strings that it allows for `convert_to`? If so, it'd be useful to type this property so that it only allows acceptable formats.

damassi:
@jiahaodavid Would you mind updating the PR title to be proper sentence case, following repo conventions? (Noticed there's a `_` between each word.)

src/Components/RelatedCollectionsRail/RelatedCollectionEntity.tsx
```diff
@@ -36,9 +36,11 @@ export class RelatedCollectionEntity extends React.Component<CollectionProps> {
       title,
     } = this.props.collection
     const artworks = artworks_connection.edges.map(({ node }) => node)
-    const bgImages = compact(artworks.map(({ image }) => image && image.url))
+    const bgImages = compact(
+      artworks.map(({ image }) => image && image.resized.url)
```
damassi: If resized is null, this code will fail. Need to guard, similar to @pepopowitz 's comment up above.

src/Utils/resizer.ts
xtina-starr: I just did a quick look and it seem gemini doesn't have specified set of formats.

src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
jiahaodavid: Thanks for the comment. @pepopowitz  @damassi 
I did not take the "undefined" case into account when I was working on the ticket. 

Also, I have a question about the changing you suggested. 
On this one,
```javascript
const bgImages = compact(
      artworks.map(({ image }) => image && image.resized && image.resized.url)
```
Why do you ask to chain `image.resized`?  
In my understanding, for `image && image.resized && image.resized.url`, if one of these return `undefined`, the final result will get `undefined`, then we used `compact` from lodash to remove the `undefined` or `null`. We will get an empty array. On line 54, if the array were empty, we used the header image in the carousel.

jiahaodavid:
> @jiahaodavid Would you mind updating the PR title to be proper sentence case, following repo conventions? (Noticed there's a `_` between each word.)

Oh, Sorry. It looks weird when there are too many "_" in a sentence.

src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
pepopowitz: In the scenario I'm thinking of, we would not get to line 54, because we'd hit a `Cannot read property 'url' of undefined` exception. 

The existing code in the PR works great for artworks that have one of these shapes:

```
{
  "image": undefined
}

{
  "image": {
    "resized": undefined
  }
}

{
  "image": {
    "resized": "some-url"
  }
}
```

But, if the shape looks like this next example, the code would throw an exception:

```
{
  "image": { }
}
```

Because `image.resized` is undefined, and we'd be trying to reference a property on undefined. Adding the `&& image.resized ` in the middle of your mapping predicate will protect us from getting those exceptions.

I don't know that our back-end services would ever return that shape....but it doesn't hurt us at all to be cautious about it.

jiahaodavid: Oh, I see. It makes sense.
Thank you for explaining. 🤝 🙇

pepopowitz:
@jiahaodavid this looks great! I reassigned it to you - feel free to merge next time you're working.

jiahaodavid:
> @jiahaodavid this looks great! I reassigned it to you - feel free to merge next time you're working.

Yes. Thank you.

artsyit:
:rocket: PR was released in v21.7.1 :rocket:

PR #2884, opened 2019-10-07
Add hooks for cypress in the PaymentPicker

sweir27:
In order to properly test the payment flow with cypress, we need to be able to hook into these elements on the payment page.

Cypress recommends [`data-test`](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements) as the selector to use for this purpose.


alloy:
:ceiling-cat: Is there an opportunity to use attributes related to accessibility? If we need to add identifiers to elements anyways then we may as well make them helpful?

artsyit:
:rocket: PR was released in v21.2.0 :rocket:

PR #2885, opened 2019-10-07
[Refactor] Update Utils/time to use `fetchQuery`

damassi:
This replaces our use of `metaphysics` with `fetchQuery`, from `relay-runtime`. It piggybacks on our network layer.

### DIFF
--- a/src/Components/WithCurrentTime.tsx
+++ b/src/Components/WithCurrentTime.tsx
@@ -1,5 +1,6 @@
+import { useSystemContext } from "Artsy"
 import { DateTime } from "luxon"
-import React from "react"
+import React, { useEffect, useState } from "react"
 import { getCurrentTimeAsIsoString } from "Utils/getCurrentTimeAsIsoString"
 import { getOffsetBetweenGravityClock } from "Utils/time"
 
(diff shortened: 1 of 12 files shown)


src/Utils/time.ts
```diff
@@ -1,46 +1,49 @@
...
 
-interface SystemTimeResults {
-  data: {
-    system: {
-      time: {
-        unix: number
+export async function getOffsetBetweenGravityClock(
+  relayEnvironment: Environment
+): Promise<number> {
+  interface SystemTimeResults {
```
alloy: No need to define this manually, just use the typing emitted by relay-compiler as normal.

src/Components/WithCurrentTime.tsx
```diff
@@ -24,55 +25,45 @@ import { getOffsetBetweenGravityClock } from "Utils/time"
...
 
-  componentDidMount() {
-    if (this.props.syncWithServer) {
-      getOffsetBetweenGravityClock().then(timeOffsetInMilliseconds => {
-        this.setState({ timeOffsetInMilliseconds })
-      })
+  useEffect(() => {
+    if (syncWithServer) {
+      getOffsetBetweenGravityClock(relayEnvironment).then(
+        _timeOffsetInMilliseconds => {
```
alloy: Perhaps rename this to `offset` or something similarly short?

alloy:
LGTM 👌 Just a few cleanup opportunities.

artsyit:
:rocket: PR was released in v21.2.1 :rocket:

src/Utils/time.ts
```diff
@@ -1,46 +1,40 @@
...
-          time {
-            unix
-          }
-        }
-      }
-    `,
-  })
-}
+  const fetchSystemTime = () => {
+    return fetchQuery<{ data: timeQueryResponse }>(relayEnvironment, query, {})
```
alloy: @damassi Realized an important detail here, this needs to specify a cache config that forces a refetch. Even if we don’t cache responses in Reaction yet, there will come a time and this would break.

damassi: Will update 👍

alloy: Thanks!

damassi: Looking at this, there doesn't seem to be an option to configure cache (maybe because of its low-level-ness?) See https://relay.dev/docs/en/fetch-query.

alloy: Can you look at the source instead? Perhaps the docs are wrong, otherwise the typings are.

damassi: ❗️ Always look at source: https://github.com/facebook/relay/blob/master/packages/relay-runtime/query/fetchQuery.js#L32 -- it's the last slot. I'll update the docs as well.

alloy: I’m pretty sure the types In Emission has this, so perhaps it’s just a matter of updating them during the migration?

The docs need a fix indeed 👍

damassi: https://github.com/facebook/relay/pull/2881

Looking at DT it's there, just need to update reaction.

PR #2874, opened 2019-10-03
Collection hubs QA

xtina-starr:
Addresses: [GROW-1574](https://artsyproduct.atlassian.net/browse/GROW-1574) & [GROW-1575](https://artsyproduct.atlassian.net/browse/GROW-1575)

- Fixes OtherCollectionsEntity when there is a missing thumbnail image (ensure a fixed height)
Screen Shot 2019-10-03 at 5 43 36 PM

- Truncate long titles for ArtistSeriesEntity component
Screen Shot 2019-10-03 at 5 44 26 PM

PR #2886, opened 2019-10-07
use truncator for long titles

xtina-starr:
Related to https://github.com/artsy/reaction/pull/2874, we use the truncator component to truncate long titles. This PR refactors the FeaturedCollectionsRail to the truncator as well.

**Before:**
Screen Shot 2019-10-07 at 6 03 13 PM

**After:**
Screen Shot 2019-10-07 at 6 09 04 PM


zephraph:
@xtina-starr can we add a little more context to the PR title (like "in featured collection entities")? That might be too long... Ha, anyway, auto uses the PR title as a changelog entry which is why I brought it up.

artsyit:
:rocket: PR was released in v21.3.0 :rocket:

PR #2887, opened 2019-10-07
[Refactor] Replace mobile LoginForm `metaphysics` calls with `fetchQuery`

damassi:
Addresses https://github.com/artsy/reaction/pull/2887
Related https://github.com/artsy/reaction/pull/2885

Updates `LoginForm` and `helpers` to use `fetchQuery` instead of `metaphysics`, and deletes `metaphysics` fetch utility. 

login


alloy:
Ace :sparkles:

artsyit:
:rocket: PR was released in v21.2.2 :rocket:

damassi:
@alloy - I presume the `cacheConfig: { force: true }` is needed here as well?

alloy:
Unsure. In this case the input variables change, which is a thing that you would normally take into account when caching responses.

damassi:
Ahh right right

PR #2889, opened 2019-10-08
Add secure payment trust signal component

zephraph:
Fixes [PURCHASE-1459](https://artsyproduct.atlassian.net/browse/PURCHASE-1459).

Adds a `Secure Payment` component to any NSO work. 

image

I've laid the groundwork to make creating the other two components easy by building out a top level `TrustSignals` component.


src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -16,13 +17,25 @@ import { ArtworkSidebar_artwork } from "__generated__/ArtworkSidebar_artwork.gra
...
 import { SystemQueryRenderer as QueryRenderer } from "Artsy/Relay/SystemQueryRenderer"
+import { SecurePaymentFragmentContainer as SecurePayment } from "../TrustSignals/SecurePayment"
 
 export interface ArtworkSidebarProps {
   artwork: ArtworkSidebar_artwork
 }
 
 const ArtworkSidebarContainer = Box
 
+const TrustSignalsContainer = styled.div`
```
sweir27: Overall this looks great!

Out of curiosity, why do we need this CSS vs. using conditional logic/adding spacers directly in our components?

zephraph: It's mainly due to the complex display states of these components paired with their spacing needs. 

The components above have 30px mb and the separator below has nothing on the top. If any one of these trust signals are rendered, then we have to attach a 30px mb to ensure the separator isn't jutting up against the components. We can't just add mb to the trust signals though, because they may and or may not render and the trust signals are spaced out by 20px from each other.

This was the easiest way to preserve the rendering relationships and spacing.

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { createFragmentContainer } from "react-relay"
+import { graphql } from "react-relay"
+import { TrustSignal, TrustSignalProps } from "./TrustSignal"
+
+interface SecurePaymentProps
+  extends Omit<TrustSignalProps, "Icon" | "label" | "description"> {
+  artwork: SecurePayment_artwork
+}
+
+export const SecurePayment = ({ artwork, ...other }: SecurePaymentProps) =>
```
damassi: Implicit returns always end up having to be refactored and can be buggy. Can this component include a `return` statement?

src/Apps/Artwork/Components/TrustSignals/TrustSignal.tsx
```diff
@@ -0,0 +1,27 @@
...
+  label: string
+  description: string | JSX.Element
+}
+
+export const TrustSignal: FC<TrustSignalProps> = ({
+  Icon,
+  label,
+  description,
+  ...other
+}) => (
```
damassi: Same here

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
zephraph: I'm pretty aggressive in never having `return` statements if I can help it 😄.

I get the sentiment and reasoning though. This should just be a lint rule.

damassi: It really is a headache for future devs. They always get unwound. For example, if a user needed to add a console log, or a new variable, or any number of things. Its considerate.

src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
sweir27: Ah gotcha. I guess if this top-level component knew which of the child components would be rendered then it could use a `Join`, but that does seem like it would make the logic more complex.

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
zephraph: Totally, I get it. Personally I lean heavily on vscode's refactoring feature, but I understand not everyone uses that. I'll update it.

src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
zephraph: Yeah, I used join to start with and thought about making a PR for it to serve before/after components, but this seems more straight forward.

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
damassi: Thank you 👍

src/Apps/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -53,6 +66,10 @@ export class ArtworkSidebar extends Component<ArtworkSidebarProps> {
           </React.Fragment>
         )}
 
+        <TrustSignalsContainer>
```
sweir27: Can we add a test that verifies that the correct trust signal shows up based on criteria?

zephraph: Can I follow up on this? I want to explore `createTestEnv` and some ways to make setting up these tests a little simpler. Right now the logic is fairly trivial so having a unit test for the actual component itself is likely low value. Ensuring that it renders in the sidebar on expected works _is_ a test that I want to implement for all of these though.

sweir27: 👍 agreed on the higher-value sidebar tests.

artsyit:
:rocket: PR was released in v21.7.0 :rocket:

PR #2893, opened 2019-10-10
Swap name prompt in address form based on billing prop

dblandin:
If we're rendering a billing form, prompt for "Name on card" otherwise prompt for "Full name".


ashkan18:
Awesome! thanks for quick fix! 💜

PR #2896, opened 2019-10-10
Trust signals 2 

sepans:
Screen Shot 2019-10-10 at 4 57 07 PM

TODO:
- [x] Update MP schema
- [x] Update flag name to `isVerifiedSeller`
- ~[ ] Add tests~ (will do a follow up PR: https://artsyproduct.atlassian.net/browse/PURCHASE-1589)


src/Apps/Artwork/Components/TrustSignals/AuthenticityCertificate.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { createFragmentContainer } from "react-relay"
+import { graphql } from "react-relay"
+import { TrustSignal, TrustSignalProps } from "./TrustSignal"
+
+interface AuthenticityCertificateProps
+  extends Omit<TrustSignalProps, "Icon" | "label" | "description"> {
+  artwork: AuthenticityCertificate_artwork
+}
+
+export const AuthenticityCertificate = ({
```
damassi: Can use React's FC to get more accurate type info:
```ts
export const AuthenticityCertificate: React.FC<AuthenticityCertificateProps> = ...
```

src/Apps/Artwork/Components/TrustSignals/AuthenticityCertificate.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface AuthenticityCertificateProps
+  extends Omit<TrustSignalProps, "Icon" | "label" | "description"> {
+  artwork: AuthenticityCertificate_artwork
+}
+
+export const AuthenticityCertificate: React.FC<
+  AuthenticityCertificateProps
+> = ({ artwork, ...other }) => {
+  return (
+    artwork.certificateOfAuthenticity &&
```
sepans: @sweir27 do you know if I need to check for the content of `certificateOfAuthenticity` here? On CMS it is only a checkbox.

sweir27: Good question!

This field takes the form:
```
certificateOfAuthenticity {
  details
}
```

And can have 3 different responses:
```
certificateOfAuthenticity: null
certificateOfAuthenticity: {
  details: "Included"
}
certificateOfAuthenticity: {
  details: "Not included"
}
```

In this case, I believe we _only_ want to show this trust signal if `details === "Included"`. (here's an example artwork: https://staging.artsy.net/artwork/banksy-sale-ends-v-dot-2)

src/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarCommercial.test.tsx
```diff
@@ -75,7 +75,7 @@ describe("ArtworkSidebarCommercial", () => {
     expect(wrapper.text()).not.toContain("VAT included in price")
   })
 
-  it("displays singe editioned hidden availability inquire work", async () => {
```
sweir27: 🔥

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
```diff
@@ -20,7 +23,10 @@ export const SecurePayment = ({ artwork, ...other }: SecurePaymentProps) => {
           <>
             {"Secure transactions by credit card through Stripe."}
             <br />
-            <Link href="https://stripe.com/docs/security/stripe">
+            <Link
+              href="https://stripe.com/docs/security/stripe"
+              target="_blank"
```
sweir27: Thank you!

src/Apps/Artwork/Components/TrustSignals/VerifiedSeller.tsx
```diff
@@ -0,0 +1,41 @@
...
+  extends Omit<TrustSignalProps, "Icon" | "label" | "description"> {
+  artwork: VerifiedSeller_artwork
+}
+
+export const VerifiedSeller: React.FC<VerifiedSellerProps> = ({
+  artwork,
+  ...other
+}) => {
+  return (
+    artwork.partner.isVerifiedSeller && (
```
sweir27: We should probably be safe here `artwork.partner && artwork.partner.isVerifiedSeller`

src/Apps/Artwork/Components/TrustSignals/AuthenticityCertificate.tsx
sepans: Found it: https://github.com/artsy/metaphysics/blob/08ed13430fdd636af6c536f157432ec721ed16b7/src/schema/v1/artwork/index.ts#L922 I need to check. Chatted with @zephraph decided to add a `boolean` field to MP like `hasCertificateOfAuthenticity` to be able to check this rather than comparing the value of current `certificateOfAuthenticity` which has the risk of the value being updated in MP breaking this flag.

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
zephraph:
```suggestion
              target="_blank"
              rel=”noopener noreferrer”
```

Out of an abundance of caution, it's always good to add this `rel` to `target="_blank"` links, unless we specifically want to allow them to have some limited ability to control our site. 

Read more here: https://developers.google.com/web/tools/lighthouse/audits/noopener. We should probably bake this into the link as a prop (like `newTab` or something).

src/Apps/Artwork/Components/TrustSignals/VerifiedSeller.tsx
zephraph:
```suggestion
    artwork.partner && artwork.partner.isVerifiedSeller && (
```

src/Apps/Artwork/Components/TrustSignals/AuthenticityCertificate.tsx
zephraph: Yeah, more on that...

Generally it feels like we should _avoid_ cases where we have to string compare on clients for these sorts of features. For limited states we should either lean on booleans or enums from GraphQL's side to ensure that clients have as little redundant behavior as possible.

sweir27: 👍 Makes a lot of sense to add to metaphysics!

src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx
sepans: 🙀 `window.opener.location = newURL`

sepans:
MP change (`hasCertificateOfAuthenticity`) has been deployed to production so this should be good to merge.

src/Apps/Artwork/Components/TrustSignals/VerifiedSeller.tsx
```diff
@@ -0,0 +1,42 @@
...
+  artwork: VerifiedSeller_artwork
+}
+
+export const VerifiedSeller: React.FC<VerifiedSellerProps> = ({
+  artwork,
+  ...other
+}) => {
+  return (
+    artwork.partner &&
+    artwork.partner.isVerifiedSeller && (
```
sweir27: We should probably also check for `isBiddable` here! Just in the case of a gallery auction.

sweir27:
One comment, otherwise looks good! Feel free to merge when ready.

src/Apps/Artwork/Components/TrustSignals/VerifiedSeller.tsx
sepans: 🙌  @sweir27. Added it in https://github.com/artsy/reaction/pull/2896/commits/3bac79ea1e8c6d76751b2259cdadf8f67e5bd111. Just wanted to double check I got that right.

artsyit:
:rocket: PR was released in v21.8.0 :rocket:

PR #2897, opened 2019-10-11
[Publising/Text] Update react-html-parser to use DOMParser

alloy:
Relies on https://github.com/artsy/react-html-parser/pull/1

PR #2900, opened 2019-10-11
Drop htmlparser2 and parse5 deps from runtime deps

alloy:
Alright, re-applying https://github.com/artsy/reaction/pull/2897 and adding an additional change to drop usage of cheerio, which depends on parse5, in favour of DOMParser.


alloy:
#mergeongreen

artsyit:
:rocket: PR was released in v21.7.4 :rocket:

PR #2901, opened 2019-10-11
[Publishing] Do not include test fixtures in main exports and bundles

alloy:
We were exporting test fixtures through modules that are used in client bundles.

### Before

article-after-removing-cheerio


artsyit:
:rocket: PR was released in v21.7.5 :rocket:

eessex:
Looks like this broke 11 tests in Positron, where we use these for mocking tests -- possible to update these? https://circleci.com/gh/artsy/positron/5246

PR #2903, opened 2019-10-14
Add DOMParser polyfill for the server

alloy:
This so clients don’t need HTML parser deps, which the browser already provides, but for some components we obviously need to render on both server and client.


damassi:
LGTM 👍 Mergin...

artsyit:
:rocket: PR was released in v21.7.7 :rocket:

PR #2904, opened 2019-10-14
Fixes margins on XS default collection headers 

ashleyjelks:
This PR addresses 2 issues reported in the default collection header:

1) The `HorizontalPadding` component wrapped around the Collections App caused the Default Header component have 20px of padding on the left and right. This PR adds an xs breakpoint to override this on the header.

2) As the Default Header should extend beyond the horizontal padding wrapping the Collections page, this PR removes any margin-left space from the first image in the header, and any margin-right space from the last image in the header to create the effect of the header bleeding into the edge of the page on all screen sizes.

[Links to GROW-1589](https://artsyproduct.atlassian.net/browse/GROW-1589)

Screen Shot 2019-10-14 at 6 34 44 PM
Screen Shot 2019-10-14 at 6 34 52 PM
Screen Shot 2019-10-14 at 6 35 00 PM
Screen Shot 2019-10-14 at 6 35 13 PM


src/Apps/Collect2/Routes/Collection/Components/Header/DefaultHeader.tsx
```diff
@@ -72,8 +72,14 @@ export const CollectionDefaultHeader: FC<Props> = ({
...
                 }}
               >
+                {/**
+                 * Don't add 5px of padding to left margin of the first image
+                 * nor to the right marging of the last image in the header
+                 * because the header should bleed to the edge of the screen
+                 */}
                 <Image
-                  mx={0.5}
+                  ml={i === 0 ? 0 : 0.5}
```
pepopowitz: This feels like something that might be better handled with `&:first-child` and `&:last-child` selectors applied to the parent element (defined as a styled component).

[Example](https://github.com/artsy/reaction/blob/5a8c9dffbcd4cee4b0074f7af23b20188a184731/src/Components/Publishing/RelatedArticles/Canvas/RelatedArticlesCanvas.tsx#L104)

src/Apps/Collect2/Routes/Collection/Components/Header/DefaultHeader.tsx
```diff
@@ -149,6 +155,11 @@ export const CollectionDefaultHeaderFragmentContainer = createFragmentContainer(
 const DefaultHeaderContainer = styled(Box)`
   background-color: ${color("black5")};
   overflow: hidden;
+
```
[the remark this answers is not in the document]
ashleyjelks: I _think_ it should be declared before we don't want to apply this padding rule to then entire CollectionsApp component that `<HorizontalPadding>` wraps, rather it should only be applied to the Header component since we want to break the standard flow of the page by allowing the Header to extend beyond the entire Collections app's margins.

This rule is being applied to the single image header component as well: https://github.com/artsy/reaction/blob/7c8cab2985e7446dc03c3eb97f4f3dbc92c9da03/src/Apps/Collect2/Routes/Collection/Components/Header/index.tsx#L303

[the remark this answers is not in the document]
ashleyjelks: The other reason I'm hesitant to make that change is that `HorizontalPadding` is a child of `FrameWithRecentlyViewed` which is being used in several places throughout the `Collect` and `Collections` pages In order to handle the conditional padding in the  `HorizontalPadding` component used in `FrameWithRecentlyViewed` then we'd have to pass a prop to indicate when to render padding there and when not to.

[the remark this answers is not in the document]
pepopowitz: Makes sense.

src/Utils/Hooks/useWindowSize.ts
```diff
@@ -1,27 +1,19 @@
-import { useEffect, useState } from "react"
-import { getViewportWidth } from "../viewport"
+import { useLayoutEffect, useState } from "react"
```
pepopowitz: Neat! TIL about `useLayoutEffect`.

src/Utils/Hooks/useWindowSize.ts
```diff
@@ -1,27 +1,19 @@
...
-    const handleResize = () => {
-      setWindowSize(viewportWidth)
+      setSize({ width, height })
     }
-
-    window.addEventListener("resize", handleResize)
-
-    return () => window.removeEventListener("resize", handleResize)
+    window.addEventListener("resize", resize)
+    resize()
```
pepopowitz: For my understanding - is this needed for the initial mount of the component?

src/Utils/viewport.ts
```diff
@@ -1,4 +1,4 @@
-export const getViewportWidth = () => {
+export const getViewportDimensions = () => {
```
pepopowitz: 💪 Solid rename.

src/Utils/Hooks/useWindowSize.ts
ashleyjelks: Nope, good catch!

artsyit:
:rocket: PR was released in v21.8.4 :rocket:

PR #2906, opened 2019-10-14
Trust signal followup

sepans:
Follow up on https://github.com/artsy/reaction/pull/2896 to fix Verified Seller logic.

Also adds stories for the other trust signal components


artsyit:
:rocket: PR was released in v21.8.1 :rocket:

PR #2908, opened 2019-10-15
Use h1 element for page title

eessex:
Uncovered while adding integration tests, we should use an `h1` for page title.


eessex:
Looks like there was a docker outage, but all green now

artsyit:
:rocket: PR was released in v21.8.2 :rocket:

PR #2909, opened 2019-10-15
Define redirects on Confirm Bid route

erikdstock:
[Jira Ticket](https://artsyproduct.atlassian.net/browse/AUCT-655) 🔐
This PR sets up some basic redirects with tests for the new confirm bid route. It also reorganizes some of our redirect handling code.

@damassi suggested that we should consider doing redirects in the route components themselves rather than a separate `render` prop on our route config. I'm not sure the best way to organize some of the shared types in this case, but open to it.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -48,5 +48,5 @@
     "**/*.snap",
     "**/*.story.tsx"
   ],
-  "debug.node.autoAttach": "off"
+  "debug.node.autoAttach": "on"
 }
--- a/src/Apps/Auction/__tests__/routes.test.ts
+++ b/src/Apps/Auction/__tests__/routes.test.ts
@@ -5,8 +5,12 @@ import { Resolver } from "found-relay"
 import getFarceResult from "found/lib/server/getFarceResult"
 import { Environment, RecordSource, Store } from "relay-runtime"
 import {
+  BidQueryResponse,
+  BidQueryResponseFixture,
+} from "../__fixtures__/routes_BidQuery"
+import {
   DeFragedRegisterQueryResponse,
-  RegisterQueryResponseFixture as Fixture,
+  RegisterQueryResponseFixture,
 } from "../__fixtures__/routes_RegisterQuery"
 
 describe("Auction/routes", () => {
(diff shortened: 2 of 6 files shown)


src/Apps/Auction/__tests__/routes.test.ts
```diff
@@ -24,84 +28,208 @@ describe("Auction/routes", () => {
...
+        `/auction/${fixture.artwork.saleArtwork.sale.id}/bid/${
+          fixture.artwork.id
+        }`,
+        fixture
+      )
 
-    expect(status).toBe(200)
-  })
+      expect(redirect.url).toBe(
+        `/auction/${fixture.artwork.saleArtwork.sale.id}/confirm-registration`
```
erikdstock: My assumption here is that this will re-trigger the 'registration pending' modal because the user has somehow tried to bid on an artwork when they should not have been able to.

src/Apps/Auction/__tests__/routes.test.ts
```diff
@@ -24,84 +28,208 @@ describe("Auction/routes", () => {
...
-    )
+      const { redirect } = await render(
+        `/auction/${fixture.artwork.saleArtwork.sale.id}/bid/${
+          fixture.artwork.id
+        }`,
+        fixture
+      )
+      expect(redirect.url).toBe(
+        `/auction/${fixture.artwork.saleArtwork.sale.id}/registration-flow`
+      )
```
erikdstock: This was a case that I wasn't sure how to handle. Currently the user would see the registration form right below their bid select component on the confirm bid page but we haven't rebuilt this part yet and it's currently ticketed as separate work. Maybe it makes sense to do both cases at once.

Another option would be to implement some post-registration `?redirectTo` logic in our `Register` reaction app so that we can send them to the bid form after they have successfully registered (this sounds like it would get very complicated fast to me).

.vscode/settings.json
```diff
@@ -48,5 +48,5 @@
     "**/*.snap",
     "**/*.story.tsx"
   ],
-  "debug.node.autoAttach": "off"
+  "debug.node.autoAttach": "on"
```
erikdstock: hmmm, not sure why this was off or if i should not be turning it back on.

src/Apps/Auction/__tests__/routes.test.ts
dblandin: I think that still might be okay. Happy to pair on it. I don't think it's blocking this PR.

.vscode/settings.json
alloy: On is a good default imo 👍

erikdstock:
#mergeongreen

artsyit:
:rocket: PR was released in v21.9.7 :rocket:

PR #2910, opened 2019-10-15
Remove snapping effect from mobile carousels because it causes weirdness in iOS 13.

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/GROW-1591.

This PR removes the snapping effect from all instances of `<SmallCarousel>`, which is effectively all carousels in Reaction at the "xs" breakpoint.

In iOS 13, the snapping action made it really difficult to swipe through the carousel (see below video). [This is a known issue in the flickity library](https://github.com/metafizzy/flickity/issues/959). We're going to add a follow-up ticket to try to address it, but for now we're removing the snapping effect to move forward with a launch of collection hubs.

---

## Before this PR

snapping-ios12

---

## The issue in iOS 13

ios13-snapping

---

## After this PR

remove-snapping

### DIFF
--- a/src/Components/v2/Carousel.tsx
+++ b/src/Components/v2/Carousel.tsx
@@ -136,7 +136,7 @@ export const SmallCarousel: React.FC<CarouselProps> = props => {
       options={{
         cellAlign: "left",
         draggable: hasMultipleSlides,
-        freeScroll: false,
+        freeScroll: true,
         contain: true,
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in v21.8.3 :rocket:

annajiganti:
When can we get this build for NPM?

PR #2911, opened 2019-10-15
[PURCHASE-1589] Trust signals test

sepans:
Fixes [PURCHASE-1589](https://artsyproduct.atlassian.net/browse/PURCHASE-1589)

### DIFF
--- a/src/Apps/Artwork/Components/TrustSignals/__tests__/VerifiedSeller.test.tsx
+++ b/src/Apps/Artwork/Components/TrustSignals/__tests__/VerifiedSeller.test.tsx
@@ -1,0 +1,72 @@
+import { VerifiedSeller_artwork } from "__generated__/VerifiedSeller_artwork.graphql"
+import { renderRelayTree } from "DevTools"
+import React from "react"
+import { graphql } from "react-relay"
+import { ExtractProps } from "Utils/ExtractProps"
+import { VerifiedSellerFragmentContainer } from "../VerifiedSeller"
+
+jest.unmock("react-relay")
+
+const render = (
+  artwork: Omit<VerifiedSeller_artwork, " $refType">,
+  extraProps?: Partial<ExtractProps<typeof VerifiedSellerFragmentContainer>>
+) =>
+  renderRelayTree({
+    Component: (props: any) => (
+      <VerifiedSellerFragmentContainer
+        artwork={{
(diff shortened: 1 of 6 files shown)


src/Apps/Artwork/Components/TrustSignals/__tests__/VerifiedSeller.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { renderRelayTree } from "DevTools"
+import React from "react"
+import { graphql } from "react-relay"
+import { ExtractProps } from "Utils/ExtractProps"
+import { VerifiedSellerFragmentContainer } from "../VerifiedSeller"
+
+jest.unmock("react-relay")
+
+const render = (
+  artwork: Omit<VerifiedSeller_artwork, " $refType">,
```
sweir27: Whoa I haven't seen this `Omit` thing before! Does it just say to ignore that field?

sepans: Yes! A nicer use case of `Omit` @zephraph showed me:

https://github.com/artsy/reaction/blob/8b0f10e13406cf8e6406a9f36d676bd3232af55c/src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx#L9

So the `SecurePayment` has the same props as `TrustSignal` except `"Icon" | "label" | "description"` so it wouldn't be possible to pass `label` etc. to `SecurePayment` and overwrite the ones that are defined in `SecurePayment` (because all props are passed down to `TrustSignal` here:

https://github.com/artsy/reaction/blob/8b0f10e13406cf8e6406a9f36d676bd3232af55c/src/Apps/Artwork/Components/TrustSignals/SecurePayment.tsx#L36
)

zephraph: We do this enough that we should probably make a `WithoutRefType` type... What do you think @ds300?

ds300: sure why not 👍 

so glad they finally made `Omit` a built-in type 💆

artsyit:
:rocket: PR was released in v21.9.0 :rocket:

PR #2914, opened 2019-10-16
SEO Improvements to the Artist page

ashkan18:
# Changes
- Show Artist's full bio and related categories by removing "Read More": https://artsyproduct.atlassian.net/browse/PURCHASE-1569

- Update consignment links to mention consign instead of learn more: 
https://artsyproduct.atlassian.net/browse/PURCHASE-1579

image

cc: @samchieng

### DIFF
--- a/src/Apps/Artist/Routes/Overview/index.tsx
+++ b/src/Apps/Artist/Routes/Overview/index.tsx
@@ -100,7 +100,7 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, State> {
                       href="/consign"
                       onClick={this.handleConsignClick.bind(this)}
                     >
-                      Learn more
+                      Consign with Artsy
                     </a>
                     .
                   </Sans>
--- a/src/Apps/Artist/Routes/Overview/index.tsx
+++ b/src/Apps/Artist/Routes/Overview/index.tsx
@@ -47,13 +47,6 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, State> {
     // no-op
   }
 
-  maybeShowGenes() {
-    const { isReadMoreExpanded } = this.state
-    const hasNoBio = !this.props.artist.biography_blurb.text
-
-    return isReadMoreExpanded || hasNoBio
-  }
-
   render() {
     if (!this.props) {
       return null
@@ -74,7 +67,6 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, State> {
 
     // TODO: Hide right column if missing current event. Waiting on feedback
     const colNum = 9 // artist.currentEvent ? 9 : 12
-    const showGenes = this.maybeShowGenes()
 
     const isClient = typeof window !== "undefined"
     const showRecommendations =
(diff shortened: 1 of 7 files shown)


src/Components/v2/ArtistBio.tsx
```diff
@@ -16,20 +15,15 @@ export const MAX_CHARS = {
...
   render() {
+    const { bio } = this.props
     return (
       <Serif size="3">
-        <Media at="xs">{this.renderReadMore("xs")}</Media>
-        <Media greaterThan="xs">{this.renderReadMore("rest")}</Media>
+        <span
+          dangerouslySetInnerHTML={{
+            __html: bio.biography_blurb.text,
+          }}
```
zephraph: Hmmm... do we have to do this? Does the bio have html contents?

zephraph: I think my concern is this opens us up to persisted XSS attacks if a partner can add bio text.

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -94,13 +78,11 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, State> {
...
-                  <Spacer mb={1} />
-                  <Genes artist={artist} />
-                  <Spacer mb={1} />
-                </>
-              )}
+              <>
+                <Spacer mb={1} />
+                <Genes artist={artist} />
+                <Spacer mb={1} />
+              </>
```
zephraph: 👍

src/Components/v2/ArtistBio.tsx
```diff
@@ -16,20 +15,15 @@ export const MAX_CHARS = {
...
-    )
-  }
   render() {
+    const { bio } = this.props
     return (
       <Serif size="3">
-        <Media at="xs">{this.renderReadMore("xs")}</Media>
-        <Media greaterThan="xs">{this.renderReadMore("rest")}</Media>
+        <span
+          dangerouslySetInnerHTML={{
```
ashkan18: @zephraph copied this from Palette's `ReadMore`, have no clue what it does, let me know if you have any concerns with this.

zephraph: I talked to @sweir27 about this. We should probably follow up later to make sure we're covering the XSS cases.

samchieng:
Looks good – one spacing thing, I think the extra padding above the bio was always there, but wanted to make it slightly smaller to have 20px between the divider and the bio copy

Zeplin spec here: https://zpl.io/bAnYWj6

ashkan18:
good call @samchieng , fixed it 
image

src/Components/v2/ArtistBio.tsx
mzikherman: I think this is about accepting markdown as input (from Artsy admins or partners), and then generating and rendering the HTML (with links, etc.).

Sanitizing the generated HTML is probably well worth a follow up, likely in https://github.com/artsy/metaphysics/blob/ac9c2bbadf84b00ad3bcb5bf271039f7909b334f/src/schema/v1/fields/markdown.ts

ashkan18:
thanks all going to merge 💜

artsyit:
:rocket: PR was released in v21.10.0 :rocket:

PR #2915, opened 2019-10-16
Use `name` from MP for artist series rail, instead of hardcoded text.

pepopowitz:
We were using a hardcoded string for the name of the artist series rail on collection hubs pages. This PR utilizes the `name` field that was imported to KAWS, which was likely the original intention.

## After
image

## Before

image


src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/index.tsx
```diff
@@ -41,7 +41,7 @@ export const ArtistSeriesRail: React.FC<ArtistSeriesRailProps> = ({
   return (
     <Content mt={2} py={3}>
       <Serif size="5" mb={1}>
-        Trending Artist Series
+        {name}
```
anandaroop: Good catch 🙌 !

artsyit:
:rocket: PR was released in v21.9.4 :rocket:

PR #2918, opened 2019-10-16
Auct 465 pricing stub

erikdstock:
This PR sets up a skeleton component with hard-coded values for pricing transparency. It has an accompanying view in storybook and can be triggered by adding `pt=1` to the query string (once the view is mounted in force).

It also includes some #minor cleanup.

### Storybook with pricing transparency:
image

### Without pricing transparency:
image

### DIFF
--- a/src/Apps/Auction/Components/BidForm.tsx
+++ b/src/Apps/Auction/Components/BidForm.tsx
@@ -8,6 +8,7 @@ import {
   Serif,
 } from "@artsy/palette"
 import { BidForm_saleArtwork } from "__generated__/BidForm_saleArtwork.graphql"
+import { PricingTransparency } from "Apps/Auction/Components/PricingTransparency"
 import { ConditionsOfSaleCheckbox } from "Components/Auction/ConditionsOfSaleCheckbox"
 import { Formik, FormikActions, FormikValues } from "formik"
 import { dropWhile } from "lodash"
@@ -16,6 +17,7 @@ import { createFragmentContainer, graphql } from "react-relay"
 import Yup from "yup"
 
 interface Props {
+  showPricingTransparency?: boolean
   saleArtwork: BidForm_saleArtwork
   onSubmit: (values: FormikValues, actions: FormikActions<object>) => void
 }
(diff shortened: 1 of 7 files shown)


src/Apps/Auction/Components/PricingTransparency.tsx
```diff
@@ -0,0 +1,67 @@
...
+        Summary
+      </Serif>
+
+      <Flex
+        flexDirection="row"
+        justifyContent="space-between"
+        alignItems="baseline"
+        pb={1}
+      >
+        <div>
```
damassi: We generally avoid `<div>` these days and use `<Box>` (which is universal / copy / pasteable to RN).

src/Apps/Auction/__fixtures__/routes_BidQuery.ts
```diff
@@ -0,0 +1,57 @@
+import { LotInfo_saleArtwork } from "__generated__/LotInfo_saleArtwork.graphql"
+import { BidForm_saleArtwork } from "../../../__generated__/BidForm_saleArtwork.graphql"
```
damassi: These should be absolute paths

src/Apps/Auction/getRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+import { routes_RegisterQueryResponse } from "__generated__/routes_RegisterQuery.graphql"
+
+export interface Redirect {
+  path: string
+  reason: string
+}
+
+export function registerRedirect(
+  sale: routes_RegisterQueryResponse["sale"],
+  me: routes_RegisterQueryResponse["me"]
```
damassi: It's a little friendlier to use objects as arguments, similar to React's `props` pattern. That way things aren't order dependent, while also being easily introspectable by logging the single arg and seeing object keys.

src/Apps/Auction/Components/PricingTransparency.tsx
erikdstock: Hmm yea, i just copied this directly from our order app so assumed that was intentional to convey that this was strictly a row

src/Apps/Auction/getRedirect.ts
erikdstock: Agreed, & i think this may change eventually as I work towards consolidating this redirect logic elsewhere. It only has one callsite right now so going to keep as is.

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -33,7 +35,11 @@ const validationSchema = Yup.object().shape({
   ),
 })
 
-export const BidForm: React.FC<Props> = ({ onSubmit, saleArtwork }) => {
+export const BidForm: React.FC<Props> = ({
+  onSubmit,
+  saleArtwork,
+  showPricingTransparency = false,
```
erikdstock: I'm not trying to suggest here that this prop should be permanent, just wanted an easy toggle for it while developing in storybook and staging. Open to other ideas!

artsyit:
:rocket: PR was released in v21.10.6 :rocket:

PR #2921, opened 2019-10-17
[Collection] Pass COLLECT_HUBS a/b test to context

damassi:
(Pair sesh with @mbilokonsky)

This fixes an issue where the CollectionHubs nav component wasn't properly SSR rendering due to a configuration issue around A/B tests in force.

Comments inline.

### DIFF
--- a/src/Apps/Collect2/Routes/Collect/index.tsx
+++ b/src/Apps/Collect2/Routes/Collect/index.tsx
@@ -18,6 +18,7 @@ import { getMetadataForMedium } from "./Components/CollectMediumMetadata"
 
 import { Collect_marketingHubCollections } from "__generated__/Collect_marketingHubCollections.graphql"
 import { collectRoutes_ArtworkFilterQueryResponse } from "__generated__/collectRoutes_ArtworkFilterQuery.graphql"
+import { useSystemContext } from "Artsy"
 import { ArtworkFilter } from "Components/v2/ArtworkFilter"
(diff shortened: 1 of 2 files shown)


src/Apps/__stories__/Collect.story.tsx
```diff
@@ -5,7 +5,17 @@ import { collectRoutes } from "../Collect2/collectRoutes"
...
   .add("Collect", () => {
-    return <MockRouter routes={collectRoutes} initialRoute="/collect" />
+    return (
+      <MockRouter
+        routes={collectRoutes}
+        initialRoute="/collect"
+        context={{
+          // FIXME: Remove after A/B test completes
+          // @ts-ignore
+          COLLECTION_HUBS: "experiment",
```
damassi: When working in Storybooks, we need to inject the ENV var in similar to how we do pass it into context on the Force side of things via A/B test

src/Apps/Collect2/Routes/Collect/index.tsx
```diff
@@ -41,15 +42,18 @@ export const CollectApp = track({
   const { description, breadcrumbTitle, title } = getMetadataForMedium(medium)
   const { trackEvent } = useTracking()
 
+  // FIXME: Remove after A/B test completes
+  // @ts-ignore
+  const { COLLECTION_HUBS } = useSystemContext()
```
damassi: Pluck things from our context, as passed in from force

artsyit:
:rocket: PR was released in v21.10.2 :rocket:

PR #2922, opened 2019-10-17
Grow-1677 Analytics tracking for hub entry points

jiahaodavid:
Address: [Grow-1577](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1577).

This PR : 

1. Adding the `onClick` property to `imagelink`, then the clicking action becomes trackable.
2. For the `CollectionsHubsHomepageNav`. Adding the `track()` as the higher(higher than the relay) order component. Because `useTracking ` still need to wrap the component with decorator or hoc, this is the FC, so here I used hoc.
3. For the `CollectionHubsNav`. He has a parent component(which is `CollectApp`, it already wrapped up with `track()` ), so, I just added `useTracking` in `CollectionHubsNav`.
4. I found out these two nav components do not have test suites.

### DIFF
--- a/src/Components/v2/CollectionsHubsHomepageNav.tsx
+++ b/src/Components/v2/CollectionsHubsHomepageNav.tsx
@@ -1,8 +1,10 @@
 import { CSSGrid } from "@artsy/palette"
 import { Serif } from "@artsy/palette"
 import { CollectionsHubsHomepageNav_marketingHubCollections } from "__generated__/CollectionsHubsHomepageNav_marketingHubCollections.graphql"
-import React, { FC } from "react"
+import { AnalyticsSchema } from "Artsy/Analytics"
+import React, { useEffect } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import track, { useTracking } from "react-tracking"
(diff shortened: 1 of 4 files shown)


src/Components/v2/CollectionsHubsHomepageNav.tsx
```diff
@@ -1,41 +1,64 @@
...
+  (props: CollectionsHubsHomepageNavProps) => {
+    const { trackEvent } = useTracking()
+    useEffect(() => {
+      trackEvent({
+        action_type: AnalyticsSchema.ActionType.Impression,
+        context_page: AnalyticsSchema.PageName.HomePage,
+        context_module: AnalyticsSchema.ContextModule.CollectionHubEntryPoint,
+        subject: AnalyticsSchema.Subject.FeaturedCategories,
+      })
+    }, [])
```
anandaroop: Is this mystery `[]` argument the thing that keeps this impression event from firing on subsequent renders _after_ mounting?

artsyit:
:rocket: PR was released in v21.11.2 :rocket:

PR #2923, opened 2019-10-17
Don't overwrite the User-Agent header for relay operations

zephraph:
Fixes [PURCHASE-1140](https://artsyproduct.atlassian.net/browse/PURCHASE-1140).

It was reported that many of our orders were being created with a user agent of something like `Reaction/16.7.0; Metaphysics`. This user agent is first set in reaction (see below) then appended to in metaphysics. 

https://github.com/artsy/reaction/blob/501d18091180abed728680661456b57f82ad27a8/src/Artsy/Relay/createRelaySSREnvironment.ts#L34

Essentially we just needed to prepend what's in reaction with the UA from the request (or the window if it's on the client).

### DIFF
--- a/src/Artsy/Router/buildServerApp.tsx
+++ b/src/Artsy/Router/buildServerApp.tsx
@@ -36,7 +36,7 @@ export function buildServerApp(config: RouterConfig & { userAgent?: string }) {
     getRelayEnvironment: () => {
       const relayEnvironment =
         (config.context && config.context.relayEnvironment) ||
-        createRelaySSREnvironment({ user })
+        createRelaySSREnvironment({ user, userAgent: config.userAgent })
 
       return relayEnvironment
     },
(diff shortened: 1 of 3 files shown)


src/Artsy/Router/buildServerApp.tsx
```diff
@@ -36,7 +36,7 @@ export function buildServerApp(config: RouterConfig & { userAgent?: string }) {
     getRelayEnvironment: () => {
       const relayEnvironment =
         (config.context && config.context.relayEnvironment) ||
-        createRelaySSREnvironment({ user })
+        createRelaySSREnvironment({ user, userAgent: config.userAgent })
```
alloy: So it’s Force that would set this on the server?

alloy:
Yup, this looks good to me 👍 

#mergeongreen

src/Artsy/Router/buildServerApp.tsx
damassi: It's set in this function: https://github.com/damassi/force/blob/4683a3e889529213dfa9c357495ac1a0b9cc886b/src/desktop/lib/buildServerAppContext.ts#L20

alloy: Gotcha, thanks 👍

artsyit:
:rocket: PR was released in v21.10.2 :rocket:

PR #2924, opened 2019-10-17
Fixes carousel for iOS 13

kierangillen:
- iOS introduced an issue with how Flickity is handling touch events
- Uses this fix: https://github.com/metafizzy/flickity/issues/740#issuecomment-485562201

### DIFF
--- a/src/Components/v2/Carousel.tsx
+++ b/src/Components/v2/Carousel.tsx
@@ -187,6 +187,19 @@ export class BaseCarousel extends React.Component<
     }
   }
 
+  allowPreventDefault = () => {
+    // Fix for flickity bug introduced in iOS 13 https://github.com/metafizzy/flickity/issues/740
+    let startX
+    this.carouselRef.ontouchstart = e => {
+      startX = e.touches[0].clientX
+    }
+    this.carouselRef.ontouchmove = e => {
+      if (Math.abs(e.touches[0].clientX - startX) > 5 && e.cancelable) {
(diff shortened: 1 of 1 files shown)


src/Components/v2/Carousel.tsx
```diff
@@ -206,13 +206,23 @@ export class BaseCarousel extends React.Component<
...
+          }
+        }
       }
     )
   }
 
   componentWillUnmount() {
     if (this.flickity) {
       this.flickity.off("select")
+      this.flickity.flickity("destroy")
```
kierangillen: This is unrelated but should be done for house keeping.

src/Components/v2/Carousel.tsx
```diff
@@ -206,13 +206,23 @@ export class BaseCarousel extends React.Component<
...
         this.flickity.on("select", this.handleSlideChange)
+        let startX
+        this.carouselRef.ontouchstart = e => {
+          startX = e.touches[0].clientX
+        }
+        this.carouselRef.ontouchmove = e => {
+          if (Math.abs(e.touches[0].clientX - startX) > 5 && e.cancelable) {
+            e.preventDefault()
+          }
+        }
```
damassi: Can this be moved into a function so that it's clearer what's happening?

kierangillen: 👍

artsyit:
:rocket: PR was released in v21.10.3 :rocket:

PR #2926, opened 2019-10-18
[createRelaySSREnvironment] Lazy load deps so they can be stripped.

alloy:
This fixes https://github.com/artsy/force/pull/4698 so it can be applied again.

### DIFF
--- a/src/Artsy/Relay/createRelaySSREnvironment.ts
+++ b/src/Artsy/Relay/createRelaySSREnvironment.ts
@@ -49,9 +49,16 @@ export interface RelaySSREnvironment extends Environment {
 export function createRelaySSREnvironment(config: Config = {}) {
   const { cache = {}, checkStatus, user, relayNetwork, userAgent } = config
 
+  /**
+   * Lazy load these here so we can safely ignore the server module from client
+   * bundles without that leading to an exception trying to import it
+   * unconditionally at the top-level.
+   */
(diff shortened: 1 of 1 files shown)


src/Artsy/Relay/createRelaySSREnvironment.ts
```diff
@@ -49,9 +49,16 @@ export interface RelaySSREnvironment extends Environment {
...
 
+  /**
+   * Lazy load these here so we can safely ignore the server module from client
+   * bundles without that leading to an exception trying to import it
+   * unconditionally at the top-level.
+   */
   const relaySSRMiddleware = isServer
-    ? new RelayServerSSR()
-    : new RelayClientSSR(cache)
+    ? new (require("react-relay-network-modern-ssr/node8/server")).default()
```
dblandin: Do we need to switch to this lazy-loading approach everywhere we reference `RelayServerSSR` in this file? I see a few other references and I wonder if they would also cause an exception within the client-side bundle.

alloy: Good question! We only need to lazy load in the places where we are using actual values, the other places in this file are type references only. After compiling this file you can see in the `dist` folder that this file has no top-level import.

dblandin: Awesome. Thanks for clarifying that!

PR #2927, opened 2019-10-18
[Carousel] Fix incorrect invocation of `Flickity.prototype.destroy()`

alloy:
https://staging.artsy.net/artist/simon-linke/cv is currently broken:

Screenshot 2019-10-18 at 16 05 17

This PR adds the typings for Flickity, but that surfaced another few places where, according to the types at least, we’re incorrectly using the API. Either the typings are incorrect and should be updated or we actually have other issues.

Further work that should be done is to use the Flickity typings in other places where Flickity API/Options are referenced.

@kierangillen Can you take this to completion?

### DIFF
--- a/src/Components/v2/Carousel.tsx
+++ b/src/Components/v2/Carousel.tsx
@@ -1,5 +1,5 @@
 import { Box, ChevronIcon, color, Flex, space } from "@artsy/palette"
-import { Options as FlickityOptions } from "flickity"
+import Flickity, { Options as FlickityOptions } from "flickity"
 import React, { Fragment } from "react"
 import styled from "styled-components"
 import { left, LeftProps, right, RightProps } from "styled-system"
@@ -167,7 +167,7 @@ export class BaseCarousel extends React.Component<
   /**
    * A reference to the Flickity instance
    */
-  flickity = null
+  flickity: Flickity = null
   carouselRef = null
 
   /**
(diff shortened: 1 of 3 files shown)


src/Components/v2/Carousel.tsx
```diff
@@ -227,7 +226,7 @@ export class BaseCarousel extends React.Component<
   componentWillUnmount() {
     if (this.flickity) {
       this.flickity.off("select")
-      this.flickity.flickity("destroy")
+      this.flickity.destroy()
```
alloy: This was the actual bug.

damassi:
@kierangillen - reminder to run code before pushing it up! 

I'll take this PR since Kieren is OOO.

src/Components/v2/Carousel.tsx
```diff
@@ -240,6 +239,9 @@ export class BaseCarousel extends React.Component<
   checkLastItemVisible = () => {
     if (this.flickity && this.flickity.selectedElements) {
       const lastItemVisible = this.flickity.selectedElements.includes(
+        // FIXME: Undocumented API. Is there a way this can be achieved with
+        // something public and commonly available?
+        // @ts-ignore
```
damassi: @kierangillen - Where did you discover `getLastCell`? It's not in the docs but I've confirmed it is indeed in the source; was it from inspecting the `flickity` instance? Is there a way to use a public API method for this, something in the docs? If not can we open a PR to flickity updating the docs and then update the corresponding type on DefinitelyTyped?

kierangillen:
@damassi thanks, I did QA it and didn’t see any issues.

artsyit:
:rocket: PR was released in v21.10.9 :rocket:

alloy:
Thanks @damassi!

PR #2928, opened 2019-10-18
Submit bid via new ConfirmBid page

bhoggard:
https://artsyproduct.atlassian.net/browse/AUCT-645

paired with @erikdstock 

Screen Shot 2019-10-22 at 12 16 20 PM

**Acceptance Criteria**

- ensure that the confirm bid page is moved to reaction
- current page: https://github.com/artsy/force/blob/3e8e7cc0b0a12a3fd39df59e5e4609b744ecd188/src/mobile/apps/feature/templates
- A developer is able to place a bid in Reaction’s storybooks
- Reaction app has tests

### DIFF
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -1,11 +1,22 @@
 import { Box, Separator, Serif } from "@artsy/palette"
+import {
+  ConfirmBidCreateBidderPositionMutation,
+  ConfirmBidCreateBidderPositionMutationResponse,
+} from "__generated__/ConfirmBidCreateBidderPositionMutation.graphql"
 import { BidFormFragmentContainer as BidForm } from "Apps/Auction/Components/BidForm"
 import { LotInfoFragmentContainer as LotInfo } from "Apps/Auction/Components/LotInfo"
 import { AppContainer } from "Apps/Components/AppContainer"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
+import { FormikActions } from "formik"
 import React from "react"
 import { Title } from "react-head"
-import { createFragmentContainer, RelayProp } from "react-relay"
+import {
+  commitMutation,
+  createFragmentContainer,
+  graphql,
+  RelayProp,
+} from "react-relay"
+import { data as sd } from "sharify"
 import createLogger from "Utils/logger"
 
 const logger = createLogger("Apps/Auction/Routes/ConfirmBid")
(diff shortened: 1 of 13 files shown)


src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+const mockLocation: any = {
+  search: "",
+}
+
+const setupTestEnv = () => {
+  return createTestEnv({
+    TestPage: ConfirmBidTestPage,
+    Component: (props: routes_BidQueryResponse) => (
+      <ConfirmBidRouteFragmentContainer location={mockLocation} {...props} />
+    ),
```
erikdstock: This little bit of extra prop injection is necessary because of the way that I included a [location query string toggle for a pricing transparency preview](https://github.com/artsy/reaction/pull/2918/files#diff-732fcbc97d5a2306a5a052ccf1852df4R40). This addition was only intended to clear a path for developers to work on the pricing transparency component + queries, so now it feels a bit contrived to pass an extra `Location` object (as a prop from the router, not `window`- we do use `window.location.assign()` elsewhere currently).

Maybe acceptable in the short term but kind of a pain.

bhoggard:
Our storybooks connects to staging (which is still down), so I was unable to get a screenshot this time.

bhoggard:
It should still look like the screenshots in @erikdstock's original PR:

https://github.com/artsy/reaction/pull/2918

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -28,6 +39,80 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
...
+    if (Array.isArray(error)) {
+      errorMessages = error.map(e => e.message)
+    } else if (typeof error === "string") {
+      errorMessages = [error]
+    } else if (error.message) {
+      errorMessages = [error.message]
+    }
+
+    // TODO: add tracking with errorMessages
+    logger.error(errorMessages) // remove once tracking is implemented
```
yuki24: We are calling `logger.error` twice and I would remove the second one since `errorMessages` is always a string whereas `error` could be an error object that has more information such as backtrace.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -28,6 +39,80 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
...
+    logger.error(error)
+
+    let errorMessages: string[]
+    if (Array.isArray(error)) {
+      errorMessages = error.map(e => e.message)
+    } else if (typeof error === "string") {
+      errorMessages = [error]
+    } else if (error.message) {
+      errorMessages = [error.message]
+    }
```
yuki24: I wonder if we should be this defensive about the type of `error` here. It's only used in one place and Promise's `catch` should only inject an `Error` object.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
bhoggard: The 2nd one will get removed and replaced with a call to tracking as soon as this is merged and I can work on [that ticket](https://artsyproduct.atlassian.net/browse/AUCT-667). I only added it to prevent a compile error from an unused variable.

erikdstock:
I realized that there is one feature not tracked elsewhere which we missed here- the query string should allow choosing a different starting price. This would be pretty simple to add and test for later but we should decide/ticket if we aren't going to do it in this one.

artsyit:
:rocket: PR was released in v21.13.0 :rocket:

PR #2931, opened 2019-10-18
Return default object on server from useWindowSize, instead of null

pepopowitz:
This should resolve the "Cannot read property 'width' of null" errors currently in staging, which were a result of returning `null` on the server in #2919.


artsyit:
:rocket: PR was released in v21.10.10 :rocket:

PR #2932, opened 2019-10-18
[GROW-1598] Excluded featured artist ids

xtina-starr:
Addresses: [GROW-1598](https://artsyproduct.atlassian.net/browse/GROW-1598)

Collaborated with @pepopowitz 

Currently, our featured artists on hub collections are based on an automated gene, which is based on the creation years of artworks. This causes strange things like modern artists appearing as featured artists on the Post-War hub. As a work around, we've added a new `featuredArtistExclusionIds` property on the Collection model that allows the marketing team to enter in artist ids that should be excluded from the FeaturedArtist module. The `featuredArtistExclusionIds` get filtered out.

**Before:** _Miro and Picasso are included and should't be_
Screen Shot 2019-10-18 at 5 31 13 PM

Post-War now includes Miro and Picasso in the `featuredArtistExclusionIds` and are filtered out.
Screen Shot 2019-10-18 at 5 35 45 PM

**After:**
Screen Shot 2019-10-18 at 5 30 35 PM


xtina-starr:
@pepopowitz as a part of this PR I'd like to increase the number of artworks we initially grab in the [artworks query](https://github.com/artsy/reaction/blob/875377b6ad1d6a738b9c5037c8249ea9f0340506/src/Apps/Collect2/Routes/Collection/index.tsx#L233) from 12 to 20. My thinking is that if we are filtering out the `featuredArtistExclusionIds` we'd want to increase the number we initially pull. Thoughts?

xtina-starr:
Yeah I think we should promote metaphysics first.

artsyit:
:rocket: PR was released in v21.11.1 :rocket:

PR #2933, opened 2019-10-18
[Carousel] Ensure we only import flickity on client

damassi:
Fixes `window is not defined` issue when SSR'ing in Force, which reverts [this line from #2927](https://github.com/artsy/reaction/pull/2927/files#diff-46847d5408e9ceb7bf8d5b5f1c59f0b5L208)

#trivial


artsyit:
:rocket: PR was released in v21.10.11 :rocket:

src/Components/v2/Carousel.tsx
```diff
@@ -206,6 +206,7 @@ export class BaseCarousel extends React.Component<
    */
```
alloy: This comment should be improved to say that the lib cannot even be loaded in environments that lack a global `window` object.

PR #2934, opened 2019-10-21
[Carousel] Type `new Flickity(…)` and document SSR import issue.

alloy:
Follow-up to #2933

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
@@ -62,7 +62,12 @@ export class LargeArtworkImageBrowser extends React.Component<
             <Col sm={1}>
               <ArrowButton
                 direction="left"
-                onClick={() => flickity.previous(false, true)}
+                onClick={() => {
+                  // FIXME: Flickity.prototype.previous typing is missing second
+                  // `isInstant` method.
+                  // @ts-ignore
+                  flickity.previous(false, true)
+                }}
               />
             </Col>
           )}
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -62,15 +62,25 @@ export class LargeArtworkImageBrowser extends React.Component<
...
           renderRightArrow={({ flickity }) => (
             <Col sm={1}>
               <ArrowButton
                 direction="right"
-                onClick={() => flickity.next(false, true)}
+                onClick={() => {
+                  // FIXME: Flickity.prototype.next typing is missing second
+                  // `isInstant` method.
+                  // @ts-ignore
+                  flickity.next(false, true)
```
alloy: These two ignores really need to be fixed in the DT types.

src/Components/v2/Carousel.tsx
```diff
@@ -1,10 +1,17 @@
...
 import { left, LeftProps, right, RightProps } from "styled-system"
 import { Media } from "Utils/Responsive"
 
+/**
+ * Do NOT use `FlickityType` to create instances of Flickity. Only refer to it
+ * for type purposes.
+ *
+ * @see BaseCarousel.prototype.componentDidMount()
+ */
+import FlickityType, { Options as FlickityOptions } from "flickity"
```
alloy: This will not emit a top-level `require` _as long_ as it is not referred to as a runtime value.

src/Components/v2/Carousel.tsx
```diff
@@ -202,12 +206,16 @@ export class BaseCarousel extends React.Component<
...
+   *
+   * In fact, the flickity library can't even be loaded in environments that do
+   * not have a global `window` object. Thus we need to lazy load it here with
+   * an inline `require` statement.
    */
   componentDidMount() {
     const { setCarouselRef } = this.props
-    const Flickity = require("flickity")
 
+    const Flickity = require("flickity") as typeof FlickityType
```
alloy: Because using `require("Flickity")` doesn't provide the class typings, we cast it here instead.

artsyit:
:rocket: PR was released in v21.11.0 :rocket:

PR #2936, opened 2019-10-21
Require login for bid confirm route

bhoggard:
https://artsyproduct.atlassian.net/browse/AUCT-685

paired with @erikdstock

### DIFF
--- a/src/Apps/Auction/__tests__/routes.test.ts
+++ b/src/Apps/Auction/__tests__/routes.test.ts
@@ -122,7 +122,7 @@ describe("Auction/routes", () => {
 
     it("redirects to the login (plus redirect_uri of sale artwork bid page) if user is not signed in", async () => {
       const fixture: BidQueryResponse = mockConfirmBidResolver()
-      ;(fixture as any).me = null
+      fixture.me = null
 
       const { redirect } = await render(
         `/auction/${fixture.artwork.saleArtwork.sale.id}/bid/${
--- a/src/Apps/Auction/__tests__/routes.test.ts
+++ b/src/Apps/Auction/__tests__/routes.test.ts
@@ -120,6 +120,21 @@ describe("Auction/routes", () => {
       )
     })
 
+    it("redirects to the login (plus redirect_uri of sale artwork bid page) if user is not signed in", async () => {
+      const fixture: BidQueryResponse = mockConfirmBidResolver()
+      ;(fixture as any).me = null
+
+      const { redirect } = await render(
+        `/auction/${fixture.artwork.saleArtwork.sale.id}/bid/${
+          fixture.artwork.id
+        }`,
+        fixture
+      )
+      expect(redirect.url).toBe(
+        "/log_in?redirect_uri=%2Fauction%2Fsaleslug%2Fbid%2Fartworkslug"
+      )
+    })
+
     it("redirects to the sale artwork page if user is not registered and registration is closed", async () => {
       const fixture: BidQueryResponse = mockConfirmBidResolver({
         sale: {
--- a/src/Apps/Auction/getRedirect.ts
+++ b/src/Apps/Auction/getRedirect.ts
@@ -6,10 +6,10 @@ export interface Redirect {
   reason: string
 }
 
-export function registerRedirect(
-  sale: routes_RegisterQueryResponse["sale"],
-  me: routes_RegisterQueryResponse["me"]
-): Redirect | null {
+export function registerRedirect({
+  me,
+  sale,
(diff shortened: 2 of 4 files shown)


src/Apps/Auction/getRedirect.ts
```diff
@@ -6,10 +6,10 @@ export interface Redirect {
   reason: string
 }
 
-export function registerRedirect(
-  sale: routes_RegisterQueryResponse["sale"],
-  me: routes_RegisterQueryResponse["me"]
-): Redirect | null {
+export function registerRedirect({
```
bhoggard: Switched to types rather than positional arguments for 2 functions.

src/Apps/Auction/getRedirect.ts
```diff
@@ -35,14 +35,22 @@ export function registerRedirect(
   return null
 }
 
-export function confirmBidRedirect(
-  artwork: routes_BidQueryResponse["artwork"],
-  me: routes_BidQueryResponse["me"]
-): Redirect | null {
+export function confirmBidRedirect({
```
bhoggard: Should we pass in location here, and redirect to the current URL, rather than the `bidPath`?

dblandin: I would love for us to adopt a redirect structure similar to what's in the OrderApp today: https://github.com/artsy/reaction/blob/master/src/Apps/Order/redirects.tsx#L163-L174

I love the path matching and explicitly named redirect rules. :heart: 

This certainly looks clear enough for now though!

src/Apps/Auction/__tests__/routes.test.ts
```diff
@@ -120,6 +120,21 @@ describe("Auction/routes", () => {
       )
     })
 
+    it("redirects to the login (plus redirect_uri of sale artwork bid page) if user is not signed in", async () => {
+      const fixture: BidQueryResponse = mockConfirmBidResolver()
+      ;(fixture as any).me = null
```
dleve123: instead of casting to `any`, thoughts on editing the type of the `BidQueryResponse` to permit a `null` at the `me` attribute? I think we can use this as an opportunity to make our test data more realistic without breaking type safety.

src/Apps/Auction/getRedirect.ts
dleve123: huh - didn't know you can do type destructing this way, very cool

src/Apps/Auction/getRedirect.ts
dleve123: > Should we pass in location here, and redirect to the current URL, rather than the bidPath?

To clarify: and redirect to the log_in page with the `redirect_uri` being equal to the current path? If so, then yes, I think that's more robust.

src/Apps/Auction/__tests__/routes.test.ts
dleve123: Pardon the nooby question, what's the deal with the  `;(...)` syntax? I have been struggling to google around for it.

dleve123:
@bhoggard and @erikdstock plz review a438ffbe6fa8e462ef341f6bcc518248e27e047d – if y'all feel comfortable with it, then this is good to merge in my book pending we resolve the question about whether bid confirmation is accessible via multiple entry-points.

src/Apps/Auction/__tests__/routes.test.ts
bhoggard: Without it, prettier gets confused and tries to join it with the line above.

erikdstock:
I think using the actual location is more robust- otherwise we may end up getting into some weird cases of needing to reconstruct larger and larger urls. @bhoggard , I think we can do that by passing the `location` prop from the `routes.tsx` file in and finding whatever property will give us the correct redirect url - probably `location.href`?

On the types part @dleve123, was the build failing due to the use of `any`? This is a cool solution but i feel like it is causing us to drift further from real-world values in our tests and takes a little longer to grok. No strong opinion either way otherwise.

dleve123:
> On the types part @dleve123, was the build failing due to the use of any? This is a cool solution but i feel like it is causing us to drift further from real-world values in our tests and takes a little longer to grok. No strong opinion either way otherwise.

Nope, nothing was failing, but was just trying to avoid using an `any` as we loose type safety as soon as we do (if I understand correctly).

bhoggard:
I've never seen this approach before with a type, so I can't really provide an opinion. I will update the code to use the request URL.

artsyit:
:rocket: PR was released in v21.12.0 :rocket:

PR #2938, opened 2019-10-21
Loosen rules on phone numbers in AddressForm for international numbers.

dleve123:
Problem:

It was discovered that users weren't able to enter country codes or
white-space when providing a phone number when registering for an
auction. Moreover, due to the way that HTML5 validations work, users were
getting generic "Please match the requested format" messages, which is a
confusing UX.

Solution:

* Change the `pattern` attribute for the phone `Input` so that only
letters are not allowed in client side validations.
  + This is in-sufficient to ensure that all phone numbers are
    deliverable, but notably there are multiple ways to specify a phone
    number on Artsy, including the settings page on www.artsy.net, so
    additional constraints should likely be implemented at the API layer.
* Set the input type to `tel` to make the DOM as semantic and HTML5 nice
* Add a placeholder that includes a country code and will pass client
side validations.

Jira 🔒: https://artsyproduct.atlassian.net/browse/AUCT-687

---

new-empty-state-for-phone-number
new-phone-number-validation

### DIFF
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -212,7 +212,7 @@ export class AddressForm extends React.Component<
                 type="tel"
                 description={this.phoneNumberInputDescription()}
                 placeholder="+1-222-222-2222"
-                pattern="[^a-z]"
+                pattern="[^a-z]+"
                 value={this.props.value.phoneNumber}
                 onChange={this.changeEventHandler("phoneNumber")}
                 error={this.getError("phoneNumber")}
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -209,9 +209,10 @@ export class AddressForm extends React.Component<
               <Input
                 id="AddressForm_phoneNumber"
                 title="Phone number"
+                type="tel"
                 description={this.phoneNumberInputDescription()}
-                placeholder="Add phone"
-                pattern="[0-9]*"
+                placeholder="+1-222-222-2222"
+                pattern="[^a-z]"
                 value={this.props.value.phoneNumber}
                 onChange={this.changeEventHandler("phoneNumber")}
                 error={this.getError("phoneNumber")}
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -211,7 +211,7 @@ export class AddressForm extends React.Component<
                 title="Phone number"
                 type="tel"
                 description={this.phoneNumberInputDescription()}
-                placeholder="+1-222-222-2222"
+                placeholder="Add phone"
                 pattern="[^a-z]+"
                 value={this.props.value.phoneNumber}
                 onChange={this.changeEventHandler("phoneNumber")}
--- a/src/Apps/Auction/Routes/__tests__/Register.test.tsx
+++ b/src/Apps/Auction/Routes/__tests__/Register.test.tsx
@@ -91,7 +91,7 @@ describe("Routes/Register ", () => {
     jest.resetAllMocks()
   })
 
-  it("emits a RegistrationSubmitFailed analtyics event and halts submission", async () => {
+  it("emits a RegistrationSubmitFailed analytics event and halts submission", async () => {
     const env = setupTestEnv()
     const page = await env.buildPage()
 


dleve123:
Curious if there's a good pattern for us to test client-side validations like this? I attempted to extend https://github.com/artsy/reaction/blob/79194751b3a3bd20aa931890b11c64b3fee69220/src/Apps/Auction/Routes/__tests__/Register.test.tsx#L190-L204 with a relevant test case, but it didn't fail without the code fix, as I believe this test file doesn't run a browser-based test. Is there an approach we have here?

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -209,9 +209,10 @@ export class AddressForm extends React.Component<
               <Input
                 id="AddressForm_phoneNumber"
                 title="Phone number"
+                type="tel"
                 description={this.phoneNumberInputDescription()}
-                placeholder="Add phone"
-                pattern="[0-9]*"
+                placeholder="+1-222-222-2222"
+                pattern="[^a-z]+"
```
dblandin: This looks like a good iteration!

Agreed that we should ask about potential validations at the API level. I also wonder if we can add some more structure around country codes. How do other sites support international phone numbers? Can we use a react component that nicely masks the input while sending up clean, structured values?

dblandin:
> Curious if there's a good pattern for us to test client-side validations like this? I attempted to extend [code] with a relevant test case, but it didn't fail without the code fix, as I believe this test file doesn't run a browser-based test. Is there an approach we have here?

I think the answer here might be our Cypress-backed integration framework: https://github.com/artsy/integrity :lock: but we could look into other headless-browser approaches.

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -209,9 +209,10 @@ export class AddressForm extends React.Component<
               <Input
                 id="AddressForm_phoneNumber"
                 title="Phone number"
+                type="tel"
                 description={this.phoneNumberInputDescription()}
-                placeholder="Add phone"
-                pattern="[0-9]*"
+                placeholder="+1-222-222-2222"
```
sweir27: Is this desired from a design perspective? It's quite different from our other placeholders.

src/Apps/Order/Components/AddressForm.tsx
dleve123: >  How do other sites support international phone numbers? Can we use a react component that nicely masks the input while sending up clean, structured values?

While doing some research, I did find some elegant looking ways to structure international phone number input, but want to focus on the acute problem here for now

dleve123:
@ashfurrow 

Re https://github.com/artsy/reaction/pull/2938#pullrequestreview-305207318

Here's are some videos of my local storybooks server connected to my iPhone via my local network.

1. Not entering any phone number

validation-gif

2. Entering a valid phone number **(note that we are able to select the + button)**
phone-entry-gif

---

> my only concern is making sure that the phone number field still brings up a phone number bad-like UI on iOS.

Do we specifically want this "bad" UI? Is there anything wrong about exposing some more "phone-y" characters?

src/Apps/Order/Components/AddressForm.tsx
dleve123: I just choose this a bit randomly. I didn't see anything in palette here, but happy to replace with anything more consistent. Could you point me in a direction if you have an idea in mind here?

sweir27: I believe `Add phone` (what was there) was intentional (and matches the other placeholders).

I recommend asking the design team if you'd like to change!

ashfurrow:
@dleve123 what you posted looks _great_, I made a typo:

```diff
- my only concern is making sure that the phone number field still brings up a phone number bad-like UI on iOS.
+ my only concern is making sure that the phone number field still brings up a phone number pad-like UI on iOS.
```

Looks great to me!

src/Apps/Order/Components/AddressForm.tsx
dleve123: Gotcha – I guess I presumed that this wasn't a deliberate decision, but good to know that it was. I'll revert the placeholder change as that isn't the thrust of this PR and don't want to slow down the fix.

dleve123:
Re-assigning to myself so that we squash and merge this, but @sweir27 would like your sign-off before merging.

dleve123:
@sweir27 followed up with me on Slack approving from her end. Merging.

sweir27:
Looks good!

artsyit:
:rocket: PR was released in v21.14.0 :rocket:

src/Apps/Order/Components/AddressForm.tsx
dblandin: FYI, I brought this up at the Design Systems meeting today and we're going to revisit the placeholder copy which I'm hoping will either clarify the intent or open a discussion towards alternatives: https://artsyproduct.atlassian.net/browse/DSG-15 :lock:

dleve123: Very cool, thanks for doing that!

PR #2941, opened 2019-10-22
this hopefully fixes the broken login flow on mobile

mbilokonsky:
There was a place where we were expecting an object to exist but it didn't. I've added a guardrail check and a unit test to cover this edge case, I think we should be good now?

Diff Analysis: 
```
{
  "total_files_changed": 2,
  "test_files_changed": 1,
  "by_type": {
    "ts": 2
  }
}
```


sweir27:
Not sure how to remove that peril warning, but this is effectively 🍏 so I'll merge.

artsyit:
:rocket: PR was released in v21.12.2 :rocket:

PR #2944, opened 2019-10-23
Create typeSupport & simplify relay fixture creation

erikdstock:
This PR makes it easier to define fixtures in the Auction app by simplifying our ConfirmBid mock query resolver code to use deep merge. In order to do that we brought in a mapped type from the [`ts-essentials`](https://github.com/krzkaczor/ts-essentials/commit/9536fd57a0c89ba358a16dd657054cf7960bdb59#diff-f0110f871bbb78393553f89381b8a4fdR9) package (we did not add it to node_modules because of its educational value).

Extra context: This allows us to define `{me: null}` when calling the fixture factory function rather than saying `fixture.me = null` (see https://github.com/artsy/reaction/commit/a438ffbe6fa8e462ef341f6bcc518248e27e047d) after creation, which in turn allows us to _not_ make the fixture a special `DeepWriteable` type to avoid `readonly` errors. It also drastically simplifies some nested fixture merging code that all really just amounted to a deep merge.

Refactor auction tests to use deepmerge in fixture creation
- Create Utils/typeSupport for colocating utility/mapped types
- remove DeepWriteable<T> type (not needed after switching to deepmerge)
- copy DeepPartial from krzkaczor/ts-essentials
- remove redundant `type Omit`s as they are now included in our ts version

Paired on this with @dleve123

### DIFF
--- a/src/Apps/Auction/__tests__/routes.test.ts
+++ b/src/Apps/Auction/__tests__/routes.test.ts
@@ -7,11 +7,13 @@ import {
   RegisterQueryResponseFixture,
 } from "Apps/Auction/__fixtures__/routes_RegisterQuery"
 import { routes } from "Apps/Auction/routes"
+import deepMerge from "deepmerge"
 import { createMockNetworkLayer2 } from "DevTools/createMockNetworkLayer"
 import { createRender } from "found"
 import { Resolver } from "found-relay"
 import getFarceResult from "found/lib/server/getFarceResult"
 import { Environment, RecordSource, Store } from "relay-runtime"
+import { DeepPartial } from "Utils/typeSupport"
 
 describe("Auction/routes", () => {
   async function render(url, mockData) {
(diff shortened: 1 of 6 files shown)


src/Components/v2/__tests__/ArtistBio.test.tsx
```diff
@@ -10,8 +10,6 @@ import {
 
 jest.unmock("react-relay")
 
-type Omit<T, K> = { [k in Exclude<keyof T, K>]: T[k] }
```
zephraph: 🙌 thanks for cleaning these up

src/Apps/Auction/__tests__/routes.test.ts
```diff
@@ -35,35 +37,13 @@ describe("Auction/routes", () => {
...
-      ...me,
-    },
-  })
+  const mockConfirmBidResolver = (
+    data: DeepPartial<ConfirmBidQueryResponse> = {}
+  ): ConfirmBidQueryResponse =>
+    deepMerge<ConfirmBidQueryResponse, DeepPartial<ConfirmBidQueryResponse>>(
+      ConfirmBidQueryResponseFixture,
+      data
+    )
```
alloy: Yup, nice 👌

src/Utils/typeSupport.ts
```diff
@@ -0,0 +1,35 @@
+/** Remove the special relay properties from an interface */
+export type DeFraged<T> = Omit<T, " $refType" | " $fragmentRefs">
```
zephraph: I was thinking we needed a centralized way to do this. I was manually doing this https://github.com/artsy/reaction/blob/7b712fc8390f0aca42f3ac53f292a4a6a753db20/src/Apps/Artwork/Components/TrustSignals/__tests__/SecurePayment.test.tsx#L11

It might be worth doing a search for `" $refType"` and updating any tests/examples that use this. (That could be a follow up)

artsyit:
:rocket: PR was released in v21.15.0 :rocket:

PR #2946, opened 2019-10-23
[Router] Remove double render

damassi:
This removes the double render in our server-side render pass.


zephraph:
Should this be a `patch` or a `minor` 🤔. Generally it'd be a `minor`, but we're restoring functionality that was there before which feels like a `patch`... 🤷‍♂

alloy:
Minor is for new features, so this should just be a patch.

artsyit:
:rocket: PR was released in v21.14.2 :rocket:

PR #2947, opened 2019-10-23
Pre-populate the selected bid with a value from the query string

erikdstock:
This PR adds the ability to pre-select a bid in the form using the `?bid=xxxx` query parameter. It follows the current behavior of falling back to the first value if a matching increment is not available in the select input. Has tests, WIP because I need to update storybooks.

TODO: 
- [x] Update storybooks and create some stories for the bidform not tied to relay or staging data.


src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -11,14 +11,15 @@ import { BidForm_saleArtwork } from "__generated__/BidForm_saleArtwork.graphql"
 import { PricingTransparency } from "Apps/Auction/Components/PricingTransparency"
 import { ConditionsOfSaleCheckbox } from "Components/Auction/ConditionsOfSaleCheckbox"
 import { Form, Formik, FormikActions, FormikValues } from "formik"
-import { dropWhile } from "lodash"
+import { dropWhile, findLast } from "lodash"
```
erikdstock: should probably change this to `find`, was previously selecting the last increment below the amount so 501 would fall back to the 500 increment.

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -35,19 +36,42 @@ const validationSchema = Yup.object().shape({
...
+}): string => {
+  if (!initialSelectedBid) {
+    return displayIncrements[0].value
+  } else {
+    const selectedNum = Number(initialSelectedBid)
+    const lastGoodIncrement = findLast(
+      displayIncrements,
+      i => Number(i.value) === selectedNum
+    )
+    return (lastGoodIncrement || displayIncrements[0]).value
```
damassi: This is #superminor DX thing but it's often a good habit to assign the return value to a variable (and return that), so that the variable can be logged later on down the line by future devs or a breakpoint inserted inbetween.

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -17,15 +21,26 @@ jest.mock("sharify", () => ({
...
+
+const setupTestEnv = ({
+  location = mockLocation,
+  data = {},
+}: { location?: Partial<Location>; data?: any } = {}) => {
   return createTestEnv({
     TestPage: ConfirmBidTestPage,
     Component: (props: routes_ConfirmBidQueryResponse) => (
-      <ConfirmBidRouteFragmentContainer location={mockLocation} {...props} />
+      <ConfirmBidRouteFragmentContainer location={location as any} {...props} />
```
bhoggard: Minor, but does this have to be `any` instead of a `Location`, as we did in the require login PR? 

https://github.com/artsy/reaction/pull/2936/files#diff-42d3becca5be832d10c7b1e055bef46dR40

erikdstock: I could do that and was going to just now, but on second look my argument against is is that this isn't a _real_ location object, it only has a `search` value for the query string. I think the `any` does a better job communicating to the dev that this is not a `Location` - it will jump out at you when a test mysteriously breaks later - while still satisfying the type checker.

erikdstock: Update: after talking with folks about Ts and testing philosophies I adopted your change. Thanks for raising!

src/Apps/Auction/Components/__stories__/LotInfo.story.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Section>
+      <LotInfo
+        relay={null as any}
+        artwork={{
+          " $refType": null as any,
+          _id: "artworkid",
+          date: "1938/1939",
+          imageUrl:
+            "https://d32dm0rphc51dk.cloudfront.net/klloxkFxMYsIt2d36Oviog/square.jpg",
```
dblandin: :peach:

erikdstock:
#mergeongreen

erikdstock:
Sorry- I accidentally added the merge on green label, thought i removed it (by deleting the commend quickly) but peril must have still caught it.

artsyit:
:rocket: PR was released in v21.16.2 :rocket:

PR #2949, opened 2019-10-24
Grow 1563 Add analytics tracking to Nav Bar

jiahaodavid:
Address: [Grow 1563](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1563)

This PR is going to add tracking "Hover" action to our code. It already tracked the "Click" action. So, I set up a condition when `haver && doHoverTracking` is true, then they will do the hover tracking.

### DIFF
--- a/src/Components/NavBar/NavBar.tsx
+++ b/src/Components/NavBar/NavBar.tsx
@@ -83,13 +83,13 @@ export const NavBar: React.FC = track(
         */}
         <NavSection display={["none", "none", "flex"]}>
           <NavSection>
-            <NavItem href="/collect" callHoverTracking>
+            <NavItem href="/collect" doHoverTracking>
               Artworks
             </NavItem>
-            <NavItem href="/auctions" callHoverTracking>
+            <NavItem href="/auctions" doHoverTracking>
               Auctions
             </NavItem>
-            <NavItem href="/galleries" callHoverTracking>
+            <NavItem href="/galleries" doHoverTracking>
               Galleries
             </NavItem>
 
@@ -99,11 +99,11 @@ export const NavBar: React.FC = track(
             <NavItem
               href="/art-fairs"
               display={["none", "none", "none", "none", "block"]}
-              callHoverTracking
+              doHoverTracking
             >
               Fairs
             </NavItem>
-            <NavItem href="/articles" callHoverTracking>
+            <NavItem href="/articles" doHoverTracking>
               Magazine
             </NavItem>
             <NavItem
@@ -114,7 +114,7 @@ export const NavBar: React.FC = track(
                   </Box>
                 )
               }}
-              callHoverTracking
+              doHoverTracking
             >
               More
             </NavItem>
--- a/src/Components/NavBar/NavBar.tsx
+++ b/src/Components/NavBar/NavBar.tsx
@@ -83,9 +83,15 @@ export const NavBar: React.FC = track(
         */}
         <NavSection display={["none", "none", "flex"]}>
           <NavSection>
-            <NavItem href="/collect">Artworks</NavItem>
-            <NavItem href="/auctions">Auctions</NavItem>
-            <NavItem href="/galleries">Galleries</NavItem>
+            <NavItem href="/collect" callHoverTracking>
+              Artworks
+            </NavItem>
+            <NavItem href="/auctions" callHoverTracking>
+              Auctions
+            </NavItem>
+            <NavItem href="/galleries" callHoverTracking>
+              Galleries
+            </NavItem>
 
             {/**
               Only show Fairs at `xlg`
@@ -93,10 +99,13 @@ export const NavBar: React.FC = track(
             <NavItem
               href="/art-fairs"
               display={["none", "none", "none", "none", "block"]}
+              callHoverTracking
             >
               Fairs
             </NavItem>
-            <NavItem href="/articles">Magazine</NavItem>
+            <NavItem href="/articles" callHoverTracking>
+              Magazine
+            </NavItem>
             <NavItem
               Menu={() => {
                 return (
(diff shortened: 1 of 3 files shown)


src/Components/NavBar/NavBar.tsx
```diff
@@ -83,20 +83,29 @@ export const NavBar: React.FC = track(
         */}
         <NavSection display={["none", "none", "flex"]}>
           <NavSection>
-            <NavItem href="/collect">Artworks</NavItem>
-            <NavItem href="/auctions">Auctions</NavItem>
-            <NavItem href="/galleries">Galleries</NavItem>
+            <NavItem href="/collect" doHoverTracking>
```
damassi: Why add the `doHoverTracking` prop if all nav items are tracked on hover?

jiahaodavid: @damassi 
Yes. Because there are somethings also using `<NavItem>`. Such as, "works for you" and some small view. For "work for you", it already has a hover tracking. I think only adding `doHoverTracking` props to those items will minimus the changing of the code. What do you think?

damassi: I would suggest creating a new prop `onHover` that can be passed into `NavItem`, and for  that single nav item you mention that already tracks we can place the hover tracking in that callback function. Inside of NavItem, if that `onHover` prop has been passed, guard against the default hover tracking (default = what happens `doHoverTracking` is passed, at which point we can remove the `doHoverTracking` prop completely).

Basically, if most of  the items now require hover tracking, then a prop shouldn't have  to be passed because one nav item already has it. Things can be inverted. Revise the item that already has hover tracking and make it more suitable for the many. 

Does this make sense? Down to pair if needed!

jiahaodavid: @damassi 
Hi, Chris. I made some updates. Take a look when you have time.

src/Components/NavBar/NavBar.tsx
```diff
@@ -116,6 +116,7 @@ export const NavBar: React.FC = track(
             {isLoggedIn && (
               <>
                 <NavItem
+                  onHover={false}
```
damassi: One should be able to exclude this prop since `false` and `undefined` (missing the prop) both evaluate to false in the end.

src/Components/NavBar/NavItem.tsx
```diff
@@ -16,6 +16,7 @@ interface NavItemProps extends BoxProps {
   className?: string
   href?: string
   onClick?: () => void
+  onHover?: boolean
```
damassi: Props with the `onSomething` form are always callback functions since the `onSomeEvent` form implies some action is happening, like a hover or a mouse click (etc).

src/Components/NavBar/NavItem.tsx
```diff
@@ -27,6 +28,7 @@ export const NavItem: React.FC<NavItemProps> = ({
   display = "block",
   href,
   onClick,
+  onHover = true,
```
damassi: This isn't needed

src/Components/NavBar/NavItem.tsx
```diff
@@ -94,6 +104,8 @@ export const NavItem: React.FC<NavItemProps> = ({
         </Sans>
       </Link>
 
+      {hover && onHover ? trackHover() : null}
```
damassi: This can [be an effect](https://reactjs.org/docs/hooks-effect.html), and we can call the tracking function when `hover` is true.

```tsx
useEffect(() => {
  if (hover) {
    trackHover()
  }
}, [hover])
```

src/Components/NavBar/NavBar.tsx
```diff
@@ -147,6 +147,12 @@ export const NavBar: React.FC = track(
                 </NavItem>
                 <NavItem Menu={UserMenu}>
                   {({ hover }) => {
+                    if (hover) {
+                      trackEvent({
+                        action_type: AnalyticsSchema.ActionType.Hover,
+                        subject: "User",
+                      })
+                    }
```
damassi: 👍

src/Components/NavBar/NavItem.tsx
```diff
@@ -59,6 +61,14 @@ export const NavItem: React.FC<NavItemProps> = ({
     }
   }
 
+  const trackHover = () => {
+    if (isString(children))
```
damassi: Can we add a comment above this line mentioning that `children` refers text that's passed into `<NavItem>`.

src/Components/NavBar/NavItem.tsx
damassi: Another thought is functions shouldn't be called like this inside of JSX, unless the function returns more JSX. Why call in here instead of around line 72?

jiahaodavid:
@damassi 
I also made some updates. Thanks for the feedback. 
https://github.com/artsy/reaction/blob/eadc3b68e36998aa5f325ba38cb0ed97b6b326c6/src/Components/NavBar/NavItem.tsx#L93-L95
I have a question. It is not related to this PR, but just for curiosity.
Why do you use `onClick && onClick()` here?

damassi:
@jiahaodavid - this is shorthand for 

```jsx
if (onClick) {
  onClick()
}
```
"if x is true, then y" -- same kind of thing as 
```jsx
const isNiceDay = sunny && checkIfRaining()
```

However, I don't recommend (and typically never write) shorthand if statements and not sure why that is there! They can be buggy, sacrafice readability for "clever" shorthand, and also come with some edge-cases. Always use an `if`!

artsyit:
:rocket: PR was released in v21.17.2 :rocket:

PR #2950, opened 2019-10-24
Ensure makesOffer isn't null before getting length

zephraph:
Fix a minor error discovered in force acceptance tests where `makesOffer` could  be null if no artworks are contained in the payload. 

https://circleci.com/gh/artsy/force/33987


artsyit:
:rocket: PR was released in v21.16.3 :rocket:

PR #2951, opened 2019-10-24
[PURCHASE-1578]  Adds `<Tags>` component for genes

sepans:
Fixes [PURCHASE-1578](https://artsyproduct.atlassian.net/browse/PURCHASE-1578)

[Design](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5d93c61cbd37a5180747902b)

Screen Shot 2019-10-24 at 4 40 03 PM

Requires follow up:
Noticed the slide down requires an animation. need to follow up on that in Palette but it would not affect reaction code.

### DIFF
--- a/src/Components/NavBar/NavItem.tsx
+++ b/src/Components/NavBar/NavItem.tsx
@@ -5,7 +5,6 @@ import { isFunction, isString } from "lodash"
 import React, { useState } from "react"
 import { animated, config, useSpring } from "react-spring"
 import styled from "styled-components"
-import { NavBarHeight } from "./NavBar"
 
 interface NavItemProps extends BoxProps {
   Menu?: React.FC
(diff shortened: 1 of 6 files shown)


sweir27:
This looks so good!!!

src/Components/NavBar/NavItem.tsx
```diff
@@ -109,7 +108,6 @@ export const NavItem: React.FC<NavItemProps> = ({
 
 const MenuContainer = styled(Box)`
   position: absolute;
-  top: ${NavBarHeight}px;
```
sepans: `top` was already being set by https://github.com/artsy/reaction/pull/2951/commits/38c58cf8335e1606a1b1babd07e758bfdfc6e998#diff-298555bc0dcd6937d9db0a76325953adR98 (`<MenuContainer top={space(6)}>`) and checked it in story book and confirmed that the correct `top` is `space(6)` and not `NavBarHeight` which is `space(6) - 1`:

Screen Shot 2019-11-01 at 1 19 24 PM

artsyit:
:rocket: PR was released in v21.17.7 :rocket:

PR #2952, opened 2019-10-25
Be defensive about list price accessors

zephraph:
cc @sepans @joeyAghion

We're seeing an increased rate of errors in production because not all of the `listPrice` ranges have a min or max value? Not sure why that would happen, but this guard should put a band-aid on the issue until we can dig deeper.


artsyit:
:rocket: PR was released in v21.16.4 :rocket:

mzikherman:
Yea, I believe some artworks can have a `nil` on either the lower or upper bound. So excluding those sorts of works entirely from the structured data seems 👍, esp. if it's not obvious how to represent those yet.

PR #2953, opened 2019-10-25
remove pricing transparency query param, make a story for it

erikdstock:
This PR creates a rudimentary story for the pricing transparency component stub and removes the query parameter that can trigger it so that it won't accidentally be visible (in theory) in force's `bid2` route.

### DIFF
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -126,7 +126,7 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
         <Separator />
         <BidForm
           initialSelectedBid={getInitialSelectedBid(props.location)}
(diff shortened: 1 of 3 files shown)


src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -126,7 +126,7 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
         <Separator />
         <BidForm
           initialSelectedBid={getInitialSelectedBid(props.location)}
-          showPricingTransparency={Boolean(/pt=1/.test(props.location.search))}
+          showPricingTransparency={false}
```
dblandin: Looks like this is the default value but I imagine you're wanting to be explicit here?

erikdstock: Yeah, for now that was my thought.

artsyit:
:rocket: PR was released in v21.16.6 :rocket:

PR #2954, opened 2019-10-25
[Collection Hubs Entry Points] Moves Experiment Viewed to collect app

xtina-starr:
This PR moves the 'Experiment Viewed' event to the base Collect component to ensure we fire it for _both_ the control and the experiment  group.

For more context see: https://artsy.slack.com/archives/C9SATFLUU/p1572022915428900?thread_ts=1571949171.406700&cid=C9SATFLUU

### DIFF
--- a/src/Components/v2/CollectionsHubsNav.tsx
+++ b/src/Components/v2/CollectionsHubsNav.tsx
@@ -1,31 +1,18 @@
 import { CSSGrid, Serif } from "@artsy/palette"
 import { CollectionsHubsNav_marketingHubCollections } from "__generated__/CollectionsHubsNav_marketingHubCollections.graphql"
+import { useTracking } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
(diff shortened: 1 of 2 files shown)


src/Components/v2/CollectionsHubsNav.tsx
```diff
@@ -1,32 +1,19 @@
...
-    trackEvent({
-      action_type: Schema.ActionType.ExperimentViewed,
-      experiment_id: experiment,
-      experiment_name: experiment,
-      variation_id: props.collectionHubTestVariation,
-      variation_name: props.collectionHubTestVariation,
-      nonInteraction: 1,
-    })
-  }, [])
-
```
xtina-starr: When we fired the Experiment Viewed event here and it was only getting called for the experiment and excluding the control. We want it to be called for both so we moved the call higher up (i.e. in the  Collect component).

zephraph:
Looks good!

artsyit:
:rocket: PR was released in v21.16.7 :rocket:

PR #2956, opened 2019-10-25
Add analytics calls to ConfirmBid route

bhoggard:
https://artsyproduct.atlassian.net/browse/AUCT-667

### DIFF
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -8,6 +8,8 @@ import { BidFormFragmentContainer as BidForm } from "Apps/Auction/Components/Bid
 import { LotInfoFragmentContainer as LotInfo } from "Apps/Auction/Components/LotInfo"
 import { AppContainer } from "Apps/Components/AppContainer"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
+import { track } from "Artsy"
+import * as Schema from "Artsy/Analytics/Schema"
 import { FormikActions } from "formik"
 import qs from "qs"
 import React from "react"
@@ -18,21 +20,23 @@ import {
   graphql,
   RelayProp,
 } from "react-relay"
+import { TrackingProp } from "react-tracking"
 import { data as sd } from "sharify"
 import { get } from "Utils/get"
 import createLogger from "Utils/logger"
 
 const logger = createLogger("Apps/Auction/Routes/ConfirmBid")
 
-interface BidProps {
+interface ConfirmBidProps {
   artwork: routes_ConfirmBidQueryResponse["artwork"]
   me: routes_ConfirmBidQueryResponse["me"]
   relay: RelayProp
   location: Location
+  tracking: TrackingProp
 }
 
-export const ConfirmBidRoute: React.FC<BidProps> = props => {
-  const { me, artwork } = props
+export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
+  const { me, artwork, tracking } = props
   const { saleArtwork } = artwork
   const { sale } = saleArtwork
   logger.log({
@@ -41,6 +45,13 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
     artwork,
     saleArtwork,
   })
+
+  const commonProperties = {
+    auction_slug: sale.id,
+    artwork_slug: artwork.id,
+    sale_id: sale._id,
+    user_id: me.id,
+  }
 
   function createBidderPosition(maxBidAmountCents: number) {
     return new Promise(async (resolve, reject) => {
@@ -59,6 +70,9 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
             ) {
               createBidderPosition(input: $input) {
                 result {
+                  position {
+                    id
+                  }
                   status
                   message_header
                   message_description_md
@@ -78,7 +92,11 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
     })
   }
 
-  function handleMutationError(actions: FormikActions<object>, error: Error) {
+  function handleMutationError(
+    actions: FormikActions<object>,
+    error: Error,
+    bidderId: string
+  ) {
     logger.error(error)
 
     let errorMessages: string[]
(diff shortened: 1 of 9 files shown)


src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -8,6 +8,8 @@ import { BidFormFragmentContainer as BidForm } from "Apps/Auction/Components/Bid
 import { LotInfoFragmentContainer as LotInfo } from "Apps/Auction/Components/LotInfo"
 import { AppContainer } from "Apps/Components/AppContainer"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
+import { track } from "Artsy"
+import * as Schema from "Artsy/Analytics/Schema"
```
damassi: import { AnalyticsSchema } from 'Artsy/Analytics'

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -18,21 +20,23 @@ import {
...
   me: routes_ConfirmBidQueryResponse["me"]
   relay: RelayProp
   location: Location
+  tracking: TrackingProp
 }
 
-export const ConfirmBidRoute: React.FC<BidProps> = props => {
-  const { me, artwork } = props
+export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
+  const { me, artwork, tracking } = props
```
damassi: Better to use the `useTracking` hook rather than pluck from props

erikdstock: does this mean removing the tracking prop?

damassi: Meaning the interface? Don't need it if it's not a prop. The `useTracking` value comes from our context.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
dleve123: @damassi Briefly, can you elaborate on the "why?" here? Generally attempting to avoid global imports? Taking a look at src/Artsy/Analytics/index.ts, pretty sure they are structurally equivalent. Looks like this is some relevant context: https://github.com/artsy/reaction/pull/2405

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -18,21 +20,23 @@ import {
...
   const { saleArtwork } = artwork
   const { sale } = saleArtwork
   logger.log({
```
dleve123: @bhoggard Do we think this log is intentional? We should probably remove if not.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -90,26 +108,51 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
...
 
-    // TODO: add tracking with errorMessages
-    logger.error(errorMessages) // remove once tracking is implemented
+    trackConfirmBidFailed(bidderId, errorMessages)
 
     actions.setSubmitting(false)
     actions.setStatus("submissionFailed")
   }
 
+  function trackConfirmBidFailed(bidderId: string, errors: string[]) {
```
dleve123: thoughts on moving the `commonProperties` variable to be just above or below these track function definitions and rename to `commonTrackingProperties`? Personally, I think that would better clarify their intent in code.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -90,26 +108,51 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
...
+    tracking.trackEvent({
+      action_type: Schema.ActionType.ConfirmBidFailed,
+      bidder_id: bidderId,
+      error_messages: errors,
+      ...commonProperties,
+    })
+  }
+
+  function trackConfirmBidSuccess(positionId: string, bidderId: string) {
```
dleve123: minor: thoughts on renaming to `trackConfirmBidSubmitted` for parity between the `action_type` and the function name?

src/Apps/Auction/Routes/ConfirmBid/index.tsx
dleve123: @damassi I believe (but could be wrong) that Erik and I tried to this more hooks oriented approach when implementing tracking for the Register component (https://github.com/artsy/reaction/pull/2727), but ran into issues with Formik.

@bhoggard I'd timeboxing ~1 hour to trying out this refactoring, because it seems worth it if it doesn't take too long!

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -90,26 +108,51 @@ export const ConfirmBidRoute: React.FC<BidProps> = props => {
...
+      bidder_id: bidderId,
+      ...commonProperties,
+    })
+  }
+
   function handleSubmit(
     values: { selectedBid: number },
     actions: FormikActions<object>
   ) {
+    const bidderId = sale.registrationStatus.id
```
dleve123: Just want to surface that it's possible for the underlying gravity endpoint, [`POST /api/v1/me/bidder_position`](https://github.com/artsy/gravity/blob/b38e5a5dbf5c181f9ab3b86a34091817a7f6a03d/app/api/v1/me_bidder_positions_endpoint.rb#L64-L68), to _create_ a `Bidder` document in Gravity's mongo, so is it safe to assume that we'll always have a bidder at this point?

Maybe for the "just bid" flow, instead of the "bid+register" flow, we'll feel free confident that we always have a preexisting `Bidder` document, but I think we'll have to re-visit this when implementing the "bid+register" flow

Did you guys discuss this during implementation?

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -107,13 +115,61 @@ describe("Routes/Register ", () => {
...
+  it("send an error event to analytics if the mutation fails", async () => {
+    const env = setupTestEnv()
+    const page = await env.buildPage()
+
+    env.mutations.useResultsOnce(createBidderPositionFailed)
+
+    await page.agreeToTerms()
+    await page.submitForm()
+
+    expect(env.mutations.mockFetch).toHaveBeenCalledWith(
```
dleve123: Is this expectation needed? We stub [the mutation above](https://github.com/artsy/reaction/pull/2956/files#diff-97e6a9881aac57a032e2b138e21443fcR141), so while correct, it seems a bit extraneous to allow expect the mutation to be called, and increases the size of the test.

bhoggard: We need it to confirm that we're correctly passing parameters to the GraphQL call.

dleve123:
Discussed IRL, but merge at your discretion here after the changes we paired through.

bhoggard:
#mergeongreen

artsyit:
:rocket: PR was released in v21.17.0 :rocket:

PR #2958, opened 2019-10-28
Collection SeoProducts description 

eessex:
Addresses https://sentry.io/organizations/artsynet/issues/1239198076/?project=28316

Stops using `!` for null checks.

### DIFF
--- a/src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
+++ b/src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
@@ -51,6 +51,9 @@ export class SeoProducts extends React.Component<SeoProductsProps> {
           ? toSentence(artists.map(artist => artist!.name))
           : null
         const isInstitution = partner && partner.type === "Institution"
+        const partnerImg = get(partner, p => {
+          return p!.profile && p.profile.icon && p.profile.icon.url
+        })
 
         return (
           <Product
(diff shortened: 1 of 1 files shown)


src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
```diff
@@ -83,8 +83,12 @@ export class SeoProducts extends React.Component<SeoProductsProps> {
                       availability: availability && AVAILABILITY[availability],
                       seller: {
                         "@type": "ArtGallery",
-                        name: partner && partner!.name,
-                        image: partner && partner!.profile!.icon!.url,
+                        name: partner && partner.name,
+                        image:
+                          partner &&
```
mzikherman: This one might benefit from its own variable and the `get` helper #minor

src/Apps/Collect2/Components/SeoProductsForArtworks.tsx
```diff
@@ -51,14 +51,17 @@ export class SeoProducts extends React.Component<SeoProductsProps> {
           ? toSentence(artists.map(artist => artist!.name))
           : null
         const isInstitution = partner && partner.type === "Institution"
+        const partnerImg = get(partner, p => {
```
mzikherman: so with `get` you don't need to be defensive with the `&&` stuff, that's why it can be kind of cleaner for this sort of thing.

can just be:

```
const partnerImage = get(partner, p => p.profile.icon.url)
```

artsyit:
:rocket: PR was released in v21.17.0 :rocket:

PR #2959, opened 2019-10-29
Article meta component

eessex:
Ports [existing jade template](https://github.com/artsy/force/blob/master/src/desktop/apps/article/templates/meta.jade) to react, part of https://artsyproduct.atlassian.net/browse/PLATFORM-1876

Force side is here: https://github.com/artsy/force/pull/4757

### DIFF
--- a/src/Components/Publishing/ArticleMeta.tsx
+++ b/src/Components/Publishing/ArticleMeta.tsx
@@ -5,6 +5,8 @@ import { data as sd } from "sharify"
 import { crop } from "Utils/resizer"
 import { createMediaStyle } from "Utils/Responsive"
 
+const responsiveCss = createMediaStyle()
+
 export const ArticleMeta: React.SFC<{
   article: ArticleData
 }> = props => {
@@ -20,7 +22,6 @@ export const ArticleMeta: React.SFC<{
   const keywords = (article.keywords && article.keywords.join(", ")) || ""
   const sailthruKeywords = [`article${keywords && `, ${keywords}`}`]
   const emailMetadata = article.published && article.email_metadata
-  const responsiveCss = createMediaStyle()
 
   if (article.featured) {
     sailthruKeywords.push("magazine")
--- a/src/Components/Publishing/ArticleMeta.tsx
+++ b/src/Components/Publishing/ArticleMeta.tsx
@@ -1,15 +1,14 @@
 import { getFullEditorialHref } from "Components/Publishing/Constants"
 import { ArticleData } from "Components/Publishing/Typings"
 import React from "react"
+import { data as sd } from "sharify"
 import { crop } from "Utils/resizer"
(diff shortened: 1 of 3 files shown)


src/Components/Publishing/ArticleMeta.tsx
```diff
@@ -0,0 +1,114 @@
+import { getFullEditorialHref } from "Components/Publishing/Constants"
+import { ArticleData } from "Components/Publishing/Typings"
+import React from "react"
+import { crop } from "Utils/resizer"
+
+export const ArticleMeta: React.SFC<{
+  article: ArticleData
+  customMetaContent: ArticleData
+  sd: any // res.locals.sd passed directly to component as sd
```
eessex: Not sure what the best way to deal with `sd` values is in this context, I based this on an older stitch implementation used on the [artist page here](https://github.com/artsy/force/blob/89e6d5f9b7928ee312abc863fc7f585ede9e94d1/src/desktop/apps/artist/server.tsx#L62).

alloy: There’s a few options.

* It looks like there’s only 2 values being used, perhaps those can just be distinct props without needing all of `sd` passed in?
* Are these values global (env) values or do they differ per request? If global, you can just expose types for them [here](https://github.com/artsy/reaction/blob/master/typings/sharify.d.ts) and then import sharify, rather than accepting it as a prop.

alloy: I see that the `RESPONSIVE_CSS` value is the CSS generated by fresnel and currently _is_ being [generated per request](https://github.com/artsy/force/blob/5144face4cfa4859e111b87b22035521f88c60bc/src/desktop/apps/article/routes.tsx#L133), but it actually should only be generated once at the file level. Also, in case it is _only_ used by this meta component, perhaps it can just import that helper here removing the need to even pass it in from Force.

eessex:
Your suggestions are now incorporated into this PR as well as the Force side.

src/Components/Publishing/ArticleMeta.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const url = getFullEditorialHref(article.layout, article.slug)
+  const socialTitle = article.social_title || article.thumbnail_title || ""
+  const searchDescription = article.search_description || article.description
+  const socialDescription = article.social_description || article.description
+  const socialImage = article.social_image || article.thumbnail_image
+  const authors = article.authors || article.contributing_authors || []
+  const keywords = (article.keywords && article.keywords.join(", ")) || ""
+  const sailthruKeywords = [`article${keywords && `, ${keywords}`}`]
+  const emailMetadata = article.published && article.email_metadata
+  const responsiveCss = createMediaStyle()
```
alloy: Can you move this out of the component to the file level?

PR #2961, opened 2019-10-29
WIP: Add retargetting tracking to confirm bid

dleve123:
See commits:
- 07af125

```
* Create a ProductInfo interface in analytics schema
  + I have no idea if this is the right place to place this type, but
    wanted to start somewhere.
* Add tests for product info for bid success case
* Still need to get this working for submission failed

Jira: https://artsyproduct.atlassian.net/browse/AUCT-667
```

- 3f12b2b

```
Why?

* This allows us to use the type schema, better protecting ourselves
incorrect tracking code.

* It also allows better factor common tracking components by replacing
dependency injection with composition with common properties defined
at the component level.

Details

- add common tracking properties to HOC call and delete commonProperties
- Remove incorrect assertion about `order_id` in failing case
- Extend analytics schema with Criteo and Auctions info
  + Add `CriteoInfo` interface to describe core information about
    integrating with Criteo
  + Add `AuctionInfo` interface to describe assorted auction-related
    properties.
- Use p instead of props in track-wrapping to avoid “props shadowing”
  error.
```

### DIFF
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -39,12 +39,6 @@ export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
   const { me, artwork, tracking } = props
   const { saleArtwork } = artwork
   const { sale } = saleArtwork
-  logger.log({
-    me,
-    sale,
-    artwork,
-    saleArtwork,
-  })
 
   function createBidderPosition(maxBidAmountCents: number) {
     return new Promise(async (resolve, reject) => {
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -10,7 +10,7 @@ import { AppContainer } from "Apps/Components/AppContainer"
 import { trackPageViewWrapper } from "Apps/Order/Utils/trackPageViewWrapper"
 import { track } from "Artsy"
 import * as Schema from "Artsy/Analytics/Schema"
-import { ProductInfo } from "Artsy/Analytics/Schema/Interaction"
+import { useTracking } from "Artsy/Analytics/useTracking"
 import { FormikActions } from "formik"
 import qs from "qs"
 import React from "react"
@@ -21,7 +21,6 @@ import {
   graphql,
   RelayProp,
 } from "react-relay"
-import { TrackingProp } from "react-tracking"
 import { data as sd } from "sharify"
 import { get } from "Utils/get"
 import createLogger from "Utils/logger"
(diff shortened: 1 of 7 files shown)


src/Artsy/Analytics/Schema/CriteoInfo.ts
```diff
@@ -0,0 +1,35 @@
...
+   */
+  product_id: string
+
+  /**
+   * The number of items considered to be within the "production"
+   *
+   * When used in the context of an auction lot, the quantity should be 1 for
+   * each bid.
+   */
+
```
dleve123: cleanup whitespace here

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -115,6 +115,21 @@ describe("Routes/ConfirmBid", () => {
...
     )
 
+    expect(window.location.assign).toHaveBeenCalledWith(
+      `https://example.com/auction/${
+        ConfirmBidQueryResponseFixture.artwork.saleArtwork.sale.id
+      }/artwork/${ConfirmBidQueryResponseFixture.artwork.id}`
+    )
+  })
+
+  it("tracks a success event to Segment including product information", async () => {
```
dleve123: Including Criteo in docstring

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -107,28 +102,31 @@ export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
...
       action_type: Schema.ActionType.ConfirmBidSubmitted,
       bidder_position_id: positionId,
       bidder_id: bidderId,
-      ...commonEventProperties,
+      order_id: bidderId,
+      products: [
+        {
+          product_id: artwork._id,
+          quantity: 1,
+          price: selectedBidAmountCents / 100,
```
mzikherman: This jumps out as something that might not respect currencies with a different multiplier. Is it possible to pass the cents amount itself to the analytics call? (presumably any calculation or conversion could be done there).

dleve123: This was lifted from the corresponding Force code. Agreed that it seems not robust to other currencies, but then again, it’s been in prod on Force for a few months. I think @cavvia is best equipped to comment here.

cavvia: Thanks for catching that! This data is just used by our marketing partner Criteo right now I think... yes, it is probably affected by currency variations but it's not a high priority area to fix for us right now, as it has been like this for a long time.

artsyit:
:rocket: PR was released in v21.17.5 :rocket:

artsyit:
:rocket: PR was released in v21.17.5 :rocket:

artsyit:
:rocket: PR was released in v21.17.5 :rocket:

artsyit:
:rocket: PR was released in v21.17.5 :rocket:

PR #2963, opened 2019-10-30
Conditions of sale --> Conditions of Sale in Auction FAQ

sweir27:
#minor Noticed this while QAing the artwork view.


artsyit:
:rocket: PR was released in v21.17.4 :rocket:

PR #2964, opened 2019-11-01
Track ad impressions and clicks on 3rd party served ads

ashleyjelks:
[Ad Server Analytics Tracking--GROW-1545](https://artsyproduct.atlassian.net/browse/GROW-1545)

Adds click and impression tracking to ads being served through the 3rd party ad server.

### DIFF
--- a/src/Components/Publishing/Display/DisplayAd.tsx
+++ b/src/Components/Publishing/Display/DisplayAd.tsx
@@ -3,10 +3,9 @@ import * as AnalyticsSchema from "Artsy/Analytics/Schema"
 import { useTracking } from "Artsy/Analytics/useTracking"
 import { is300x50AdUnit } from "Components/Publishing/Display/DisplayTargeting"
 import { AdDimension, AdUnit } from "Components/Publishing/Typings"
-import React, { SFC, useState } from "react"
+import React, { SFC, useEffect, useState } from "react"
 import { Bling as GPT } from "react-gpt"
 import styled from "styled-components"
-import { useDidMount } from "Utils/Hooks/useDidMount"
 
 export interface DisplayAdProps extends FlexProps {
   adUnit: AdUnit
@@ -43,9 +42,8 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
   const [isAdEmpty, setAdEmpty] = useState(null)
   const isMobileLeaderboardAd = is300x50AdUnit(adDimension)
   const { trackEvent } = useTracking()
-  const isMounted = useDidMount()
 
-  if (isMounted) {
+  useEffect(() => {
     trackEvent({
       action_type: AnalyticsSchema.ActionType.Impression,
       context_page: AnalyticsSchema.PageName.ArticlePage,
@@ -54,7 +52,7 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
       context_page_owner_slug: props.articleSlug,
       context_page_owner_type: AnalyticsSchema.OwnerType.Article,
     })
-  }
+  }, [])
 
   const onClickAd = () => {
     trackEvent({
--- a/src/Components/Publishing/Display/DisplayAd.tsx
+++ b/src/Components/Publishing/Display/DisplayAd.tsx
@@ -1,9 +1,12 @@
 import { Box, color, Flex, FlexProps, Sans } from "@artsy/palette"
+import * as AnalyticsSchema from "Artsy/Analytics/Schema"
+import { useTracking } from "Artsy/Analytics/useTracking"
 import { is300x50AdUnit } from "Components/Publishing/Display/DisplayTargeting"
 import { AdDimension, AdUnit } from "Components/Publishing/Typings"
 import React, { SFC, useState } from "react"
 import { Bling as GPT } from "react-gpt"
 import styled from "styled-components"
+import { useDidMount } from "Utils/Hooks/useDidMount"
 
 export interface DisplayAdProps extends FlexProps {
   adUnit: AdUnit
@@ -16,6 +19,7 @@ export interface DisplayAdProps extends FlexProps {
   }
   isSeries?: boolean
   isStandard?: boolean
+  articleSlug: string
 }
 
 export interface DisplayAdContainerProps extends FlexProps {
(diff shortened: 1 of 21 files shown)


src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -28,10 +32,27 @@ export interface DisplayAdContainerProps extends FlexProps {
...
+
+  const onClickAd = () => {
+    trackEvent({
+      context_page: AnalyticsSchema.PageName.ArticlePage,
+      context_module: AnalyticsSchema.ContextModule.AdServer,
+      context_page_owner_id: props.targetingData.post_id,
+      context_page_owner_slug: props.articleSlug,
+      context_page_owner_type: AnalyticsSchema.OwnerType.Article,
+    })
+  }
```
xtina-starr: For the [click event](https://github.com/artsy/reaction/blob/e92baf0ba0cb3aa4d25f448969d6ec926f24fd73/src/Apps/Collect2/Routes/Collection/Components/CollectionsHubRails/OtherCollectionsRail/OtherCollectionEntity.tsx#L25), I think we'd want to `action_type: Schema.ActionType.Click` here.

src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -49,8 +70,19 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
...
   }
 
+  if (useDidMount) {
+    trackEvent({
+      context_page: AnalyticsSchema.PageName.ArticlePage,
+      context_module: AnalyticsSchema.ContextModule.AdServer,
+      context_page_owner_id: props.targetingData.post_id,
+      context_page_owner_slug: props.articleSlug,
+      context_page_owner_type: AnalyticsSchema.OwnerType.Article,
+    })
```
xtina-starr: And `action_type: AnalyticsSchema.ActionType.Impression` here.

xtina-starr:
I think this is good to merge once you add the `action_type`.

src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -49,8 +70,19 @@ export const DisplayAd: SFC<DisplayAdProps> = props => {
     return null
   }
 
+  if (useDidMount) {
```
pepopowitz: This looks like incorrect usage of `useDidMount`. I'd expect this statement to always return true, because `useDidMount` is a function.

Looking at the code in `useDidMount`, I'd expect the usage to look like this: 

```js
const didMount = useDidMount()

...

if (didMount) {
  ...
}
```

I also think the initial call to `useDidMount()` should be moved up to line 46, to sit with the other hooks calls. One critical thing about hooks is that they are called in the same order every single time. Right now, that's going to happen in this function even if we call `useDidMount()` on line 73...but grouping the hooks calls at the top of the function prevents us from inserting an early-return in between them in the future. (Which would cause a conditional hooks call, which would not work.)

src/Components/Publishing/Display/__tests__/DisplayAd.test.tsx
```diff
@@ -9,53 +10,91 @@ import "jest-styled-components"
...
-    page_type: "article",
-    post_id: "594a7e2254c37f00177c0ea9",
-    tags: "Art Market",
+    expect(trackEvent).toBeCalledWith({
+      context_page: "Article",
+      context_module: "AdServer",
+      context_page_owner_type: "Article",
+      context_page_owner_id: StandardArticle.id,
+      context_page_owner_slug: StandardArticle.slug,
+    })
```
pepopowitz: We should add an `action_type` prop here to align with [@xtina-starr 's above comments](https://github.com/artsy/reaction/pull/2964/files#r342234920)

src/Components/Publishing/__tests__/Article.test.tsx
```diff
@@ -31,6 +31,17 @@ global.fetch = jest.fn(() =>
 jest.mock("../ToolTip/TooltipsDataLoader", () => ({
   TooltipsData: props => props.children,
 }))
+jest.mock("Artsy/Analytics/useTracking")
+
+const trackEvent = jest.fn()
+
+beforeEach(() => {
+  ;(useTracking as jest.Mock).mockImplementation(() => {
```
pepopowitz: For as often as we're repeating this block, it might make sense to extract a common function, so these all get reduced to something like 

```
beforeEach(() => {
  mockTracking(trackEvent)
})
```

It would be less repetition and less code to look at in each `beforeEach`.

ashleyjelks: I actually tried creating an imported module with the repetitive code but it kept failing...when I ran in in the debugger the error thrown was  something like `the no- import-unmocked-threw an error` [which looks like an issue others have run into when importing mocks.](https://github.com/facebook/jest/issues/4262) I'd love to pair on this if you have some tijme, but I time-boxed and wasn't able to find a working solution during that allotted period.   Maybe its a good candidate for a Wednesday pairing session?

src/Components/Publishing/Display/DisplayAd.tsx
```diff
@@ -28,10 +32,40 @@ export interface DisplayAdContainerProps extends FlexProps {
...
+    ...otherProps
+  } = props
   const [width, height] = adDimension.split("x").map(a => parseInt(a))
   const [isAdEmpty, setAdEmpty] = useState(null)
   const isMobileLeaderboardAd = is300x50AdUnit(adDimension)
+  const { trackEvent } = useTracking()
+  const isMounted = useDidMount()
+
+  if (isMounted) {
+    trackEvent({
```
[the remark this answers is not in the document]
ashleyjelks: Hmm, yes, and actually I wonder if the hook should be updated to handle this since I would assume it's intended to only be called once after the component mounts as well.  I'll look into it and make the changes.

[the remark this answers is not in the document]
xtina-starr: I guess I thought that's what `useDidMount` was doing as well because it looks like it's just a wrapper for `useEffect` which receives the empty array to only fire once.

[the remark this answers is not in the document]
xtina-starr: Maybe it's worth checking in with @kierangillen to see what the use case for `useDidMount` is but I feel like both ways of writing this are correct.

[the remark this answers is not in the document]
pepopowitz: The way I read the code for `useDidMount`, it is firing an effect once, but that effect is to update a state property that tracks whether the component has "mounted" or not. This would be useful for a situation where you didn't want something to happen in your component until you were sure the component had mounted, but it doesn't do anything to restrict things from happening every re-render after it's mounted. Our scenario is different - we want something to happen once the component is first rendered, *but never again*.

artsyit:
:rocket: PR was released in v22.1.0 :rocket:

PR #2967, opened 2019-11-04
Moved artist insights to the right PURCHASE-1608

ashkan18:
# Change
https://artsyproduct.atlassian.net/browse/PURCHASE-1608
-  Move Artist Insights to the right side of bio
- Remove current events

# Selfie
image

cc: @samchieng

### DIFF
--- a/src/Apps/Artist/Routes/Overview/index.tsx
+++ b/src/Apps/Artist/Routes/Overview/index.tsx
@@ -18,7 +18,6 @@ import { TrackingProp } from "react-tracking"
 import { get } from "Utils/get"
 import { ArtistArtworkFilterRefetchContainer as ArtworkFilter } from "./Components/ArtistArtworkFilter"
 import { ArtistRecommendationsQueryRenderer as ArtistRecommendations } from "./Components/ArtistRecommendations"
-import { CurrentEventFragmentContainer as CurrentEvent } from "./Components/CurrentEvent"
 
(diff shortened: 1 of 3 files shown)


samchieng:
@ashkan18 `col 3` might be a little too tight since the insights are now longer than the bio + categories, can we try `col 4`?

ashkan18:
@samchieng 👍 this is with 4

image

sepans:
Looks great 💯 . A couple of UI/design questions:

I am guessing for artists without career achievement block we want to use the full width for bio. Is that correct @samchieng ?

Also noticed title of the block has been changed and moved outside of the box in [zeplin file](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5db361940c012583408ff90b). Wondering if that is part of this ticket?

And finally, seems the margins being a bit off on mobile comparing to [the design](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5db36194f0f387550b9a56ff). Seems the page should stay the way it is now for mobile (except for the title being moved outside the box?)

samchieng:
If there's no career achievement block, we can keep the empty space, so the bio & categories should still use `col 8`

And for moving the title outside the box I will create a separate ticket, so don't need for this PR!

For mobile can you send a screenshot? Should be the same layout we have right now

artsyit:
:rocket: PR was released in v22.0.6 :rocket:

PR #2977, opened 2019-11-06
[MP v2] Remove mock resolvers and add more typings to tests

alloy:
🚗


artsyit:

PR #2980, opened 2019-11-06
User should be redirected to artwork page after confirmation that a bid is winning

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-712

**Do not squash the commits as each commit has a meaningful message.** 

Adds polling logic that checks the bidding status and redirects the user to the artwork page when the user is winning.

### Placing a user as a registered user

AUCT-712

### 

Screen Shot 2019-11-06 at 12 17 05 PM

### DIFF
--- a/src/Apps/Auction/Components/BidForm.tsx
+++ b/src/Apps/Auction/Components/BidForm.tsx
@@ -7,6 +7,7 @@ import {
   Separator,
   Serif,
 } from "@artsy/palette"
+import { BidForm_me } from "__generated__/BidForm_me.graphql"
 import { BidForm_saleArtwork } from "__generated__/BidForm_saleArtwork.graphql"
 import { PricingTransparency } from "Apps/Auction/Components/PricingTransparency"
 import { ConditionsOfSaleCheckbox } from "Components/Auction/ConditionsOfSaleCheckbox"
@@ -17,10 +18,11 @@ import { createFragmentContainer, graphql } from "react-relay"
 import Yup from "yup"
 
 interface Props {
-  showPricingTransparency?: boolean
-  saleArtwork: BidForm_saleArtwork
   initialSelectedBid?: string
+  me: BidForm_me
   onSubmit: (values: FormikValues, actions: FormikActions<object>) => void
+  saleArtwork: BidForm_saleArtwork
+  showPricingTransparency?: boolean
 }
 
 interface FormValues {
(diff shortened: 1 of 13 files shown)


yuki24:
I don't believe the peril comment is relevant as we auto-update the CHANGELOG upon each release.

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -57,18 +63,38 @@ const getSelectedBid = ({
...
+  bidder: BidForm_saleArtwork["sale"]["registrationStatus"],
+  me: BidForm_me
+) => {
+  const isRegistered = !!bidder
+
+  return {
+    requiresCheckbox: !isRegistered,
+    requiresPaymentInformation: !(isRegistered || me.hasQualifiedCreditCards),
+  }
+}
```
erikdstock: Nice, I like capturing these points in one area.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -34,8 +37,12 @@ interface ConfirmBidProps {
   location: Location
 }
 
+const MAX_POLL_ATTEMPTS = 20
+
 export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
-  const { artwork } = props
+  let pollCount = 0
```
erikdstock: minor question here whether there is a better way to do this using a react hook.

bhoggard:
I don't see a branch where we display the credit card form. Is that handled by a redirect case somewhere?

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -93,189 +102,220 @@ describe("Routes/ConfirmBid", () => {
...
-          quantity: 1,
-          price: 50000,
-        },
-      ],
+      mockBidderPositionQuery.mockReset()
+      mockBidderPositionQuery.mockResolvedValue(
+        confirmBidBidderPositionQueryWithWinning
+      )
+
+      setTimeout(() => {
```
yuki24: This is a bit regrettable but our test setup is a bit heavier than I initially anticipated, so I couldn't figure out how to use the Jest timer in tandem with the page object. But there must be a better way than `setTimeout`.

cc @ds300

yuki24:
I'm still a bit concerned about the `setTimeout`, but let's re-visit once we have a point-person who can look into it. Merging.

src/Apps/Auction/Components/BidForm.tsx
dleve123: +1 – really readable code!

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
ds300: I think it should be possible to use jest's mock timers with `createTestEnv`. Do you remember what issues you ran into? Happy to pair on this if you have some spare minutes soon!

PR #2990, opened 2019-11-07
Allow users to register+bid

dleve123:
Jira: https://artsyproduct.atlassian.net/browse/AUCT-716

### DIFF
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -67,6 +67,13 @@ export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
                 result {
                   position {
                     id
+                    sale_artwork {
+                      sale {
+                        registrationStatus {
+                          id
+                        }
+                      }
+                    }
                   }
                   status
                   message_header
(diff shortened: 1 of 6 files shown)


src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -141,21 +148,30 @@ export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
     values: { selectedBid: number },
     actions: FormikActions<object>
   ) {
-    const bidderId = sale.registrationStatus.id
     const selectedBid = Number(values.selectedBid)
+    const possibleExistingBidderId: string | null = sale.registrationStatus
```
dleve123: I'm a bit confused here. The type for the mutation response states that the `registrationStatus` can be `null`, but seems like TS isn't picking up on that in this file? Check out the below screenshot:

Screen Shot 2019-11-07 at 3 26 30 PM

Even when manually declaring that this var can be null-able...

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -268,6 +269,41 @@ describe("Routes/ConfirmBid", () => {
...
+      const page = await env.buildPage()
+      env.mutations.useResultsOnce(createBidderPositionSuccessfulAndBidder)
+      mockBidderPositionQuery.mockResolvedValue(
+        confirmBidBidderPositionQueryWithWinning
+      )
+
+      await page.submitForm()
+
+      setTimeout(() => {
+        // TODO: What's up with it being called 2 times?
```
yuki24: This was caused by the previous test that had a mock with bidding bidding status: https://github.com/artsy/reaction/pull/2990/commits/c07a458efc5b21f978ae66c508f8f7c6de74bd19#diff-97e6a9881aac57a032e2b138e21443fcR251
Because the bidding status was pending the polling logic was getting kicked in, but the test finishes before the delayed logic is executed. I changed the mock data to the bidding status of `winning` so it'll immediately call the tracking code, which makes the test execution order more predictable.

artsyit:
:rocket: PR was released in v22.2.0 :rocket:

PR #3007, opened 2019-11-14
Grow 1378 Add AggregateOffer schema for Collections pages

jiahaodavid:
Address:
 [Grow 1378](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1378)

This PR adds the aggregate offer schema to collections pages.

It queries for the most expensive artwork within the collection to determine the max price of the aggregate offer, and the least expensive artwork to determine the min price. It handles individual prices or price ranges on those artworks. 

Tests are added to catch many edge-cases; please let me know if I missed any!


src/Apps/Collect2/Components/SeoProductsForCollection.tsx
```diff
@@ -0,0 +1,132 @@
...
+  } else if (currencyCode === "EUR") return price * 1.1
+}
+
+export const getMaxMinPrice = priceableArtworks => {
+  const priceRange = [null, null]
+  priceableArtworks.map(artwork => {
+    if (artwork.node.is_price_range && artwork.node.listPrice) {
+      const { maxPrice, minPrice } = artwork.node.listPrice
+      const convertedPrice = {
+        maxPrice: simpleConversionToUSD(maxPrice.major, maxPrice.currencyCode),
```
sepans: I think our best bet here is to ignore currencies other than `USD` because there are tens of other currencies that we accept and could lead to very off ranges. (cc: @zephraph do you have other suggestions?)

src/Apps/Collect2/Components/SeoProductsForCollection.tsx
```diff
@@ -0,0 +1,132 @@
...
+export function simpleConversionToUSD(price: number, currencyCode: string) {
+  if (currencyCode === "USD") {
+    return price
+  } else if (currencyCode === "GBP") {
+    return price * 1.28
+  } else if (currencyCode === "EUR") return price * 1.1
+}
+
+export const getMaxMinPrice = priceableArtworks => {
+  const priceRange = [null, null]
```
sepans: #minor I have slight preference having an object instead of array here to make the following lines more readable (`priceRange.min` instead of `priceRange[0]`)

Also it is possible to set the initial values to `-Number.MAX_VALUE`/`Number.MAX_VALUE` to avoid the checks like `!priceRange[1] || ...`

```js
const priceRange = { min: Number.MAX_VALUE, max: - Number.MAX_VALUE } 
```

However this is just a stylistic suggestion so if you are more comfortable with the current implementation it is perfectly fine.

src/Apps/Collect2/Components/SeoProductsForCollection.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+export const SeoProductsForCollection = createFragmentContainer(SeoProducts, {
+  artworks: graphql`
+    fragment SeoProductsForCollection_artworks on FilterArtworks
+      @argumentDefinitions(
+        first: { type: "Int", defaultValue: 30 }
+        after: { type: "String", defaultValue: "" }
+      ) {
+      artworks_connection(first: $first, after: $after) {
```
sepans: Artist metadata uses [these parameters](https://github.com/artsy/reaction/blob/713a65c548d8c76421d17c80c13822c1687d44fc/src/Apps/Artist/Components/ArtistMeta.tsx#L258-L260) to make sure works are published and for sale. Maybe it could be used here as well? 

With that I think `getFilteredArtworks` would not be needed.

src/Apps/Collect2/Components/SeoProductsForCollection.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return (
+      <>
+        <Product
+          data={{
+            name: collectionName,
+            description: collectionDescription,
+            url: collectionURL,
+            offers: {
+              "@type": "AggregateOffer",
+              lowPrice: getMaxMinPrice(priceAbleArtworks)[0],
```
sepans: This is probably a very rare case but `lowPrice` is required so if it has an invalid value no `Product` metadata should be added to page.

src/Apps/Collect2/Components/SeoProductsForCollection.tsx
jiahaodavid: @sepans Thanks for the feedback. 🤝 
I made some changes based on your feedback. 
You mentioned I should use  [these parameters](https://github.com/artsy/reaction/blob/713a65c548d8c76421d17c80c13822c1687d44fc/src/Apps/Artist/Components/ArtistMeta.tsx#L258-L260), but that one is from `artist`, this case is using `filterArtworks`. From the `filterArtworks`, it only has `sort, after, first, before, last`.
I am not familiar with Relay. Should I make a new query? or I did try to modify the query, but I got some errors. Something like, `said it must have the same query as his parent has`. It seems I can not modify the query.

sepans: Yes you are absolutely right. sorry about misleading you 😓 😆

jiahaodavid: @sepans 
It's ok.🙇 Thanks a lot.
I have another idea. I try to use two fragments. One is sorted by price (desc), and another is sorted by price(asc). Then I just get the first index of each one. When I try to modify Relay, I got many errors. It might need more time to figure it out. 
The existing PR, and the idea I mentioned above. Comparing two ideas. Is it worth to figure it out? or This PR is good enough. There are two reasons that I came up with another idea. 
First,  I saw there is a ticket on our board, it about improving the performance on the collection page. I guess this PR might hurt the performance.
Second, I found out the query only gets 30 artworks, not all of them. If there are more than 30 artworks on the page, the result is wrong.
What do you think?

jiahaodavid:
There is more information in the comment area in this [ticket](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1378).

anandaroop:
As an update, David and I have been working on currency conversion within MP here: https://github.com/artsy/metaphysics/pull/2087

src/Apps/Collect2/Components/SeoProductsForCollections.tsx
```diff
@@ -0,0 +1,148 @@
...
+    fragment SeoProductsForCollections_descending_artworks on FilterArtworksConnection {
+      edges {
+        node {
+          id
+          availability
+          listPrice {
+            __typename
+            ... on PriceRange {
+              minPrice {
+                major
```
pepopowitz: Discussing with @anandaroop : Based on the changes in MP, we think this should look like `major(convertTo:"USD")`. This should apply to everywhere we query `major`.

There may be issues with Relay compilation that prevent us from even getting this far, though.

pepopowitz:
Pending re-review, I believe this PR is ready to go.

anandaroop:
Alrighty I added that spec I was suggesting, and as expected it was 🍏 

Merge on green!

xtina-starr:
Thanks for adding the additional tests @anandaroop. I think this is good to merge.

artsyit:
:rocket: PR was released in v24.0.22 :rocket:

src/Apps/Collect2/Components/__tests__/SeoProductsForCollections.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+    )
+    props.ascending_artworks = buildAscendingArtworks(buildIndividualPrice(15))
+    const wrapper = renderProducts()
+
+    const html = wrapper.html()
+    expect(html).toContain('"lowPrice":15')
+    expect(html).toContain('"highPrice":30')
+  })
+
+  describe("no prices for descending artwork", () => {
```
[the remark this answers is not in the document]
pepopowitz: (I know this was merged already, but following up on this convo...)

I don't know the data well enough to know whether all the scenarios tested are "possible", but I also feel like we miss edge cases fairly often with our tests. These were added for completeness - not to say that I think they _would_ happen, but if for some reason they _did_, they're accounted for.

PR #2935, opened 2019-10-21
[Revert] Move arc router code back into reaction

damassi:
Reverts https://github.com/artsy/reaction/pull/2863 and https://github.com/artsy/reaction/pull/2866

After trying to build a few more examples in Arc this weekend I realized that there are too many edge-cases and unique configuration aspects to make this more than an artsy-specific project. For SSR, people are gonna have to look at their environments and build their own abstractions; there's no silver bullet for this unfortunately. 

Follow up PR will be to remove Arc from force and I'll delete the arc repo.  

- Force PR: https://github.com/artsy/force/pull/4717

PR #3010, opened 2019-11-14
Ensure user agent gets passed from SSR Relay Environment

zephraph:
This should (hopefully) fix [PURCHASE-1140](https://artsyproduct.atlassian.net/browse/PURCHASE-1140).

This was originally introduced in https://github.com/artsy/reaction/pull/2923 and accidentally reverted in https://github.com/artsy/reaction/pull/2935. 

For context, this is required so that the original user agent is passed all the way through the system. Otherwise it gets dropped and isn't recorded in our analytics.


artsyit:
:rocket: PR was released in v22.0.3 :rocket:

PR #3011, opened 2019-11-14
User without an existing credit card should be able to bid and register

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-465
paired with @bhoggard 

This PR adds a way for unregistered users without credit cards to big and register at the same time. Most of the code is taken from the `Register.tsx` component and there may be a good refactoring opportunity, but in terms of functionality this is good to go.

## Screenshot

AUCT-465

### DIFF
--- a/src/Apps/Auction/Components/BidForm.tsx
+++ b/src/Apps/Auction/Components/BidForm.tsx
@@ -17,6 +17,9 @@ import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import Yup from "yup"
 
+import { Address, AddressForm } from "Apps/Order/Components/AddressForm"
+import { CreditCardInput } from "Apps/Order/Components/CreditCardInput"
+
 interface Props {
   initialSelectedBid?: string
   me: BidForm_me
@@ -26,8 +29,10 @@ interface Props {
 }
 
 export interface FormValues {
-  selectedBid: string
+  address?: Address
   agreeToTerms: boolean
+  creditCard?: string
+  selectedBid: string
 }
 
 const validationSchemaForRegisteredUsers = Yup.object().shape({
@@ -63,7 +68,7 @@ const getSelectedBid = ({
   return selectedIncrement.value
 }
 
-const determineDisplayRequirements = (
+export const determineDisplayRequirements = (
   bidder: BidForm_saleArtwork["sale"]["registrationStatus"],
   me: BidForm_me
 ) => {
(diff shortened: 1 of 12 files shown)


src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -137,6 +152,48 @@ export const BidForm: React.FC<Props> = ({
                   {showPricingTransparency && <PricingTransparency />}
                 </Flex>
 
+                {requiresPaymentInformation && (
+                  <Box>
+                    <Separator mb={3} />
+
+                    <Serif size="4" color="black100">
```
dleve123: this block is identical to:

https://github.com/artsy/reaction/blob/6222a8ad9ee209f5ec4974b2d40d51d35de2cb2c/src/Apps/Auction/Components/RegistrationForm.tsx#L224

– thoughts on extracting some form of shared component? Particularly with content, we would they blocks to change together.

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -137,6 +152,48 @@ export const BidForm: React.FC<Props> = ({
...
+                      Please enter your credit card information below. The name
+                      on your Artsy account must match the name on the card, and
+                      a valid credit card is required in order to bid.
+                    </Serif>
+                    <Serif size="4" mt={2} color="black100">
```
dleve123: identical to the above

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -137,6 +152,48 @@ export const BidForm: React.FC<Props> = ({
...
+                    >
+                      Card Information
+                    </Serif>
+
+                    <CreditCardInput
+                      error={{ message: errors.creditCard } as stripe.Error}
+                    />
+
+                    <Box mt={2}>
+                      <AddressForm
```
dleve123: Our registration form is implemented with Formik, but here we are using the AddressForm directly without Formik? Why the difference? Is there a way we can make the approach here more consistent?

dleve123: Nvm! I see now that this is within a larger Formik component.

src/Apps/Auction/Components/BidForm.tsx
yuki24: Yes, there is a large opportunity for refactoring. @bhoggard are I looked at some other stuff like GraphQL calls, small functions, validation schemas, and even consolidating the `RegistrationForm` and `BidForm` , but it seemed like a lot of refactoring.

We might want to loop @dblandin and @erikdstock in to get it done well, but I don't think that should block merging this PR.

artsyit:
:rocket: PR was released in v22.2.0 :rocket:

PR #3014, opened 2019-11-15
PaymentForm __id -> id

ds300:
further proof, if any were needed, that noImplicitAny is a win.


artsyit:
:rocket: PR was released in v22.0.7 :rocket:

PR #3017, opened 2019-11-16
Upgrade to relay v7

alloy:
📈


src/__generated__/Accept_order.graphql.ts
```diff
@@ -24,6 +24,11 @@ export type Accept_order = {
     readonly " $fragmentRefs": FragmentRefs<"TransactionDetailsSummaryItem_order" | "ArtworkSummaryItem_order" | "ShippingSummaryItem_order" | "CreditCardSummaryItem_order">;
     readonly " $refType": "Accept_order";
 };
+export type Accept_order$data = Accept_order;
+export type Accept_order$key = {
+    readonly " $data"?: Accept_order$data;
+    readonly " $fragmentRefs": FragmentRefs<"Accept_order">;
+};
```
alloy: All of these additions are to support the experimental hooks (which needs `react-relay` to be installed from 1pm’s `experimental` dist-tag).

patches/relay-compiler+7.1.0.patch
```diff
@@ -25,17 +25,14 @@ index d0ae8f5..3d47c54 100755
        sources.push(source.body);
        astDefinitions.push.apply(astDefinitions, (0, _toConsumableArray2["default"])(ast.definitions));
      });
-@@ -15352,9 +15352,10 @@ function find(tagFinder, text, absPath) {
+@@ -17478,9 +17478,7 @@ function find(tagFinder, text, absPath) {
```
damassi: What do these two patches do and why are they needed?

mzikherman:
That seemed easy! (besides the patch).

patches/relay-compiler+7.1.0.patch
ds300: This relay-compiler patch is to give us 'clickable' file names with correct `line:column` numbers in error messages.

artsyit:
:rocket: PR was released in v22.1.0 :rocket:

patches/relay-compiler+7.1.0.patch
damassi: @ds300 so to confirm, if the two patches in this PR were to suddenly go away it would have no effect on application behavior?

ds300: right :+1:

alloy: Although you would have to clear the relay-compiler cache in `$TMPDIR`

PR #3018, opened 2019-11-18
Add modal to "Certificate of Authenticity" Badges with explanation of COA

lilyfromseattle:
[PURCHASE-1605]
__Motivation__
COA is a non homogenous standard across the art Industry. In order to help our collector make the best decision, it is important to show more information about how COAs are created etc.

__Story__
As a user, When I view an artwork that has a certificate of authenticity, If I click on the Certificate of Authenticity link in the COA Trust Signal text, I see a modal that provides more explanation about certificates of authenticity.

__Acceptance Criteria__
COA Info Modal upon click on COA Trust Signals link
Change copy on the Badge to display: “A certificate of authenticity is included with purchase”

Screen Shot 2019-11-18 at 2 03 00 PM

[PURCHASE-1605]: https://artsyproduct.atlassian.net/browse/PURCHASE-1605

### DIFF
--- a/src/Apps/Artwork/Components/TrustSignals/__tests__/AuthenticityCertificate.test.tsx
+++ b/src/Apps/Artwork/Components/TrustSignals/__tests__/AuthenticityCertificate.test.tsx
@@ -71,65 +71,7 @@ describe("AuthenticityCertificate", () => {
     expect(component.find("TrustSignal").length).toBe(1)
   })
 
-  it("Click on certificate of authenticity link opens modal", async () => {
-    // const component = await render({
-    //   id: "opaque-cert-id",
-    //   hasCertificateOfAuthenticity: true,
-    //   is_biddable: false,
-    // })
+  it.todo("Click on certificate of authenticity link opens modal")
 
-    const component = mount(
-      <AuthenticityCertificate
-        artwork={{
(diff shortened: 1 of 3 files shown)


src/Apps/Artwork/Components/TrustSignals/__tests__/AuthenticityCertificate.test.tsx
```diff
@@ -65,4 +70,66 @@ describe("AuthenticityCertificate", () => {
     })
     expect(component.find("TrustSignal").length).toBe(1)
   })
+
+  it("Click on certificate of authenticity link opens modal", async () => {
```
zephraph: My concern with this test is that potentially the expect clauses aren't actually executing. 

Could you add `expect.assertions(2)` on 75 just to make sure everything is being executed as expected?

lilyfromseattle: Damn you're right

lilyfromseattle: Can we pair on this?

artsyit:

PR #3020, opened 2019-11-19
Create gene links view

zephraph:
Adds the related links view from the gene page to reaction. 

image

Currently the markup structure is preserved from what exists on the gene page and no additional styling was added. This is largely due to the fact that the gene page isn't in line with our current design system and updating the whole gene page to match feels out of scope for this ticket.

### DIFF
--- a/src/Components/Gene/GeneRelatedLinks.tsx
+++ b/src/Components/Gene/GeneRelatedLinks.tsx
@@ -1,0 +1,91 @@
+import { Box } from "@artsy/palette"
+import {
+  GeneRelatedLinksQuery,
+  GeneRelatedLinksQueryResponse,
+} from "__generated__/GeneRelatedLinksQuery.graphql"
+import { useSystemContext } from "Artsy"
+import { renderWithLoadProgress } from "Artsy/Relay/renderWithLoadProgress"
+import { SystemQueryRenderer as QueryRenderer } from "Artsy/Relay/SystemQueryRenderer"
+import React from "react"
+import { graphql } from "react-relay"
+
+/**
+ * Used on the Gene page to show related artists and categories.
+ */
+export const GeneRelatedLinks: React.FC<GeneRelatedLinksQueryResponse> = ({
+  gene,
+}) => {
+  return (
+    <>
+      <Box className="related-genes related-links bisected-header-cell-section is-fade-in">
+        <h2>Related Categories</h2>
+        <div className="related-genes-links">
+          {gene.similar.edges.map(({ node: similarGene }, index) => {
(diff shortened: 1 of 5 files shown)


src/Components/Gene/GeneRelatedLinks.tsx
```diff
@@ -0,0 +1,91 @@
...
+ */
+export const GeneRelatedLinks: React.FC<GeneRelatedLinksQueryResponse> = ({
+  gene,
+}) => {
+  return (
+    <>
+      <Box className="related-genes related-links bisected-header-cell-section is-fade-in">
+        <h2>Related Categories</h2>
+        <div className="related-genes-links">
+          {gene.similar.edges.map(({ node: similarGene }, index) => {
```
sweir27: Is it possible that `gene.similar` would return null and we should be more defensive about the presence of `edges`?

src/Components/Gene/GeneRelatedLinks.tsx
```diff
@@ -0,0 +1,91 @@
...
+                {separator}
+              </React.Fragment>
+            )
+          })}
+        </div>
+      </Box>
+      <Box className="related-artists related-links bisected-header-cell-section is-fade-in">
+        <h2>Related Artists</h2>
+        <div className="artists">
+          {gene.artists.edges.map(({ node: artist }, index) => {
```
sweir27: Same here re: safe accessing.

zephraph:
@sepans I suspect not a lot of places. I certainly can duplicate the required styles over in reaction and remove the unused styles on Force's side. That might be better generally.

src/Components/Gene/GeneRelatedLinks.tsx
zephraph: Certainly possible. I'll wrap it in a check.

sweir27:
Are you planning to follow up with a test to verify the display logic + an update to add the markup here?

Other than those notes, it looks 👍 .

zephraph:
I’ve got tests, just haven’t pushed yet. Will do that in a bit.

src/Components/Gene/GeneRelatedLinks.tsx
```diff
@@ -0,0 +1,106 @@
...
+              name
+            }
+          }
+        }
+      }
+    `,
+  }
+)
+
+// export const GeneRElatedArtistLinksFragmentContainer
```
zephraph:
```suggestion
```

artsyit:
:rocket: PR was released in v22.3.0 :rocket:

PR #3021, opened 2019-11-19
Remove entry points impression event from CollectionsHubsNav

xtina-starr:
This removes the entry points impression event from `src/Components/v2/CollectionsHubsHomepageNav.tsx`. We remove it here because it fires too early and should be fired it from [force](https://github.com/artsy/force/pull/4784) upon the user scrolling within a specific area.

### DIFF
--- a/src/Components/v2/CollectionsHubsHomepageNav.tsx
+++ b/src/Components/v2/CollectionsHubsHomepageNav.tsx
@@ -14,15 +14,10 @@ interface CollectionsHubsHomepageNavProps {
 }
 
 export const CollectionsHubsHomepageNav = track(
-  {
-    context_page: AnalyticsSchema.PageName.HomePage,
-    context_module: AnalyticsSchema.ContextModule.CollectionHubEntryPoint,
-    subject: AnalyticsSchema.Subject.FeaturedCategories,
-  },
+  {},
   { dispatch: data => Events.postEvent(data) }
 )((props: CollectionsHubsHomepageNavProps) => {
   const { trackEvent } = useTracking()
-  trackEvent({})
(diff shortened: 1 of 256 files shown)


xtina-starr:
I'm not sure if these generated files in this PR should be added. @pepopowitz @ashleyjelks thoughts?

pepopowitz:
> I'm not sure if these generated files in this PR should be added. @pepopowitz @ashleyjelks thoughts?

Yeah, something seems up with that. I'm not sure I understand where those came from.

src/Components/v2/CollectionsHubsHomepageNav.tsx
```diff
@@ -18,7 +18,6 @@ export const CollectionsHubsHomepageNav = track(
     context_page: AnalyticsSchema.PageName.HomePage,
     context_module: AnalyticsSchema.ContextModule.CollectionHubEntryPoint,
     subject: AnalyticsSchema.Subject.FeaturedCategories,
-    action_type: AnalyticsSchema.ActionType.Impression,
```
pepopowitz: Is this change removing the actual impression event, or just removing metadata and the event is still firing? (It's possible that there's a second file that changed, I'm not seeing it right now with the relay noise though.)

ashleyjelks: Plus one that we can remove these generated files from the PR. I was looking at the PR in Force and I'm wondering why both of these tracking events are needed, both here in Reaction and there in Force?

xtina-starr:
Closing this in favor of https://github.com/artsy/reaction/pull/3025.

PR #3025, opened 2019-11-20
Remove impression event

xtina-starr:
This removes the entry points impression event from `src/Components/v2/CollectionsHubsHomepageNav.tsx`. We remove it here because it fires too early and should be fired from [force](https://github.com/artsy/force/pull/4784) upon the user scrolling within a specific area.

I closed https://github.com/artsy/reaction/pull/3021 in favor of this PR to not include the generated files.

### DIFF
--- a/src/Components/v2/CollectionsHubsHomepageNav.tsx
+++ b/src/Components/v2/CollectionsHubsHomepageNav.tsx
@@ -14,16 +14,10 @@ interface CollectionsHubsHomepageNavProps {
 }
 
 export const CollectionsHubsHomepageNav = track(
-  {
-    context_page: AnalyticsSchema.PageName.HomePage,
-    context_module: AnalyticsSchema.ContextModule.CollectionHubEntryPoint,
-    subject: AnalyticsSchema.Subject.FeaturedCategories,
-    action_type: AnalyticsSchema.ActionType.Impression,
-  },
+  {},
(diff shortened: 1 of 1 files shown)


src/Components/v2/CollectionsHubsHomepageNav.tsx
```diff
@@ -14,16 +14,10 @@ interface CollectionsHubsHomepageNavProps {
 }
 
 export const CollectionsHubsHomepageNav = track(
-  {
-    context_page: AnalyticsSchema.PageName.HomePage,
```
pepopowitz: I'm not sure how accurately I understand the tracking hooks, so I raise this as more of a question than a suggestion.

I _think_ we still want to leave lines 17 to 20. 

My understanding of how the `track(...)` function works here is that any properties we set in this object will be used for any calls to `trackEvent` inside this function. The `context_page`, `context_module`, and `subject` all seem important enough to apply to all tracked events in this component. 

By removing these lines, I _think_ that the event tracked on line 46 would end up sending blank values for each of those properties, which I don't think is what we want.

Does that sound correct, or do you think my understanding of the tracking helpers is incorrect?

pepopowitz:
Thank you for making changes! Looks great :shipit:

artsyit:
:rocket: PR was released in v22.3.1 :rocket:

PR #3027, opened 2019-11-21
Grow-1479 Blank thumbnail on collections with generated banner 

jiahaodavid:
Address : [Grow 1479](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1479)

Screen Shot 2019-11-21 at 13 46 00

Object: If there is no custom image header, it shows a gray circle. 

Those circle things are using `EntityHeader` from Palette. In Palette, we have [this](https://github.com/artsy/palette/blob/0ce2ed7da2cdce43bbad51709ae4bebe4545bf03/packages/palette/src/elements/EntityHeader/EntityHeader.tsx#L49). And then, the `EntityHeader` actually import `avatar`. Inside the avatar, it has condition checking(see [here](https://github.com/artsy/palette/blob/0ce2ed7da2cdce43bbad51709ae4bebe4545bf03/packages/palette/src/elements/Avatar/Avatar.shared.tsx#L66)), if it has imgURL, then it shows headerIMG, otherwise, it shows `initials string`. 
So, I just pass an empty space to `initials string`. It will render a gray circle.

### DIFF
--- a/src/Apps/Collect2/Routes/Collections/Components/CollectionsGrid.tsx
+++ b/src/Apps/Collect2/Routes/Collections/Components/CollectionsGrid.tsx
@@ -59,6 +59,7 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
                   <Media greaterThan="xs">{index < 3 && <Separator />}</Media>
 
                   <EntityHeader
+                    initials=" "
                     py={2}
                     href={`/collection/${collection.slug}`}
                     imageUrl={imageUrl || undefined}
--- a/src/Apps/Collect2/Routes/Collections/Components/CollectionsGrid.tsx
+++ b/src/Apps/Collect2/Routes/Collections/Components/CollectionsGrid.tsx
@@ -58,6 +58,7 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
                   <Media at="xs">{index === 0 && <Separator />}</Media>
                   <Media greaterThan="xs">{index < 3 && <Separator />}</Media>
 
+                  {/* A empty space in initials to generate a blank avatar when there's no imageUrl*/}
                   <EntityHeader
                     initials=" "
                     py={2}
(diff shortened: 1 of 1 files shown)


src/Apps/Collect2/Routes/Collections/Components/CollectionsGrid.tsx
```diff
@@ -59,6 +59,7 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
                   <Media greaterThan="xs">{index < 3 && <Separator />}</Media>
 
                   <EntityHeader
+                    initials=" "
```
pepopowitz: Neat! 👍 

Could we add a comment above line 61, to make it clear to future readers of this code that we are passing `initials=" "` so that the gray circle shows up? We do have this PR to help them figure that out, but this code "seems strange" - enough that I could easily see someone deleting it without investigating why it was written.

Commenting JSX is kind of tricky, if you haven't seen it before it looks like this: 

```jsx
{/* A JSX comment */}
<EntityHeader />
...
```

It'd be great if we could put it directly above line 62, but I don't think we can put it inside an element...so right before the `<EntityHeader` line would be adequate.

jiahaodavid: Indeed. That empty space looks strange. Thanks for pointing out.  🤝

src/Apps/Collect2/Routes/Collections/Components/CollectionsGrid.tsx
```diff
@@ -58,7 +58,9 @@ export class CollectionsGrid extends Component<CollectionsGridProps> {
                   <Media at="xs">{index === 0 && <Separator />}</Media>
                   <Media greaterThan="xs">{index < 3 && <Separator />}</Media>
 
+                  {/* A empty space in initials to generate a blank avatar when there's no imageUrl*/}
```
pepopowitz:
```suggestion
                  {/* An empty space in initials to generate a blank avatar when there's no imageUrl */}
```

Completely minor nit-pick for grammar & consistency of spacing. Feel free to disregard _or_ accept.

artsyit:
:rocket: PR was released in v22.7.2 :rocket:

PR #3028, opened 2019-11-21
Track validation errors and registration submitted events

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-465

It adds the `<OnSubmitValidationError>` component to rack validation errors from Formik.

@dblandin @erikdstock I'm not sure if I'm using it right, but let me know if there's anything I should update.

## Tracking validation errors

Screen Shot 2019-11-21 at 2 27 59 PM

## Tracking `registration submitted`

Screen Shot 2019-11-21 at 3 08 32 PM

### DIFF
--- a/src/Apps/Auction/Components/RegistrationForm.tsx
+++ b/src/Apps/Auction/Components/RegistrationForm.tsx
@@ -7,20 +7,13 @@ import { ErrorModal } from "Components/Modal/ErrorModal"
 import { Form, Formik, FormikActions, FormikProps } from "formik"
 import React, { useEffect, useState } from "react"
 import {
-  CardElement,
   Elements,
   injectStripe,
   ReactStripeElements,
   StripeProvider,
 } from "react-stripe-elements"
 import { data as sd } from "sharify"
-import styled from "styled-components"
 import * as Yup from "yup"
-
-export const StyledCardElement = styled(CardElement)`
-  width: 100%;
-  padding: 9px 10px;
-`
 
 export interface FormResult {
   token: stripe.Token
@@ -125,7 +118,7 @@ const validationSchema = Yup.object().shape({
   ),
 })
 
-type TrackErrors = (errors: string[]) => void
+export type TrackErrors = (errors: string[]) => void
 
 /*
   This component exists only to capture formik's renderProps and track form
@@ -139,9 +132,9 @@ type TrackErrors = (errors: string[]) => void
   Background:
     https://github.com/jaredpalmer/formik/issues/1484#issuecomment-490558973
  */
-const OnSubmitValidationError: React.FC<{
+export const OnSubmitValidationError: React.FC<{
   cb: TrackErrors
-  formikProps: FormikProps<FormValues>
+  formikProps: Partial<FormikProps<FormValues>>
 }> = props => {
   const { cb, formikProps } = props
 
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -20,6 +20,7 @@ import { track } from "Artsy"
 import * as Schema from "Artsy/Analytics/Schema"
 import { useTracking } from "Artsy/Analytics/useTracking"
 import { FormikActions } from "formik"
+import { isEmpty } from "lodash"
 import qs from "qs"
 import React, { useEffect, useState } from "react"
 import { Title } from "react-head"
@@ -54,6 +55,7 @@ const MAX_POLL_ATTEMPTS = 20
 
 export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
   let pollCount = 0
+  let registrationTracked = false
 
   const { artwork, me, relay, stripe } = props
   const { saleArtwork } = artwork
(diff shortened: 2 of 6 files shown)


src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -394,8 +394,17 @@ describe("Routes/ConfirmBid", () => {
       await page.agreeToTerms()
       await page.submitForm()
 
-      expect(mockPostEvent).toHaveBeenCalledTimes(1)
-      expect(mockPostEvent).toHaveBeenCalledWith({
+      expect(mockPostEvent).toHaveBeenCalledTimes(2)
+      expect(mockPostEvent.mock.calls[0][0]).toEqual({
```
dleve123: small thing: I would recommend updating the doc string to reflect that we're tracking 2 events: bid and registration

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -438,8 +447,17 @@ describe("Routes/ConfirmBid", () => {
       await page.agreeToTerms()
       await page.submitForm()
 
-      expect(mockPostEvent).toHaveBeenCalledTimes(1)
-      expect(mockPostEvent).toBeCalledWith({
+      expect(mockPostEvent).toHaveBeenCalledTimes(2)
+      expect(mockPostEvent.mock.calls[0][0]).toEqual({
```
dleve123: oh this is interesting, so we're simulating the case where registration succeeds, but bidding fails. I'd recommend updating the docstring here to make this more clear.

dleve123: Also, are we testing the case when registration fails when attempting to bid+register?

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -65,6 +67,9 @@ export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
     me as any
   )
 
+  let bidderId: string | null =
```
dleve123: nice refactoring.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -54,6 +55,7 @@ const MAX_POLL_ATTEMPTS = 20
 
 export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
   let pollCount = 0
+  let registrationTracked = false
```
dleve123: probably a newbie question - but why the need to maintain state here like this?

src/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -139,9 +132,9 @@ type TrackErrors = (errors: string[]) => void
   Background:
     https://github.com/jaredpalmer/formik/issues/1484#issuecomment-490558973
  */
-const OnSubmitValidationError: React.FC<{
+export const OnSubmitValidationError: React.FC<{
   cb: TrackErrors
-  formikProps: FormikProps<FormValues>
+  formikProps: Partial<FormikProps<FormValues>>
```
erikdstock: I wonder if this form should be moved up to a more common/shared directory- it seems like other forms may want to use it as well.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
erikdstock: Also wondering this (I think it has already been explained to me in some other context but why not use `useState`?

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
yuki24: We can't track registration failed because it's all done [by Gravity](https://github.com/artsy/gravity/blob/5985f0f43078699d7ff81d963e08286f6d3db32f/app/api/v1/bidder_positions_endpoint.rb#L65-L72) in the single bidder position endpoint and errors from Gravity will be translated to a GraphQL error in Metaphysics.

src/Apps/Auction/Routes/ConfirmBid/index.tsx
yuki24: > why the need to maintain state here like this?

The user would stay on the same page if the user wasn't registered, tried to bid and register but the user was outbid. This way the user will be able to place another bid without a lot of distraction (and also the simplest we could implement). We want to avoid calling the `registration submitted` event more than once in such a case.

> why not use `useState`?

We don't want the whole component to get re-rendered on every internal state update so components like `<OnSubmitValidationError>` won't fire the callback too many times and React won't run unnecessary VDOM diff-ing.

src/Apps/Auction/Components/RegistrationForm.tsx
yuki24: Agreed - there are so many things shared across `Register` and `ConfirmBid` and they have imports from each other, which makes it harder to trace dependencies. Though that would probably be a number of incremental changes, so I wanted to focus on shipping first. cc @dblandin

yuki24:
updated test descriptions and added one more test to cover the case explained [here](https://github.com/artsy/reaction/pull/3028#discussion_r349679423), merging.

artsyit:

PR #3029, opened 2019-11-21
Include partner's name on copy stating bidders agree to Artsy's COS

bhoggard:
Jira: https://artsyproduct.atlassian.net/browse/AUCT-736 🔒 

Screenshot for benefit auction:

Screen Shot 2019-11-21 at 3 43 14 PM

----

Screenshot for commercial auction:

Screen Shot 2019-11-21 at 3 16 32 PM


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarExtraLinks.tsx
```diff
@@ -213,6 +226,10 @@ export const ArtworkSidebarExtraLinksFragmentContainer = createFragmentContainer
         }
         sale {
           is_closed: isClosed
+          is_benefit: isBenefit
```
sweir27: Why not just use `isBenefit` here? Doesn't seem like we gain much by aliasing. (I believe `is_closed: isClosed` was written that way for convenience during the migration 😄).

src/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarExtraLinks.test.tsx
```diff
@@ -32,13 +33,30 @@ describe("ArtworkSidebarExtraLinks", () => {
...
+      )
+    })
+  })
+
   describe("for work in an auction", () => {
     beforeAll(async () => {
       wrapper = await getWrapper(LiveAuctionArtwork)
     })
-    it("displays proper text", () => {
+    it("displays proper conditions of sale text", () => {
```
sweir27: Since you have a bit of logic around the partner name ending with `s`, it would be good to add a test exercising that case!

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarExtraLinks.tsx
yuki24: Yeah I agree. Given that we have already migrated to MP v2 there isn't a lot of reasons we should continue to use lower case fields.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarExtraLinks.tsx
```diff
@@ -104,10 +104,23 @@ class ArtworkSidebarExtraLinksContainer extends React.Component<
...
   }
 
+  conditionsOfSaleText() {
+    const first = "By placing your bid you agree to Artsy's "
+    if (!this.props.artwork.sale.isBenefit) {
+      const partnerName = this.props.artwork.sale.partner.name
+      const possessivePartnerName = partnerName.endsWith("'s")
+        ? partnerName
+        : partnerName + "'s"
+      return first + "and " + possessivePartnerName + " "
```
dblandin: Small grammar nit but I think that in this case we want to treat Artsy and the partner as a single unit having joint possession over the Conditions of Sale.

For example:

> By placing this bid you agree to Artsy and Seuol Auction's Conditions of Sale.

https://brians.wsu.edu/2016/05/19/joint-possessives/

bhoggard: I talked this over with Nicole B, and decided to leave it this way. Effectively they're agreeing the the _respective_ COS of 2 organizations.

dblandin: Gotcha. Makes sense! In that case, do we link to the partner's COS as well?

bhoggard: No we don't. We have a blob of text on some auction pages if the partner wants us to do that.

artsyit:

PR #3033, opened 2019-11-26
User can see buyer's premium cost on web

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-675

This PR wires up the `calculatedCost` field and the `<PriceTransparency>` component to displays the costs to the user. It's a bit tricky to SSR the entire page since the `selectedBid` isn't always known due to direct links (e.g. https://staging.artsy.net/auction/barry-lot-placed-on-block-mocktion/bid2/sanjay-theodore-transcendence) so I just pulled it to a client-only call. Testing is also tricky since we need "re-fetch" behaviour on the max bid amount update, and I slightly updated the test setup.

## Todos

 * [ ] Add a feature flag `ENABLE_PRICE_TRANSPARENCY` to Force staging and production

## Screenshot

AUCT-675


src/DevTools/createTestEnv.tsx
```diff
@@ -147,13 +149,14 @@ class TestEnv<MutationNames extends string, TestPage extends RootTestPage> {
...
     this.mutations.mockFetch.mockImplementation(wrappedFetchQuery)
+    this.mockQuery.mockImplementation(wrappedFetchQuery)
 
     // Switch on mutation/query when making requests to help make assertions
     // Seems we only make assertions about mutations right now
     const mockNetwork = Network.create((operation, variableValues) => {
       return operation.operationKind === "mutation"
         ? this.mutations.mockFetch(operation, variableValues)
-        : wrappedFetchQuery(operation, variableValues)
+        : this.mockQuery(operation, variableValues)
```
yuki24: @ds300  I wasn't sure how we should be testing "re-fetch" behavior so I just added the `mockQuery` variable to give us flexibility to update mocks in an `it` block. I wasn't sure if this is the intended approach and would like your eyes on this so we can make the testing foundation better while maintaining the original design and philosophy.

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -111,13 +113,15 @@ export const determineDisplayRequirements = (
...
+  artworkSlug,
   initialSelectedBid,
   me,
   onSubmit,
+  relay,
   saleArtwork,
-  showPricingTransparency = false,
   trackSubmissionErrors,
 }) => {
+  const { ENABLE_PRICE_TRANSPARENCY } = sd as any
```
dblandin: You should be able to use the `GlobalData` type, defined here once you add the `ENABLE_PRICE_TRANSPARENCY` mapping: https://github.com/artsy/reaction/blob/master/typings/sharify.d.ts

src/Apps/Auction/Components/PricingTransparency.tsx
```diff
@@ -65,3 +78,53 @@ export const PricingTransparency: React.FC<{}> = () => {
     </Flex>
   )
 }
+
+export const PricingTransparency = withSystemContext(
```
dblandin: I think the pattern used elsewhere is to call this `PricingTransparencyQueryRenderer` and alias it if neccessary at the call site, such as:

```typescript
import { PricingTransparencyQueryRenderer as PricingTransparency } from "Apps/Auction/Components/PricingTransparency"
```

I think that be a bit more explicit than this and the underscore prefixed `_PricingTransparency` naming. What do you think?

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -124,6 +132,8 @@ const setupTestEnv = ({
 
 describe("Routes/ConfirmBid", () => {
   beforeEach(() => {
+    ;(sd as any).ENABLE_PRICE_TRANSPARENCY = false
```
dblandin: Similar comment to above. Once we add the mapping of `ENABLE_PRICE_TRANSPARENCY` to the `GlobalData` type, we might not need this `as any`.

https://github.com/artsy/reaction/blob/master/typings/sharify.d.ts

src/DevTools/createTestEnv.tsx
ds300: Looks great! Thanks!

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
yuki24: I've also updated the tests as part of https://github.com/artsy/reaction/pull/3033/commits/4a5361fd563dc066cc0f74b3da60791e19a851d5 since now it's a `readonly` field that can't be mutated, and mutating it isn't good anyway, so I just switched to a jest mock.

src/Apps/Auction/Components/PricingTransparency.tsx
yuki24: done: https://github.com/artsy/reaction/pull/3033/commits/ce0c2b3bc2db8027a1099b20c466af794f7e2217

src/Apps/Auction/Components/__stories__/PricingTransparency.story.tsx
```diff
@@ -6,7 +6,23 @@ import { PricingTransparency } from "../PricingTransparency"
...
+              buyersPremium: {
+                display: "$1,000",
+              },
+              subtotal: {
+                display: "$6,000",
+              },
+            },
+          },
+        }}
+      />
```
yuki24: @erikdstock we might want to update this storybook to use a query renderer as well, but it's been too long since I touched storybooks in Reaction. Let me know if you want to pair on this.

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -201,7 +201,7 @@ export const BidForm: React.FC<Props> = ({
                     </Sans>
                   )}
 
-                  {ENABLE_PRICE_TRANSPARENCY && (
+                  {!ENABLE_PRICE_TRANSPARENCY && (
```
yuki24: Wait, I didn't mean to add this here...

artsyit:
:rocket: PR was released in v22.8.0 :rocket:

PR #3034, opened 2019-11-27
Track "selected max bid" events on the confirm bid page

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-773

## Screenshot

Screen Shot 2019-11-27 at 3 36 57 PM

### DIFF
--- a/src/Artsy/Analytics/Schema/AuctionInfo.ts
+++ b/src/Artsy/Analytics/Schema/AuctionInfo.ts
@@ -12,10 +12,19 @@ export interface AuctionInfo {
    * but occasionally it is a fixed bid amount.
    */
   bidder_position_id: string
+
   /**
    * List of reasons why there's a failure event
    *
    *  Used by the auction registration flow.
    */
   error_messages: string[]
+
+  /**
+   * The amount of max bid the user selected in cents. For example, if the user
+   * selected $5,000.00, it will be reported as 500000.
+   *
+   *  Used by the auction confirm bid flow.
+   */
+  selected_max_bid_minor: string
 }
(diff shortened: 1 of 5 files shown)


src/Artsy/Analytics/Schema/AuctionInfo.ts
```diff
@@ -12,10 +12,19 @@ export interface AuctionInfo {
...
    */
   error_messages: string[]
+
+  /**
+   * The amount of max bid the user selected in cents. For example, if the user
+   * selected $5,000.00, it will be reported as 500000.
+   *
+   *  Used by the auction confirm bid flow.
+   */
+  selected_max_bid_minor: string
```
yuki24: @dblandin calling this `_minor` to align with the rest of Artsy's naming conventions for currencies. Not sure analytics also applies but just making sure we are not adding any more diverse names.

dblandin: Sounds good but I think there's an open question of whether we want the convention to be `minor` or `minor_units`. @alloy made a good point towards being more explicit with `minor_units`.

I don't think we should block this PR but we should decide one way or another soon and be consistent going forward.

cc/ @starsirius

artsyit:
:rocket: PR was released in v22.9.0 :rocket:

src/Artsy/Analytics/Schema/AuctionInfo.ts
alloy: https://www.notion.so/artsy/Currencies-Task-Force-6c91c37091304189890f90466b72463f#470f00ee9a814c9390b29670c8d7ed07

PR #3035, opened 2019-11-27
Fixes a bug where your max bid is not pre-selected on the new confirm bid page

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-774

It seems like the actual prop in [`found`](https://github.com/4Catalyzer/found) looks like `props.match.location` instead of `props.location`. The test was passing because it was injecting to `props.location`, but in the actual browser it was actually missing. I updated it to properly confirm the interface. I also dropped the `qs` dependency since `found` already does this and we don't need to parse it out by ourselves.


artsyit:
:rocket: PR was released in v22.9.1 :rocket:

PR #3037, opened 2019-12-02
PURCHASE-1649: Adds data-test attribute to the sale message on artwork view

sweir27:
Adds a specific identifier to this element so we can select it in our integrity tests!

image


sepans:
🙏 

@lilyfromseattle had a comment in the [ticket](https://artsyproduct.atlassian.net/browse/PURCHASE-1649) that sale message appears next to radio buttons for edition set artworks. Is that going to be in a different PR?

sweir27:
Updated! Will #mergeOnGreen

artsyit:
:rocket: PR was released in v22.9.2 :rocket:

PR #3038, opened 2019-12-03
Modifies waypoint method call

ashleyjelks:
Ad server analytics tracking still wasn't firing as expected. This PR modifies the tracking event to fire as the waypoint leaves the viewport, rather than on entering, with an offset added to push the waypoint's boundaries up now that the event is firing on leave.

[Links to GROW-1678](https://artsyproduct.atlassian.net/browse/GROW-1678)

Firing Events Before Fix:
Kapture 2019-12-03 at 14 52 15

Firing Events After Fix:
Kapture 2019-12-03 at 14 47 58


artsyit:
:rocket: PR was released in v22.9.5 :rocket:

PR #3039, opened 2019-12-03
Handle missing partner on artwork page

sweir27:
Noticed while looking into Google Search Console issues that we were throwing a 500 on a few pages due to this error: https://sentry.io/organizations/artsynet/issues/1247032632/?project=28316&query=is%3Aunresolved&statsPeriod=14d

This should fix! (So would null checks cc @zephraph 😄)


artsyit:
:rocket: PR was released in v22.9.3 :rocket:

PR #3036, opened 2019-12-02
Remove dots from the artist images

ashkan18:
# Problem

https://artsyproduct.atlassian.net/browse/PURCHASE-1678

# Selfie
image

Removed dots by passing `pageDots` options from artists page and changed the height to 180 to reflect the new size.

cc: @samchieng

PR #3044, opened 2019-12-05
Bring back arrows removed by #3036

ashkan18:
# Problem
Arrows for artwork images on artist page are gone after #3036 

# Cause
We removed `HorizontalPadding` around carousel to remove padding from carousel on mobile, but removing that we also removed it from desktop which didn't change the look when we looked at the previous PR results but looks like it makes carousel going over the arrow icon.

# Solution
Wrap desktop version of the carousel with `HorizontalPadding`.

# Selfie
image


artsyit:
:rocket: PR was released in v22.9.7 :rocket:

PR #3049, opened 2019-12-09
Bump version of lazy-load lib to fix SSR issue

zephraph:
Fixes an issue w/ `react-lazy-load-image-component` where lazy loading is broken when using intersection observers because of a faulty SSR check. 

cc @pepopowitz


artsyit:
:rocket: PR was released in v22.10.2 :rocket:

PR #3050, opened 2019-12-10
[GROW-1633] remove duplicate hover analytics event

xtina-starr:
Addresses: [GROW-1633](https://artsyproduct.atlassian.net/browse/GROW-1633)

When a logged in user hovers over the notification bell we fire two hover analytics events. This PR removes the hover event located in the NavBar component and relies solely on the hover tracking event in the NotificationsBadge component. The event in the NavBar is dependent on a cookie that is calculated and set in the NotificationsBadge so I think we can assume that's the best source of truth.

### DIFF
--- a/src/Components/NavBar/NotificationsBadge.tsx
+++ b/src/Components/NavBar/NotificationsBadge.tsx
@@ -1,9 +1,8 @@
 import { Box, color, Flex, Sans } from "@artsy/palette"
 import { NotificationsMenuQueryResponse } from "__generated__/NotificationsMenuQuery.graphql"
-import { AnalyticsSchema, SystemContext } from "Artsy"
-import { useTracking } from "Artsy/Analytics/useTracking"
+import { SystemContext } from "Artsy"
 import cookie from "cookies-js"
-import React, { useContext, useEffect } from "react"
+import React, { useContext } from "react"
 import styled from "styled-components"
 import { get } from "Utils/get"
 import createLogger from "Utils/logger"
@@ -11,12 +10,7 @@ import { NotificationsQueryRenderer } from "./Menus"
 
 const logger = createLogger("Components/NavBar")
 
-export const NotificationsBadge: React.FC<{
-  /**
-   * If hovering over the nav item, `hover` is passed into the badge (Overlay)
-   */
-  hover?: boolean
-}> = ({ hover }) => {
+export const NotificationsBadge: React.FC<{}> = () => {
   const isClient = typeof window !== "undefined"
   return isClient ? (
     <NotificationsQueryRenderer
@@ -73,11 +67,7 @@ export const NotificationsBadge: React.FC<{
 
         return (
           <Box>
-            <CircularCount
-              count={displayCount}
-              rawCount={totalUnread}
-              hover={hover}
-            />
+            <CircularCount count={displayCount} />
           </Box>
         )
       }}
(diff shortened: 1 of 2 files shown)


anandaroop:
Makes sense, but looking over this a question occurs to me —  does the `NotificationsBadge` component actually render at all if there are not any pending notifications? (I'm looking at [this short-circuit return](https://github.com/artsy/reaction/blob/Grow_1378/src/Components/NavBar/NotificationsBadge.tsx#L110-L112))

If not, I think that means we'll go from double-firing events to not firing events at all in the case where there's no badge / msg count. 

If that's correct maybe we want the `NavBar` hover event to be the source of truth?

xtina-starr:
@anandaroop that's a good catch. I just tested the PR'd code and the event does not fire a hover analytics event if there are no notifications. In this case does we think it's better to have a double fire or should I remove the track event in the NotificationBadge?

cc @pepopowitz

pepopowitz:
Good catch, @anandaroop!

@xtina-starr If the tracking is consistent and accurate when it is fired from `NavBar`, let's invert to tracking there instead of in `NotificationsBadge`. If it doesn't fire consistently or accurately there, we're going to have to do more work....and I think this would then be a case where we should be using the React pattern of elevating state to the nearest common ancestor. In this case, that would mean moving the notification count management into the `NavBar` (and possibly extracting it to a custom hook in doing so, if that makes sense to do).

xtina-starr:
@pepopowitz @anandaroop do you all have any feedback on the latest changes to this PR?

pepopowitz:
I think it looks good! I'll wait for Roop's review before merging.

src/Components/NavBar/NotificationsBadge.tsx
```diff
@@ -92,35 +82,16 @@ const CircularCount: React.FC<{
...
   if (!notificationsLabel) {
     return null
   }
 
-  useEffect(() => {
-    if (hover) {
-      trackEvent({
-        action_type: AnalyticsSchema.ActionType.Hover,
-        subject: AnalyticsSchema.Subject.NotificationBell,
-        new_notification_count: rawCount,
```
anandaroop: Assuming it's ok to lose this flavor of event altogether, this LGTM 👍

xtina-starr: I tried the event with and without notifications. So I think so.

pepopowitz: I'm trying to think of a scenario where the notification count tracked by `NavBar` would be different than what the user actually saw, due to the cookie not being updated yet when they hovered. 

I *think* it's possible that the server could render a bell with 0 notifications, and while the [client request](https://github.com/artsy/reaction/pull/3050/files#diff-dbe27717b00f02b52447098fe8e92d13L22) was still waiting for a response from the server, the user hovered....in which case we would track that the user hovered over the bell with 0 notifications. But that would still be a correctly tracked event, since there were 0 listed *when they hovered*.

Are there any other scenarios either of you can dream up where communicating state between `NotificationsBadge` and `NavBar` via cookies would cause inaccurate events to be tracked?

pepopowitz: In general I think the way these components **should** be communicating state is with the parent (NavBar) managing it and passing it down. What we have right now, communicating state via cookies, seems to work...and it's less work than rewriting to manage state at the higher level. I'm good with this if we don't think we're opening ourselves up for weird bugs in terms of tracking the wrong thing.

anandaroop: Ah, I hadn't even considered these subtleties. 

(Above I was responding to the [mistaken, I see now] impression that were were losing the `AnalyticsSchema.Subject.NotificationBell` type of hover event. But this is in fact still fired by the remaining code.)

@pepopowitz I think your take (i.e. that the edge case-y event would "still be a correctly tracked event") makes sense — I'm good with merging this.

xtina-starr: Okay in that case I'm going to go ahead and merge this.

artsyit:
:rocket: PR was released in v23.0.2 :rocket:

PR #3051, opened 2019-12-10
Coverage reporting

eessex:
Adds coverage reporting via the codecov orb, which is uploaded on test runs here: https://codecov.io/gh/artsy/reaction

This app has no specific configuration for codecov, so settings will be inherited from our org settings yml at codecov.io.

Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-1983


eessex:
Coverage is looking pretty good on the first run, we are at 82.1% -- detail visible here: https://codecov.io/gh/artsy/reaction/branch/coverage

artsyit:
:rocket: PR was released in v23.0.5 :rocket:

PR #3052, opened 2019-12-10
Updates to display ad tracking

pepopowitz:
This PR corrects the impression tracking of display ads.

Impressions of ads will only be tracked once, when the user first views them.

I accomplished this by removing our use of waypoints, and relying on Google to manage when the ad is viewed, with the `onImpressionViewable` event. The rules Google uses to identify if an ad is viewable are [in their docs](https://support.google.com/admanager/answer/4524488?hl=en&visit_id=637116147923101672-4119934528&rd=1). 

If a user scrolls very quickly past an ad, it is possible that we *won't* track an impression for it, but that seems correct to me, since they didn't truly look at it.

Along the way, I refactored the DisplayAd component to be a functional component instead of a class, as I found that easier to work with. 

~Tests are also broken; I'll fix those before this PR is ready for merge.~ Tests are fixed now!

Note: Support for tracking *clicks* on ads will be covered in a future PR.

### DIFF
--- a/src/Components/Publishing/Display/DisplayAd.tsx
+++ b/src/Components/Publishing/Display/DisplayAd.tsx
@@ -1,10 +1,10 @@
 import { Box, color, Flex, FlexProps, Sans } from "@artsy/palette"
 import { track } from "Artsy/Analytics"
 import * as AnalyticsSchema from "Artsy/Analytics/Schema"
+import { useTracking } from "Artsy/Analytics/useTracking"
 import { is300x50AdUnit } from "Components/Publishing/Display/DisplayTargeting"
 import { AdDimension, AdUnit } from "Components/Publishing/Typings"
-import { once } from "lodash"
-import React from "react"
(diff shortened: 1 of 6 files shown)


ashleyjelks:
This looks good so far!


For the click tracking I took a look at the react-gpt docs and I wonder if a possible approach is to use the library's event listeners to detect a click event and then to call the tracking method in the callback?



```
removeListener(eventType, cb) - Removes the specified listener from the listener array for the specified eventType.
removeAllListeners([eventType]) - Removes all listeners, or those of the specified eventType.
```

pepopowitz:
@ashleyjelks This one's all set, pending review! I'm going to work on tracking clicks from ads as a separate PR.

ashleyjelks:
LGTM! Let's merge when you're ready @pepopowitz!

artsyit:
:rocket: PR was released in v23.0.3 :rocket:

PR #3054, opened 2019-12-11
Add user agents to buildClientApp and SystemContext

zephraph:
This is a minor update to add more hooks for passing user agents to the relay connection constructor. It will _not_ completely fixed PURCHASE-1140, but it should ensure client side relay requests have the proper UA. 

To completely fix PURCHASE-1140, we need to make sure that all usages of `SystemContextProvider` eventually get the correct prop when called in force.


sweir27:
@zephraph Sorry if I dropped the ball here! I don't remember this PR, but feel free to merge if you think it's still an improvement.

zephraph:
Will revisit when we look at this issue again.

PR #3056, opened 2019-12-12
Update "artist summary" on artist page

lilyfromseattle:
__Before:__
Desktop:
Screen Shot 2019-12-12 at 12 32 58 PM

Mobile:
Screen Shot 2019-12-12 at 12 32 50 PM

__After:__

Desktop with both auction record and representation status:
Screen Shot 2019-12-12 at 11 15 34 AM

Desktop with only one indicator:
Screen Shot 2019-12-12 at 11 10 25 AM

Mobile with only one indicator:
Screen Shot 2019-12-12 at 11 10 39 AM


sweir27:
Would you mind posting a mobile `After` version? (So Sam C can take a look!)

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -269,11 +328,115 @@ const handleOpenAuth = (mediator, artist) => {
...
+
+const renderAuctionHighlight = artist => {
+  if (
+    !artist.auctionResultsConnection ||
+    artist.auctionResultsConnection.edges.length < 1
+  ) {
+    return null
+  }
+  const topAuctionResult =
+    artist.auctionResultsConnection.edges[0].node.price_realized.display
```
sweir27: This would benefit from a `get` statement (using our internal helper). Might mean you can also get rid of/move the check above.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -144,25 +157,38 @@ export class LargeArtistHeader extends Component<Props> {
...
+                  handleOpenAuth(props.mediator, props.artist)
+                }
+              >
+                Follow
+              </FollowArtistButton>
+            </Flex>
           </Flex>
         </Box>
+        <Spacer mb={1} />
+        <Flex flexDirection="row">
```
sweir27: Does this create extra/unwanted spacing if there is no auction highlight or representation status?

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -144,25 +157,38 @@ export class LargeArtistHeader extends Component<Props> {
...
-              user={user}
-              onOpenAuthModal={() =>
-                handleOpenAuth(props.mediator, props.artist)
-              }
-            >
-              Follow
-            </FollowArtistButton>
+            <Flex justifyContent="space-between">
+              {props.artist.counts.follows > 50 && (
+                <Flex flexDirection="column">
```
sweir27: In the spec it looks like these should be centered (but they don't appear that way in the screenshot).

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -242,20 +268,53 @@ export class SmallArtistHeader extends Component<Props> {
...
               }
             >
               Follow
             </FollowArtistButton>
-          </Box>
-        </Box>
+          </Flex>
+        </Flex>
+        <Spacer mb={1} />
+        <Flex flexDirection="row" justifyContent="center">
```
sweir27: Ditto to the question above on spacing.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -269,11 +328,115 @@ const handleOpenAuth = (mediator, artist) => {
...
+}
+
+const CATEGORIES = {
+  "blue-chip": "Blue chip",
+  "top-established": "Established",
+  "top-emerging": "Emerging",
+}
+
+const ICON_MAPPING = {
+  BLUE_CHIP: BlueChipIcon,
```
sweir27: Feels odd to use this format instead of just `"blue-chip"`, etc. as keys here.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -269,11 +328,115 @@ const handleOpenAuth = (mediator, artist) => {
...
+      mr={5}
+    >
+      <AuctionIcon pr={5} />
+      <Sans size="2">{topAuctionResult} Auction Record</Sans>
+    </RoundedFlex>
+  )
+}
+
+const CATEGORIES = {
+  "blue-chip": "Blue chip",
```
sweir27: This looks like it's replicated in [another file](https://github.com/artsy/reaction/blob/29703cd9e891b85b9f13b5907528a3e900bb328c/src/Components/v2/SelectedCareerAchievements.tsx#L19). That's only 2 places but you could easily just export that constant and use it here.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -269,11 +328,115 @@ const handleOpenAuth = (mediator, artist) => {
...
+  if (
+    partnersConnection &&
+    partnersConnection.edges &&
+    partnersConnection.edges.length > 0
+  ) {
+    const highCategory = highestCategory(partnersConnection.edges)
+    const type = highCategory.toUpperCase().replace("-", "_")
+
+    return (
+      <RoundedFlex background={color("black5")} width="auto" py={5} px={10}>
```
sweir27: It seems worth refactoring out this component that shows the icon+label (it may be something that eventually shows up in palette).

It could accept the output of your `ICON_MAPPING` const above and a label and render accordingly.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -269,11 +328,115 @@ const handleOpenAuth = (mediator, artist) => {
...
+const renderRepresentationStatus = artist => {
+  const { highlights } = artist
+  const { partnersConnection } = highlights
+  if (
+    partnersConnection &&
+    partnersConnection.edges &&
+    partnersConnection.edges.length > 0
+  ) {
+    const highCategory = highestCategory(partnersConnection.edges)
+    const type = highCategory.toUpperCase().replace("-", "_")
```
sweir27: If you're only using this to map the icons, you can remove in favor of a different key-value strategy in your constant above.

src/Apps/Artist/Components/__tests__/ArtistHeader.test.tsx
```diff
@@ -39,4 +39,16 @@ describe("ArtistHeader", () => {
...
+    const wrapper = await getWrapper()
+    const html = wrapper.html()
+    expect(html).toContain("Blue chip")
+  })
+
+  it("renders auction record indicator", async () => {
+    const wrapper = await getWrapper()
+    const html = wrapper.html()
+    expect(html).toContain("Auction Record")
+  })
```
sweir27: Can you add tests verifying that if the correct data isn't present, it does not show these indicators?

src/Apps/Artist/Components/ArtistHeader.tsx
lilyfromseattle: I accidentally uploaded the wrong screenshot for mobile my bad

src/Apps/Artist/Components/ArtistHeader.tsx
lilyfromseattle: this actually exists somewhere else. Is it better to import it and leave the mapping as is or change it like you suggest?

src/Apps/Artist/Components/ArtistHeader.tsx
lilyfromseattle: Idk why it doesn't say `outdated` but I removed this spacer

artsyit:
:rocket: PR was released in v24.0.1 :rocket:

PR #3058, opened 2019-12-17
Updates artist page design to match vision

sweir27:
This PR:
- Removes the `shows`, `cv`, and `articles` tabs in favor of putting them on the overview page
- Removes the artworks filter from the overview tab and puts it on a separate `/works-for-sale` page
- Adds some redirect logic:
  - If the artist has no content (bio, insights, etc.), redirect to the `/works-for-sale` tab and show the zero state
  - Adds a new route: `/works-for-sale`
  - All existing sub routes (i.e. `/artist/andy-warhol/cv`) re-route to the root

There are a few things that aren't quite working:
- ~Filters 😭~
- ~An error which I suspect is also related to some incorrect filter setup: ~
image

Turns out the filters were broken due to a local issue with my relay cache, and that warning that I was seeing is expected (cc @mzikherman ).

I will probably set this up as a review app for easier UI testing, but other than those issues above I believe it's good to go!

### Lots of content
image

image

### Some content
image

image

### No content
image


damassi:
@sweir27 - I've noticed that error due to mismatches between case / form wen passing query params and / or route params (e.g., `path: '/artist/:id'`. The casing is often passed verbatim over to elastic search. 

This PR shows what I mean:
- [Here](https://github.com/artsy/reaction/pull/2820/files#diff-6d6dfa2da310c5900f2ced293fc897aeL32), where `artistID` -> `artist_id`. 
- Because [here](https://github.com/artsy/reaction/pull/2820/files#diff-1d31f38d87470c693f29382aad7cbf1aR104), our filter endpoint expects `artist_id` to be passed. 

This took me a sec to figure out when I was refactoring the filter. Down to help brainstorm.

src/Apps/Artist/routes.tsx
```diff
@@ -1,20 +1,63 @@
...
-import { CVRouteFragmentContainer as CVRoute } from "./Routes/CV"
-import {
-  ArtistOverviewQuery,
-  OverviewRouteFragmentContainer as OverviewRoute,
-} from "./Routes/Overview"
-import { ShowsRouteFragmentContainer as ShowsRoute } from "./Routes/Shows"
+import { OverviewRouteFragmentContainer as OverviewRoute } from "./Routes/Overview"
+import { WorksRouteFragmentContainer as WorksRoute } from "./Routes/Works"
+
+graphql`
```
damassi: ?

src/Apps/Artist/Routes/Works/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { ArtistRecommendationsQueryRenderer as ArtistRecommendations } from "Apps/Artist/Routes/Overview/Components/ArtistRecommendations"
+import React, { Component } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { get } from "Utils/get"
+
+export interface WorksRouteProps {
+  artist: Works_artist
+}
+
+export class WorksRoute extends Component<WorksRouteProps> {
```
damassi: `React.FC`

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -36,16 +41,26 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, {}> {
     // no-op
   }
 
+  @track<OverviewRouteProps>(props => ({
+    action_type: Schema.ActionType.Click,
+    subject: "Clicked browse works for sale",
+    destination_path: props.artist.href,
```
sweir27: TODO: This should probably be `/artist/${artistSlug}/works-for-sale`

src/Apps/Artist/routes.tsx
sweir27: This is needed in the `render()` function below in order to determine whether we need to redirect or not.

src/Apps/Order/__tests__/OrderApp.test.tsx
```diff
@@ -67,7 +67,7 @@ describe("OrderApp routing redirects", () => {
     expect(redirect).toBe(undefined)
   })
 
-  it("redirects to the status route if the order is not pending", async () => {
+  it.only("redirects to the status route if the order is not pending", async () => {
```
sweir27: TODO: oops

sweir27:
Forgot to mention that after this PR is in, I'll follow up with another to refactor.

Right now we still have `CV`, `Shows` and `Articles` as "Routes" under the `Artist` folder.

sweir27:
Agh my commit history got all mixed up. Will fix tomorrow!

sweir27:
`Shop works for sale` button:
image

And on mobile
image

src/Apps/Artist/Routes/Overview/index.tsx
```diff
@@ -94,9 +115,17 @@ export class OverviewRoute extends React.Component<OverviewRouteProps, {}> {
 
           {showArtistInsights && (
             <Col sm={4}>
-              <Box pl={2}>
-                <SelectedCareerAchievements artist={artist} />
-              </Box>
+              <Media greaterThan="xs">
+                <Box pl={2}>
+                  <SelectedCareerAchievements artist={artist} />
```
sepans: The difference is only padding?

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -34,6 +35,16 @@ const H1 = styled.h1`
 
 const H2 = H1.withComponent("h2")
 
+export const WorksForSaleButton = styled.div`
```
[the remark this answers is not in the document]
mzikherman: Yea, should be using the Palette primitives over `div` and things like that.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -106,25 +134,39 @@ export class LargeArtistHeader extends Component<Props> {
...
+                      />
+                    </a>
+                  )
+                }}
+              />
+              <WorksForSaleButton>
+                <Button
+                  variant="primaryWhite"
+                  size="small"
+                  onClick={() => this.handleBrowseWorksClick(props.artist.slug)}
```
mzikherman: This came up with client side routing (and we were discussing with @joeyAghion in lunch and learn), but we should try to avoid using custom click handlers, to handle navigation and routing when possible.

This feels like a link to the /works-for-sale tab of an artist page, yet uses `window.location.assign` in an `onClick` (which will break client-side routing), rather than an anchor tag with `href` pointing to that (for users w/o javascript).

src/Apps/Artist/Routes/Overview/index.tsx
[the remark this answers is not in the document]
sweir27: OMG I totally forgot about that! You are so right.

src/Apps/Artist/Components/ArtistHeader.tsx
sweir27: Oh!

I did this so that we could get custom tracking around the button. Is there a solution that satisfies both of these needs?

src/Apps/Artist/__tests__/routes.test.tsx
```diff
@@ -0,0 +1,181 @@
+import { routes_ArtistTopLevelQueryRawResponse } from "__generated__/routes_ArtistTopLevelQuery.graphql"
```
mzikherman: Very nice tests!

src/Apps/Artist/Components/ArtistHeader.tsx
mzikherman: You can still have a no-op `onClick` just for analytics tracking purposes, but the component should probably just be a [styled] anchor tag with `href`.

src/Apps/Artist/Components/NavigationTabs.tsx
```diff
@@ -80,16 +83,50 @@ export class NavigationTabs extends React.Component<Props> {
...
     )
   }
 }
 
+export const hasOverviewContent = ({
+  statuses,
+  related,
+  biographyBlurb,
+}: {
+  statuses: { shows: boolean; cv: boolean; articles: boolean }
```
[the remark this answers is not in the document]
sweir27: I don't believe so (removing them makes the type `any`). I could pull in specific types from the fragment, but this also makes it very clear which arguments are needed which I think is helpful.

damassi:
@mzikherman / @sweir27  - can y'all elaborate on the relay cache issue mentioned in the description?

mzikherman:
@damassi There were two separate issues. One is the expected warnings about edges lacking an `id`, which is a Relay runtime warning.

At the same time @sweir27 was also experiencing issues where the Relay compiler was error'ing out, but the error message didn't make any sense. That was:

```
ERROR:
Parse error: TypeError: Cannot read property 'length' of undefined in […]
```

and the solution was removing the compiler cache
```
$ rm -rf $TMPDIR/RelayFindGraphQLTags-*
```

courtesy of @alloy in https://artsy.slack.com/archives/C2TQ4PT8R/p1572959259008600

alloy:
This warning is I believe a Relay bug. It doesn’t negatively affect things, because these edges never have IDs, but it also shouldn’t complain about it. Iirc, I wrote some of my thoughts on it in the ‘reaction mpv2 migration’ doc, so somebody should take some time to debug and trace the issue in relay-runtime and if indeed a bug submit an upstream patch.

damassi:
Thanks @alloy, @mzikherman  - will give that a look 👍

src/Apps/Artist/ArtistApp.tsx
```diff
@@ -26,21 +37,52 @@ export interface ArtistAppProps {
...
+
   return (
     <AppContainer>
       <ArtistMeta artist={artist} />
-      <Row>
-        <Col>
-          <ArtistHeader artist={artist} />
-        </Col>
-      </Row>
+      {route.displayNavigationTabs && (
```
zephraph: Why is the navigation conditionally rendered?

sweir27:
@zephraph @mzikherman @sepans @lilyfromseattle 

Mind giving this another pass? The main difference between the last version and this one is:
- Previously, I had removed the `/cv`, `/shows`, and `/articles` routes altogether. The desire is to _keep_ them, but just remove the tabs.
- Those ☝️ sub pages only show a back link instead of tabs, so I made that change as well.
- Added a couple of modules on the overview page:
  - Works for Sale (pulls a few artworks and sticks them in a standard carousel)
  - Current shows
  - Articles

sweir27:
Check out artist-big-bang.artsy.net to actually play around!

Some screenshots:
image


image

src/Apps/Artist/ArtistApp.tsx
sweir27: Based on the designs, we either show the navigation tabs or a back-link.

zephraph:
Is the `browse all works for sale` button supposed to be full width on desktop?

src/Apps/Order/OrderApp.tsx
```diff
@@ -22,7 +22,7 @@ declare global {
   }
 }
 
-const findCurrentRoute = ({ routes, routeIndices }: Match) => {
+export const findCurrentRoute = ({ routes, routeIndices }: Match) => {
```
sepans: #minor Worth moving this into a helper file?

src/Apps/Artist/Routes/Works/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+export interface WorksRouteProps {
+  artist: Works_artist
+}
+
+export const WorksRoute: React.FC<WorksRouteProps> = props => {
+  const { artist } = props
+  const { sidebarAggregations } = artist
+
+  const isClient = typeof window !== "undefined"
```
mzikherman: This isn't great since it hardcodes `window`, isn't portable to RN, etc.

Instead, should use a `componentDidMount` if a class method, or a hook otherwise.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -115,25 +145,43 @@ export class LargeArtistHeader extends Component<Props> {
...
+                        width={slide.resized.width}
+                        height={slide.resized.height}
+                        preventRightClick={!isAdmin}
+                      />
+                    </a>
+                  )
+                }}
+              />
+              <WorksForSaleButton pl={4}>
+                <Link
```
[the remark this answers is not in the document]
sweir27: Ahh! That's awesome-- didn't know about the `RouterLink`. Thanks!

src/Apps/Artist/Routes/Overview/Components/WorksForSaleRail.tsx
```diff
@@ -0,0 +1,124 @@
...
+    `,
+  }
+)
+
+export const WorksForSaleRailQueryRenderer: React.FC<{
+  artistID: string
+}> = ({ artistID }) => {
+  const { relayEnvironment } = useContext(SystemContext)
+  return (
+    <QueryRenderer<WorksForSaleRailRendererQuery>
```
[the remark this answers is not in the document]
zephraph: I would be interested in what we could do to help educate / catch instances of this earlier. 

Could we add a warning lint rule?

mzikherman:
High-level meta comment, which only just occurred to me as I was looking over this PR today, which I'll post for posterity, although I'm not sure how applicable this is now.

Basically, this seems like a large change to the current production artist app, and is being done in one shot, which requires a large PR, review app, QA'ing. Also, when ready, it seems like this will replace the current app, so wouldn't be AB tested?

Below is how I think I might approach this, with an eye on making some of those steps easier (and possibly enabling an AB test).

## Altered sequence proposal

- in Reaction, duplicate the `Artist` app into `Artist2`. Then, all updates could be made into `Artist2`. Since this is a new app that wouldn't replace anything, there could then be a series of smaller/WIP PR's here.

- in the Force artist app, pull in `Artist2` instead of `Artist` from Reaction based on some conditional (maybe a lab feature?).

- eventually, the conditional above could change to a proper AB test (if we want to test this)

- in Reaction, once we resolve the AB test and assuming we want to continue with the `Artist2` version, we can delete `Artist` completely and rename `Artist2` to `Artist`.

Again, apologies for the really late comments/feedback. I'm sure I'm missing a bunch of nuance in the above (when you duplicate the `Artist` app initially in Reaction, prob need to append a `2` to the Relay queries, etc.) or if you'd already considered it.

sweir27:
Thanks @mzikherman!

In general I agree with you about incremental rollouts, AB tests, etc.

This is sort of a special case. 😅We'd decided not to AB test because we were confident in wanting to ship the new version as it matches our vision for what the artist page should be.

The original version just moved content from the tabs onto the overview page, so doing it in one go made a bit more sense. However, after viewing it in the review app, we realized that we needed to revisit the drawing board and re-think the design. 

I wanted to put the ✨ new vision ✨ up in a review app quickly so that we can see if we're confident in how it looks (didn't take much work given it was based on prior work + the patterns that you all have set up in this project 😄).

src/Apps/Artist/Routes/Overview/Components/WorksForSaleRail.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+interface WorksForSaleRailProps {
+  artist: WorksForSaleRail_artist
+}
+const HEIGHT = 150
+
+const WorksForSaleRail: React.FC<WorksForSaleRailProps & {
+  onArtworkClicked: () => void
+}> = ({ artist, onArtworkClicked }) => {
+  const { user, mediator } = useContext(SystemContext)
```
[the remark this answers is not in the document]
sweir27: It gets passed into `FillWidthItem` for the save button (perhaps to control a potential auth flow?)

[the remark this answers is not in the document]
mzikherman: Ah, likely. Cool!

src/Apps/Artist/Routes/Overview/Components/WorksForSaleRail.tsx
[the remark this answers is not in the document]
mzikherman: good call, I think you may have brought up before too (when this component was first added).

I think a lint rule makes sense! I guess we're saying that we'd pretty much always prefer `SystemQueryRenderer` and so should have a lint error if someone imports `QueryRenderer`?

mzikherman:
Thanks for the explanation @sweir27 . With an AB test, I think you'd probably have to do it in the slightly more redundant/incremental approach, in order to be able to actually have both versions existing side by side.

But without the need for that, and with the ease today of composing and moving these components around, this is fine too! 👍

src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionEntity.tsx
```diff
@@ -37,7 +37,14 @@ export class ArtistCollectionEntity extends React.Component<CollectionProps> {
       title,
       artworksConnection,
     } = this.props.collection
-    const artworks = artworksConnection.edges.map(({ node }) => node)
+    const artworks = get(artworksConnection, ac =>
```
sweir27: Added this to fix the crash loop we're seeing due to the current elasticsearch issues on staging (would be an error on prod too, so good to fix!)

src/Apps/Artist/ArtistApp.tsx
```diff
@@ -45,21 +59,57 @@ export const ArtistApp: React.FC<ArtistAppProps> = props => {
...
```
sweir27: We store some metadata (`displayNavigationTabs`) on the route so we can use it here to decide whether or not to display the "Back to artist" link or the full set of tabs.

src/Apps/Artist/ArtistApp.tsx
```diff
@@ -87,6 +137,15 @@ export const ArtistApp: React.FC<ArtistAppProps> = props => {
   )
 }
 
+export const StyledLink = styled(RouterLink)`
```
sweir27: This gets re-used throughout the artist app, which is why I chose to export it from here.

We use `RouterLink` so we can avoid full page refreshes when moving between different sub-artist-pages.

src/Apps/Artist/Components/ArtistCollectionsRail/ArtistCollectionsRail.tsx
```diff
@@ -42,16 +43,22 @@ export class ArtistCollectionsRail extends React.Component<
...
-    const { collections } = this.props
+    const { collections, includeTopSpacer } = this.props
     if (collections.length > 3) {
       return (
         <Box>
           <Waypoint onEnter={once(this.trackImpression.bind(this))} />
-          <Sans size="3" weight="medium">
-            Browse by iconic collections
+
+          {includeTopSpacer && <Separator my={3} />}
```
sweir27: Doesn't feel great to have to add this here, but we don't want to show the separator/spacer unless content _loads_, and we only know that from within this element (otherwise we'd have to fetch for data in the parent and avoid any performance benefits we'd get from loading client-side).

src/Apps/Artist/Routes/Works/index.tsx
sweir27: 👍this follows existing patterns in the artist app so will follow up afterwards and fix other instances too.

src/Apps/Artist/routes.tsx
```diff
@@ -24,36 +70,51 @@ export const routes: RouteConfig[] = [
...
+        if (!artist) {
+          return null
+        }
+
+        const showArtistInsights =
+          showMarketInsights(artist) ||
+          (artist.insights && artist.insights.length > 0)
+        const hasArtistContent = hasOverviewContent(artist)
+
+        const alreadyAtWorksForSalePath = match.location.pathname.includes(
```
sweir27: This is where we determine our zero-state routing logic.

If the artist has no "overview-worthy" content, we redirect to the `/works-for-sale` path _unless_ you're already at that path (to avoid infinite looping).

src/Components/v2/SelectedCareerAchievements.tsx
```diff
@@ -100,26 +94,13 @@ export class SelectedCareerAchievements extends React.Component<
       return null
     }
 
-    const Container = props => {
```
sweir27: Found this wasn't used anywhere and added some complexity, so removed for now.

artsyit:
:rocket: PR was released in v25.5.0 :rocket:

PR #3063, opened 2019-12-18
Fix scrolling issue after modal is closed

sepans:
fixes [PURCHASE-1686](https://artsyproduct.atlassian.net/browse/PURCHASE-1686)

**Problem:**

Closing modal leaves the `overflow-y: hidden` on `Body` tag:

Screen Shot 2019-12-06 at 2 18 38 PM

**Solution:**

It turned out the component is unmounted before `this.prop.show` (coming from `state.ts`) is being set to false when closing the modal
In general it is not a bad idea to clean up side effects so added the functions cleaning up to `componentWillUnmount`

Also add a check to avoid setting multiple key event listeners when other props change

### DIFF
--- a/src/Components/Modal/ModalWrapper.tsx
+++ b/src/Components/Modal/ModalWrapper.tsx
@@ -64,17 +64,23 @@ export class ModalWrapper extends React.Component<
   }
 
   componentDidMount() {
-    this.updateBodyScrollBlock()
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in v24.0.2 :rocket:

PR #3064, opened 2019-12-18
Remove linking artifact

eessex:
Checked the history and it looks like this might be an artifact from the `yarn integrate reaction` process in force-- shouldn't be here!


artsyit:
:rocket: PR was released in v24.0.4 :rocket:

PR #3065, opened 2019-12-18
PURCHASE-1700: Artist Page Tweaks

lilyfromseattle:
[PURCHASE-1700] Artist Page Tweaks

Center-align the number of followers with the label

Blue chip / emerging / established badge should always come before the auction record badge

Add “Representation” after Blue chip / established / emerging on both the top level badges and in career achievements

i.e. “Blue Chip Representation” (Title case on the badge)

“Blue chip representation” (sentence case in career achievements)

Adjust the icon size from 18px to 20px in badges

Add padding-top: 2px to text within badge to better align with the icon

Screen Shot 2019-12-18 at 3 56 27 PM
Screen Shot 2019-12-18 at 3 56 15 PM

[PURCHASE-1700]: https://artsyproduct.atlassian.net/browse/PURCHASE-1700

### DIFF
--- a/src/Apps/Artist/Components/ArtistHeader.tsx
+++ b/src/Apps/Artist/Components/ArtistHeader.tsx
@@ -5,7 +5,6 @@ import { Mediator, SystemContextConsumer } from "Artsy"
 import { track, Track } from "Artsy/Analytics"
 import * as Schema from "Artsy/Analytics/Schema"
 import { FollowArtistButtonFragmentContainer as FollowArtistButton } from "Components/FollowButton/FollowArtistButton"
-import { CATEGORIES } from "Components/v2/ArtistMarketInsights"
 import { Carousel } from "Components/v2/Carousel"
 import React, { Component, Fragment } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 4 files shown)


src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -44,6 +43,12 @@ interface Props {
   mediator?: Mediator
 }
 
+const Categories = {
```
sweir27: This casing seems odd for a constant-- would probably make it `CATEGORIES`.

Also, below you are adding "Representation" to the string. Why not just include that here? (i.e. `"Blue Chip Representation"`)

src/Components/v2/SelectedCareerAchievements.tsx
```diff
@@ -68,12 +68,13 @@ export class SelectedCareerAchievements extends React.Component<
     ) {
       const highCategory = highestCategory(partnersConnection.edges)
       const type = highCategory.toUpperCase().replace("-", "_")
+      const label = CATEGORIES[highCategory] + " representation"
```
sweir27: Same question here as above-- is `CATEGORIES` used anywhere else? If not can you just change the values to include `"representation"`?

artsyit:
:rocket: PR was released in v24.0.6 :rocket:

PR #3070, opened 2019-12-23
[AUCT-812] Fix link in buy now benefit auctions to link to sale, not partner

bhoggard:
sub-task of https://artsyproduct.atlassian.net/browse/AUCT-663


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarPartnerInfo.tsx
```diff
@@ -13,6 +13,15 @@ export class ArtworkSidebarPartnerInfo extends React.Component<
   ArtworkSidebarPartnerInfoProps
 > {
   renderPartnerName() {
+    const sale = this.props.artwork.sale
+    if (sale) {
```
yuki24: Wouldn't this end up always displaying the sale name even when the sale is not a benefit auction (e.g. Phillips auctions)?

bhoggard: No, because `ArtworkSidebarAuctionPartnerInfo` is used for auctions. This case is only for "buy now" sales which are not marked as auctions.

yuki24:
I've resolved the merge conflict and this should be good now. #mergeongreen

artsyit:
:rocket: PR was released in v24.0.10 :rocket:

PR #3075, opened 2020-01-02
PURCHASE-1656: Add phone number field to Address for pickup

lilyfromseattle:
# [PURCHASE-1656] Add phone number field to Address for pickup
https://www.figma.com/file/LieDNuIuKAfhDeEPkjan4p/Checkout-Flow?node-id=0%3A1

This field should be required.

This is a refactor of [this PR](https://github.com/artsy/reaction/compare/master...lilyfromseattle:new_phone_number_field?expand=1) as was suggested in the comments

__Before:__
Screen Shot 2020-01-02 at 11 49 45 AM

__After:__
Screen Shot 2020-01-02 at 11 49 30 AM

[PURCHASE-1656]: https://artsyproduct.atlassian.net/browse/PURCHASE-1656

### DIFF
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -39,6 +39,7 @@ export interface AddressFormProps {
   domesticOnly?: boolean
   showPhoneNumberInput?: boolean
   shippingCountry?: string
+  phoneNumberOnly?: boolean
   errors: AddressErrors
   touched: AddressTouched
 }
@@ -85,7 +86,9 @@ export class AddressForm extends React.Component<
   }
 
   phoneNumberInputDescription = (): string | null => {
-    if (this.props.billing && this.props.showPhoneNumberInput) {
+    if (this.props.phoneNumberOnly) {
+      return "After your order is confirmed, a specialist will contact you within 2 business days to coordinate pickup."
+    } else if (this.props.billing && this.props.showPhoneNumberInput) {
       return null
     } else {
       return "Required for shipping logistics"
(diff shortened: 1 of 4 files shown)


src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -85,7 +86,9 @@ export class AddressForm extends React.Component<
   }
 
   phoneNumberInputDescription = (): string | null => {
-    if (this.props.billing && this.props.showPhoneNumberInput) {
+    if (this.props.phoneNumberOnly) {
```
sweir27: Thoughts on adding a separate component for this phone number input instead of trying to reuse `AddressForm`? There's a lot of conditional logic added, seems like it would be simpler/easier to reason about if it were separate.

lilyfromseattle: I did it that way at first, but I found myself duplicating so much that this seemed like a better option

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -392,6 +416,7 @@ export const ShippingFragmentContainer = createFragmentContainer(
         state
         requestedFulfillment {
           __typename
+
```
sweir27: This seems like it will only fetch the `phoneNumber` for shipping orders-- should `phoneNumber` be here instead of nested under that type?

src/Apps/Order/Components/AddressForm.tsx
sweir27: What was duplicated? It seems like you can keep the input fields the same but create a version of this component that is pared down to just the phone number input. Would that work?

Since this component is used across the site where we need people to add shipping/billing addresses for credit cards (auctions, user settings payments, orders), it feels odd to extend it for this purpose.

We should probably (not as part of this PR 😄) re-name this component and extract it from the `Order` sub-app.

src/Apps/Order/Routes/Shipping/index.tsx
lilyfromseattle: Possibly...but I get errors when I move it

sweir27: You should be able to add:
```
... on CommercePickup {
  phoneNumber
}
```

lilyfromseattle: ```Unknown field 'phoneNumber' on type 'CommercePickup'```

src/Apps/Order/Components/AddressForm.tsx
lilyfromseattle: Like the handlers and stuff. It just felt like a lot of duplication when I started doing it that way

lilyfromseattle: I talked it over with Ashkan and he thought this way was fine but can do it that way if you think its better

sweir27:
@lilyfromseattle I think this one can be closed. 😄

PR #3077, opened 2020-01-03
Add "no-switch-case-fall-through" tslint rule

pepopowitz:
While working on #3007, I ran into a tricky bug that was a result of accidentally falling through the `case` of a `switch`. It makes sense to me to disallow fall-through, as it has only ever caused me headaches. I'm PR'ing this change instead of making it directly in master, to allow others to speak up against disallowing fall-through.

There was only one place (CollectMediumMetadata) that was taking advantage of fall-through, and it might have been an accident.


artsyit:
:rocket: PR was released in v24.0.13 :rocket:

PR #3080, opened 2020-01-03
Fix artwork grid preloading too many images

zephraph:
In digging more into [PURCHASE-1658](https://artsyproduct.atlassian.net/browse/PURCHASE-1658) I noticed something very strange. Many more images were being  preloaded than what was specified via the top level artwork grid. We usually default to 6, but I was seeing 11 (on mobile). 

Was scratching my head trying to figure out what prop was being passed down improperly, but then I realized I just goofed up the math. 

## Visualization

ll-example

### DIFF
--- a/src/Components/ArtworkGrid/ArtworkGrid.tsx
+++ b/src/Components/ArtworkGrid/ArtworkGrid.tsx
@@ -117,7 +117,7 @@ export class ArtworkGridContainer extends React.Component<
             artwork={artwork}
             key={artwork.id}
             mediator={this.props.mediator}
-            lazyLoad={i + j >= preloadImageCount}
+            lazyLoad={i * columnCount + j >= preloadImageCount}
             onClick={() => {
               if (this.props.onBrickClick) {
                 this.props.onBrickClick(artwork)
--- a/src/Components/ArtworkGrid/ArtworkGrid.tsx
+++ b/src/Components/ArtworkGrid/ArtworkGrid.tsx
@@ -108,16 +108,24 @@ export class ArtworkGridContainer extends React.Component<
     }
     const sections = []
 
-    for (let i = 0; i < columnCount; i++) {
+    for (let column = 0; column < columnCount; column++) {
       const artworkComponents = []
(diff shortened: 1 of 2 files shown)


src/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -117,7 +117,7 @@ export class ArtworkGridContainer extends React.Component<
             artwork={artwork}
             key={artwork.id}
             mediator={this.props.mediator}
-            lazyLoad={i + j >= preloadImageCount}
+            lazyLoad={i * columnCount + j >= preloadImageCount}
```
mzikherman: Since i,j both start at 0, plus there's a `>=`, so you're doing a bit of math, and as such seems ripe for a subtle 'off-by-1' error or what have you (not unlike the current issue).

It might be clearer to have an `imageCount` that you instantiate to 0 outside the loops and increment inside the inner loop. Then the prop is `lazyLoad={imageCount > preloadImageCount}`

This might remove any math altogether!

This is an overly pedantic suggestion btw, nice find.

pepopowitz: I second what @mzikherman says above. This does currently look like code I could easily mess up; a named variable would help me better understand it. 

It also might be helpful to introduce test coverage of this counting, somehow.

zephraph: I started going down this road, but as the loops are now this wouldn't work. When it's laying  out the grid, it lays out the columns first. So it'd fill up the left-most column then it'd fill up the right-most column. We could add it to a variable if we flipped the loops. 

@pepopowitz yeah, I'll look into adding a test too.

zephraph:
I've updated the loop variables to provide more clarity generally and documented the math part. I'll add some tests to fully wrap this up.

zephraph:
All right. Updated variables to make it easier to follow. Added a comment for clarification. Also added a test. I used test on the old code and it did fail as expected.

artsyit:
:rocket: PR was released in v24.0.17 :rocket:

PR #3062, opened 2019-12-17
[GROW-1698] Fix whitespace in global nav

anandaroop:
https://artsyproduct.atlassian.net/browse/GROW-1698

Minor pre-existing Q/A issue that turned up while developing the [soon-to-be launched a/b test](https://github.com/artsy/force/pull/4860)

### Before

before

### After

after

(Pretty subtle — check the spacing to the left of Artworks and to the right of More 🔍 )

PR #3082, opened 2020-01-06
[GROW-1725] Fix global nav bar spacing

anandaroop:
https://artsyproduct.atlassian.net/browse/GROW-1725

Follow-up to https://github.com/artsy/reaction/pull/3062

One more minor tweak to the nav bar. This time accounts for all four permutations, as follows:

story

This^ comes from the story that is also added in this PR.


artsyit:
:rocket: PR was released in v24.0.18 :rocket:

PR #3083, opened 2020-01-06
Prevent right click on auction results images

sweir27:
We want to prevent users from right-clicking on images in the auction results section of the artist page.


artsyit:
:rocket: PR was released in v24.0.16 :rocket:

PR #3084, opened 2020-01-06
Remove react-responsive-media and add fresnel

zephraph:
Turns out reaction was still using react-responsive-media! Whoops. This updates it to use fresnel (which is the same library, just renamed).


artsyit:
:rocket: PR was released in v24.0.19 :rocket:

PR #3085, opened 2020-01-06
Only render one breakpoint for mobile googlebot

zephraph:
Fixes [PURCHASE-1712](https://artsyproduct.atlassian.net/browse/PURCHASE-1712).

This updates our device detection to check for googlebot's UA. If the mobile UA is present it only sends one breakpoint down the wire.

It's only detecting mobile because we've been switched over to mobile first indexing by Google. Also, the UA isn't _exact_ so there's potential that some device or service that's spoofing or using a partial of Google's UA gets swept up in this. We have access to what the full UA would be [here](https://support.google.com/webmasters/answer/1061943?hl=en). 

I'm fine with it being loose, but I'm interested to hear other's thoughts.

### DIFF
--- a/src/Artsy/Router/Utils/matchingMediaQueriesForUserAgent.ts
+++ b/src/Artsy/Router/Utils/matchingMediaQueriesForUserAgent.ts
@@ -11,6 +11,12 @@ import {
 export function matchingMediaQueriesForUserAgent(
   userAgent: string
 ): MatchingMediaQueries {
+  // Only return xs breakpoint for mobile googlebot
+  const ua = userAgent.toLowerCase()
+  if (ua.includes("googlebot") && ua.includes("mobile")) {
+    return ["notHover", "xs"]
+  }
+
   const device = findDevice(userAgent)
   if (!device) {
     return undefined


joeyAghion:
> ...we've been switched over to mobile first indexing by Google...

True, but in practice Google is many-tentacled and I expect we'll continue to see legitimate traffic from non-mobile googlebot crawlers as well. I think that's fine w.r.t. this change, but just wanted that expectation to be clear.

A bigger concern: does this mean that some of our page content will no longer be in the HTML source _at all_? I worry about this because content that was previously not "visible" but _was_ in the source _at least_ would be indexed. Is there a risk that we damage how our pages match for some of the search terms that might not be rendered in the `xs` version?

zephraph:
> A bigger concern: does this mean that some of our page content will no longer be in the HTML source at all?

It means that anything not in the xs breakpoint won't be rendered for mobile googlebot, yes. Google vastly discounts hidden content, so even though that hidden content was previously in the markup, it wasn't doing much for ranking. 

Here's a pretty good hidden text experiment: 
https://www.rebootonline.com/blog/hidden-text-experiment/

I don't believe we'll see much, if any, of a ranking change. Less duplicate content would be beneficial in that it increases the amount of pages that can be covered under our crawl budget. There's a slight perf improvement likely due to less components being rendered (though I imagine that would largely be negligible). 

---

Bottom line: We're now predominantly mobile ranked. If it's not in XS, it doesn't help us. We've gotta keep that in mind.

joeyAghion:
Merging!

I can't tell from [PURCHASE-1712](https://artsyproduct.atlassian.net/browse/PURCHASE-1712), but it sounds like the primary motivation for this is to improve our crawl budget utilization. We should monitor whatever diagnostics best reflect that once this is released.

artsyit:
:rocket: PR was released in v24.0.20 :rocket:

PR #3089, opened 2020-01-10
Add `data-test` attribute to Order code so we can use it in integrity

ashkan18:
# Problem
We want to be able to find the newly created order's `code` in Integrity so we can test proper emails.

# Solution
Wrap code around a `span` and add `data-test` for it so we can find this element in the UI during integrity.

### DIFF
--- a/src/Apps/Order/Routes/Status/index.tsx
+++ b/src/Apps/Order/Routes/Status/index.tsx
@@ -246,7 +246,7 @@ export class StatusRoute extends Component<StatusProps> {
             {title}
           </Serif>
           <Sans size="2" weight="regular" color="black60" mb={[2, 3]}>
-            {flowName} #{order.code}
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/Status/index.tsx
```diff
@@ -246,7 +246,7 @@ export class StatusRoute extends Component<StatusProps> {
             {title}
           </Serif>
           <Sans size="2" weight="regular" color="black60" mb={[2, 3]}>
-            {flowName} #{order.code}
+            {flowName} <span data-test="OrderCode">#{order.code}</span>
```
sepans: camel case or snake case? 😆

artsyit:
:rocket: PR was released in v24.0.23 :rocket:

PR #3090, opened 2020-01-13
Reset pagination on artwork filter updates

zephraph:
Fixes [1692](https://artsyproduct.atlassian.net/browse/PURCHASE-1692)

If a user paginates in the artwork grid then applies a filter it preserved the paginated state and could sometimes show no results. This update resets the pagination whenever a filter is set or unset.


artsyit:
:rocket: PR was released in v24.0.24 :rocket:

PR #3095, opened 2020-01-15
MX-104 - Fix double capitalized letters in iOS inputs

kierangillen:
- There's an existing iOS bug which uppercases the first two letters when `autoCapitalization="word"` is enabled, so we are for now removing autoCapitalization on mobile
- WebKit bug report from 2016: https://bugs.webkit.org/show_bug.cgi?id=157410
- [MX-104]
- https://artsyproduct.atlassian.net/browse/MX-104

## Before fix:

Screen Shot 2020-01-15 at 3 22 07 PM

[MX-104]: https://artsyproduct.atlassian.net/browse/MX-104

### DIFF
--- a/src/Apps/Order/Components/AddressForm.tsx
+++ b/src/Apps/Order/Components/AddressForm.tsx
@@ -94,6 +94,7 @@ export class AddressForm extends React.Component<
 
   render() {
     const lockCountryToOrigin = !this.props.billing && this.props.domesticOnly
+
     return (
       <Join separator={<Spacer mb={2} />}>
         <Flex flexDirection="column">
@@ -101,7 +102,6 @@ export class AddressForm extends React.Component<
             id="AddressForm_name"
             placeholder="Add full name"
             title={this.props.billing ? "Name on card" : "Full name"}
-            autoCapitalize="words"
             autoCorrect="off"
             value={this.props.value.name}
             onChange={this.changeEventHandler("name")}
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -94,14 +94,27 @@ export class AddressForm extends React.Component<
 
   render() {
     const lockCountryToOrigin = !this.props.billing && this.props.domesticOnly
+
+    // FIXME: There's an existing iOS bug which uppercases the first two letters when autoCapitalization="word" is enabled
+    // so we are for now disabling autoCapitalization on mobile
+    // https://bugs.webkit.org/show_bug.cgi?id=157410
+    let autoCapitalizeWord = null
```
ashfurrow: I think setting this to `undefined` here might be more semantically correct (in terms of being the same as not passing the prop in at all).
```suggestion
    let autoCapitalizeWord = undefined
```

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -94,14 +94,27 @@ export class AddressForm extends React.Component<
...
+    if (
+      typeof window !== "undefined" &&
+      window.innerWidth &&
+      window.innerWidth > 768
+    ) {
+      autoCapitalizeWord = "words"
+    }
```
ashfurrow: This makes sense, though I'm not sure if we need something so complex. I think removing the `autoCapitalize="words"` altogether might be an acceptable solution, since `autoCapitalize` isn't required for desktop browsers (and iPad would likely have the same bug as iPhone). I think removing `autoCapitalize="words"` in all cases would be a better approach, but I'm open to feedback on this 👍

zephraph: I agree w/ this.

kierangillen:
@ashfurrow Ok made the suggested change.

artsyit:
:rocket: PR was released in v24.0.28 :rocket:

PR #3096, opened 2020-01-16
My purchases #hackaton2020

sepans:
#Hackaton2020 work by @lilyfromseattle @ashkan18 and @sepans 

### Screenshots:
Screen Shot 2020-01-17 at 11 09 38 AM

TODOs:
- Responsiveness. Will be in a separate PR. 
- Add pagination. Will be in a separate PR.


src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,139 @@
...
+  Sans,
+  Serif,
+  StackableBorderBox,
+} from "@artsy/palette"
+import { PurchaseHistory_orders } from "__generated__/PurchaseHistory_orders.graphql"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface OrderRowProps {
+  order: any
```
sweir27: We should properly type this! Something like `PurchaseHistory_orders["edges"][number]["node"] I _think_ can work.

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+const OrderRow = (props: OrderRowProps) => {
+  const { order } = props
+  const artwork = order.lineItems.edges[0].node.artwork
```
sweir27: Let's use optional chaining or `get` here to make this safer.

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,139 @@
...
+    artwork && (
+      <Box p={1}>
+        <StackableBorderBox
+          flexDirection="row"
+          alignItems="center"
+          justifyContent="space-between"
+        >
+          <Flex width="350px">
+            <Flex height="auto" alignItems="center">
+              {artwork.image.resized.url && (
```
sweir27: Ditto here (to the `get` comment).

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,139 @@
...
+              flexDirection="column"
+              justifyContent="center"
+              style={{ overflow: "hidden" }}
+            >
+              <Link
+                href={`/orders/${order.internalID}/status`}
+                underlineBehavior="hover"
+              >
+                <Serif size="2" weight="semibold">
+                  {artwork.artist_names}
```
sweir27: Why alias this field below? We should just be able to use `artistNames` across the board.

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+interface PurchaseHistoryProps {
+  orders: PurchaseHistory_orders
+}
+
+const PurchaseHistory: React.FC<PurchaseHistoryProps> = (
+  props: PurchaseHistoryProps
+) => {
+  const { orders } = props
+  const myOrders = orders.edges.map(x => x.node)
```
sweir27: Will this fail if you have no orders? May want to be safer accessing `.edges` here.

src/Apps/Purchase/PurchaseApp.tsx
```diff
@@ -0,0 +1,55 @@
...
+        </SafeAreaContainer>
+      </AppContainer>
+    )
+  } else {
+    // not an admin
+    return <ErrorPage code={404} />
+  }
+}
+
+const SafeAreaContainer = styled(Box)`
```
sweir27: Do we need this? (and the `Meta` tag above?)

I believe it was added to the order app to make the `StickyFooter` work, but we don't have that on this page so I think it can be removed.

src/Apps/Purchase/__tests__/PurchaseApp.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <HeadProvider>
+          <SystemContextProvider user={user}>{children}</SystemContextProvider>
+        </HeadProvider>
+      </MockBoot>
+    ),
+  })
+
+describe("Purchase app", () => {
+  describe("User with admin privilages", () => {
+    it("renders orders with view details button", async () => {
```
sweir27: Can we add a test for when an admin views this page but there are 0 orders?

src/Apps/Purchase/Components/PurchaseHistory.tsx
sepans: Just to be able to use `OrderApp` fixtures. We can switch back if we ended up using the new relay fixture generators. In the spirit of hackathon I may try it out if I get some extra time.

src/Apps/Purchase/Components/PurchaseHistory.tsx
sepans: I _think_ graphql returns empty array for edges but no harm to add the check

src/Apps/Purchase/Components/PurchaseHistory.tsx
sweir27: I'd vote for copy-pasting and updating those fixtures (within your test file) and making this change vs. learning the new way or leaving as-is!

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+interface OrderRowProps {
+  order: PurchaseHistory_orders["edges"][number]["node"]
+}
+const OrderRow = (props: OrderRowProps) => {
+  const { order } = props
+  const artwork = get(order, o => o.lineItems.edges[0].node.artwork)
+
+  return (
+    artwork && (
```
damassi: rather than placing this in the return, maybe 

```jsx
if (!artwork) {
  return null
}
```
just above?

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,144 @@
...
+            >
+              <Link
+                href={`/orders/${order.internalID}/status`}
+                underlineBehavior="hover"
+              >
+                <Serif size="2" weight="semibold">
+                  {artwork.artist_names}
+                </Serif>
+              </Link>
+              <div style={{ lineHeight: "1" }}>
```
zephraph: You don't need this wrapper to adjust line height. You can technically pass that directly to the Serif component...

Though I'm not sure if the types reflect that 🤔

src/Apps/Purchase/PurchaseApp.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { AppContainer } from "Apps/Components/AppContainer"
+import { SystemContext } from "Artsy"
+import { ErrorPage } from "Components/ErrorPage"
+import React, { useContext } from "react"
+import { Title } from "react-head"
+import { createFragmentContainer, graphql } from "react-relay"
+import { userIsAdmin } from "Utils/user"
+import { PurchaseHistoryFragmentContainer as PurchaseHistory } from "./Components/PurchaseHistory"
+
+export interface Props {
```
zephraph:
```suggestion
export interface PurchaseAppProps {
```

It's a minor suggestion, but having this more explicit prevents accidental name collisions when doing `export * from 'blah'` in places. While technically it's very unlikely for this particular module to ever be referenced in that way, I still think it's somewhat of a best practice.

src/Apps/Purchase/PurchaseApp.tsx
```diff
@@ -0,0 +1,41 @@
...
+  orders: PurchaseApp_orders
+}
+
+export const PurchaseApp = (props: Props) => {
```
zephraph:
```suggestion
export const PurchaseApp = (props: PurchaseAppProps) => {
```

src/Apps/Purchase/Components/PurchaseHistory.tsx
sepans: 👍  Looked at the designs and the line-height is `18px` there so the correct relative line height should be `1.3` but I always thought `<Serif>`'s `size` property automatically sets line height but it wasn't there. Checked order pages and noticed over there the line height is also explicitly set. 🤔

zephraph: Hmmmm... sounds like a bug.

sepans:
Going to merge this. Thanks everyone for comments.

artsyit:
:rocket: PR was released in v24.1.0 :rocket:

src/Apps/Purchase/Components/PurchaseHistory.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  artwork {
+                    date
+                    image {
+                      resized(width: 55) {
+                        url
+                      }
+                    }
+                    internalID
+                    title
+                    artist_names: artistNames
```
mzikherman: #minor why the alias'ing to snake case?

mzikherman:
Awesome! What a much-needed feature.

src/Apps/Purchase/Components/PurchaseHistory.tsx
sepans: I put a TODO [here](https://github.com/artsy/reaction/pull/3096/files#diff-794b682df99624bdb3a6cd26e54af282R52) to clean it up. This is just to be able to use the `OrderApp` fixtures for now.

PR #3097, opened 2020-01-16
Wrap main page header in header tags

eessex:
Noticed that the new react header doesn't use a `<header>` html tag -- this caused some issues with Integrity's login tests, where we couldn't find the 'log in' button on fully react pages. 

Adding header will have effects on accessibility, making nav clearer/more semantic for screen readers.


mzikherman:
Cool!

artsyit:
:rocket: PR was released in v24.0.27 :rocket:

PR #3099, opened 2020-01-17
Add UI to request lot condition from artwork page

dblandin:
This adds a button next to the "Condition" row in additional artwork details when the work is in an auction.

### Screen Recordings

**Authenticated / Successful request**

Peek 2020-01-23 10-56

---

**Authenticated / Successful request**

Peek 2020-01-23 10-58

---

**Unauthenticated**

Peek 2020-01-23 10-58-2


dleve123:
@dblandin Can you add some screenshots/screencaptures of the introduced UI? Perhaps the storybook?

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,205 @@
...
+    } else if (error.message) {
+      errorMessages = [error.message]
+    }
+
+    setRequestingConditionReport(false)
+    trackRequestConditionReportFailure(errorMessages)
+    setShowErrorModal(true)
+  }
+
+  const commonProperties = {
```
dleve123: Is it possible to promote any of these properties to be arguments to [`track()`](https://github.com/artsy/reaction/pull/3099/files#diff-578cbd16d0663ec14ecf363c0f808a66R23-R25)?

dblandin: Ahhh, likely. I'll give that a try.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
```diff
@@ -24,17 +26,28 @@ export class ArtworkDetailsAdditionalInfo extends React.Component<
...
 
     const listItems = [
       {
         title: "Condition",
-        value: conditionDescription ? conditionDescription.details : null,
+        value:
+          isBiddable && sd.ENABLE_REQUEST_CONDITION_REPORT ? (
+            <RequestConditionReport
+              saleArtworkID={saleArtwork.internalID}
+              relay={this.props.relay}
```
mzikherman: Rather than passing the `relay` prop around, you can get `relayEnvironment` directly from the `useSystemContext` call.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,205 @@
...
+          input: { saleArtworkID },
+        },
+      })
+    })
+  }
+
+  const handleMutationError = error => {
+    logger.error(error)
+
+    let errorMessages: string[]
```
dleve123: This stanza is the same as the one in the [Register app](https://github.com/artsy/reaction/blob/master/src/Apps/Auction/Routes/Register/index.tsx#L156-L163).

Have you considered abstracting this a bit? Seems like this could be a function named something like `standardizeError(error: Error): string[]`. Just wanted to surface the idea!

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
dleve123: I believe the ConfirmBid component is the latest understanding how we can track common properties. https://github.com/artsy/reaction/blob/master/src/Apps/Auction/Routes/ConfirmBid/index.tsx#L361-L367

Here's a commit that should have some helpful context here: 926455ac268e13e8661e2dee5905169064016d17

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
dblandin: The shared bit is:

```typescript
    if (Array.isArray(error)) {
      errorMessages = error.map(e => e.message)
    } else if (typeof error === "string") {
      errorMessages = [error]
    } else if (error.message) {
      errorMessages = [error.message]
    }
```

We actually don't need most of it. We're not dealing with multiple errors and we're not passing strings as errors either. I'll go through and simplify this.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,205 @@
...
+    setShowConditionReportRequestedModal,
+  ] = useState(false)
+
+  const [showErrorModal, setShowErrorModal] = useState(false)
+
+  const isLoggedIn = Boolean(user)
+  const { saleArtworkID, relay } = props
+
+  const requestConditionReport = () => {
+    return new Promise(async (resolve, reject) => {
```
dleve123: Have you considered typing the promise returned here, similar to what we do in the [`ConfirmBid` component?](https://github.com/artsy/reaction/blob/f53215c82577ef229b6dec192211d8a9175b5f14/src/Apps/Auction/Routes/ConfirmBid/index.tsx#L79). I _believe_ that would remove needing the cast to `any` on L122 below.

dblandin: Definitely thinking about tightening up the types. Haven't gotten to it yet but will definitely take a look at this promise while I'm doing so.

src/Apps/Artwork/Components/ArtworkDetails/__tests__/ArtworkDetails.test.tsx
```diff
@@ -184,5 +208,84 @@ describe("ArtworkDetails", () => {
...
+
+      wrapper = await getWrapper({
+        ...ArtworkDetailsFixture,
+        isBiddable: true,
+        saleArtwork: {
+          internalID: "sale-artwork-id",
+          id: "sale-artwork-id",
+        },
+      })
+      expect(wrapper.html()).toMatch("to request")
```
dleve123: Any reason to not assert upon the full text "Log in to request report"? Would there be a way to add a data attribute to an element to make it easier to target a subset of text on the page?

dblandin: The design comps have "Log in" in the button and "to request" next to the button. In the rendered html, you won't see the full text of "Log in to request" as one chunk. I tried matching upon the presence of certain components like `UnauthenticatedContent` to complement the string matching but was having some trouble there.

Not sure I follow the data attribute idea. Can you elaborate?

src/Apps/__tests__/Fixtures/Artworks.ts
```diff
@@ -815,8 +815,11 @@ export const AllAdditionalDetailsPresent = {
   additional_information:
     "In Francesca DiMattio’s The Fox and the Hound, 2018, lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullam DiMattio’s The Fox and the Hound, 2018, lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullam. Tempor incididunt ut labore et dolore magnas.",
   sale: null,
+  isInAuction: false,
```
dleve123: What motivated these changes? Did you just notice that the stub was incomplete? Or is it related to some stories being changed in this diff?

typings/sharify.d.ts
```diff
@@ -22,6 +22,7 @@ declare module "sharify" {
       readonly ARTIST_COLLECTIONS_RAIL_IDS: string[]
       readonly CMS_URL: string
       readonly ENABLE_PRICE_TRANSPARENCY: string
+      readonly ENABLE_REQUEST_CONDITION_REPORT: string
```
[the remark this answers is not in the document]
dblandin: Unless we're coercing the value, these attributes should come in as strings. I don't think this attribute, coming from Force's environment, will be of type `boolean` so typing it such would be incorrect.

src/Apps/Artwork/Components/ArtworkDetails/__tests__/ArtworkDetails.test.tsx
erikdstock: I believe he means something like
```
div.wrapper data-type='condition-report-request'
  button
    Log in
  p
    to request
```

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,205 @@
...
+  relay?: RelayProp
+}
+
+export const RequestConditionReport: React.FC<RequestConditionReportProps> = track(
+  {
+    context_module: AnalyticsSchema.ContextModule.RequestConditionReport,
+  }
+)(props => {
+  const { mediator, user } = useSystemContext()
+  const { trackEvent } = useTracking()
```
erikdstock: Whenever I come back to this the amount of boilerplate around tracking is surprising. Are both a HOC with `track()()` and the `useTracking` hook necessary?

src/Apps/Artwork/Components/ArtworkDetails/__tests__/ArtworkDetails.test.tsx
dleve123: > Not sure I follow the data attribute idea. Can you elaborate?

Sure – there might not be a comparable pattern with React, but when testing Rails apps, there's often a pattern of doing something like

```html
<div data-test="some-selector">
  <p>foo</p>
  <p>bar</p>
</div> 
<p>baz</p>
```

so that one can select the text within some selector like the following:

(pseudo-ruby / capybara)
```ruby
within '.some-selector' do
  text = page.text()
end
```

---

There might something similar here could be employed!

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <Spinner size="small" />
+      ) : (
+        "Request condition report"
+      )}
+    </Button>
+  )
+
+  const RequestedConditionReportModal: React.FC = () => (
+    <Modal
+      title="Condition report requested"
```
erikdstock: Do we have a good sense of all the different places a modal can surface on this page? I'm thinking of registration with checkbox, confirmation, login, this... I wish there were a single interface to manage all these though I guess that is out of scope of this app (and I guess it still wouldn't on its own answer the question of what happens if multiple components try to trigger a modal)

src/Apps/Artwork/Components/ArtworkDetails/__tests__/ArtworkDetails.test.tsx
dblandin: Ahh, gotcha. Yes, I'm familiar with "within" for capybara in ruby test suites. I'll see what's available here to better scope the assertion.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
dblandin: The login flow is the only modal triggered by the `mediator` event hub here. I don't think the other modals would conflict. The event handling should be isolated enough not to cause issues. I actually prefer that isolation rather than having a global state of modals.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
dblandin: Good point. Might be able to lose that HOC with `track()` here.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
dblandin: Added that promise type :+1:

src/Apps/__tests__/Fixtures/Artworks.ts
dblandin: I rolled back these changes. Not sure if they're still needed.

.storybook/webpack.config.js
```diff
@@ -52,6 +53,7 @@ const sharifyPath = sharify({
   ADMIN_URL,
   APP_URL,
   CMS_URL,
+  ENABLE_REQUEST_CONDITION_REPORT,
```
dblandin: This allows you to set `ENABLE_REQUEST_CONDITION_REPORT=true` in your `.env` file to enable the UI to render on staging+relay-backed artwork page/component stories.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,244 @@
...
+  )
+})
+
+export const RequestConditionReportQueryRenderer: React.FC<{
+  saleArtworkID: string
+}> = ({ saleArtworkID }) => {
+  const { relayEnvironment } = useSystemContext()
+
+  return (
+    <QueryRenderer<RequestConditionReportQuery>
```
yuki24: What are the SSR implications here? Is it okay to add a `QueryRenderer` deep down the component tree? (which, in this case, is reasonable to me since we are conditionally showing the `<RequestConditionReport>` component) It might make it difficult to track what requests are being made in a single page and also could slow down the page load a bit.

cc @zephraph

dblandin: Good question! I might need to explicitly defer rendering to the client so that this isn't rendered server-side.

mzikherman: Use `SystemQueryRenderer` instead of `QueryRenderer` and it will take care of deferring.

dblandin: Ahh, fantastic. Thanks @mzikherman!

dblandin:
@mzikherman @dleve123 @erikdstock @yuki24 

I'm still working on the specs utilizing `createTestEnv` but the implementation should be close to its final state now and ready for another phase of review.

There are now screen recordings in the PR description and if run locally via `yarn start`, you can see working storybook entries at http://localhost:9001/?path=/story/apps-artwork-components-artistdetails--requestconditionreport (but your host/port may differ). Thanks to @ds300 for his help figuring that out (see second commit description for more context).

The other aspect that may change a bit is around analytic events after speaking with @mikehrom and @abhitip on the Data team.

src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+    return authentication.login(mediator)
+  }
+
+  const handleRequestConditionReportClick = () => {
+    setRequesting(true)
+
+    requestConditionReport()
+      .then(data => {
+        const theData = data as any
```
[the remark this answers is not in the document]
dblandin: Hopefully not! I'll circle back to that. :+1:

src/Apps/Artwork/Components/ArtworkDetails/__tests__/ArtworkDetails.test.tsx
dblandin: This spec is skipped because because, as a QueryRenderer, this component's render is deferred to the client. I think this would be better tested as a full relay and interaction-enabled spec backed by our `createTestEnv` tooling.

.storybook/webpack.config.js
```diff
@@ -142,18 +141,23 @@ module.exports = async ({
     })
   }
 
-  config.module.rules.push({
+  config.module.rules.push(
+    {
```
erikdstock: Looks like some prettier diff here- just want to be sure this is ok and not due to an out-of-sync config.

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
```diff
@@ -76,7 +85,11 @@ export class ArtworkDetailsAdditionalInfo extends React.Component<
               </Col>
               <Col xs={12} sm={6} md={6} lg={9}>
                 <Sans size="2" weight="regular" color="black60">
-                  <ReadMore maxChars={140} content={value} />
+                  {React.isValidElement(value) ? (
+                    value
+                  ) : (
+                    <ReadMore maxChars={140} content={value} />
```
erikdstock: Why the need for this guard?

artsyit:
:rocket: PR was released in v24.2.0 :rocket:

artsyit:
:rocket: PR was released in v24.2.0 :rocket:

artsyit:
:rocket: PR was released in v24.2.0 :rocket:

PR #3101, opened 2020-01-21
Add `data-test` tags to offer response radio btns

sepans:
Add `data-test` tags to offer response radio btns


ashkan18:
🙌 #mergeongreen

artsyit:
:rocket: PR was released in v24.1.2 :rocket:

PR #3102, opened 2020-01-21
Fix double image load issue by updating lazy load library

zephraph:
See https://github.com/artsy/palette/pull/628 for more context. 

TL;DR sometimes (esp on chrome) images were downloading twice when lazy loading. Updating the library resolves that issue.


artsyit:
:rocket: PR was released in v24.1.4 :rocket:

PR #3104, opened 2020-01-22
PURCHASE-1656: Add phone number field to Address for pickup 

lilyfromseattle:
# [PURCHASE-1656] Add phone number field to Address for pickup
https://www.figma.com/file/LieDNuIuKAfhDeEPkjan4p/Checkout-Flow?node-id=0%3A1

This field should be required.

This is a refactor of [this PR](https://github.com/artsy/reaction/pull/3075) as was suggested in the comments

__Before:__
Screen Shot 2020-01-02 at 11 49 45 AM

__After:__
Screen Shot 2020-01-02 at 11 49 30 AM

[PURCHASE-1656]: https://artsyproduct.atlassian.net/browse/PURCHASE-1656


src/Apps/Order/Components/PhoneNumberForm.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  render() {
+    return (
+      <Flex flexDirection="column">
+        <Input
+          id="PhoneNumberForm_phoneNumber"
+          title="Phone number"
+          type="tel"
+          description={this.props.label}
+          placeholder="Add phone"
```
zephraph: Should this be `Add phone number`?

src/Apps/Order/Components/PhoneNumberForm.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  getError = (phoneNumber): string => {
+    return (this.props.touched && this.props.errors) || ""
+  }
+
+  render() {
+    return (
+      <Flex flexDirection="column">
+        <Input
```
zephraph: So the `Input` component here is wrapped in a Flex, but the only sibling that lives with it is a `Spacer`. 

It seems like we could just add an `mb` to `Input` and delete both the `Flex` and `Spacer`, right? Alternatively the parent component could add the spacer where necessary.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -238,6 +259,16 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
     }
   }
 
+  private validatePhoneNumber(phoneNumber: string) {
+    const error = validatePresence(phoneNumber)
+    const hasError = error != null
```
zephraph:
```suggestion
    const hasError = error !== null
```

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -268,10 +308,17 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+      addressErrors,
+      addressTouched,
+      phoneNumber,
+      phoneNumberError,
+      phoneNumberTouched,
+    } = this.state
     const artwork = get(
       this.props,
-      props => order.lineItems.edges[0].node.artwork
+      props => props.order.lineItems.edges[0].node.artwork
```
zephraph: Ohhh, was this broken before?

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -340,6 +387,25 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
                     onChange={this.onAddressChange}
                     domesticOnly={artwork.onlyShipsDomestically}
                     shippingCountry={artwork.shippingCountry}
+                    showPhoneNumberInput={false}
```
ashkan18: since this is `shipping` shouldn't this be `true` and remove following `PhoneNumberForm` and let `AddressForm` handle it? alternatively maybe we should just move `phoneNumber` out of `AddressForm` altogether and in places we show `PhoneNumber` we add it as a separate component?
My main question is, when `showPhoneNumber` will be `true` if we don't use it here either.

src/Apps/Order/Routes/Shipping/index.tsx
zephraph: Oh, hmm, I see. Yeah, you're right. That would work, but it isn't necessarily how it's intended to be used I think. This whole thing could also be something like

```
    const artwork = get(
      order,
      o => o.lineItems.edges[0].node.artwork
```

Really works either way though. I'll resolve this.

src/Apps/Order/Routes/Shipping/index.tsx
damassi: Maybe a typo but good to know that `!= null` is a weird JS shortcut that means: not equal `null` _and_ not equal `undefined`.

src/Apps/Order/Components/PhoneNumberForm.tsx
lilyfromseattle: Input doesn't take `mb` .but I can add it to the flex

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -76,6 +86,14 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
     address: this.startingAddress,
     addressErrors: {},
     addressTouched: {},
+    phoneNumber:
+      this.props.order.requestedFulfillment &&
+      (this.props.order.requestedFulfillment.__typename === "CommerceShip" ||
```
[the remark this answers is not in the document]
lilyfromseattle: It was causing type errors for some reason so this is what Sepand and I came up with but I;'m open to alternate suggestions

src/Apps/Order/Routes/Shipping/index.tsx
sweir27: @ashkan18 it's needed in the bid flow!

sweir27: I just created: https://artsyproduct.atlassian.net/browse/PURCHASE-1750

It's caused a lot of confusion that the `AddressForm` code exists in the `Order` app, since it's used in a few places. We should move it somewhere more generic!

src/Apps/Order/Components/AddressForm.tsx
```diff
@@ -203,7 +203,7 @@ export class AddressForm extends React.Component<
             />
           </Flex>
         </TwoColumnSplit>
-        {(!this.props.billing || this.props.showPhoneNumberInput) && (
+        {this.props.showPhoneNumberInput && (
```
lilyfromseattle: Hey @dblandin you added this check right? Is it ok if I remove it?

src/Apps/Order/Routes/Shipping/index.tsx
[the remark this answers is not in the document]
lilyfromseattle: @zephraph @ds300 Any ideas on why this is causing type errors without the check?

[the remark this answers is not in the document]
sepans: not sure if I am looking at the right place but is this [`other`](https://github.com/artsy/reaction/blob/073b55f5a30888c2f595010a518dec917387ab67/src/__generated__/ShippingOrderAddressUpdateMutation.graphql.ts#L46) causing it?

data/schema.graphql
```diff
@@ -6054,6 +6213,22 @@ type RelatedArtworkGrid implements ArtworkContextGrid {
   ): ArtworkConnection
 }
 
+# Autogenerated input type of RequestConditionReport
+input RequestConditionReportInput {
```
dblandin: I think this schema diff should go away with a rebase.

src/Apps/Order/Components/AddressForm.tsx
dblandin: @lilyfromseattle I think this might be still used. We show this form in user profile payment methods as well.

data/schema.graphql
```diff
@@ -1600,6 +1600,25 @@ type CalculatedCost {
   subtotal: Money
 }
 
+# Autogenerated input type of CaptureHold
+input CaptureHoldInput {
```
dblandin: This diff should go away with a rebase with master.

src/Apps/Order/Routes/Shipping/index.tsx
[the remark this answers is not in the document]
ds300: *tldr* we still need this check and yes @sepans it's to do with that `"%other"`

`"%other"` is a sentinel value that the typescript relay plugin adds to `__typename` union types to force you to handle 'unexpected' cases. Not 100% sure but I think the idea is that in the future you could very possibly add another type to e.g. the union `CommerceRequestedFulfillment`, which wouldn't technically be a breaking schema change, but it could break your code if it's only ever expecting there to be two possible values for `__typename` on `CommerceRequestedFulfillment` and suddenly we add a third.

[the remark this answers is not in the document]
lilyfromseattle: So is it best to leave the check in for now?

src/Apps/Order/Components/AddressForm.tsx
lilyfromseattle: From what I can see it's used in BidForm, RegistrationForm, PaymentPicker, and PaymentForm. In all places `billing` is true, so removing the `!this.props.billing` check shouldn't affect the other usages.

src/Apps/Order/Routes/Shipping/index.tsx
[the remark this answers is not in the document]
ds300: Yep

src/Apps/Order/Components/AddressForm.tsx
dblandin: Yeah, I'm having trouble finding an example of when `billing` is et to `false` as well. One thing @sweir27 and I had in mind when consolidating the payment forms was not showing "Required for shipping logistics" on those other forms. So feel free to remove this condition but I would check those other places to make sure everything looks okay.

src/Apps/Order/Routes/Shipping/index.tsx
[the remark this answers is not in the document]
ds300: although now you can change this bit from 

```ts
(this.props.order.requestedFulfillment.__typename === "CommerceShip" ||
        this.props.order.requestedFulfillment.__typename === "CommercePickup")
```

to just

```ts
this.props.order.requestedFulfillment.__typename != "%other"
```

I think.

src/Apps/Order/Components/AddressForm.tsx
ashkan18: could it be add new payment method under user's setting or during bidding registration?

src/Apps/Order/Components/PhoneNumberForm.tsx
lilyfromseattle: It was like this before (on the shipping page) so I kept it as is

artsyit:
:rocket: PR was released in v24.3.0 :rocket:

PR #3109, opened 2020-01-27
[FX-1692] Reorder mobile nav items to match desktop

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-1692

This brings the mobile web nav menu inline with the desktop nav, so that they have the same items in the same order:

comp

This also similarly updates the A/B experiment version of the nav bar — which is currently disabled but not definitively retired yet.


artsyit:
:rocket: PR was released in v24.3.1 :rocket:

PR #3110, opened 2020-01-28
Use open:auth with redirectTo option to skip signup onboarding

dblandin:
When creating a new account for a user via a request for a condition report on an artwork in a sale, skip the onboarding flow so that the user can quickly return to the artwork page to complete their condition report request.


artsyit:
:rocket: PR was released in v24.4.0 :rocket:

PR #3112, opened 2020-01-30
User should not be able to advance to the payment page without entering a phone number with their shipping address

lilyfromseattle:
# [PURCHASE-1757] User should not be able to advance to the payment page without entering a phone number with their shipping address

This is a follow up to fix a regression that I accidentally caused with my phone number field PR.

When I place an order
if I select shipping
and enter a shipping address
I should be required to enter a phone number

Currently, I'm able to advance without entering my phone number.

__After:__
Screen Shot 2020-01-30 at 9 47 39 AM


sweir27:
Nice! #mergeongreen

ashkan18:
Thanks for quick fix! 🙌

artsyit:
:rocket: PR was released in v24.4.1 :rocket:

PR #3117, opened 2020-01-31
[FX-1773] Adjust default jpeg quality for cropped images

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-1773

We have multiple approaches for requesting on-the-fly resized images. One of them is to use these `resize`  and `crop` utility functions from within Reaction code.

While working on image optimizations related to collections, I noticed that the `crop` default was still set to 95, even though [we had settled on 80 a while back](https://github.com/artsy/force/pull/2226), and even though `resize` in this file itself had been [similarly updated previously](https://github.com/artsy/reaction/pull/1491).

This just carries that change through to `crop` as well.


anandaroop:
Tests fixed 😸

artsyit:
:rocket: PR was released in v25.4.1 :rocket:

PR #3118, opened 2020-01-31
Update the auction result layout

zephraph:
This PR:

- Removes the sidebar (which will be re-added with filters)
- Updates the results to be on top of the auction results list
- Slightly updates the result language

image

image

A few learnings from this

- When we add the filtering we'll have to pay special attention to how we do the `Showing X of Y results`
- We need to make sure we're properly co-locating components

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
@@ -1,5 +1,6 @@
 import { Col, Row } from "@artsy/palette"
 import { ArtistAuctionResults_artist } from "__generated__/ArtistAuctionResults_artist.graphql"
+import { AuctionResultsCount_results } from "__generated__/AuctionResultsCount_results.graphql"
 import { PaginationFragmentContainer as Pagination } from "Components/v2"
 import React, { Component } from "react"
 import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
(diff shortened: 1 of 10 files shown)


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -102,47 +104,54 @@ class AuctionResultsContainer extends Component<
...
-
-                  <Box pt={0.5}>
-                    <Separator />
-                  </Box>
-
-                  <ArtistAuctionDetailsModal
-                    auctionResult={state.selectedAuction}
+                */}
+
+                {/* <Col sm={10}> */}
```
sepans: ✂️ ? also [here](https://github.com/artsy/reaction/pull/3118/files#diff-643ca957db8afbe569c78cea5a1bb9ffR154)? or you are keeping them here to be added back when the filters are added?

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -102,47 +104,54 @@ class AuctionResultsContainer extends Component<
...
+                <Flex
```
sepans: 👍

zephraph:
I did given that we may/may not re-use it in the next steps

artsyit:
:rocket: PR was released in v25.2.0 :rocket:

PR #3119, opened 2020-02-01
Avoid performance hit by not importing all components from v2

zephraph:
TL;DR we were accidentally importing a lot of modules we didn't need to in many of our apps. 

---

Reaction currently compiles to commonjs and does not benefit from any treeshaking.  Anytime we use a named import (like `import { foo } from 'bar'`) we import everything from the module `bar` even though we only use `foo`. It's exactly the same as doing `const { foo } = require("bar)`. 

There's a common practice of having an `index.ts` file which exports things from other files. You'll see this in a lot of other projects as it's generally a good way to reduce the verbosity of importing a bunch of stuff. Unfortunately, this pattern doesn't play well if you're compiling to commonjs because you'd get _all_ the dependencies in the index file. 

A prime example of that is `src/Components/v2/index.ts` which is deleted in this PR. 

I discovered this while trying to determine why `openseadragon` was being hoisted into our `common` bundle. After turning off common bundle generation I noticed it was utilized in almost all of our new apps. Using [webpack analyze](http://webpack.github.io/analyse) I was able to find the chain of `openseadragon.js` -> `Lightbox/index.ts` -> `Components/v2/index.ts` -> `ArtworkInfo.tsx`. Because `ArtworkInfo`was using `import { } from 'Components/v2'` it ended up pulling all the dependencies. Likewise with all these other files that I've updated.

---

There are four other `index` files that could be yielding some bloat:
image

I'm not going through the process of unrolling them though, because generally I see unrolling these imports as a band-aid instead of a fix. Ultimately we need to update Reaction's build process to not transpile to common.js modules. That will enable us to properly treeshake and we can essentially undo this PR.


zephraph:
It’s somewhat complicated.

I mean, if we weren’t transpiling to commonjs it wouldn’t have been a problem because a bundle would only include the things it imported. Because we are doing commonjs it gets everything. 

We could set a budget, but the thing that’s likely to grow is common.js itself. I think we should likely focus a lot of effort on slimming that down. 

To your question, it’s a hard thing to prevent currently. Maybe a lint rule. The issue is, fixing it (by remove imports) makes importing modules more ergonomically difficult. We really need to transition to not transpiling. I’ve already got a WIP branch for that, but it’s certainly non-trivial.

mzikherman:
Amazing sleuthing!

artsyit:
:rocket: PR was released in v25.1.2 :rocket:

joeyAghion:
> We could set a budget, but the thing that’s likely to grow is common.js itself. I think we should likely focus a lot of effort on slimming that down.

What I mean (I _think_) is to set a budget for the overall common.js bundle. Wouldn't that surface a future instance of a developer importing a lot of unnecessary code?

xtina-starr:
Wow @zephraph nice job continuing to chase this down!

Can you post a link to your WIP branch when you're in a good place to share?

anandaroop:
Good stuff!

I added the post-game report as a comment on https://artsyproduct.atlassian.net/browse/FX-1676 (and I'm linking this PR to that ticket for posterity's sake).

PR #3120, opened 2020-02-03
[WIP] Switch to Me.orders

ashkan18:
# Change
During https://github.com/artsy/metaphysics/pull/2149 we added `orders` to `Me` from being a root field.  Here we catch up with that change.

@sepans @lilyfromseattle having trouble with with fixing the tests :(


ashkan18:
closing since this was done as part of #3123

PR #3122, opened 2020-02-03
Use es modules

zephraph:
A follow up to #3119, this PR intends to preserve the esmodule module syntax from reaction so that force (and other apps) can apply proper treeshaking w/ webpack depending on usage. 

This is considered a major change because it requires upstream consumers to be ESM compatible (all our projects that use webpack should be). 

I did make some updates to use `import()` which will asynchronously download some files as async chunks but it's configured to _only_ do this on the client. We shouldn't have to mess with something like [react-loadable](https://github.com/jamiebuilds/react-loadable) yet as we're not pulling any async bundles in on the server. 

This does leave us with a few `require` statements, but for simplicity's sake I've left those alone.


babel.config.js
```diff
@@ -0,0 +1,59 @@
...
+      },
+    },
+    presets: [
+      "@babel/typescript",
+      "@babel/react",
+      [
+        "@babel/env",
+        {
+          useBuiltIns: "usage",
+          modules: isTest ? "commonjs" : false,
```
zephraph: The static version of this file should support switching between these two, but I wasn't sure about how the merge order would be handled, so I broke it out into this file format.

src/Components/v2/Lightbox/Lightbox.tsx
```diff
@@ -206,8 +206,7 @@ class LightboxComponent extends React.Component<LightboxProps, LightboxState> {
     if (element) {
       this.setState({
         element,
-        // FIXME: convert to import('openseadragon) once force supports it
-        promisedDragon: Promise.resolve(require("openseadragon")),
+        promisedDragon: import("openseadragon"),
```
zephraph: Because this is in `componentDidMount` this chunk will only be pulled in clientside.

zephraph:
Given that this is a relatively complex change, I'm going to create a canary of this package (after the tests are passing) and set up a review app w/ that so we can poke at it.

PR #3124, opened 2020-02-03
Add missing title to sort select on auction results

zephraph:
Adds `Sort:` in front of the select as per the design specs. Also updates some of the language to match design spec. 

**Before**

image

**After**

image


src/Apps/Artist/Routes/AuctionResults/Components/SortSelect.tsx
```diff
@@ -23,6 +23,7 @@ export const SortSelect = () => {
     <Subscribe to={[AuctionResultsState]}>
       {(filters: AuctionResultsState) => (
         <SelectSmall
+          title="Sort"
```
damassi: Does this still look good on mobile? Part of the reason it was removed was to make consistent with responsive. @oxaudo would know more.

damassi:
@zephraph - can this be closed?

PR #3126, opened 2020-02-04
[Tracking] Consolidate page-view tracking within new trackingMiddleware 

damassi:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2209

This PR cleans up our pageview tracking by consolidating things behind a `trackingMiddleware` function, which is piped through to our router. This fires whenever our route changes, minus query params. This allows us to move various tracking calls from the route config, router links and individual apps into one location. 

Inside of this function I've also moved our experimental A/B test call. 

As things were at one point all spread out, and we were possibly double firing or under firing depending on component mount or re-render, I think this might have a significant impact on our metrics. 

This also adds referrer tracking.

tracking


src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -0,0 +1,77 @@
...
+ * PageView tracking middleware for use in our router apps.
+ *
+ * @see Artsy/Router/buildClientApp.tsx for mount point
+ */
+export function trackingMiddleware() {
+  return _store => next => action => {
+    const { type, payload } = action
+
+    switch (type) {
+      case ActionTypes.UPDATE_LOCATION: {
```
damassi: Any app that is built using our routing framework will now have automatic page-view tracking on URL changes.

src/Apps/Order/Routes/__tests__/Accept.test.tsx
```diff
@@ -18,11 +18,6 @@ import {
 import { AcceptFragmentContainer } from "../Accept"
 import { OrderAppTestPage } from "./Utils/OrderAppTestPage"
 
-jest.mock("Artsy/Analytics/trackPageView", () => ({
-  trackPageView: jest.fn(),
-}))
-const mockTrackPageView = require("Artsy/Analytics/trackPageView")
-  .trackPageView as jest.Mock
```
damassi: This test code was removed because we're now handling things one level up and no longer need individual route checks.

src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -0,0 +1,77 @@
...
+  // REPLACE: "@@farce/REPLACE",
+  // TRANSITION: "@@farce/TRANSITION",
+  // GO: "@@farce/GO",
+  // CREATE_HREF: "@@farce/CREATE_HREF",
+  // CREATE_LOCATION: "@@farce/CREATE_LOCATION",
+  // DISPOSE: "@@farce/DISPOSE",
+}
+
+/**
+ * PageView tracking middleware for use in our router apps.
```
damassi: All page-level tracking should exclusively be handled in this file

src/Artsy/Router/buildClientApp.tsx
```diff
@@ -65,6 +66,7 @@ export function buildClientApp(config: RouterConfig): Promise<Resolve> {
           parse: queryStringParsing,
           stringify: qs.stringify,
         }),
+        trackingMiddleware(),
```
damassi: In the future, other middleware can go here. It conforms to Redux's [specification](https://redux.js.org/advanced/middleware/).

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -96,7 +79,9 @@ export class ArtworkApp extends React.Component<Props> {
       artwork: { listPrice, availability, is_offerable, is_acquireable },
     } = this.props
 
-    // Pageview
+    // FIXME: This breaks our global pageview tracking in the router level.
+    // Can these props be tracked on mount using our typical @track() or
+    // trackEvent() patterns as used in other apps?
```
damassi: @mzikherman - How should we handle this? Maybe reach out to Ani?

damassi:
@eessex - how do I go about bumping the Node version in Circle?

package.json
```diff
@@ -7,7 +7,7 @@
   "author": "Art.sy Inc <it@artsymail.com>",
   "license": "MIT",
   "engines": {
-    "node": "^10.13.0"
+    "node": "^12.14.0"
```
mzikherman: What happens if this is bumped, with respect to Palette, Force, Volt (which I believe uses Reaction stuff), etc.?

damassi: Actually, I'm going to remove this constraint so that the consuming apps can set the version.

src/Apps/Artwork/ArtworkApp.tsx
mzikherman: Yea, we should figure out if we need these in the page events, or if they can be inferred by the team later on.

The `trackProductView` call above here is also an ad-hoc tracking call that we want to fire on all navigations of an artwork page (such as from an artwork to another artwork), _if_ the resulting artwork is ecommerce.

In the tracking middleware, is it possible to reach into the props or Relay store or something, in order to figure out if we're navigating from an artwork to another artwork, and if so..include these additional properties/events?

src/Artsy/Analytics/trackingMiddleware.ts
mzikherman: This whole file seems 👍 to me.

package.json
```diff
@@ -6,9 +6,6 @@
   "repository": "https://github.com/artsy/reaction.git",
   "author": "Art.sy Inc <it@artsymail.com>",
   "license": "MIT",
-  "engines": {
-    "node": "^10.13.0"
-  },
```
zephraph: It'd probably be best if we didn't group this change in this PR. Also, I wonder if we shouldn't just introduce a range while we're migrating.  

```
"engines": {
	"node": ">=10.13.0 <13.0.0"
}
```

damassi: I don't think it's necessary to have a lock here, we're not doing anything special other than running storybooks and compiling files with babel. If for some reason it causes problems we can add it back, but what i think this will do is remove problems and friction as client apps consume reaction.

damassi: (Will move this to differnt PR tho!)

zephraph: Sounds good to me.

src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -0,0 +1,67 @@
...
+  return store => next => action => {
+    const { type, payload } = action
+
+    switch (type) {
+      case ActionTypes.UPDATE_LOCATION: {
+        const { pathname } = payload
+        const referrer = get(
+          store.getState(),
+          state => state.found.match.location.pathname
+        )
```
damassi: Pluck referrer from previous redux state. (Redux FTW.)

mzikherman:
I'll pull this down and link to Force and check it out.

Looking at the code (and I see the convo in Slack), I think the `product viewed` event still isn't fully hooked up. It remains as-is, so it's something that fires `track` on mount for artworks: https://github.com/artsy/reaction/blob/ed7f6c3fd52102becb8bec90968add28fc5766c9/src/Apps/Artwork/ArtworkApp.tsx#L44

And, without client side routing and full page reload, that works. But navigating from an artwork to another artwork, doesn't re-mount and so doesn't fire that subsequent event.

It's more of a general thing- we've placed some custom analytics in `componentDidMount`. In the client-side routing world...that doesn't work when the app is already mounted.

src/Apps/Artwork/ArtworkApp.tsx
mzikherman: I guess we should remove this pageview tracking call (currently there's two that fire when navigating to an artwork, this one and the middleware one).

mzikherman:
One initial observation: the 'timing' of the pageview event feels a bit early. I wouldn't think much of it, but Ani mentioned it as possibly being significantly different than the control group for routing. With client side routing, the pageview event and URL seem to update, while the purple bar is still loading and before the page is truly visible and ready.

This might be fine! But, it may be different enough to make testing vs. control on pageviews possibly suspect.

Other than that it looks great locally!

damassi:
@mzikherman 

> One initial observation: the 'timing' of the pageview event feels a bit early

Since so much other stuff was was cleaned up I think we should roll with it. If things are significantly still off we can revise further. 

I've rigged up artwork page tracking / product tracking in [this commit](https://github.com/artsy/reaction/pull/3126/commits/c724f982c766d693a91d8a7d4589f26c88aaa5fb), which also excludes it from the middleware via a new `excludePaths: [string]` middleware option.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -22,29 +22,75 @@ import { track } from "Artsy/Analytics"
...
-  // TODO: Move the below tracking, which consists of:
-  //
-  //  * a custom `track` event when the artwork is acquireable or in an auction
-  //  * a custom pageview event including extra metadata
-  //
-  // into an appropriate wrapper HOC.
+  /**
+   * On mount, trigger a page view and product view
+   *
+   * FIXME: We're manually invoking pageView tracking here, instead of within
```
mzikherman: Cool, thanks.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -22,46 +22,73 @@ import { track } from "Artsy/Analytics"
...
     this.trackPageview()
     this.trackProductView()
-    this.trackABTest()
   }
 
-  // TODO: Remove after AB test ends.
-  trackABTest() {
-    const { tracking } = this.props
-    const { CLIENT_NAVIGATION_V2 } = sd
+  componentDidUpdate(prevProps) {
```
mzikherman: Sweet, this seems to be good locally. I recall running into an issue where the `match: { location }` prop didn't seem to update as you navigated (it always reflected the _first_ route you hit, so starting in /collect, then clicking on an artwork, and then on another artwork, would wind up being unexpected with /collect still coming thru as the prop).

Anyway! Usual caveat with `componentDidUpdate` applies, this might unexpectedly trigger more than you'd expect.

mzikherman:
Merge when ready (I guess there's no breaking change or Force update needed).

src/Apps/Artwork/ArtworkApp.tsx
damassi: Ah yeah, this is actually resolved by https://github.com/artsy/reaction/pull/3126/commits/c8368e1f8a16df523391e9785759d2a173dcb50d#diff-1eac5195ddbe10a2fe914a51461798d1R240-R246, which by tapping into the router context directly will always keep things perfectly in synch, as if it were a part of the middleware. (If for some reason the speediness of the analytics call remains an issue this is the direction we'd have to go.)

artsyit:
:rocket: PR was released in v25.4.0 :rocket:

PR #3127, opened 2020-02-04
Resize category rail

xtina-starr:
Addresses: [FX-1775](https://artsyproduct.atlassian.net/browse/FX-1775)

The "Browse by category" rail is another instances where we were pulling in larger images than are needed. This PR uses the resize util to resize the collection thumbnail to 60x60.

Screen Shot 2020-02-04 at 9 54 16 AM


artsyit:
:rocket: PR was released in v25.3.0 :rocket:

PR #3128, opened 2020-02-04
Remove Node verson requirement

damassi:
Reaction isn't doing anything special that would require a constraint like this -- just running storybooks and compiling files with babel. However, when upgrading other apps that consume reaction, this constrain often gets in the way and requires more coordination than necessary. This removes that friction.


artsyit:
:rocket: PR was released in v25.3.1 :rocket:

PR #3129, opened 2020-02-04
Fixes a bug where unregistered users do not get feedback on bid + rergister

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-835

This bug was caused by the presence of the `registrationStatus` field in `src/Apps/Auction/routes.tsx`:

https://github.com/artsy/reaction/blob/ed7f6c3fd52102becb8bec90968add28fc5766c9/src/Apps/Auction/routes.tsx#L49-L52

Because of that field, Relay monitors any change in that specific field and [re-renders the entire page](https://github.com/artsy/reaction/blob/ed7f6c3fd52102becb8bec90968add28fc5766c9/src/Apps/Auction/routes.tsx#L20-L33) when we receive a new `registartionStatus` object from the results of the `createBidderPosition` mutation:

https://github.com/artsy/reaction/blob/ed7f6c3fd52102becb8bec90968add28fc5766c9/src/Apps/Auction/Routes/ConfirmBid/index.tsx#L94-L97

For some reason, when there is an update, all the components are re-mounted (completely thrown away and re-initialized from scratch) as opposed to just receiving prop updates. As a result of this the reference to the Formik object in the `setTimeout` would be stale and end up causing a warning from React:

Screen Shot 2020-01-30 at 4 43 20 PM

This PR works around the issue by removing the `registrationStatus` field from the top-level query. Generally we should avoid specifying any fields and instead spcify fragments so the top-level query can stay as a place to coordinate queries. [We request the field in `BidForm` anyway](https://github.com/yuki24/reaction/blob/8a632b2eb094f4cfc13d2a9035cc9a88e7a2453a/src/Apps/Auction/Components/BidForm.tsx#L317-L320) so the actual HTTP response from MP won't be affected.

I also added a `clearTimeout` to clean up unnecessary execution call in case the user leaves the page between polling calls.

~I'm trying to clean up the GraphQL definitions to reduce confusion around this behaviour, but it's a bit difficult to get the types right. I'm going to send another PR.~ This may require a bit larger refactoring than I initially anticipated. There is room for improvement for sure, but we probably need someone who is more familiar with https://github.com/4Catalyzer/found.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@artsy/reaction",
-  "version": "25.5.7",
+  "version": "25.5.8",
   "description": "Force’s React Components",
   "main": "dist/index.js",
   "repository": "https://github.com/artsy/reaction.git",


src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
```diff
@@ -875,6 +871,27 @@ describe("Routes/ConfirmBid", () => {
...
+        confirmBidBidderPositionQueryWithOutbid
+      )
+
+      await page.fillFormWithValidValues()
+      await page.agreeToTerms()
+      await page.submitForm()
+
+      expect(window.location.assign).not.toHaveBeenCalled()
+      expect(page.text()).toContain("Your bid wasn’t high enough.")
+    })
```
yuki24: Sadly this test passes without the fix because of the difference between the actual impl and [the test setup](https://github.com/artsy/reaction/blob/ed7f6c3fd52102becb8bec90968add28fc5766c9/src/DevTools/createTestEnv.tsx#L222).

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -59,21 +59,23 @@ interface ConfirmBidProps extends ReactStripeElements.InjectedStripeProps {
 const MAX_POLL_ATTEMPTS = 20
 
 export const ConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
+  let timer = null
```
erikdstock: I'm thinking this is okay since these variables (timer, pollcount) aren't used to render the form, but thought I'd mention anyways - is it okay in react to be declaring and mutating objects in the closure of a FC? I'm thinking yes, no need to use a state hook, but wanted to mention it.

yuki24:
The tests are failing and I'm working on them.

yuki24:
I just noticed that this would introduce a bug where unqualified users could open the confirm bid page rather than getting redirected to `/auction/:sale_slug/confirm-registration`. This needs to be taken care of before merging.

src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx
dleve123: Can you elaborate a bit? Is there a way we can get the test to fail legitimately?

src/Apps/Auction/getRedirect.ts
```diff
@@ -43,7 +43,7 @@ export function confirmBidRedirect(
   const { saleArtwork } = artwork
 
   const { sale } = saleArtwork
-  const { registrationStatus } = sale
+  const { registrationStatus } = sale as any
```
dleve123: is the `any` needed here? Seems like `sale` is appropriately typed

src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -59,21 +59,23 @@ interface ConfirmBidProps extends ReactStripeElements.InjectedStripeProps {
...
   let pollCount = 0
   let registrationTracked = false
 
   const { artwork, me, relay, stripe } = props
   const { saleArtwork } = artwork
   const { sale } = saleArtwork
+  const { registrationStatus } = sale as any
```
dleve123: similar question to below. I know that in the red diff `sale` was cast to `any` – but pulling down the code in VScode indicates that `saleArtwork` is typed. Have we tried to remove the `any`?

src/Apps/Auction/Components/BidForm.tsx
```diff
@@ -315,6 +315,7 @@ export const BidFormFragmentContainer = createFragmentContainer(BidForm, {
       sale {
         slug
         registrationStatus {
+          internalID
```
dleve123: why is `internalID` necessary here?

dleve123:
@yuki24 and I discussed this IRL and I understand and to provide context to others:

The latest blocker here is that without `registrationStatus` in src/Apps/Auction/routes.tsx, the `getRedirect` functional redirect approach isn't working.

To implement redirecting with Found and Relay in a way that we think will work, we'll need render [Found `<Redirect>` components](https://github.com/4Catalyzer/found#redirects) with data needs separate from the Route component.

@yuki24 can you confirm this?

PR #3135, opened 2020-02-06
[Tracking Bugfix] Fix url and referrer tracking

damassi:
Fixes an issue that was observed by @anipetrov: 

From [Slack message](https://artsy.slack.com/archives/CA8SANW3W/p1581028242462600?thread_ts=1580919019.365800&cid=CA8SANW3W): 

1. tracking pageview referrers from within reaction is producing referrers that are paths eg referrer: "/collection/post-war". in our transformations, we expect those to be urls, eg. referrer: "https://www.artsy.net/collection/post-war" we use external libraries to parse pageview referrers to define when a session is external and give proper channel attribution (eg if someone is coming from google, we parse that to search). those libraries expect urls.
2. the url property on reaction pages is inconsistent, eg in this case, the path is correct, referring to collection/roy-lihtenstein-landscapes but the url is the url of the previous page. seems to only happen upon a consecutive pageview of the same type, so collection > collection 

So the fix entailed bringing incorporating `sd.APP_URL` from force, and explicitly setting the `url` property to avoid the race.

### DIFF
--- a/src/Artsy/Analytics/trackingMiddleware.ts
+++ b/src/Artsy/Analytics/trackingMiddleware.ts
@@ -1,5 +1,6 @@
 import { trackExperimentViewed } from "Artsy/Analytics/trackExperimentViewed"
 import ActionTypes from "farce/lib/ActionTypes"
+import { data as sd } from "sharify"
 import { get } from "Utils/get"
 import { getENV } from "Utils/getENV"
 import createLogger from "Utils/logger"
@@ -43,13 +44,34 @@ export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
           if (!foundExcludedPath) {
             logger.warn("Tracking PageView:", pathname)
 
-            analytics.page(
-              { path: pathname, referrer },
-              { integrations: { Marketo: false } }
-            )
+            const trackingData: {
+              path: string
+              referrer?: string
+              url: string
+            } = {
+              path: pathname,
+              url: sd.APP_URL + pathname,
+            }
+
+            // TODO: Remove after EXPERIMENTAL_APP_SHELL AB test ends.
+            if (getENV("EXPERIMENTAL_APP_SHELL")) {
+              trackingData.referrer = sd.APP_URL + referrer
+            }
+
+            // TODO: Remove after EXPERIMENTAL_APP_SHELL AB test ends.
+            if (
+              ["/collect", "/collections", "/collection/"].some(path =>
+                pathname.includes(path)
+              ) &&
+              referrer
+            ) {
+              trackingData.referrer = sd.APP_URL + referrer
+            }
+
+            analytics.page(trackingData, { integrations: { Marketo: false } })
           }
 
-          // TODO: Remove after AB test ends.
+          // TODO: Remove after EXPERIMENTAL_APP_SHELL AB test ends.
           if (getENV("EXPERIMENTAL_APP_SHELL")) {
             trackExperimentViewed("client_navigation_v2")
           }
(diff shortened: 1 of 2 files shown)


src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -43,13 +44,34 @@ export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
...
             logger.warn("Tracking PageView:", pathname)
 
-            analytics.page(
-              { path: pathname, referrer },
-              { integrations: { Marketo: false } }
-            )
+            const trackingData: {
+              path: string
+              referrer?: string
+              url: string
```
mzikherman: It doesn't look like `url` was provided at all before. Does that automatically come thru as a relative path, and that's no good anymore? (and we prefer a full URL)

src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -43,13 +44,34 @@ export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
...
+            if (getENV("EXPERIMENTAL_APP_SHELL")) {
+              trackingData.referrer = sd.APP_URL + referrer
+            }
+
+            // TODO: Remove after EXPERIMENTAL_APP_SHELL AB test ends.
+            if (
+              ["/collect", "/collections", "/collection/"].some(path =>
+                pathname.includes(path)
+              ) &&
+              referrer
```
mzikherman: What's `referrer` here? I don't fully understand this.

mzikherman: Oh nevermind!

src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -43,13 +44,34 @@ export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
...
+            }
+
+            // TODO: Remove after EXPERIMENTAL_APP_SHELL AB test ends.
+            if (getENV("EXPERIMENTAL_APP_SHELL")) {
+              trackingData.referrer = sd.APP_URL + referrer
+            }
+
+            // TODO: Remove after EXPERIMENTAL_APP_SHELL AB test ends.
+            if (
+              ["/collect", "/collections", "/collection/"].some(path =>
```
mzikherman: So this basically ensures this for the collect suite of pages, which is always client-side routed.

mzikherman:
Merge when ready!

src/Artsy/Analytics/trackingMiddleware.ts
damassi: referrer means the previous path. This too is inferrered from segment if not explicitly provided. Since these pages are already client-side routing enabled, we check to see if there's already a referer in the router state on the client. This means we've navigated from say collection -> collection. Use that if so. If not, we've performed an SSR render (first init) so use the referrer from the server.

src/Artsy/Analytics/trackingMiddleware.ts
damassi: `url` is automatically inferrered by segment if not provided. The issue is that we were firing the tracking call before the url was completely transitionioned, so segment would immediately invoke the window.location object and populate.

src/Artsy/Analytics/trackingMiddleware.ts
damassi: Yup

damassi:
Sending it to staging, and if all looks good going to deploy to prod 👍

artsyit:
:rocket: PR was released in v25.5.2 :rocket:

src/Artsy/Analytics/trackingMiddleware.ts
mzikherman: Ah ok, saw your other comment too. The fact the Segment lib automatically infers and includes this (if not specified) makes sense with this change, I hadn't realized that.

PR #3136, opened 2020-02-07
[Tracking Bugfix] Make path absolute

damassi:
Missed the requirement that the path, too, be absolute.  cc @mzikherman


artsyit:
:rocket: PR was released in v25.5.3 :rocket:

PR #3137, opened 2020-02-07
[Refactor] Implement route-level bundle splitting

damassi:
JIRA: https://artsyproduct.atlassian.net/browse/PLATFORM-2222
Force PR: https://github.com/artsy/force/pull/5024

This PR implements code-splitting via [webpack](https://webpack.js.org/guides/code-splitting/) (on the force side) and [`@loadable/components`](https://loadable-components.com/docs/server-side-rendering/). 

Something similar was previously implemented when the v1 router implementation was created, but was subsequently removed because we didn't yet have client-side routing wired up and the library introduced some tooling bloat. 

Now that we've finished client-side routing, we need to bring it back so that we don't bundle all of our app code inside of a single common.js bundle.


.babelrc
```diff
@@ -2,36 +2,57 @@
...
+      {
+        "useBuiltIns": "usage"
+      }
+    ]
   ],
   "plugins": [
-    ["@babel/plugin-proposal-decorators", {
-      "legacy": true
-    }],
+    "@loadable/babel-plugin",
```
damassi: Babel plugin handles a lot of the magic

src/Apps/Artist/ArtistApp.tsx
```diff
@@ -153,3 +153,6 @@ export const ArtistAppFragmentContainer = createFragmentContainer(
     `,
   }
 )
+
+// Top-level route needs to be exported for bundle splitting in the router
+export default ArtistAppFragmentContainer
```
damassi: `@loadable/components` has a hard requirement that split bundles export a default. There's currently a [PR open](https://github.com/gregberge/loadable-components/pull/483) to fix this.

src/Apps/Artist/routes.tsx
```diff
@@ -68,7 +68,7 @@ graphql`
 export const routes: RouteConfig[] = [
   {
     path: "/artist/:artistID",
-    Component: ArtistApp,
+    getComponent: () => loadable(() => import("./ArtistApp")),
```
damassi: For things that we want to split, we implement this interface, which wraps ES6 dynamic `import()` statements.

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -28,6 +29,23 @@ import { RouterConfig } from "./"
...
+import { getENV } from "Utils/getENV"
+
+// This is the stats file generated by webpack loadable plugin
+let statsFile
+try {
+  statsFile = path.resolve(
+    process.cwd(),
+    "public/assets",
+    "loadable-stats.json"
+  )
```
damassi: On the force side of things, when we compile our app Webpack is responsible for creating bundle splits wherever it sees `import()` keyword. Loadable components then uses the stats file to parse which components have been dynamically split and creates a manifest to synchronously load on the server and rehydrate on the client.

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -107,14 +130,38 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
...
-                sheet.collectStyles(<ServerApp tags={headTags} />)
-              )
+              // Extract bundle split chunks for SSR
+              let bundleJSX
+
+              // FIXME: Remove once EXPERIMENTAL_APP_SHELL A/B test completes
+              if (getENV("EXPERIMENTAL_APP_SHELL")) {
+                const extractor = new ChunkExtractor({
+                  statsFile,
+                  entrypoints: ["experimental-app-shell"],
```
damassi: Currently, the only webpack chunk that supports SSR bundle-splitting is code within the new app shell hot path. Since all new apps will eventually be mounted here, we can start aggressively bundle splitting other code, and further reduce main bundle sizes. For example -- subroutes within the artist page (the tabs) can fetched only when the user clicks the tab, reducing the main bundle size.

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -107,14 +130,38 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
...
+                const bundleScriptTags = extractor.getScriptTags()
+                scripts.concat(bundleScriptTags)
+
+                // Wrap component tree in library contexts to extract usage
+                bundleJSX = extractor.collectChunks(
+                  sheet.collectStyles(<ServerApp tags={headTags} />)
+                )
+
+                // TODO: Wire up if A/B test is successful
+                // extractor.getLinkTags()
```
damassi: This is in support for "fetch bundle on hover" style behavior, similar to gatsby or instant-page.

zephraph:
@damassi, you’re my hero

src/Apps/Artist/ArtistApp.tsx
zephraph: It’s a fine tradeoff for now 👍

src/Apps/Artist/routes.tsx
```diff
@@ -68,7 +68,8 @@ graphql`
 export const routes: RouteConfig[] = [
   {
     path: "/artist/:artistID",
-    Component: ArtistApp,
+    getComponent: () =>
+      loadable(() => import(/* webpackChunkName: "ArtistApp" */ "./ArtistApp")),
```
zephraph: 👍

src/Artsy/Router/buildServerApp.tsx
```diff
@@ -107,14 +132,48 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
...
+              if (getENV("EXPERIMENTAL_APP_SHELL")) {
+                const extractor = new ChunkExtractor({
+                  statsFile,
+                  entrypoints: [
+                    "ArtistApp",
+                    "ArtworkApp",
+                    "CollectApp",
+                    "CollectionApp",
+                    "CollectionsApp",
+                    "SearchApp",
```
damassi: @zephraph - it looks like i had to explicitly set the bundles here, based on the webpack identifier strings. Maybe you can poke and see if there's a better way to automate this...

damassi: Yeah looking at this more, this is funny -- it's loading all of these bundles up front, versus just the bundle for the page we're on. We need to play with the configuration.

src/Apps/Artist/Components/StyledLink.tsx
```diff
@@ -0,0 +1,12 @@
+import { RouterLink } from "Artsy/Router/RouterLink"
```
zephraph: This file was pulled out because importing it from the `ArtistApp` directly causes the `ArtistApp` to be bundled with the app shell. We'll have to use a linter or other mechanism to  help people avoid this pitfall.

damassi:
@zephraph - I know you mentioned some chunk cleanup; otherwise this is ready to go from my side. (Also note that I just invited you to the A/B test meeting on Monday.) 

Thanks again for digging in with me 💯

zephraph:
I was looking to do some optimization of how the scripts were delivered, but after investigating prod I realized that all our assets are being served via http2. 

I think this might be good to ship.

damassi:
Aighttt lets do this

artsyit:
:rocket: PR was released in v25.7.0 :rocket:

PR #3138, opened 2020-02-07
[Tracking Bugfix] Fix artwork page 1-off

damassi:
Caught a small bug on staging related to how we handle page-view tracking differently in the Artwork app. Had to update with the full path.


artsyit:
:rocket: PR was released in v25.5.4 :rocket:

PR #3139, opened 2020-02-07
As a user I see the updated list view of auction results and i can expand the result

lilyfromseattle:
# [PURCHASE-1732] As a user I see the updated list view of auction results and i can expand the result

This does not involve adding any new filters or anything. It's just updating the list view item to match the new UI.

AC: 
- UI is its responsive 
- Result can expand
- Matches the spec
https://zpl.io/agB7doM

__Before:__

Mobile: 
Screen Shot 2020-02-07 at 4 14 08 PM

Desktop: 
Screen Shot 2020-02-07 at 4 14 52 PM

__After:__

Mobile: 
Screen Shot 2020-02-07 at 4 14 43 PM

Desktop:
Screen Shot 2020-02-07 at 4 15 03 PM

[PURCHASE-1732]: https://artsyproduct.atlassian.net/browse/PURCHASE-1732


CHANGELOG.md
```diff
@@ -20,8 +20,7 @@
 
 - Brian Beckerle ([@brainbicycle](https://github.com/brainbicycle))
 
----
-
+--
```
zephraph:
```suggestion
---

```

We should probably leave this how it was given that this document is automatically updated.

zephraph:
So I've been debugging a hydration issue that @lilyfromseattle was seeing. 

The console warning is 

```
Warning: Expected server HTML to contain a matching <a> in <div>.
```

I traced it down to the carousel hydration behavior, so it's _not_ directly related to this PR (but we do need to figure out how to solve it before we merge)

zephraph:
In fixing the rehydration/media issue, I noticed that the logged out state still  needs some updates

image

spec: 

image

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -320,19 +320,21 @@ const renderPricing = (salePrice, user, mediator, size) => {
     )
   } else {
     const btnSize = size === "xs" || "sm" ? "small" : "large"
+    const buttonMargin = size === "xs" ? "10px" : "0px"
```
zephraph:
```suggestion
    const buttonMargin = size === "xs" ? 1 : 0
```

It's good to use the spacing units from palette instead of exact pixel sizes when it matches up.

artsyit:
:rocket: PR was released in v25.6.0 :rocket:

PR #3140, opened 2020-02-09
[Bugfix] Ensure nav closed on clienside routes

damassi:
Noticed a little bug in the mobile nav: When the new app shell is enabled, the mobile nav doesn't close on `/collect` link click because there's not a full page transition.


artsyit:
:rocket: PR was released in v25.5.7 :rocket:

PR #3141, opened 2020-02-09
[Clientside Routing] Update mobile page loader to be more obvious

damassi:
We're speculating that this might be what impacted inquiries on mobile during the A/B test -- meaning, a user has clicked a link, and they can't see that a loader is going by on the top of the screen and they think the page may have timed out? This puts it in the middle, and dims the background a bit. After the A/B test results we'll need to go back to design for something a more permanent. 

loader


mzikherman:
Does this need any corresponding Force CSS update?

damassi:
Nope, no css update.

artsyit:
:rocket: PR was released in v25.5.8 :rocket:

PR #3143, opened 2020-02-10
Remove deprecated fonts from classic articles, fix bylines

eessex:
Some updates to enable displaying classic articles in React
- Use palette fonts
- Show channel name when provided for byline
- add share components

Screen Shot 2020-02-10 at 11 55 26 AM

### DIFF
--- a/src/Components/Publishing/Byline/ClassicByline.tsx
+++ b/src/Components/Publishing/Byline/ClassicByline.tsx
@@ -1,9 +1,13 @@
-import { color } from "@artsy/palette"
-import { avantgarde } from "Assets/Fonts"
+import { Box, color, Sans } from "@artsy/palette"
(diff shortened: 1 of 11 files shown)


src/Components/Publishing/Byline/ClassicByline.tsx
```diff
@@ -12,33 +16,55 @@ interface ClassicBylineProps {
...
+      author,
+      published_at,
+      slug,
+      social_title,
+      thumbnail_title,
+    },
     date,
   } = props
 
   const contributors = getAuthorByline(contributing_authors, false)
```
pepopowitz: There's a good amount of logic in this component. I see that there are snapshot tests - do they cover all the branches? If not, now is probably a good time to add coverage.

artsyit:
:rocket: PR was released in v25.8.0 :rocket:

PR #3144, opened 2020-02-11
Fixes a bug where unregistered users who bid and register and don't meet the reserve do not get feedback that their bid did not meet the reserve

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-835

We were seeing the following warning as well as the issue described in the ticket:

Screen Shot 2020-01-30 at 4 43 20 PM

The reason why this is occurring is that we are dynamically creating a trackable component while rendering in `ConfirmBid/index.tsx`: 

https://github.com/artsy/reaction/blob/98f689f4bda9f53b09b92135fedb2d376806a8fc/src/Apps/Auction/Routes/ConfirmBid/index.tsx#L360-L366

And here the `Component` variable always holds a reference to a newly created object. When this component is passed into the React VDOM and a re-render is triggered, the newly created component is considered to be a "diff" even though it's essentially the same thing. As a result of this, all the components below it would get thrown away once the new trackable component gets in, making all object references stale (and thus the warning above). The solution is simply to avoid creating component definitions dynamically.

## Mitigation

It is a bit tricky to write a test. In theory, we should be able to replicate this bug with the test below:

```tsx
it("allows the user to place a bid after agreeing to terms", async () => {
  const env = setupTestEnv()
  const page = await env.buildPage({
    mockData: FixtureForUnregisteredUserWithoutCreditCard,
  })

  createTokenMock.mockResolvedValue(stripeTokenResponse)
  env.mutations.useResultsOnce(createCreditCardAndUpdatePhoneSuccessful)
  env.mutations.useResultsOnce(createBidderPositionSuccessfulAndBidder)
  mockBidderPositionQuery.mockResolvedValue(
    confirmBidBidderPositionQueryWithOutbid
  )

  await page.fillFormWithValidValues()
  await page.agreeToTerms()
  await page.submitForm()

  expect(window.location.assign).not.toHaveBeenCalled()
  expect(page.text()).toContain("Your bid wasn’t high enough.")
})
```

Unfortunately, this test passes because of the difference between the actual production code and the test setup. One way to work around this is to deliberately change the `track<ConfirmBidProps>(...)` part of the code to:

```tsx
export class TrackingWrappedConfirmBidRoute extends React.Component<ConfirmBidProps> {
  render() {
    const Component = track<ConfirmBidProps>(props => ({
      context_page: Schema.PageName.AuctionConfirmBidPage,
      auction_slug: props.artwork.saleArtwork.sale.slug,
      artwork_slug: props.artwork.slug,
      sale_id: props.artwork.saleArtwork.sale.internalID,
      user_id: props.me.internalID,
    }))(StripeWrappedConfirmBidRoute)

    return <Component {...this.props} />
  }
}
```

And then we could call the `forceUpdate()` function:

```tsx
// It is important to force-update here otherwise the test does not
// replicate the issue described in:
// https://artsyproduct.atlassian.net/browse/AUCT-835
page.find(TrackingWrappedConfirmBidRoute).instance().forceUpdate()

expect(window.location.assign).not.toHaveBeenCalled()
expect(page.text()).toContain("Your bid wasn’t high enough.")
```

At this point the test would start failing the way we want it to fail.

Finally, we could fix the test by moving the `Component` definition out of the tracking component:

```tsx
const Component = track<ConfirmBidProps>(props => ({
  context_page: Schema.PageName.AuctionConfirmBidPage,
  auction_slug: props.artwork.saleArtwork.sale.slug,
  artwork_slug: props.artwork.slug,
  sale_id: props.artwork.saleArtwork.sale.internalID,
  user_id: props.me.internalID,
}))(StripeWrappedConfirmBidRoute)

export class TrackingWrappedConfirmBidRoute extends React.Component<ConfirmBidProps> {
  render() {
    return <Component {...this.props} />
  }
}
```

This seems like deliberately written code and questionable if you have no context around what happened to the bid + register form.

So I don't think testing is a good way to capture potential problems like this. The change in this PR will solve the issue, but here are some other things we can do to avoid running into the same issue in the future:

 * Just be careful when you are using `track` :trollface: 
 * Use a class component and `@track` decorator so TS will deal with creating a new component for us
 * Write a linter that checks certain calls that could invalidate the VDOM tree are not being used at runtime

So I'd be curious to hear what you all think.

## Screenshot

AUCT-835

### DIFF
--- a/src/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -356,17 +356,13 @@ export const StripeWrappedConfirmBidRoute: React.FC<ConfirmBidProps> = props => 
   )
 }
 
-const TrackingWrappedConfirmBidRoute: React.FC<ConfirmBidProps> = props => {
-  const Component = track<ConfirmBidProps>(p => ({
-    context_page: Schema.PageName.AuctionConfirmBidPage,
-    auction_slug: p.artwork.saleArtwork.sale.slug,
-    artwork_slug: p.artwork.slug,
-    sale_id: p.artwork.saleArtwork.sale.internalID,
-    user_id: p.me.internalID,
-  }))(StripeWrappedConfirmBidRoute)
-
-  return <Component {...props} />
-}
+const TrackingWrappedConfirmBidRoute = track<ConfirmBidProps>(props => ({
+  context_page: Schema.PageName.AuctionConfirmBidPage,
+  auction_slug: props.artwork.saleArtwork.sale.slug,
+  artwork_slug: props.artwork.slug,
+  sale_id: props.artwork.saleArtwork.sale.internalID,
+  user_id: props.me.internalID,
+}))(StripeWrappedConfirmBidRoute)
 
 export const ConfirmBidRouteFragmentContainer = createFragmentContainer(
   TrackingWrappedConfirmBidRoute,


src/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -356,17 +356,13 @@ export const StripeWrappedConfirmBidRoute: React.FC<ConfirmBidProps> = props =>
...
-
-  return <Component {...props} />
-}
+const TrackingWrappedConfirmBidRoute = track<ConfirmBidProps>(props => ({
+  context_page: Schema.PageName.AuctionConfirmBidPage,
+  auction_slug: props.artwork.saleArtwork.sale.slug,
+  artwork_slug: props.artwork.slug,
+  sale_id: props.artwork.saleArtwork.sale.internalID,
+  user_id: props.me.internalID,
+}))(StripeWrappedConfirmBidRoute)
```
zephraph: Good find! That was a super tricky issue. 👍

zephraph:
I'll give the reviewers a chance to pitch in, but I'm 👍. This would, as you pointed out, be a significant challenge to test via a unit level test. It would probably be good to consider this flow for an e2e test at some point though.

dleve123:
@yuki24 I'm incredibly stoked that we fixed this specific issue! Amazing work.

With that said, I think we would be remiss to just merge this PR without something that'll help other future developers not make the same mistake.

A couple of thoughts:

- is it possible to change [`ConfirmBid.test.tsx`](https://github.com/artsy/reaction/blob/29703cd9e891b85b9f13b5907528a3e900bb328c/src/Apps/Auction/Routes/__tests__/ConfirmBid.test.tsx) to test more "end-to-end"?
- Can we say something like:

> `track` consider dangerous - use `@track` to avoid unneeded component removal

and then refactor this call-sight to use the `@track` decorator to set the appropriate pattern here? Considering you have context here, this could be a great upkeep ticket to create in our backlog.

yuki24:
@zephraph @dleve123 I think we should be able to add an e2e test to Integrity as part of this ticket: https://artsyproduct.atlassian.net/browse/AUCT-863 (just for context @zephraph, we are planning to add more auction-related tests to Integrity).

Honestly, I'm not sure how much the linter could be useful. We could use that as a way for us to learn how to write a linter to make it more meaningful, but it seems like a contrived reason.

I'll go ahead and merge this PR since writing an e2e test can't be achieved within Reaction.

artsyit:
:rocket: PR was released in v25.5.10 :rocket:

PR #3151, opened 2020-02-13
Take 2: [Refactor] Implement route-level bundle splitting

damassi:
Re-opens https://github.com/artsy/reaction/pull/3137, which was reverted in force due to some weirdness. 

Going to go the review-app-route.


artsyit:
:rocket: PR was released in v25.9.0 :rocket:

PR #3155, opened 2020-02-13
Collapse auction results filter for reasons

zephraph:
Doing a thing for reasons. Ask me about it, it's a laugh.


zephraph:
@sepans, I've temporarily disabled the test for the auction filter. It'll fail until we've defaulted to expanding the filter (or we'll have to update it to expand).

artsyit:
:rocket: PR was released in v25.8.4 :rocket:

PR #3159, opened 2020-02-14
[Bugfix] Prefix injected split tag scripts with CDN_URL

damassi:
Fixes issue where injected bundle-split scripts aren't prefixed with CDN url when `CDN_URL` is available. 

This mirrors how our [assetMiddleware](https://github.com/artsy/force/blob/master/src/lib/middleware/assetMiddleware.ts#L23) function works over in force.

A future refactor would be to pass back concrete script src elements that can be individually piped into the asset helper.

### DIFF
--- a/src/Artsy/Router/buildServerApp.tsx
+++ b/src/Artsy/Router/buildServerApp.tsx
@@ -176,7 +176,10 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
                 bundleScriptTags
                   .split("\n")
                   .map(script => {
-                    // In production, prefix injected script src with CDN endpoint
+                    /**
+                     * In production, prefix injected script src with CDN endpoint.
+                     * @see https://github.com/artsy/force/blob/master/src/lib/middleware/assetMiddleware.ts#L23
+                     */
                     if (getENV("CDN_URL")) {
                       const scriptTagWithCDN = script.replace(
                         /src="\/assets/g,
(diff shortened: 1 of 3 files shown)


src/Artsy/Router/buildServerApp.tsx
```diff
@@ -175,24 +175,35 @@ export function buildServerApp(config: ServerRouterConfig): Promise<Resolve> {
...
-                        (
-                          script.includes("/assets/runtime-manifest.") ||
-                          script.includes("/assets/common.")
-                        )
+                  .map(script => {
+                    // In production, prefix injected script src with CDN endpoint
+                    if (getENV("CDN_URL")) {
+                      const scriptTagWithCDN = script.replace(
+                        /src="\/assets/g,
+                        `src="${getENV("CDN_URL")}/assets`
```
joeyAghion: Is there a way this can use exactly the same code as is used elsewhere to render the "main" asset hostname? Otherwise we risk generating the URLs slightly differently, or allowing the 2 locations to get out-of-sync over time

damassi: Unfortunately not, since we primarly handle our script injection in force, in a jade [asset helper middleware](https://github.com/artsy/force/blob/master/src/lib/middleware/assetMiddleware.ts#L23).

Coming from reaction, we pass through different kinds of javascript, not just script tags pointing at urls. This can be refactored so that a distinction can be made, however, but that would be follow-up. 

In any case, the logic is the same in both cases: `if CDN_URL, prepend`

artsyit:
:rocket: PR was released in v25.9.2 :rocket:

PR #3160, opened 2020-02-15
Bundle split search; fix SSR search input

damassi:
Adding bundle splitting to the SearchApp, and also fix SSR rendering on the search input in the nav bar, which flashes in because on client-side load initial render is behind a QueryRenderer.


artsyit:
:rocket: PR was released in v25.9.3 :rocket:

PR #3161, opened 2020-02-15
[Bugfix] Ensure all search-bar nav items navigate properly

damassi:
Fixes issue where search result item urls not within the global router context returned 404; e.g., https://staging.artsy.net/pushkin-house.

### DIFF
--- a/src/Artsy/Router/makeAppRoutes.tsx
+++ b/src/Artsy/Router/makeAppRoutes.tsx
@@ -18,31 +18,7 @@ interface RouteList {
 }
 
 export function makeAppRoutes(routeList: RouteList[]): RouteConfig[] {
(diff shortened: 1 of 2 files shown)


src/Artsy/Router/makeAppRoutes.tsx
```diff
@@ -18,31 +18,7 @@ interface RouteList {
 }
 
 export function makeAppRoutes(routeList: RouteList[]): RouteConfig[] {
-  function removeDisabledRoutes(acc, route: RouteList) {
```
damassi: Just moved these nested functions out of the outer function for readability

damassi:
Self merging so that things are in good state for Monday

artsyit:
:rocket: PR was released in v25.9.4 :rocket:

PR #3162, opened 2020-02-15
[Clientside Routing] Wire in Order app

damassi:
Companion PR: https://github.com/artsy/force/pull/5063 (see for testing instructions) 

This wires the order app into new App shell. Haven't checked the full flow yet but things are mostly the same as this is already an SSR router app; just mounted the route config into a parent.

#### Overview of changes:
1) Bundle split all of the Order subroutes 
1) Moved the minimal nav bar directly into the app, from the server-side portion. We get all the data we need from reaction to render the link back to artwork
1) Updated the transition hook (when user cmd+r to reload) to check for `beforeunload` event in `OrderApp`, versus at the router level.

#### Todo:
- [x] Swap out top nav bar for blank nav bar on order / make offer route
- [x] Ensure Buy Now button uses router to route
- [x] Ensure Make Offer button uses router to route
- [x] Ensure route hook transitions block transitioning away with alert popup
- [x] Address server-side redirect if not logged in
- [x] Test full Buy Now flow 
- [x] Test full Make Offer flow 

bnow


damassi:
@sweir27 and @mzikherman - this is now ready to go.

src/Components/NavBar/MinimalNavBar.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface MinimalNavBarProps {
+  to: string
+  children: React.ReactNode
+}
+
+export const MinimalNavBar: React.FC<MinimalNavBarProps> = props => {
+  return (
+    <Box
+      zIndex={1000}
```
damassi: Force's nav bar is like 980, which because of some other weird elements it has to be that high, so this is higher, because this just covers the page

src/Apps/Order/routes.tsx
```diff
@@ -80,7 +59,7 @@ export const routes: RouteConfig[] = [
     children: [
       {
         path: "respond",
-        Component: RespondRoute,
+        getComponent: () => loadable(() => import("./Routes/Respond")),
```
sweir27: I'm sure this is fine, but can you explain what this change is for?

sweir27:
This looks OK to me-- feel free to merge when you're ready to keep an eye on things! A successful run of integrity with this + the force side in should give us confidence in most of the user flows.

src/Apps/Order/routes.tsx
damassi: This is bundle splitting. When webpqck encounters a dynamic `import()` keyword it will generate a new .js file, that contains all the sources for that route. That way, if a user never encounters this path, they wont need to download the JS. We're now splitting at all of our routes. The `loadable` aspect is is a wrapper around [loadable-components](https://github.com/gregberge/loadable-components), it provides some additional features but most importantly helps us manage SSR.

artsyit:
:rocket: PR was released in `v25.10.0` :rocket:

PR #3169, opened 2020-02-18
Migrate "Promoted Content" banner from Force

eessex:
Migrates Backbone "Promoted Content" banner used on articles authored to promote sales and gallery partners. 

This is the last Backbone component used in the /article app in Force and unblocks moving `classic` articles to our react app. 

Screen Shot 2020-02-18 at 1 57 52 PM
Screen Shot 2020-02-18 at 1 56 52 PM


damassi:
@eessex - just to keep this on your radar as a larger thing to think about should it come up: it would be great to move the articles app to the new router, but there's still a lot of entangled react stuff spread between force and reaction and migrating that wont really be possible until all of our react code is moved to reaction and for just handles basic routing. It would be a nice long-term win if this could eventually be solved.

eessex:
> it would be great to move the articles app to the new router
Yes definitely on my radar - we are getting close to articles living entirely in reaction, but the infinite scroll tooling poses some challenges to moving routing. There were some a/b tests last year that explored removing the infinite scroll feature, worth checking with product if there are any options here to unblock.

damassi:
Aw nice! If things are getting close to being reaction-only I feel like we could make the infinite scrolling work, however which way... Keep me in the loop if you see any more significant movement here.

src/Components/Publishing/Header/Layouts/Components/__tests__/ClassicPromotedContent.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+    const component = getWrapper()
+    expect(component.text()).toMatch("Promoted Content")
+    expect(component.text()).toMatch("Contessa Gallery")
+    expect(component.find(Button).text()).toMatch("Explore Gallery")
+    expect(component.html()).toMatch("square.jpg")
+  })
+
+  it("Renders expected content for auctions", () => {
+    const component = getWrapper()
```
pepopowitz: I'm not requesting changes, but one technique that can be useful for reading tests is to set up the data that makes this test unique inside of this test. We do it in the previous test - `props.article.partner = ....`. 

In _this_ test, my assumption is that there is a sale object in the default fixture, but I have to go read that fixture file to make sure. If we also set `props.article.sale` at the top of this test, even if it's duplication, it gives me a good sense of what data is needed for this test to pass.

artsyit:
:rocket: PR was released in `v25.11.2` :rocket:

PR #3171, opened 2020-02-19
Add integrity classname

damassi:
We're using a new `<MinimalNavBar>` component and it needs an identifier for our acceptance tests.

### DIFF
--- a/src/Components/NavBar/MinimalNavBar.tsx
+++ b/src/Components/NavBar/MinimalNavBar.tsx
@@ -21,7 +21,7 @@ export const MinimalNavBar: React.FC<MinimalNavBarProps> = props => {
     >
       <AppContainer>
         <Box height={70} px={[2, 4]}>
-          <RouterLink to={props.to}>
+          <RouterLink to={props.to} className="acceptance__logoLink">
             <ArtsyLogoBlackIcon />
           </RouterLink>
         </Box>


artsyit:
:rocket: PR was released in `v25.11.1` :rocket:

src/Components/NavBar/MinimalNavBar.tsx
```diff
@@ -21,7 +21,7 @@ export const MinimalNavBar: React.FC<MinimalNavBarProps> = props => {
     >
       <AppContainer>
         <Box height={70} px={[2, 4]}>
-          <RouterLink to={props.to}>
+          <RouterLink to={props.to} className="acceptance__logoLink">
```
sweir27: @damassi can you please use the `data-test` attribute here as we do in [other places](https://github.com/artsy/reaction/pull/3089)?

It's a small change but best to be consistent!

damassi: Oh this is already in :)

sweir27: I know! I am suggesting a quick update.

The integrity tests are primarily failing for another reason (an inability to access cms-staging), so there's not a huge rush over there. 😄

damassi: (No meaning, i've already updated reaction with `data-test`)

sweir27: Oooh 👍

PR #3173, opened 2020-02-19
Show My Conversations

ashkan18:
# Problem
As a user I can't see my current conversations anywhere in our UI and my only means of communication is good old broken email system.

# Solution
Add two **currently Admin Only** pages.

## Add a new page to show all user's conversations
Following the pattern of #3096 , adding a new `Conversations` page to show user's current conversations.

`/user/conversations`

_**Image Deleted**_

## Add a new page for conversation details where we can reply to conversations

`/user/conversations/:conversationID` 

_**Image Deleted**_

# Questions
- I didn't add conversation details as children route under `user/conversations` and added it as a separate route but maybe make sense to have it as a children route?
- In sotrybook, when showing conversation detail, i need to pass in a conversation id that is related to current user, how can we do that? basically anyone running storybook right now on this branch wouldn't get conversation details because it's pointing to one of my conversations.

# Follow Ups
- Add pagination to both pages
- Responsiveness
- Tracking

# Keep in mind
This is for now **admin only** feature

# Special Thanks To
@mzikherman who did most of this in iOS 

# Selife
image


zephraph:
This is awesome!

One note, we'll want to flip the order of the conversation. The most recent message should be closest to the input box.

damassi:
Really nice PR @ashkan18 👍 

Would you mind adding your routes to the [global router](https://github.com/artsy/reaction/blob/master/src/Apps/getAppRoutes.tsx) file as well? Then, unless there's some server-side bits needed, you won't need to worry about the force side of things.

src/Apps/Conversation/routes.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+// @ts-ignore
+import { RouteConfig } from "found"
+import { ConversationAppFragmentContainer as ConversationApp } from "./ConversationApp"
+import { DetailFragmentContainer as DetailRoute } from "./Routes/Detail"
+
+export const routes: RouteConfig[] = [
+  {
+    path: "/user/conversations",
+    Component: ConversationApp,
```
damassi: Following patterns [here](https://github.com/artsy/reaction/blob/master/src/Apps/Artist/routes.tsx#L65) can be we sure to split this bundle? Note that an additional [`export default`](https://github.com/artsy/reaction/blob/master/src/Apps/Artist/ArtistApp.tsx#L147-L148) will need to be used inside of the component that's being loaded.

ashkan18:
@damassi does this mean adding it to global route I don't have to do anything on force other than updating to reaction including this version?!

damassi:
Yup, but note that this is only for staging for right now. On monday we'll be running an A/B test for a few weeks and then after releasing to prod. If we need this in prod sooner then can add a route to force. 

But yeah, in general, unless we need specific server-side functionality things only need to be added here. If needing server-side stuff, then we create a new middleware similar to what's seen [here](https://github.com/artsy/force/blob/master/src/desktop/apps/experimental-app-shell/auction/bidderRegistrationMiddleware.ts)

ashkan18:
@damassi cool! that makes sense! Also while I have you, do you know why `type-check` and `jest` are failing with those components on Circle? works totally fine locally

damassi:
@ashkan18 - that totally looks like a linux / mac case sensitivity issue -- are you sure that the file path referenced in the import is identical to what is actually on the file system?

ashkan18:
@damassi yes! it was git not updating my rename locally on mac! had to do a specific `git mv`

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { createFragmentContainer, RelayProp } from "react-relay"
+import { graphql } from "relay-runtime"
+import { MessageFragmentContainer as Message } from "./Message"
+import { Reply } from "./Reply"
+
+interface ItemProps {
+  item: Conversation_conversation["items"][0]["item"]
+}
+
+const Item = (props: ItemProps) => {
```
damassi: `React.FC< ItemProps >` (here and below)

src/Apps/getAppRoutes.tsx
```diff
@@ -27,5 +28,8 @@ export function getAppRoutes(): RouteConfig[] {
     {
       routes: searchRoutes,
     },
+    {
+      routes: conversationRoutes,
+    },
```
damassi: #superminor this object is alpha

ashkan18: what do you mean?

damassi: Oh sorry :) meaning everything is intentionally alphabetized so that things are easier to find later.

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -0,0 +1,142 @@
...
+          <Serif size="2" weight="semibold">
+            {item.artistNames}
+          </Serif>
+          <Serif italic size="2" color="black60" lineHeight={1.3}>
+            {item.title}, {item.date}
+          </Serif>
+        </Flex>
+      </Flex>
+    )
+  } else if (item.__typename === "Show") {
```
mzikherman: Forgot about this, I guess these are inquiries on shows (do we still allow those?)

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -0,0 +1,142 @@
...
+  relay: RelayProp
+}
+
+const Conversation: React.FC<ConversationProps> = props => {
+  const { conversation, relay } = props
+  const messageCount = conversation.messages.edges.length
+  return (
+    <Box>
+      {conversation.items.map((i, idx) => (
+        <Item item={i.item} key={idx} />
```
mzikherman: The `key` here (and below in the iteration), are possibly better as `id` (which should be that unique Relay id), vs. the array index or cursor value (which aren't guaranteed to be unique).

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -0,0 +1,142 @@
...
+      ))}
+      {conversation.messages.edges.map((m, idx) => (
+        <Message
+          message={m.node}
+          initialMessage={conversation.initialMessage}
+          key={m.cursor}
+          isFirst={idx === messageCount - 1}
+        />
+      ))}
+      <Reply conversation={conversation} environment={relay.environment} />
```
mzikherman: #minor rather than pass around these kinds of props (current user, Relay environment, mediator, other shared global 'grab bag' stuff), you can do `const { relayEnvironment } = useSystemContext()` calls in the specific component that needs it. That can be cleaner vs. a situation of passing lots of props down explicitly all the time.

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }
+        to {
+          name
+          initials
+        }
+        initialMessage
+        lastMessageID
+        unread
+        messages(first: $count, after: $after, sort: DESC)
+          @connection(key: "Messages_messages", filters: []) {
```
mzikherman: fancy!

src/Apps/Conversation/Components/Conversations.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <ConversationSnippet conversation={edge.node} key={edge.cursor} />
+        ))
+      ) : (
+        <Sans size="2">No Conversations</Sans>
+      )}
+    </Box>
+  )
+}
+
+export const ConversationsFragmentContainer = createRefetchContainer(
```
mzikherman: Interesting, this is a refetch container, but I don't see where it can get refreshed?

src/Apps/Conversation/routes.tsx
```diff
@@ -0,0 +1,46 @@
...
+        first: 10,
+      }
+    },
+    cacheConfig: {
+      force: true,
+    },
+  },
+  {
+    path: "/user/conversations/:conversationID",
+    getComponent: () => loadable(() => import("./Routes/Detail")),
```
mzikherman: #minor rather than `Detail` (which sort of adds another vocabulary word), what about a `Conversation` view/route?

src/Apps/__stories__/ConversationApp.story.tsx
```diff
@@ -0,0 +1,22 @@
...
+storiesOf("Apps/Conversation", module)
+  .add("Default", () => {
+    return (
+      <MockRouter
+        routes={conversationRoutes}
+        initialRoute="/user/conversations"
+      />
+    )
+  })
+  .add("Detail", () => {
```
mzikherman: This sort of feels like a hard storybook entry to maintain, especially since we don't have consistent fixtures that live forever on staging, etc.

It's probably sufficient to just provide the storybooks entry for the Inbox as a whole (as you've done above).

src/Apps/Conversation/Components/Conversation.tsx
ashkan18: yes, im pretty sure somewhere deep in one of our pages you can still make show inquiries

src/Apps/Conversation/Components/Conversation.tsx
ashkan18: well... its basically you! I'm just a copy/paster here.

src/Apps/Conversation/routes.tsx
ashkan18: we already have a `conversation` component too i wonder if that makes it confusing, for now leaving this as is if its ok.

src/Apps/Conversation/Components/Conversations.tsx
ashkan18: i used it mainly for pagination but then ignored it for now.

src/Apps/Conversation/Components/Conversation.tsx
ashkan18: this lead to some weird issues with `getLinkedRecord`, ended up keeping passing `environment` around for now.

ashkan18:
addressed most comments and commented on the ones i didn't do, going to #mergeongreen

artsyit:
:rocket: PR was released in `v25.14.0` :rocket:

PR #3174, opened 2020-02-19
Use data-link instead of classname for integration test

damassi:
#trivial


artsyit:
:rocket: PR was released in `v25.11.3` :rocket:

PR #3175, opened 2020-02-19
[PURCHASE-1793] Fix artwork filter events being tracked twice.

sepans:
Fixes [PURCHASE-1793](https://artsyproduct.atlassian.net/browse/PURCHASE-1793)

Noticed the event is already being tracked [here](https://github.com/sepans/reaction/blob/faf119874fbe9d518cddbd7d4704f06e9c6e14ba/src/Components/v2/ArtworkFilter/index.tsx#L127) inside the component so there is no need to track it using the `onFilterClick` prop. Today during grooming realized this has been also reported by #data-team.


sepans:
Yes @damassi I was also debating where should tracking be. At least for the sake of this PR it would have been easier to remove it from one place than 4 different places. In the end I decided to do it this way for 👆 reason you mentioned + any other page that uses the artwork filters automatically get tracking, _I think_.

artsyit:
:rocket: PR was released in `v25.11.5` :rocket:

PR #3176, opened 2020-02-19
Remove dd-trace

damassi:
This removes our DataDog tracing around `buildServerApp` and removes the utility since it was only used in this file.


joeyAghion:
If `dd-trace` isn't used elsewhere within Reaction, can it be removed from the dependencies/`package.json` as well?

damassi:
Ah yeah, good call.

artsyit:
:rocket: PR was released in `v25.11.6` :rocket:

PR #3177, opened 2020-02-19
Pass data-test down through RouterLink #trivial

damassi:
Since our RouterLink component stips unauthorized props we needed to allowlist `data-test`.


artsyit:
:rocket: PR was released in `v25.11.4` :rocket:

PR #3179, opened 2020-02-20
Experiment with making CI times much faster

zephraph:
Given that jest tests are the slowest part of reaction's CI, I'm experimenting with different ways we can make it faster (before doing the deep, hard work of really doing heavy performance/profiling work on the extremely slow tests). 

My thoughts are two fold:

1. Non-master commit builds should only run the tests over files which diverge from master. This optimization is built into jest but only possible when there are no differences in global things like dependencies or setup. 
2. If there aren't real code changes, don't run the tests at all.

This is really just a thought experiment and not necessarily meant to land in prod. 

<details>
  <summary>📦 Published PR as canary version: <code>25.28.0-canary.3179.53605.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.28.0-canary.3179.53605.0
  # or 
  yarn add @artsy/reaction@25.28.0-canary.3179.53605.0
  ```
</details>


damassi:
@zephraph - one related thought I had the other day is i'd be curious to turn off `--runInBand` and see what happens... maybe previous issues fixed themselves?

zephraph:
@damassi yep, definitely something we can try. I feel like @eessex or @ds300 might have some firmer insights on why it wasn't working out. I want to say there were some relay mocking issues and we were running into circle constraints.

zephraph:
That last bump to force rebuild was successful 🎉

zephraph:
But the next one wasn't 😅

zephraph:
Okay, that last commit worked perfectly 👍

ds300:
`--runInBand` was enabled because process forking on circle is broken in a way that causes tests to time out nondeterministically.

zephraph:
I'm gonna close this PR since there's some unnecessary noise. Will re-apply these changes in another PR

PR #3181, opened 2020-02-20
[Clientside Nav] Fix double track on artwork page

damassi:
Fixes a browser-level alert guard when navigating away from completed order -> status page. Don't need it there.

Also fixes double tracking on artwork page.

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -50,24 +50,6 @@ export class ArtworkApp extends React.Component<Props> {
     this.trackLotView()
   }
 
-  componentDidUpdate(prevProps) {
-    /**
-     * If we've changed routes within the app, trigger pageView and productView.
-     *
-     * FIXME: We're manually invoking pageView tracking here, instead of within
-     * the `trackingMiddleware` file as we need to pass along additional metadata.
-     * Waiting on analytics team to decide if there's a better way to capture this
-     * data that remains consistent with the rest of the app.
-     */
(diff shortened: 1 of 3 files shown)


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -50,24 +50,6 @@ export class ArtworkApp extends React.Component<Props> {
     this.trackLotView()
   }
 
-  componentDidUpdate(prevProps) {
```
damassi: We don't need this as on route change a full page re-rerender occurs -- including componentDidMount.

mzikherman: This feels like it conflicts with the behavior we observed in the first AB test. Namely, when navigating to an artwork from another artwork (or within any of the same app), the app wouldn't get re-mounted. Instead, it's a params/props update.

Or, do you mean that with our new tracking middleware (responding to router events), we can simply drop this React lifecycle hook entirely? Cause that does make sense to me, but I feel like your explanation is a bit different than my understanding.

mzikherman: (I would also have a follow-up, that if the component doesn't re-mount but our tracking middleware will take care of the page view, maybe this should be reinstated, but _without_ the pageview (and with the custom product and lot view events)).

damassi: Adding a log to `componentDidMount` and tracking and clicking between artworks triggers a proper event from componentDidMount each time -- it rerenders the whole page. Nothing strange?

artsyit:
:rocket: PR was released in `v25.13.1` :rocket:

PR #3185, opened 2020-02-21
[QueryRenderers] Delay showing spinner

damassi:
Companion PR: https://github.com/artsy/palette/pull/640

This fixes the issue thats frequently seen on our artist, artwork and collect and search pages where we conditionally render content in query renderers on the client. For a large number of pages we never render these sections, yet we always see a preloader that flashes while we make a request, and shifts our layouts all around in an ugly unstable way. 

This adds a default delay of 800ms until we show the preloader, which covers cases when a request takes an unreasonable amount of time. 

In most cases though, the area will return empty, and no loader will show, keeping the layout stable. 

#### Before: 

before

#### After: 

after

### DIFF
--- a/src/Artsy/Relay/renderWithLoadProgress.tsx
+++ b/src/Artsy/Relay/renderWithLoadProgress.tsx
@@ -19,13 +19,6 @@ const SpinnerContainer = styled.figure`
   flex-direction: column;
   justify-content: center;
   position: relative;
-
-  /*
-    Noticed some weird behavior in force due to figure margin being set to 0
-    in global css. Adding padding of 22 gets things back to visual default.
-  */
-  margin: unset;
-  padding: 22px;
 `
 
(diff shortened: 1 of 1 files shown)


src/Artsy/Relay/renderWithLoadProgress.tsx
```diff
@@ -20,12 +20,7 @@ const SpinnerContainer = styled.figure`
   justify-content: center;
   position: relative;
 
-  /*
-    Noticed some weird behavior in force due to figure margin being set to 0
-    in global css. Adding padding of 22 gets things back to visual default.
-  */
   margin: unset;
-  padding: 22px;
```
damassi: This was also causing unnecessary shifts all over the place

artsyit:
:rocket: PR was released in `v25.17.1` :rocket:

PR #3186, opened 2020-02-21
Collector Verification App

erikdstock:
This PR sets up the identity verification flow landing page.
Related force PR: artsy/force#5090

image


src/Apps/CollectorVerifcation/routes.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+        if (
+          !identityVerification ||
+          identityVerification.userID !== me.internalID
+        ) {
+          return <ErrorPage code={404} />
+        }
+
+        return <Component {...props} />
+      }
```
damassi: Using custom render methods like this breaks the routers internal functionality slightly. 

All of these guards can move into `CollectorVerificationApp` (exit early if 404), and this `render` function can be removed.

src/Apps/CollectorVerifcation/routes.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { graphql } from "react-relay"
+import createLogger from "Utils/logger"
+import { CollectorVerificationApp } from "./CollectorVerificationApp"
+
+const logger = createLogger("Apps/CollectorVerification/routes")
+
+export const routes: RouteConfig[] = [
+  {
+    path: "/collector-verification/:id",
+    Component: CollectorVerificationApp,
```
damassi: Be sure to code-split at all routes: https://github.com/artsy/reaction/blob/master/src/Apps/Conversation/routes.tsx#L10

src/Apps/CollectorVerifcation/routes.tsx
erikdstock: ah, i think @yuki24 taught our team about this last week. Thanks for the heads up.

damassi:
@erikdstock - One more thing: for this to work with client-side routing it needs to be added to our global routes file: https://github.com/artsy/reaction/blob/master/src/Apps/getAppRoutes.tsx#L12

dleve123:
Let's discuss more, but did you confirm naming Identity Verification, instead of Collector Verification? The naming scheme on the backend in "identity" based and while we're internally referring to this process as "collector vetting" now instead of "buyer vetting", I think that's likely subject to future change.

erikdstock:
> Let's discuss more, but did you consider naming Identity Verification, instead of Collector Verification? The naming scheme on the backend in "identity" based and while we're internally referring to this process as "collector vetting" now instead of "buyer vetting"

Not a hill I'm going to die on, but my rationale for this was twofold:
- 'identity verification' is a kind of internal and unfriendly term for folks who pay attention to URIs
- this isn't a CRUD/restful `/model/edit` route, but the start of a flow, so we don't really care what the underlying model name is. It'e more like a mutation- they are starting a flow.

src/Apps/IdentityVerification/IdentityVerificationPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const clickContinueToVerification = () => {
+    trackEvent({
+      context_page_owner_id: identityVerification.internalID,
+      action_type: Schema.ActionType.ClickedContinueToIdVerification,
+      context_page: Schema.PageName.IdentityVerificationPage,
+    })
+  }
+
+  return (
+    <SystemContextConsumer>
```
damassi: Why is this needed?

damassi:
Reminder to mount this into the [global routes](https://github.com/artsy/reaction/blob/master/src/Apps/getAppRoutes.tsx).

src/Apps/IdentityVerification/IdentityVerificationPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+import React from "react"
+import { Meta, Title as HeadTitle } from "react-head"
+import { createFragmentContainer } from "react-relay"
+import { graphql } from "relay-runtime"
+
+interface Props {
+  me: IdentityVerificationPage_me
+}
+
+export const IdentityVerification: React.FC<Props> = ({ me }) => {
```
damassi: #minor - In our other apps the the main entry point file and export is named `SomeNameApp` (like `ArtistApp`, `ArtworkApp`, etc)

src/Apps/IdentityVerification/IdentityVerificationPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+              </Box>
+            </Box>
+          </AppContainer>
+        )
+      }}
+    </SystemContextConsumer>
+  )
+}
+
+const IdentityVerificationPage = createFragmentContainer(IdentityVerification, {
```
damassi: This should include what this is: `export const IdentityVerificationAppFragmentContainer = `

src/Apps/IdentityVerification/IdentityVerificationPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+      name
+      identityVerification(id: $id) {
+        internalID
+        state
+        userID
+      }
+    }
+  `,
+})
+export default IdentityVerificationPage
```
damassi: And this default export is exclusively used for [bundle splitting in the route](https://github.com/artsy/reaction/blob/master/src/Apps/Artist/ArtistApp.tsx#L147-L148).

src/Apps/IdentityVerification/IdentityVerificationPage.tsx
erikdstock: I figured it was something to include by convention.

src/Apps/IdentityVerification/__fixtures__/routes_IdentitifcationPageQuery.ts
```diff
@@ -0,0 +1,15 @@
+import { routes_IdentityVerificationPageQueryRawResponse } from "__generated__/routes_IdentityVerificationPageQuery.graphql"
+
+export const IdentityVerificationPageQueryResponseFixture: routes_IdentityVerificationPageQueryRawResponse = {
+  me: {
+    internalID: "my-user-id",
+    name: "user",
+    id: "unused-user",
+    identityVerification: {
+      internalID: "identity-verification-id",
+      id: "unused-idv",
```
dleve123: similar comment as the above

src/Apps/IdentityVerification/__tests__/IdentityVerificationPage.test.tsx
```diff
@@ -0,0 +1,76 @@
+import { IdentityVerificationPageTestQueryRawResponse } from "__generated__/IdentityVerificationPageTestQuery.graphql"
+import { routes_IdentityVerificationPageQueryResponse } from "__generated__/routes_IdentityVerificationPageQuery.graphql"
+import deepMerge from "deepmerge"
+import { createTestEnv } from "DevTools/createTestEnv"
+import { Location } from "found"
+import React from "react"
+import { graphql } from "react-relay"
+import { IdentityVerificationPageQueryResponseFixture } from "../__fixtures__/routes_IdentitifcationPageQuery"
```
dleve123: typo in routes_*Identitifcation*PageQuery

src/Apps/IdentityVerification/__tests__/IdentityVerificationPage.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <IdentityVerificationPage {...props} />
+    ),
+    query: graphql`
+      query IdentityVerificationPageTestQuery @raw_response_type {
+        me {
+          ...IdentityVerificationPage_me @arguments(id: "idv-id")
+        }
+      }
+    `,
+    defaultData: IdentityVerificationPageQueryResponseFixture as IdentityVerificationPageTestQueryRawResponse,
```
dleve123: Why do we need to cast this as another type? It's defined to be of type `routes_IdentityVerificationPageQueryRawResponse` – which seems "closer" to the production code than a type geared towards the test page. Perhaps I'm just not understanding something here!

src/Apps/IdentityVerification/__tests__/IdentityVerificationPage.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  })
+}
+
+describe("IdentityVerification route", () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+  describe("for signed-in user", () => {
+    it("allows an identity instance owner to begin identity verification", async () => {
```
dleve123: #minors:
  - thoughts on changing the docstring to only specify that a user is able to _view_ a page about the identity verification for now, and not begin (at least not yet).
  - thoughts on separating the setup / act / assert phases of this test with newlines. I, personally, think that would make the test code more scan-able.

src/Apps/IdentityVerification/__tests__/IdentityVerificationPage.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    `,
+    defaultData: IdentityVerificationPageQueryResponseFixture as IdentityVerificationPageTestQueryRawResponse,
+  })
+}
+
+describe("IdentityVerification route", () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+  describe("for signed-in user", () => {
```
dleve123: thoughts on adding a test case for when there is no signed in user?

src/Apps/IdentityVerification/__tests__/routes.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+} from "found/lib/server"
+import { Environment, RecordSource, Store } from "relay-runtime"
+
+import { routes } from "Apps/IdentityVerification/routes"
+
+import { routes_IdentityVerificationPageQueryRawResponse } from "__generated__/routes_IdentityVerificationPageQuery.graphql"
+import { createRender } from "found"
+
+describe("IdentityVerification/routes", () => {
+  const idvID = (IdentityVerificationPageQueryResponseFixture.me as any)
```
dleve123: are we sure the cast to `any` is needed here? Based on my read of the code, the types check out.

src/Apps/IdentityVerification/__tests__/routes.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { IdentityVerificationPageQueryResponseFixture } from "Apps/IdentityVerification/__fixtures__/routes_IdentitifcationPageQuery"
```
dleve123: same typo here as noted above

src/Apps/IdentityVerification/IdentityVerificationPage.tsx
damassi: All apps mounted into the router have [this tree](https://github.com/artsy/reaction/blob/master/src/Artsy/Router/Boot.tsx) at the root, which means that if you need props from the system context can use the  [useSystemContext](https://github.com/artsy/reaction/blob/master/src/Artsy/SystemContext.tsx#L118-L124) hook.

erikdstock: makes sense, will remove.

src/Apps/IdentityVerification/__fixtures__/routes_IdentitifcationPageQuery.ts
```diff
@@ -0,0 +1,15 @@
+import { routes_IdentityVerificationPageQueryRawResponse } from "__generated__/routes_IdentityVerificationPageQuery.graphql"
+
+export const IdentityVerificationPageQueryResponseFixture: routes_IdentityVerificationPageQueryRawResponse = {
+  me: {
+    internalID: "my-user-id",
+    name: "user",
+    id: "unused-user",
```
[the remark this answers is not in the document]
erikdstock: I'm not querying for it, it just appeared in the type so i added it to the fixture to appease typescript.

artsyit:
:rocket: PR was released in `v25.23.0` :rocket:

src/Apps/IdentityVerification/routes.tsx
```diff
@@ -0,0 +1,18 @@
+import loadable from "@loadable/component"
+import { RouteConfig } from "found"
+import { graphql } from "react-relay"
+
+export const routes: RouteConfig[] = [
+  {
+    path: "/identity-verification/:id",
+    getComponent: () => loadable(() => import("./IdentityVerificationApp")),
```
damassi: Sorry for all the pattern changes! One small change that was deployed on Friday is setting up a mechanism for preloading the split bundle. See [here](https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/routes.tsx#L9-L12)

dleve123: cc @erikdstock on this. @damassi Can you provide some more context here? What is the desired change and why the shift in pattern?

damassi: See https://github.com/artsy/reaction/pull/3207

PR #3187, opened 2020-02-21
Fix conversation detail link

ashkan18:
Follow up on #3173 to fix the link that was missing `/user`

going to self-merge it


ashkan18:
#mergeongreen

artsyit:
:rocket: PR was released in `v25.17.0` :rocket:

PR #3188, opened 2020-02-21
[BundleSplitting] Comment out for now

damassi:
Noticed an issue where async chunk requested from the parent of an async chunk doesn't fetch from our CDN url, only top-level chunks. 

Breadcrumbs of previous work: 

- Ensure top-level chunks are served from `CDN_URL`: https://github.com/artsy/reaction/pull/3159 

The current flow in production is: 
1) Webpack will compile out all of our chunks and create a manifest
1) We check if we're in a prod environment, and prefix a cloudfront CDN url via our [asset helper](https://github.com/damassi/force/blob/c2ce1db7b7007721d768497439ae81e64380395b/src/lib/middleware/assetMiddleware.ts#L20-L27)
1) In reaction, when loadable components dynamically injects scripts into the page *_on the first render_* [we do the same](https://github.com/artsy/reaction/blob/master/src/Artsy/Router/buildServerApp.tsx#L171-L176)
1) However, this doesn't take into account _dynamic_ requests from top-level routes, such as clicking the [works-for-you](https://staging.artsy.net/artist/andy-warhol/works-for-sale) tab on the Artist page, at which point webpack resolves the split bundle via its own resolution mechanism, according to what is set in the `publicPath` entry of our base webpack config, which handles domain: https://github.com/artsy/force/blob/master/webpack/envs/baseConfig.js#L22

#### What we need to do to fix issue: 
We need to resolve the inconsistency between our asset helper's resolution mechanism and webpacks `publicPath`.

### DIFF
--- a/src/Apps/Artist/routes.tsx
+++ b/src/Apps/Artist/routes.tsx
@@ -1,4 +1,3 @@
-import loadable from "@loadable/component"
 import { hasSections as showMarketInsights } from "Apps/Artist/Components/MarketInsights/MarketInsights"
 import {
   ArtworkFilters,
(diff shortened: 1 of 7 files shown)


damassi:
Reverting.

artsyit:
:rocket: PR was released in `v25.17.0` :rocket:

PR #3190, opened 2020-02-21
Update field used for notifications badge count

sweir27:
This PR updates the `NotificationsBadge` (just the "count") to reference the unread notifications count that we keep track of in gravity. This will correctly go to 0 when notifications are dismissed.

Note that this doesn't make any changes to the feed or the dropdown.

Old:
image

New:
image

Actual data:
image

Gif showing dismissal:
notifications

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -4701,6 +4701,26 @@ type HomePageRelatedArtistArtworkModule {
   basedOn: Artist
 }
 
+type IdentityVerification {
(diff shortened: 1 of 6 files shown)


sweir27:
@mzikherman I'm looking into why CI is exploding (looks to not just be my PR), but other than that this should be reviewable!

artsyit:
:rocket: PR was released in `v25.18.3` :rocket:

PR #3191, opened 2020-02-22
Simplify A/B test lookup via `express-http-context`

damassi:
See https://github.com/artsy/force/pull/5099 for more info. 

This updates our `getENV` isomorphic ENV lookup util with another server-side layer:
- First, check to see if `SOME_ENV` var exists within the request / response cycle, stored within express-http-context. If so, return that value. Useful for A/B testing.
- If undefined, fall back to `process.env.SOME_ENV`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -19,6 +19,8 @@
       }
     ],
     "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-proposal-optional-chaining",
+    "@babel/plugin-proposal-nullish-coalescing-operator",
     [
       "relay",
       {
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -222,7 +222,14 @@ export class ArtworkApp extends React.Component<Props> {
             </Col>
           </Row>
 
-          <div id="lightbox-container" />
+          <div
+            id="lightbox-container"
+            style={{
+              position: "absolute",
+              top: 0,
+              zIndex: 1100, // over top nav
+            }}
+          />
           <SystemContextConsumer>
             {({ mediator }) => <>{this.enableIntercomForBuyers(mediator)}</>}
           </SystemContextConsumer>
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -245,7 +245,6 @@ export class ArtworkApp extends React.Component<Props> {
 export const ArtworkAppFragmentContainer = createFragmentContainer(
   (props: Props) => {
     const {
-      match,
       match: {
         location: { pathname, state },
       },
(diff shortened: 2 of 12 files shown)


src/Utils/getENV.tsx
```diff
@@ -2,6 +2,14 @@ import { data as sd } from "sharify"
 
 export function getENV(ENV_VAR) {
   const isServer = typeof window === "undefined"
-  const envVar = isServer ? process.env[ENV_VAR] : sd[ENV_VAR]
+
+  let envVar
+  if (isServer) {
+    const httpContext = require("express-http-context")
+    envVar = httpContext.get(ENV_VAR) ?? process.env[ENV_VAR]
```
damassi: Uses new [nullish coalesing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) plugin, which always returns the left hand side of the operation unless `null` or `undefined`.

.babelrc
```diff
@@ -19,6 +19,8 @@
       }
     ],
     "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-proposal-optional-chaining",
```
damassi: Suprised this wasn't already in! We can no do 
```
const deepPropOrUndefined = foo?.bar?.baz?.bam
```
which should eliminate our need for the `get` util.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -222,7 +222,14 @@ export class ArtworkApp extends React.Component<Props> {
...
           </Row>
 
-          <div id="lightbox-container" />
+          <div
+            id="lightbox-container"
+            style={{
+              position: "absolute",
+              top: 0,
+              zIndex: 1100, // over top nav
+            }}
```
damassi: Fixes long standing issue where when a user views an artwork in full screen, the top nav bar flops out due to z-index and position.

.babelrc
mzikherman: oo sick!

src/Apps/Artwork/Components/OtherWorks/Header.tsx
```diff
@@ -16,14 +17,11 @@ export const Header: React.SFC<HeaderProps> = props => {
...
       {buttonHref && (
-        <Button
-          variant="secondaryOutline"
-          mb={3}
-          // FIXME: Move to <Link> component once https://github.com/artsy/palette/pull/130 is merged
-          onClick={() => (window.location.href = buttonHref)}
-        >
-          View all
-        </Button>
+        <RouterLink to={buttonHref}>
```
mzikherman: Ah cool, yea these custom `window.location.href` navigations in JS dont work with our router.

src/Utils/getENV.tsx
mzikherman: Interesting. I didn't fully get immediately how different is than `||`. I do now recognize the subtle difference, but I think generally any sort of conditional assignment like this, if you are expecting a false-y/null/undefined to be possible (or even valid, esp. as a default), I almost always open a console to try it out (and/or mentally unpack it in my brain for a second).

Cool! I look forward to trying this.

mzikherman:
This looks good! Since this depends on Node v12, which still hasn't been fully deployed to production, so we hold off on merging these PR's until we're happy with v12 in prod?

artsyit:
:rocket: PR was released in `v25.18.1` :rocket:

PR #3193, opened 2020-02-24
Analytics schema v2 - authImpression event

eessex:
An MVP demonstrating helper functions that create tracking args for segment events. This PR creates the `authImpression` event, which strictly types arguments that are allowed for this particular tracking action. 

The v2 analytics directory is separated into two folders:
- `/Schema`, which holds schemas for individual events and property types on events
- `/Events`, which holds helper functions like `authImpression` to generate event data that conforms to the schema.

This work is intended to be published as an NPM package, but will hold on moving this into a dependency until we have proven this event works as expected.

TODO:
- [x] Force auth triggers conform to schema
- [x] Determine downstream migration steps for Redshift value changes

Related context: https://www.notion.so/Analytics-Schema-Improvements-4e091c7186c8417b9f9fa05690570834


src/Artsy/Analytics/v2/Events/AuthImpression.ts
```diff
@@ -0,0 +1,59 @@
...
+
+interface AuthImpressionArgs {
+  copy?: string
+  contextModule: AuthContextModule
+  intent: AuthIntent
+  onboarding?: boolean
+  trigger?: AuthTrigger
+  triggerSeconds?: number
+  type: AuthModalType
+}
```
cavvia: Is there any way of avoiding some of the repetition here? The `AuthImpression` interface and the args declaration are very similar.

src/Artsy/Analytics/v2/Events/AuthImpression.ts
```diff
@@ -0,0 +1,59 @@
...
+  modal_copy?: string
+  onboarding: boolean
+  trigger: AuthTrigger
+  trigger_seconds?: number
+  type: AuthModalType
+}
+
+interface AuthImpressionArgs {
+  copy?: string
+  contextModule: AuthContextModule
```
cavvia: Good to see strict typing on standard properties like these :+1:

src/Artsy/Analytics/v2/Schema/ContextModule.ts
```diff
@@ -0,0 +1,15 @@
...
+  categoryRail = "categoryRail",
+  featuredArtistsRail = "featuredArtistsRail",
+  header = "header",
+  otherWorksByArtistRail = "otherWorksByArtistRail",
+  otherWorksByGalleryRail = "otherWorksByGalleryRail",
+  relatedArtistsRail = "relatedArtistsRail",
+  relatedWorksRail = "relatedWorksRail",
+  worksByPopularArtistsRail = "worksByPopularArtistsRail",
+  worksForSaleRail = "worksForSaleRail",
+}
```
cavvia: :+1:

src/Artsy/Analytics/v2/Events/AuthImpression.ts
eessex: We gain two things from using a different interface for args - some options can be optional and have default values, and the naming convention. In JS apps our convention uses `contextModule` for args, and this format is ideal. Segment expects `context_module`, so the keys are different on the `AuthImpression` event interface.

cavvia:
Thanks for this! I like that there is stricter typing here than is currently the case. Definitely a step in the right direction.

My general sense is that there is a fair amount of boilerplate here for the declaration of a single event. I wonder if we can streamline this somehow. I would like analysts to be able to declare events without having to understand typescript, for example. In this context I've mentioned JSON Schema in the past. 

Another concern is that the schema may potentially be spread out over many files under this scheme. e.g. adding a new context module to an `AuthImpression` event requires an update in two places here. Is there any way to streamline this by centralizing the schema a bit more? Having more of an overview of the schema is pretty useful to analysts.

src/Apps/Artist/Routes/Overview/Components/ZeroState.tsx
```diff
@@ -1,8 +1,9 @@
 import { Link, Message } from "@artsy/palette"
-import { AnalyticsSchema, useSystemContext } from "Artsy"
```
damassi, line 2: @eessex - we should avoid `import * as SchemaV2` and stick to the named export format as seen here

src/Apps/Artist/Routes/Overview/Components/WorksForSaleRail.tsx
```diff
@@ -39,6 +40,7 @@ const WorksForSaleRail: React.FC<WorksForSaleRailProps & {
         return (
           <FillwidthItem
             artwork={artwork.node}
+            contextModule={ContextModule.worksForSaleRail}
```
damassi, line 43:
```js
import { AnalyticsSchemaV2 } from "Artsy"
...
contextModule={AnalyticsSchemaV2.ContextModule.worksForSaleRail}
```
that way we have a consistent way to find and replace in the future aligned with the rest of our analytics imports.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
```diff
@@ -31,6 +31,7 @@ import {
   ShareIcon,
   Spacer,
 } from "@artsy/palette"
+import { ContextModule } from "Artsy/Analytics/v2/Schema"
```
damassi, line 34: Same here

src/Apps/Artwork/Components/OtherWorks/RelatedWorksArtworkGrid.tsx
```diff
@@ -13,6 +13,7 @@ import React, { useContext } from "react"
 import styled from "styled-components"
 import createLogger from "Utils/logger"
 
+import { ContextModule } from "Artsy/Analytics/v2/Schema"
```
damassi, line 16: (And all further imports)

src/Artsy/Analytics/v2/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as Schema from "./Schema"
+export const AnalyticsSchema = Schema
```
damassi, line 2: export const AnalyticsSchemaV2 = Schema

eessex:
I will be moving this to an NPM package next week once this is proved e2e - @damassi your concerns of imports are noted, but I think it will make more sense to move them when we update all the imports are rerouted to the package.

src/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -77,17 +78,22 @@ export class FollowArtistButton extends React.Component<Props, State> {
...
+    const trackingData: FollowTrackingData = this.props.trackingData || {}
 
     if (user && user.id) {
       this.followArtistForUser(user)
     } else if (onOpenAuthModal) {
-      const config = {
-        contextModule: "intext tooltip",
-        intent: "follow artist",
+      onOpenAuthModal(ModalType.signup, {
+        contextModule: SchemaV2.ContextModule.intextTooltip,
```
damassi, line 88: 👍

damassi:
Sounds good 👍

artsyit:
:rocket: PR was released in `v26.0.0` :rocket:

PR #3198, opened 2020-02-25
Fix issue where expanding auction result changed page

zephraph:
When on any auction results page other than the initial page, expanding an auction card would bring you back to the first page and expand whatever auction card was at that position on the first page. 

The index of the expanded card was previously stored as a filter. When we updated the filter logic to not reset the page we were puzzled when it continued to reset anyway. After some exploration, we discovered a different bug where pagination disregards the page provided by the filters: it just always resets to 1 when a filter changes. This isn't an issue normally because the default expected behavior is to go back to the first page when a filter is changed. In this case though, we wanted to keep the current page.

Instead of fixing that larger underlying bug, which very likely would require metaphysics changes, we decided to simplify the card expanding logic by pulling it out of filters. There is a slight behavior change in which multiple cards can be expanded at one time. If needed we can follow up and fix that behavior.

Given that expansion is no longer driven by filters we added a new tracking method called `Auction result item clicked`. cc @louislecluse, you might want to test this after we merge. 

<details>
  <summary>📦 Published PR as canary version: <code>25.18.2-canary.3198.52571.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.18.2-canary.3198.52571.0
  # or 
  yarn add @artsy/reaction@25.18.2-canary.3198.52571.0
  ```
</details>

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
@@ -21,7 +21,7 @@ import { useTracking } from "react-tracking"
 import { get } from "Utils/get"
 
 export interface Props extends SystemContextProps {
-  expanded: boolean
+  expanded?: boolean
   auctionResult: ArtistAuctionResultItem_auctionResult
   index: number
   mediator?: Mediator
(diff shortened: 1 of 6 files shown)


artsyit:
:rocket: PR was released in `v25.18.2` :rocket:

PR #3199, opened 2020-02-26
Add data file to describe new menu structure (for mobile and desktop)

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-1702 

Co-authored-by: Christina Thompson

This roughs in a data structure that can be re-used across the new menu components we are creating for mobile and for desktop.

The file includes data for simple urls, as well as for nested submenus. Our expectation is that the eventual components for these new menus will import this data file and transform them as needed into the appropriate UI, whether for mega dropdowns (desktop) or for animated transitions (mobile).

The file is based on work done by CCE and documented [here](https://docs.google.com/spreadsheets/d/1WLQq6xRH8fa-LribPv5NiwnSuBQHQtjmMObm2GLIGY0/edit#gid=112964129)

### DIFF
--- a/src/Components/NavBar/menuData.ts
+++ b/src/Components/NavBar/menuData.ts
@@ -1,0 +1,448 @@
+interface MenuData {
+  title: string
+  links: LinkData[]
+}
+
+type LinkData = SimpleLinkData | MenuLinkData
+
+// e.g. "Editorial"
+interface SimpleLinkData {
+  text: string
+  href: string
+  dividerBelow?: boolean
+}
+
+// e.g. "Art Movement >"
+interface MenuLinkData {
+  text: string
+  menu: MenuData
+  dividerBelow?: boolean
+}
+
+export const menuData: MenuData = {
+  title: "", // root menu, so no title
+  links: [
+    {
+      text: "Artworks",
+      menu: {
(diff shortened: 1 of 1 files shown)


anandaroop:
Ah yeah, I can see how that would be confusing. This shows what those are _meant_ to convey:

- a `link.text` is the innerText of an `<a>` element. 
- a `menu.title` is the display title of a submenu that animates in (on mobile), or the subhead in the big dropdown (on desktop)

(`title` was actually different from `text` in some versions of the mobile design, thus we accounted for it explicitly in both contexts.)

Unknown

Let me know if that's clear as mud — if so, we can still change the naming 😛

sweir27:
Ahh gotcha. Maybe I'm confused because the `text` on desktop take you to a new page, but on mobile it will take you to a new screen in the nav menu (so slightly different versions of "links").

I can't think of a better alternative right now, and if it becomes confusing as you work with it it's not too hard to change. 😄

So 👍 from me!

anandaroop:
Cool, gonna merge then, and it's totally possible we will iterate on the structure or naming here. Thanks!

artsyit:
:rocket: PR was released in `v25.21.0` :rocket:

PR #3202, opened 2020-02-26
Log recaptcha misses on signup

eessex:
Logs from https://github.com/artsy/gravity/pull/12774 indicate that signups are coming through Force regularly that do not include the `recaptcha_token`. 

I believe all signups besides the Inquiry flow use Reaction's `FormSwitcher`, which should include tokens with all signup requests. This logging (which can be removed in the future) is to help us nail down if the errant Force signups are being routed through Reaction's signup form, and if they are caused by missing `recaptcha` lib on the window or a missing key.

### DIFF
--- a/src/Utils/recaptcha.ts
+++ b/src/Utils/recaptcha.ts
@@ -1,4 +1,7 @@
 import { data as sd } from "sharify"
+import createLogger from "Utils/logger"
+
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in `v25.20.2` :rocket:

PR #3203, opened 2020-02-27
[Clientside Nav] Move to more conventional A/B test structure

damassi:
This moves things to a conventional A/B test structure. 

@mzikherman - I keep revisiting the strange production-only edgecase around the ENV var being unset when clicking `/collect` and want to play it safe because we can't afford to get incorrect metrics. We can't repo it locally. If there's a chance that things get out of sync we might be over or under reporting a given A/B test state. 

I don't know if it has anything to do with `express-http-context`, but I'd rather be 100% sure that it's safe for a change as big as this, especially if it has to do with concurrency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
   "peerDependencies": {
     "@artsy/palette": "^4.16.0 || ^5.0.0 || ^7.0.0",
     "@sentry/browser": "^4.2.3",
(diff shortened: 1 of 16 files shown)


mzikherman:
LGTM!

artsyit:
:rocket: PR was released in `v25.21.0` :rocket:

PR #3204, opened 2020-02-27
Bump Node to v12 to have parity with Force

anandaroop:
For users of nvm, this will ensure that your local Reaction sessions will use [the same version of Node that Force does](https://github.com/artsy/force/commit/d6fd1103200d04e68b1194f9728a93411e298136#diff-ae9eba3500703eaba1d5b8e13832d955).

This fixes the currently broken `yarn integrate force` command.


dleve123:
👍– thanks @anandaroop !

artsyit:
:rocket: PR was released in `v25.21.1` :rocket:

PR #3207, opened 2020-02-28
[Clientside Nav] Add async bundle preloading pattern to router

damassi:
When we added bundle splitting we had a WIP item to figure out how to ensure that the split bundles were loaded once the route transitioned so as to mitigate the split-second white-screen flash. This PR adds a new pattern to address: 

```tsx
const MyApp = loadable(() => import('./MyApp')

const routes = [{
  path: '/path/to/app',
  getComponent: () => MyApp,
  prepare: () => {
    MyApp.preload()
  },
  ... 
}]
```

Whenever routes transition in the new router, `prepare` is called, which in turn calls the `preload` fn added by [loadable-components](https://loadable-components.com/docs/prefetching/). Since data is usually being fetched before the transition, the bundle loads in parallel leading to a stable page once the transition completes. 

#### Before: 

flash

#### After: 

noflash

I need to do one more "add tests" PR to wrap up this work; will address missing coverage there.

### DIFF
--- a/src/Apps/Artist/routes.tsx
+++ b/src/Apps/Artist/routes.tsx
@@ -61,6 +61,11 @@ const OverviewRoute = loadable(() => import("./Routes/Overview"))
 const WorksForSaleRoute = loadable(() => import("./Routes/Works"))
 const AuctionResultsRoute = loadable(() => import("./Routes/AuctionResults"))
 
+// Artist pages tend to load almost instantly, so just preload it up front
+if (typeof window !== "undefined") {
+  ArtistApp.preload()
+}
+
 // FIXME:
 // * `render` functions requires casting
 // * `Redirect` needs to be casted, as it’s not compatible with `RouteConfig`
--- a/src/Apps/Artist/routes.tsx
+++ b/src/Apps/Artist/routes.tsx
@@ -56,13 +56,21 @@ graphql`
   }
 `
 
+const ArtistApp = loadable(() => import("./ArtistApp"))
+const OverviewRoute = loadable(() => import("./Routes/Overview"))
+const WorksForSaleRoute = loadable(() => import("./Routes/Works"))
+const AuctionResultsRoute = loadable(() => import("./Routes/AuctionResults"))
+
 // FIXME:
 // * `render` functions requires casting
 // * `Redirect` needs to be casted, as it’s not compatible with `RouteConfig`
 export const routes: RouteConfig[] = [
   {
     path: "/artist/:artistID",
-    getComponent: () => loadable(() => import("./ArtistApp")),
+    getComponent: () => ArtistApp,
+    prepare: () => {
+      ArtistApp.preload()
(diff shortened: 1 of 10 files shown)


src/Apps/Artist/routes.tsx
```diff
@@ -56,13 +56,26 @@ graphql`
...
 
+const ArtistApp = loadable(() => import("./ArtistApp"))
+const OverviewRoute = loadable(() => import("./Routes/Overview"))
+const WorksForSaleRoute = loadable(() => import("./Routes/Works"))
+const AuctionResultsRoute = loadable(() => import("./Routes/AuctionResults"))
+
+// Artist pages tend to load almost instantly, so just preload it up front
+if (typeof window !== "undefined") {
+  ArtistApp.preload()
+}
```
zephraph: Are we sure we want this to always happen?

src/Apps/Artist/routes.tsx
```diff
@@ -56,13 +56,26 @@ graphql`
...
 // * `render` functions requires casting
 // * `Redirect` needs to be casted, as it’s not compatible with `RouteConfig`
 export const routes: RouteConfig[] = [
   {
     path: "/artist/:artistID",
-    getComponent: () => loadable(() => import("./ArtistApp")),
+    getComponent: () => ArtistApp,
+    prepare: () => {
+      ArtistApp.preload()
+    },
```
zephraph: If we do decide to always preload, it's likely that we won't need this.

src/Apps/Artist/routes.tsx
```diff
@@ -231,6 +205,57 @@ export const routes: RouteConfig[] = [
...
+            artist(id: $artistID) {
+              ...AuctionResults_artist
+            }
+          }
+        `,
+      },
+
+      // FIXME: Remove the following unused routes
+      {
+        path: "cv",
```
zephraph: No preload for these?

damassi: No longer used

src/Apps/Artist/routes.tsx
damassi: Its a no-op if already loaded, so would rather keep things consistent. We might not want to preload immediately in the future.

src/Apps/Artist/routes.tsx
zephraph: Hmmm, I think I'm missing something. These pages still exist on the artist page, they're just behind links instead of tabs. That means the routes are still needed, right?

damassi: Oh, i didn't realize these were being used outside of the tabs. If that's the case then yup, we should load them as well in a similar fashion. Will update that now.

damassi: Done

artsyit:
:rocket: PR was released in `v25.22.0` :rocket:

PR #3208, opened 2020-02-28
Only render "RelatedCollections" that have artworks

sweir27:
Relevant ticket: https://artsyproduct.atlassian.net/browse/FX-1827

Because some of the collections pulled in the `RelatedCollectionsRail` query return `null` for `artworksConnection`, we got a cascading failure in this component that led to an infinite reload.

We should look into why those collections are returning `null` for `artworksConnection`, but it also doesn't hurt to be safe here.


src/Components/RelatedCollectionsRail/__tests__/RelatedCollectionsRail.test.tsx
```diff
@@ -38,6 +38,19 @@ describe("CollectionsRail", () => {
...
   })
 
+  it("does not render a carousel entry if it has no artworks", () => {
+    const collectionsCopy = clone(props.collections)
+    collectionsCopy.push({
+      slug: "jasper-johns-flags2",
+      headerImage: "http://files.artsy.net/images/jasperjohnsflag.png",
+      title: "Jasper Johns: Flags Part 2",
+      price_guidance: 1000,
+      artworksConnection: null,
```
anandaroop: Ah, so in the case of this bug, `artworksConnection` is coming back as null altogether rather than with `edges: []`

Curious! But this looks like a good fix.

src/Components/RelatedCollectionsRail/RelatedCollectionsRail.tsx
```diff
@@ -114,6 +118,19 @@ export const RelatedCollectionsRailFragmentContainer = createFragmentContainer(
...
+          first: 3
+          aggregations: [TOTAL]
+          sort: "-decayed_merch"
+        ) {
+          edges {
+            node {
+              id
+            }
+          }
+        }
```
anandaroop: This `artworksConnection` field has to be hoisted up here bc of Relay data masking, is that right?

artsyit:
:rocket: PR was released in `v25.21.3` :rocket:

src/Components/RelatedCollectionsRail/RelatedCollectionsRail.tsx
mzikherman: Yea, we have this pattern in a few places. The exact same snippet, copied into here, will ensure we don't kick off redundant data requests (differing slightly in some query params maybe), when the goal is to have an exist-y check on the data.

And then yea, the child component has the actual data fragment, but a parent component often needs to know whether a child component will render or not, for a layout reason.

Short of larger refactors with the component heirarchy, I don't know a great solution! This might be a reason to actually use the `@relay(mask: false)` directive, or copy the whole fragment into a string that can be shared and interpolated here (to help avoid a bug of updating one of the two copies).

sweir27: I've been encouraged to avoid relay masking wherever possible-- in this case the duplication means that if we no longer need this functionality, it's very clear to just remove it from this one spot!

PR #3214, opened 2020-02-28
[Search] Dont split bundles

damassi:
Seeing a chunk issue, so going to avoid splitting for the time being.


artsyit:
:rocket: PR was released in `v25.22.2` :rocket:

PR #3217, opened 2020-02-29
[Search] Visual QA

damassi:
Fixes a few visual qa items on search:

In the new app shell when performing a search the query stays in the input, since there's no longer a hard jump between pages to clear it out. Now it clears on enter: 

input

The search title is a function of url bar state. Now, when transitioning away from the page the url changes and the title updates, but before a full page transition. This ensures that things only update when changed: 

Before: 

title-before

After: 

title-after

### DIFF
--- a/src/Apps/Search/SearchApp.tsx
+++ b/src/Apps/Search/SearchApp.tsx
@@ -12,7 +12,7 @@ import { Footer } from "Components/v2/Footer"
 import { RecentlyViewedQueryRenderer as RecentlyViewed } from "Components/v2/RecentlyViewed"
 
 import { RouterState, withRouter } from "found"
-import React from "react"
+import React, { useEffect, useState } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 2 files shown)


src/Apps/Search/SearchApp.tsx
```diff
@@ -26,11 +26,21 @@ export interface Props extends RouterState {
...
-    {count.toLocaleString()} Result{count > 1 ? "s" : ""} for "{term}"
-  </Serif>
-)
+}) => {
+  const formatResults = () =>
+    count.toLocaleString() +
+    " Result" +
+    (count > 1 ? "s" : "") +
+    " for " +
+    `"${term}"`
```
zephraph: Would a template string be cleaner here?

damassi: Yup, much cleaner! fixed.

artsyit:
:rocket: PR was released in `v25.24.1` :rocket:

PR #3218, opened 2020-02-29
[Artist] Visual QA

damassi:
Noticed that on the artist "Works" tab the layout is unstable due to the collections rail always making a request even though the data has already been fetched, which also applies if there's no rail for a given artist, leading to a loading state to empty content. This fixes both things and ensures a stable layout.

Wondering if there was a deliberate reason for forcing a refetch every time? 

(Note that we have a relay cache expiration of 15 minutes, so unless the collections rail has rapidly changing results we should be ok.) 

### Before: 

before

### After: 

after


artsyit:
:rocket: PR was released in `v25.23.1` :rocket:

PR #3221, opened 2020-03-02
[Clientside Nav] Add more tests

damassi:
Adds a few more tests around client-side nav work.


src/Artsy/Router/buildAppRoutes.tsx
```diff
@@ -34,11 +34,12 @@ export function makeAppRoutes(routeList: RouteList[]): RouteConfig[] {
        * manifest, navigate via router versus doing a hard jump between pages.
        */
       catchLinks(window, href => {
-        const url = ROUTE_NAMESPACE + href
```
zephraph: What makes `ROUTE_NAMESPACE` no longer required?

damassi: It was always just an empty string -- I had this in place because for the POC I had apps mounted within a namespace, but no longer needed.

artsyit:
:rocket: PR was released in `v25.23.2` :rocket:

PR #3222, opened 2020-03-02
Make countries list on the Address form render only EU countries if it's marked as euOrigin and domesticOnly

oxaudo:
This is work in progress for supporting 'Euro domestic' shipping.

Want to get circle to run the specs and see how it goes.


oxaudo:
@zephraph and @damassi - @ansor4  and myself would appreciate an opinion on wether this change makes sense or if that looks like a decent way to get to the required functionality.
Basically the requirement is that 'domestic' shipping in case the original `ShippingCountry' is in Eu includes all EU countries instead of just the country where the work is located. I opted out to add a separate prop to artworkForm to signify that the  form should assume `euOrigin`. This prop is already added to artwork in MP. 

Also wondering what is `codecov` fail is about here?

zephraph:
cc @eessex, could you look at the code coverage here?

src/Components/v2/CountrySelect.tsx
```diff
@@ -18,7 +28,7 @@ const COUNTRY_SELECT_OPTIONS = [
   { text: "Albania", value: "AL" },
   { text: "Algeria", value: "DZ" },
   { text: "American Samoa", value: "AS" },
-  { text: "AndorrA", value: "AD" },
+  { text: "Andorra", value: "AD" },
```
oxaudo: This I think was a plain typo:)

src/Components/v2/CountrySelect.tsx
```diff
@@ -205,7 +216,7 @@ const COUNTRY_SELECT_OPTIONS = [
   { text: "Sao Tome and Principe", value: "ST" },
   { text: "Saudi Arabia", value: "SA" },
   { text: "Senegal", value: "SN" },
-  { text: "Serbia and Montenegro", value: "CS" },
```
oxaudo: And I believe proper way to refer to this country is 2 countries: https://en.wikipedia.org/wiki/ISO_3166-2:CS

oxaudo:
@ansor4 - updated the list of countries that are included in EU local shipping here.

oxaudo:
Going to merge that to generate the new build that we can add to force.

artsyit:
:rocket: PR was released in `v25.27.15` :rocket:

PR #3224, opened 2020-03-02
Update AB test name to CLIENT_NAVIGATION_V4

damassi:
Updates client-side A/B test name from `CLIENT_NAVIGATION_V3` to `CLIENT_NAVIGATION_V4` just so we can be sure that the data is consistent, accounting for a mid-day deploy on Thursday after test was released.


artsyit:
:rocket: PR was released in `v25.24.2` :rocket:

PR #3226, opened 2020-03-03
[FX-1758] add a fallback image when there is no headerImage

xtina-starr:
Addresses: [FX-1758](https://artsyproduct.atlassian.net/browse/FX-1758)

**Problem:**
When links for collection pages are shared on social media the display appears with a small photo. This is because for most collection pages we don't have a header image and the meta data for the image gets pulled from the images in the artwork grid.

**Fix:**
This PR introduces a new fallback image that is a larger, resized version of the first image in the artwork grid. The logic checks for the headerImage first and if it's not present uses the resized image.

### DIFF
--- a/src/Apps/Collect2/Routes/Collection/index.tsx
+++ b/src/Apps/Collect2/Routes/Collection/index.tsx
@@ -62,6 +62,7 @@ export class CollectionApp extends Component<CollectionAppProps> {
       slug,
       headerImage,
       description,
+      fallbackHeaderImage,
       artworksConnection,
       descending_artworks,
       ascending_artworks,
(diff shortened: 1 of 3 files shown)


xtina-starr:
@sweir27 my thought process if the social links still don't display as we'd like after including a fallback image with the size increase, we might want to consider adding a custom social image to the collection model.

src/Apps/Collect2/Routes/Collection/index.tsx
```diff
@@ -74,13 +75,16 @@ export class CollectionApp extends Component<CollectionAppProps> {
       : `Buy, bid, and inquire on ${title} on Artsy.`
 
     const showCollectionHubs = collection.linkedCollections.length > 0
+
+    const socialImage =
+      headerImage || fallbackHeaderImage?.edges[0]?.node?.image?.resized.url
```
sweir27: I think this line will still fail if `edges` returns `null`-- but a test can confirm!

src/Apps/Collect2/Routes/Collection/index.tsx
```diff
@@ -77,7 +77,10 @@ export class CollectionApp extends Component<CollectionAppProps> {
     const showCollectionHubs = collection.linkedCollections.length > 0
 
     const socialImage =
-      headerImage || fallbackHeaderImage?.edges[0]?.node?.image?.resized.url
+      headerImage ||
+      (fallbackHeaderImage?.edges &&
+        fallbackHeaderImage?.edges[0]?.node?.image?.resized.url)
+
```
xtina-starr: I refactored this to more explicitly check for `null`. Regarding testing, I attempted to add tests but the CollectionApp component doesn't currently have any tests and it's quite a bit of work to set up so I created a [ticket](https://artsyproduct.atlassian.net/browse/FX-1849) to do so.

damassi: @xtina-starr - I _think_ `edges[0]?` should work? Can try it out in the chrome console: 

Screen Shot 2020-03-11 at 3 51 28 PM

artsyit:
:rocket: PR was released in `v25.30.1` :rocket:

PR #3220, opened 2020-03-02
Only lint changed files

zephraph:
This sets up a package I wrote this weekend called [lint-changed](https://github.com/artsy/lint-changed). It's very similar in theory to `lint-staged` except it's intended to be useful for CI. 

The premise is that linting has a 1-to-1 relationship with files and running linting over the entire codebase is pretty wasteful. On a PR this will only lint  the files that have changed since `origin/master`. On master this will only lint the files that have changed since the last tag.

This is in the same vein of doing less duplicative work on CI. 

<details>
  <summary>📦 Published PR as canary version: <code>25.22.3-canary.3220.53079.0</code></summary>
  <br />
  
  :sparkles: Test out this PR locally via:
  
  ```sh
  npm install @artsy/reaction@25.22.3-canary.3220.53079.0
  # or 
  yarn add @artsy/reaction@25.22.3-canary.3220.53079.0
  ```
</details>

PR #3230, opened 2020-03-04
Only lint changed files

zephraph:
This restores #3220 but fixes an issue that we had where it was failing in master.  

See https://github.com/artsy/lint-changed/pull/3 for more details on the actual  fix.


artsyit:
:rocket: PR was released in `v25.24.3` :rocket:

PR #3232, opened 2020-03-04
Begin ID Verification flow from collector verification app

erikdstock:
Completes [[AUCT-894](https://artsyproduct.atlassian.net/browse/AUCT-894)]

TODO:
- [ ] Implement tracking on mutation error (skipping until we have a request for this)
- [x] Complete tests

### DIFF
--- a/src/Apps/IdentityVerification/IdentityVerificationApp.tsx
+++ b/src/Apps/IdentityVerification/IdentityVerificationApp.tsx
@@ -8,7 +8,8 @@ import { AppContainer } from "Apps/Components/AppContainer"
 import { useTracking } from "Artsy"
 import * as Schema from "Artsy/Analytics/Schema"
 import { ErrorPage } from "Components/ErrorPage"
-import React from "react"
+import { ErrorModal } from "Components/Modal/ErrorModal"
+import React, { useState } from "react"
 import { Meta, Title as HeadTitle } from "react-head"
 import {
   commitMutation,
@@ -16,12 +17,15 @@ import {
   graphql,
   RelayProp,
 } from "react-relay"
+import createLogger from "Utils/logger"
 // import { RedirectException } from "found"
+
+const logger = createLogger("IdentityVerificationApp.tsx")
 
 interface Props {
   me: IdentityVerificationApp_me
-  // match: any
   relay: RelayProp
+  // match: any
 }
 
 const IdentityVerificationApp: React.FC<Props> = ({ me, relay }) => {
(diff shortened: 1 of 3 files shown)


src/Apps/IdentityVerification/IdentityVerificationApp.tsx
```diff
@@ -1,35 +1,131 @@
...
+    const {
+      startIdentityVerification: {
+        startIdentityVerificationResponseOrError: {
+          identityVerificationFlowUrl,
+          mutationError,
+        },
+      },
+    } = result
+    if (mutationError) {
+      logger.error("FAIL in handleMutation")
```
damassi: Unless handled elsewhere, maybe log the mutation error from the logger?

erikdstock: yes, this was just debugging/understanding why the `onError` code wasn't being executed (because that is only if the xhr request itself fails, need to check for error in `onCompleted`). If there is a _better_ way of doing this than what i am doing (where onError basically takes 2 different signatures) I'm interested to hear it!

src/Apps/IdentityVerification/IdentityVerificationApp.tsx
```diff
@@ -1,35 +1,131 @@
...
     })
   }
 
+  function startIdentityVerification() {
+    const mutation = new Promise<
+      IdentityVerificationAppStartFlowMutationResponse
+    >((resolve, reject) => {
+      commitMutation<IdentityVerificationAppStartFlowMutation>(environment, {
+        onCompleted: data => resolve(data),
+        onError: error => reject({ error }),
```
erikdstock: Major question on approach is here. Do we have any very simple examples for a pattern on doing a mutation and handling errors either from metaphysics (in `onCompleted`) or request failure (`onError`)? In psuedocode, my current version looks like this:

```ts
 const mutation = new Promise<
   IdentityVerificationAppStartFlowMutationResponse
 >((resolve, reject) => {
   commitMutation<IdentityVerificationAppStartFlowMutation>(environment, {
     onCompleted: data => resolve(data),
     onError: error => reject({ error }),
     mutation: graphql`mutation {...`
   })
 })
  
const handleMutationSuccess = (
    result: IdentityVerificationAppStartFlowMutationResponse
  ) => {
    const {
      startIdentityVerification: {
        startIdentityVerificationResponseOrError: {
          identityVerificationFlowUrl,
          mutationError,
        },
      },
    } = result
    if (mutationError) {
      handleMutationError(mutationError)
    } else {
      location.assign(identityVerificationFlowUrl)
    }
  }
  const handleMutationError = (mutationError: {
    /** Message/detail from commitMutation conCompleted handler */
    message?: string
    detail?: string
    /** Error from commitMutation onError handler (also possibly in the mutation result) */
    error?: string
  }) => {
    logger.error("FAIL", mutationError)
}

  mutation.then(handleMutationSuccess).catch(handleMutationError)

  }
```

damassi: See [here](https://github.com/damassi/reaction/blob/d43843bf504b980fa8d6862cc3520c39b9c6065e/src/Apps/Search/Components/SendFeedback.tsx#L90-L92) for a possible approach to handling errors via `ErrorWithMetadata`.

damassi: (Also lots of different examples in Reaction when searching for `commitMutation`; maybe one of those results will help.)

erikdstock:
Merging, but want to call out these commits from the source branch as we may want to re-implement them later:
866a507 - tests for tracking errors
791378d - redirect to login if no `me` (didn't work as expected, was out of scope of ticket so I let it go for now)

artsyit:
:rocket: PR was released in `v25.26.2` :rocket:

PR #3233, opened 2020-03-05
Add attribute to body when app is ready to be tested

zephraph:
This adds a mechanism to tell cypress that we're ready to start testing. This is a mechanism we can use to avoid timing issues w/ rehydration. 

It will need to be followed up with an integrity update to check for this after page visits. 

<details>
  <summary>📦 Published PR as canary version: <code>25.25.0-canary.3233.53368.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.25.0-canary.3233.53368.0
  # or 
  yarn add @artsy/reaction@25.25.0-canary.3233.53368.0
  ```
</details>

### DIFF
--- a/src/Apps/Components/AppShell.tsx
+++ b/src/Apps/Components/AppShell.tsx
@@ -30,9 +30,7 @@ export const AppShell = props => {
    * Let our end-to-end tests know that the app is hydrated and ready to go
    */
   useEffect(() => {
-    if (typeof window !== "undefined") {
-      document.body.setAttribute("data-test-ready", "")
-    }
+    document.body.setAttribute("data-test-ready", "")
   }, [])
 
   return (
--- a/src/Apps/Components/AppShell.tsx
+++ b/src/Apps/Components/AppShell.tsx
@@ -26,6 +26,15 @@ export const AppShell = props => {
     }
   }, [routeConfig])
 
+  /**
+   * Let our end-to-end tests know that the app is hydrated and ready to go
+   */
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      document.body.setAttribute("data-test-ready", "")
+    }
+  }, [])
+
   return (
     <Box width="100%">
       <Box pb={6}>


src/Apps/Components/AppShell.tsx
```diff
@@ -26,6 +26,15 @@ export const AppShell = props => {
     }
   }, [routeConfig])
 
+  /**
+   * Let our end-to-end tests know that the app is hydrated and ready to go
+   */
+  useEffect(() => {
+    if (typeof window !== "undefined") {
```
damassi: Inside of `useEffect` there will always be a window

zephraph: Oh, right. Thanks!

src/Apps/Components/AppShell.tsx
```diff
@@ -26,6 +26,15 @@ export const AppShell = props => {
...
+      document.body.setAttribute("data-test-ready", "")
```
damassi: The format we've been using for data attributes includes [some assignment](https://github.com/artsy/integrity/blob/7dd7aab86a8d4c750d6bd979b76ff193c3ade927/cypress/utils/buyArtworkWithPickup.ts#L42)

src/Apps/Components/AppShell.tsx
```diff
@@ -26,6 +26,15 @@ export const AppShell = props => {
...
+    }
```
zephraph:
```suggestion
    document.body.setAttribute("data-test-ready", "")
```

zephraph: cc @damassi we can do multi-line suggestions now!

src/Apps/Components/AppShell.tsx
```diff
@@ -26,6 +26,13 @@ export const AppShell = props => {
     }
   }, [routeConfig])
 
+  /**
+   * Let our end-to-end tests know that the app is hydrated and ready to go
+   */
+  useEffect(() => {
+    document.body.setAttribute("data-test-ready", "")
```
zephraph:
```suggestion
    document.body.setAttribute("data-test", "AppReady")
```

zephraph: Something like this?

artsyit:
:rocket: PR was released in `v25.26.0` :rocket:

PR #3240, opened 2020-03-06
[Clientside Nav] Add <NetworkTimeout /> & <NetworkOfflineMonitor /> components

damassi:
Addresses 
- https://artsyproduct.atlassian.net/browse/PLATFORM-2283
- https://artsyproduct.atlassian.net/browse/PLATFORM-2284

### `<NetworkTimeout />`
This adds a new component to our routing framework that will appear if a request is taking longer than 10 seconds, showing our standard ErrorDialog component and prompting the user to retry, which will reload the page. 

alert

--- 

### `<NetworkOfflineMonitor />`

Also added a `NetworkOfflineMonitor` component that will check if the network is offline, and shows a Banner alerting the user: 

offline

(Can test this via chrome dev tools.)

### DIFF
--- a/src/Artsy/Router/NetworkOfflineMonitor.tsx
+++ b/src/Artsy/Router/NetworkOfflineMonitor.tsx
@@ -1,14 +1,15 @@
-import { ModalDialog } from "Components/Modal/ModalDialog"
+import { Banner, Box } from "@artsy/palette"
+import { debounce } from "lodash"
 import React, { useEffect, useState } from "react"
 
 export const NetworkOfflineMonitor: React.FC = () => {
   const [showOfflineModal, toggleOfflineModal] = useState(false)
 
   const setOffline = () => {
-    toggleOfflineModal(true)
+    debounce(() => toggleOfflineModal(true), 100)
   }
   const setOnline = () => {
-    toggleOfflineModal(false)
+    debounce(() => toggleOfflineModal(false), 100)
   }
 
   useEffect(() => {
@@ -26,13 +27,12 @@ export const NetworkOfflineMonitor: React.FC = () => {
   }
 
   return (
-    <ModalDialog
-      show={showOfflineModal}
-      heading="Your network is offline"
-      primaryCta={{
-        text: "Dismiss",
-        action: () => setOnline(),
-      }}
-    />
+    <Box position="fixed" top={58} width="100%" style={{ opacity: 0.9 }}>
+      <Banner
+        backgroundColor="black10"
+        textColor="black100"
+        message="Network offline"
+      />
+    </Box>
   )
 }
--- a/src/Artsy/Router/NetworkOfflineMonitor.tsx
+++ b/src/Artsy/Router/NetworkOfflineMonitor.tsx
@@ -1,0 +1,38 @@
+import { ModalDialog } from "Components/Modal/ModalDialog"
+import React, { useEffect, useState } from "react"
+
+export const NetworkOfflineMonitor: React.FC = () => {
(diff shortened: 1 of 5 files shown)


joeyAghion:
What happens now in case of other (non-timeout) errors? We should definitely handle them, but it may be healthy to just degrade to browser errors in some circumstances or at least reveal the nature of the errors (status codes, etc.). Otherwise we may end up needing to diagnose errors with very little context. I'm thinking of 5xxs from Metaphysics in particular, but also the wide variety of other types like connections closing, SSL, Cloudflare....

Is the offline modal triggered only on an attempted navigation?

src/Artsy/Router/NetworkOfflineMonitor.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  useEffect(() => {
+    window.addEventListener("offline", setOffline)
+    window.addEventListener("online", setOnline)
+
+    return () => {
+      window.removeEventListener("offline", setOffline)
+      window.removeEventListener("online", setOnline)
+    }
+  }, [])
```
zephraph: If something happens and their network triggers on and off a few times this could be _really_ annoying. Maybe we should denounce it?

src/Artsy/Router/NetworkTimeout.tsx
```diff
@@ -0,0 +1,34 @@
...
+      clearTimeout(timeout)
+    }
+  }, [])
+
+  if (!showErrorModal) {
+    return null
+  }
+
+  return (
+    <ErrorModal
```
zephraph: Is there someway that we can send that this happened to sentry?

src/Artsy/Router/NetworkOfflineMonitor.tsx
zephraph: Actually, thinking more about this, is there a subtler mechanism we could provide? 

There is a somewhat common pattern of using a banner to show offline

image

damassi:
@joeyAghion 

> What happens now in case of other (non-timeout) errors? We should definitely handle them, but it may be healthy to just degrade to browser errors in some circumstances or at least reveal the nature of the errors (status codes, etc.). Otherwise we may end up needing to diagnose errors with very little context. I'm thinking of 5xxs from Metaphysics in particular

We already handle errors via the `RenderError` component. If something throws in the tree, or some other error goes down due to `@principleField`, etc, the user will get taken to the error page. It's been in for a while. See https://github.com/artsy/reaction/blob/master/src/Artsy/Router/RenderStatus.tsx#L110-L130.

src/Artsy/Router/NetworkOfflineMonitor.tsx
damassi: Oh yeah! we can use the staging status-bar and change the color a bit. Good idea.

damassi:
@zephraph - addressed your requests and this is ready for another review!

src/Artsy/Router/NetworkOfflineMonitor.tsx
```diff
@@ -0,0 +1,38 @@
...
+import React, { useEffect, useState } from "react"
+
+export const NetworkOfflineMonitor: React.FC = () => {
+  const [showOfflineModal, toggleOfflineModal] = useState(false)
+
+  const setOffline = () => {
+    debounce(() => toggleOfflineModal(true), 100)
+  }
+  const setOnline = () => {
+    debounce(() => toggleOfflineModal(false), 100)
```
zephraph: 👍

artsyit:
:rocket: PR was released in `v25.28.0` :rocket:

PR #3243, opened 2020-03-07
[Carousel] Remove error boundary

damassi:
@sweir27 - caught a reload bug issue, which I tracked down [to here](https://sentry.io/organizations/artsynet/issues/1514677136/?project=28316&query=is%3Aunresolved+flickity&sort=freq&statsPeriod=14d). 

In a super alpha version of the shell I noticed an issue with flickity, and to be safe I added this guard that was never removed, and I haven't seen the issue reappear in many months. However, when some bad MP data came through to the carousel, it would catch in the error boundary and trigger a reload, versus failing back to error page which we now have in place. 

https://github.com/artsy/reaction/pull/3208 fixed the data issue.


artsyit:
:rocket: PR was released in `v25.27.1` :rocket:

PR #3244, opened 2020-03-08
Replace getENV with sd for now

yuki24:
`sd` is a bit smarter than `getENV` since it converts `'false'` (the string) to `false` (the boolean.) Because this was only checking the presence of the value, the string value `'false'` was always evaluating to truthy regardless of what is set to it, causing the lot condition report feature to always be on for every sale.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
+++ b/src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
@@ -9,8 +9,8 @@ import {
 import { ArtworkDetailsAdditionalInfo_artwork } from "__generated__/ArtworkDetailsAdditionalInfo_artwork.graphql"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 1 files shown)


yuki24:
#mergeongreen since this is somewhat urgent and I want to minimize extra admin costs.

artsyit:
:rocket: PR was released in `v25.27.2` :rocket:

src/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAdditionalInfo.tsx
```diff
@@ -44,7 +44,7 @@ export class ArtworkDetailsAdditionalInfo extends React.Component<
         title: "Condition",
         value:
           canRequestLotConditionsReport &&
-          getENV("ENABLE_REQUEST_CONDITION_REPORT") ? (
+          sd.ENABLE_REQUEST_CONDITION_REPORT ? (
```
damassi: @yuki24 - The problem is `sd` only renders client-side, so the SSR pass may be compromised. Any chance an update can be made to `getENV` instead?

PR #3247, opened 2020-03-09
Migrate to ESLint

zephraph:
This is pretty highly WIP. 

I used https://github.com/typescript-eslint/tslint-to-eslint-config to bootstrap the process. Afterward I spent time digging into the `Artist` try to tweak things to produce the least error level noise. There are a lot of warnings that are interesting, but I've got `--quiet` enabled for now which silences those. 

Given that we're using `lint-changed` now, the linter will only run for... well... files that have changed. So it's technically possible for us to merge this once we feel comfortable without applying changes to every file. 

I would like to make sure we tweak the levels to not be too annoying though. 

<details>
  <summary>📦 Published PR as canary version: <code>25.27.3-canary.3247.53670.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.27.3-canary.3247.53670.0
  # or 
  yarn add @artsy/reaction@25.27.3-canary.3247.53670.0
  ```
</details>


.eslintrc.js
```diff
@@ -0,0 +1,188 @@
+module.exports = {
```
damassi: That's a lot of "offs"! 

I set up [this config](https://github.com/damassi/gatsby-starter-typescript-rebass-netlifycms/blob/master/.eslintrc.js) on a personal project which matches things pretty well (in the sense that most code I write is in Reaction, and so that's the defaults I'm most familiar with). Might be something here?

.eslintrc.js
```diff
@@ -0,0 +1,188 @@
...
+  },
+  extends: [
+    "plugin:@typescript-eslint/recommended",
+    "plugin:@typescript-eslint/recommended-requiring-type-checking",
+    "plugin:jsx-a11y/recommended",
+    "plugin:react/recommended",
+    "plugin:import/errors",
+    "plugin:import/warnings",
+    "plugin:import/typescript",
+    "plugin:relay/recommended",
```
damassi: Also noticed that there's no eslint-config-prettier plugin setup here, so that might be why so many rules have been impacted below.

zephraph: Most of those I didn’t add manually, but are generated from the tslint to Eslint migration tool. 

But, yeah, still wip

damassi: We can / should do this by hand, which will be better long-term. Having every rule listed out is a maintenence burden while also being kinda confusing for other devs in the future ("why did x set this in this way?")

zephraph: I agree that having too many rules would be confusing.

This PR isn't final, it's just a good starting point. I used the tool because it's recommended and a good kicking off point to get the same coverage as we had. Optimizations like adding the prettier plugin (which is installed, but I just forgot to add) will likely reduce the amount of rules here. 

Again though, this is just a WIP PR and in no way meant to represent the final thing. Just wanted to get something up.

damassi:
Going to close this so that we can start one that's hand crafted.

PR #3248, opened 2020-03-09
Make hide grids function a bit safer

damassi:
Noticed there was an elastic search data issue at https://staging.artsy.net/artwork/chip-hughes-untitled and digging in, saw that we could add a bit more safety. See [this thread](https://artsy.slack.com/archives/C9YNS4X32/p1583767239172700).

Also noticed here that this revealed a bug in the `get` function itself -- if the initial object being passed in is null, then it will still error out, versus being safe all the way through the chain. (Right now, any _subsequent_ properties can be null / undefined). Either way, we can now use optional chaining so get is no longer nececessary.


damassi:
I don't have time to dive into `get` at the moment, but we should encourage the use of optional chaning when we see PRs that use new instances of get.

artsyit:
:rocket: PR was released in `v25.27.6` :rocket:

PR #3249, opened 2020-03-09
Update schema

damassi:
Synching schema with new date field from convection 

#trivial


artsyit:
:rocket: PR was released in `v25.27.4` :rocket:

PR #3253, opened 2020-03-09
[Artist] Fix redirect

damassi:
Noticed that the redirect was formatted incorrectly


src/Apps/Artwork/Components/ArtworkBanner/index.tsx
```diff
@@ -15,6 +15,10 @@ export interface ArtworkBannerProps {
 export const ArtworkBanner: React.SFC<ArtworkBannerProps> = props => {
   const { context, partner, sale } = props.artwork
 
+  if (!sale) {
+    return null
+  }
```
damassi: Interestingly, a test already covers this. Either way, just to be safe.

artsyit:
:rocket: PR was released in `v25.27.7` :rocket:

PR #3255, opened 2020-03-10
Remove auction routes

damassi:
Removing auction stuff from routes


artsyit:
:rocket: PR was released in `v25.27.8` :rocket:

PR #3256, opened 2020-03-10
Remove BidFlow from router

damassi:
Remove all router code from the bid flow. 

Merging to get this hotfix out.


artsyit:
:rocket: PR was released in `v25.27.9` :rocket:

dleve123:
@damassi Were there any further problems found with this code, or did you removal out of caution?

damassi:
There are two parts to the auction flow -- the redirects that happen when registering from an auction page and entering a credit card, and then the artwork-only page flow (bid, enter credit card, raise bid, etc) -- we removed the first part, but not the second, which became apparent when clicking the bid button, as we then tried to do a "soft" jump via the router to the bid form page but that had been unmounted on the force side of things. This pr removes the remainder of code.

damassi:
Its a bit of a bummer to remove all of this (versus being able to toggle off) but we have references in the future to re-add when the credit card issue is resolved.

PR #3257, opened 2020-03-10
Add Integrity-ready flag to boot

damassi:
Follow up to https://github.com/artsy/reaction/pull/3233, which expands coverage a bit for all kinds of components / pages that use `<Boot>`.


artsyit:
:rocket: PR was released in `v25.27.11` :rocket:

PR #3258, opened 2020-03-10
Move conversations app to be behind lab feature instead of admin flag

zephraph:
This is the over half of the work needed to move the inquiries view from behind admin to behind the lab feature flag. 

I didn't realize it was gated both in the menu and in the app.


artsyit:
:rocket: PR was released in `v25.27.12` :rocket:

PR #3261, opened 2020-03-11
[Consignments] Build out new Artist/Consign landing page

damassi:
Figma: https://www.figma.com/file/MFZBkqq9A5E7msdDYUkyJK/Consign-Top-20-Landing-Page?node-id=60%3A863

This is the first pass over the new consign page. (Most) desktop and mobile styles have been wired up, and dynamic data is being fetched for the bricks based on the spreadsheet. (I still need to wrap the artwork bricks with additional metadata, though.)

There's some style tweaks that need to be done in a follow up PR after I meet with Nicole, but this will allow us to get something reviewable up on staging. 

After QA, will follow up in a second PR with test coverage. 

consign
consign-small


src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
```diff
@@ -0,0 +1,2593 @@
...
+
+      // Mapped CSV columns
+      const {
+        "views to content from that artist (last month)": views,
+        "ROUNDED views to content from that artist (last month)": roundedViews,
+        "unique visitors to content from artist (last month)": uniqueVisitors,
+        "ROUNDED unique visitors to content from artist (last month)": roundedUniqueVisitors,
+        "STR (last 12 months)": str,
+        "Realized / estimate": realized,
+        "Highest realized price": highestRealized,
```
damassi: @jonallured, @pepopowitz - I'm taking the columns straight out of Joannas spreadsheet via CSV converter, see any problem with this?

pepopowitz: One thought I have is about future updates. I don't know what our thoughts are on how much this data is going to change, but given past experience with spreadsheet-driven-data, I wouldn't count on what's in the spreadsheet being final. 

Assuming the data is going to change at the source, we should have a simple & repeatable way of replacing it here. I could see that going a few ways:

* Including an actual raw CSV (or something) in the project and parsing that via this function. Replacement of data would be as simple as exporting the spreadsheet as CSV and pasting it into the file.
* Leaving it as JSON, but providing guidance on which tool to use to convert the CSV to JSON. I know this seems simple, but I think it's important to provide this guidance so that it's repeatable, and people don't have to investigate tooling.
* Including a task in the project to turn the CSV into JSON. 

Regardless, I think it'd be important to include docs in a .md about how to replace the data. I also think if we stick with the JSON route, we should extract the JSON to a file completely on its own. I find large, hardcoded data sets easier to work with when they stand alone.

damassi: All good points! Will address.

(I didn't want to include a CSV-to-JSON lib in the project because it would be a 1-off and unsure how often this data will need to change, but I do think moving data into its own file would be helpful, and adding some instructions. If this starts to change with any regularity then we can pull in a lib.)

src/Apps/Artist/Routes/Consign/Utils/staticCSVToJSONData.ts
```diff
@@ -0,0 +1,2372 @@
+/**
+ * CSV data exported from https://docs.google.com/spreadsheets/d/1EicHOCyCozsPBkVOkQGTTUtEds7qQuacUAgLy-C2-Nk/edit#gid=0
+ *
+ * To update data:
+ *   1) Open spreadsheet link above
+ *   2) File > Export > CSV
+ *   3) Copy CSV data and paste into https://www.csvjson.com/csv2json
+ *   4) Copy results and overwrite below
+ */
```
damassi: @pepopowitz - I added some update docs here. Anyone needing to modify or update this static data will encounter this file in the execution path. Does this satisfy your request?

src/Components/ErrorBoundary.tsx
```diff
@@ -13,11 +13,13 @@ interface Props {
 
 interface State {
   asyncChunkLoadError: boolean
+  genericError: boolean
```
damassi: These changes were rebased from https://github.com/artsy/reaction/pull/3272; once that's merged this can be rebased from master and will vanish from PR.

src/Apps/Artist/Routes/Consign/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { getConsignmentData } from "./Utils/getConsignmentData"
+
+interface ConsignRouteProps {
+  artist: Consign_artist
+  artworksByInternalID: routes_ArtistConsignQueryResponse["artworksByInternalID"]
+  match: Match
+  router: Router
+}
+
+const ConsignRoute: React.FC<ConsignRouteProps> = props => {
```
jonallured: This component turned out awesome, super easy to read and understand!! 💯 🏆 🎉

src/Apps/Artist/Routes/Consign/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { user } = useSystemContext()
+  const pathname = props.match.location.pathname.replace("/consign", "")
+  const artistConsignment = getConsignmentData(pathname)
+
+  // FIXME: Ungate admin-only feature when ready to launch
+  const isAuthorizedToView = userIsAdmin(user) && artistConsignment
+
+  // Redirect back to artist overview if artist not found within hand-picked data
+  useEffect(() => {
```
jonallured: Can you help me understand why the redirect is happening inside a hook? Is there something special about this context that makes the redirect work? Will the page flash some amount of the consign markup and then redirect when this hook fires?

src/Apps/Artist/routes.tsx
```diff
@@ -226,6 +232,42 @@ export const routes: RouteConfig[] = [
       },
 
       // Routes not in tabs
+
+      {
+        path: "consign",
```
jonallured: I guess I was expecting to see the redirection happening here, but maybe this is the wrong place to check the list of artist names? Like, sorta weird to load up that data here instead of inside the component, is that the idea?

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
jonallured: fwiw, I like how you addressed this @damassi 💯 

My only worry is that they'll change the column names or order or something, haha.

src/Apps/Artist/Routes/Consign/Components/SectionContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  background?: Color | "#EAE5E7"
+}
+
+export const SectionContainer: React.FC<SectionContainerProps> = ({
+  children,
+  background = "white100",
+  ...rest
+}) => {
+  // For sections with bg-colors outside of our design system
+  const bg = background === "#EAE5E7" ? "#EAE5E7" : color(background)
```
pepopowitz: A constant for `#EAE5E7` would make sense, as it's hardcoded three times.

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Passed https://docs.google.com/spreadsheets/d/1EicHOCyCozsPBkVOkQGTTUtEds7qQuacUAgLy-C2-Nk/edit#gid=0
```
pepopowitz: Do we want this URL visible to the world?

src/Apps/Artist/Routes/Consign/Utils/staticCSVToJSONData.ts
pepopowitz: Looks great!

src/Apps/Artist/routes.tsx
```diff
@@ -226,6 +232,42 @@ export const routes: RouteConfig[] = [
...
+        getComponent: () => ConsignRoute,
+        prepare: () => {
+          ConsignRoute.preload()
+        },
+        displayFullPage: true,
```
pepopowitz: I think this means that `found` will pass through any custom properties it doesn't recognize, for the route to access? Neat - I didn't know that.

src/Components/Artwork/Details.tsx
```diff
@@ -112,6 +112,10 @@ export class Details extends React.Component<Props, null> {
   }
 
   saleInfoLine() {
+    if (!this.props.showSaleLine) {
```
pepopowitz: Is this something that was missed somewhere else, or is this part of this work? I don't see `showSaleLine` anywhere else within the GitHub UI, but that doesn't mean it's not hiding it from me...

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
jonallured: I thought about that too, it does have proper permissions so that it's not actually exposing any data.

src/Apps/Artist/Routes/Consign/index.tsx
damassi: There's no reason -- and you're right! there's a small flash. I'll guard it behind an `if` and return null.

src/Apps/Artist/routes.tsx
damassi: The problem is that at this level we don't get access to `admin` properties. Once we're ready to deploy we can move the redirect up a level to here.

src/Components/Artwork/Details.tsx
damassi: This piggy backs on existing props in this UI that isn't actually being used anywhere at the moment (but did in the past). It refers to the sale state -- so `Bidding Open`, `Sold`, etc. We don't want to include that line in the artwork brick for consignments.

src/Apps/Artist/Routes/Consign/index.tsx
damassi: ahh i know why -- its because we do a server-side render pass, where this redirect isn't available (requires window.history). Will update and move to router per your suggestion when we're ready to launch and don't need to worry about guarding against `admin`.

src/Apps/Artist/routes.tsx
jonallured: Cool, sounds good! 👍

src/Apps/Artist/routes.tsx
damassi: It doesn't _totally_ work out of the box, but one can perform a [special incantation](https://github.com/damassi/reaction/blob/f7dae56f3540fa3b20f10c66ad3553bcd6f3cf63/src/Artsy/Router/Utils/findCurrentRoute.tsx#L3) and pluck it out...

src/Artsy/Router/NetworkOfflineMonitor.tsx
```diff
@@ -22,7 +22,13 @@ export const NetworkOfflineMonitor: React.FC = () => {
   }
 
   return (
-    <Box position="fixed" top={58} width="100%" style={{ opacity: 0.9 }}>
+    <Box
+      position="fixed"
+      left={0}
```
damassi: Noticed a little unrelated thing

src/Apps/Artist/Routes/Consign/Components/ArtistConsignSellArt.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+
+import { Box, Button, Sans, Serif } from "@artsy/palette"
+import { SectionContainer } from "./SectionContainer"
+
+export const ArtistConsignSellArt: React.FC = props => {
+  return (
+    <SectionContainer background="#EAE5E7">
```
jonallured: I guess you could use that `lightPurple` const that you made over in SectionContainer

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
```diff
@@ -0,0 +1,79 @@
...
+    str: string
+    uniqueVisitors: string
+    views: string
+  }
+}
+
+/**
+ * Massage data to be more usable
+ */
+export function getConsignmentData(pathname: string): ArtistConsignment {
```
jonallured: If I'm reading this right, I believe we're calling this function twice on load - once in the route file and once in the component, is that right? Any perf concerns there? Sorta thinking out loud if there would be any advantage to caching anything here. 🤔

src/Components/Artwork/Badge.tsx
```diff
@@ -14,9 +14,14 @@ const MIN_IMAGE_SIZE = 150
 
 class Badge extends React.Component<BadgeProps> {
   get stackedLayout() {
+    // During the SSR render pass we don't have access to window pixel data so
+    // default to high density screen.
+    const devicePixelRatio =
+      typeof window === "undefined" ? 2 : window.devicePixelRatio
```
jonallured: Not that it really matters, but could we do this:

```ts
const devicePixelRatio = window?.devicePixcelRatio ?? 2
```

I guess that would require the babel plugins for safe navigation and coalescing or whatever it's called?

damassi: Ah, tried this one :) `window` is as special case, where the `?` operator doesn't work as expected

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
damassi: It's probably ok for now, but we can keep that in mind.

src/Components/Artwork/Badge.tsx
jonallured: haha darn

jonallured:
Ok going to mark this one #mergeongreen to move things forward. We talked on Slack that if you have follow up items @pepopowitz then we can address them in the next step.

src/Apps/Artist/Routes/Consign/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <ArtistConsignHeader artistName={props.artist.name} />
+      <ArtistConsignRecentlySold
+        artistConsignment={artistConsignment}
+        artworksByInternalID={props.artworksByInternalID}
+      />
+      <ArtistConsignPageViews artistConsignment={artistConsignment} />
+      <ArtistConsignMarketTrends artistConsignment={artistConsignment} />
+      <ArtistConsignHowtoSell />
+      <ArtistConsignFAQ />
+      <ArtistConsignSellArt />
```
damassi: One thing to note @jonallured / @pepopowitz (if not already aware) is that due to our heavy use of relay, and the need around query names (which tie to component names by compiler architecture) we tend to namespace stuff in this way: `RouteSubRouteComponentName`. Even if there's no relay in the component, it helps for consistency reason and later adjustments.

artsyit:
:rocket: PR was released in `v25.32.0` :rocket:

PR #3263, opened 2020-03-12
the default should be country of origin if the country select is locked to EU only

oxaudo:
Realized we need to select the country where Artwork is from for the form to work properly.


artsyit:
:rocket: PR was released in `v25.27.16` :rocket:

PR #3264, opened 2020-03-12
Expand openAuthModal helpers, update NavBar triggers

eessex:
Prep work for merging #3193, adds typed helper `openAuthModal` and renames existing helper to `openAuthModalToFollowSave`. 

Also updates the docs to bring up to date with current implementation, and converts `NavBar` components to use helpers. 

<details>
  <summary>📦 Published PR as canary version: <code>25.31.1-canary.3264.54408.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.31.1-canary.3264.54408.0
  # or 
  yarn add @artsy/reaction@25.31.1-canary.3264.54408.0
  ```
</details>


docs/authentication.md
```diff
@@ -1,59 +1,83 @@
...
 ## Entry Points
+
 There are three ways a user can see authentication forms. Note that each of these are using the same underlying form infrastructure which includes validation, form submission, and analytics events.
 
 ### Modal
-Authentication modals can only be triggered with the `mediator`. The `mediator` is a global event emitter/listener in Force that is responsible for various event handling. Below is an example of how a modal might be triggered.
+
+There are helpers located in Reaction and Force to open authentication modals. Use one of Reaction's helpers if a `mediator` is available to your component.
+
+It is recommended that the helper is used rather than calling the `mediator` directly to ensure that all typing for analytics is available.
```
damassi: This is a good long-term change towards factoring out `mediator`, too, since  `mediator` depends on Backbone.

src/Utils/openAuthModal.ts
```diff
@@ -19,7 +19,13 @@ export enum AuthModalIntent {
   SaveArtwork = "save artwork",
 }
 
-export const openAuthModal = (
+export const openAuthModal = (mediator: Mediator, options: ModalOptions) => {
+  mediator.trigger("open:auth", {
+    ...options,
+  })
```
damassi: #minor - no need to create an intermediate object here:
```js
mediator.trigger("open:auth", options)
```

damassi: @eessex - As a side architectural thought, do you have any opinions about possibly plucking `mediator` from window? My thinking is:

```js
// force -- global_client_setup.tsx
import mediator from 'lib/mediator'
...
window.mediator = mediator // or some other window.varName, etc
```

Then in Reaction, inside of these helpers, we can grab mediator from there, rather than having to thread it via props throughout the app and pass it in as a parameter at each call-site. This would  then allow us to funnel all mediator references into a single location, and remove a lot of code in general. Since mediator can only ever be called on the client, this is safe.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarArtists.tsx
```diff
@@ -30,14 +30,14 @@ export class ArtworkSidebarArtists extends React.Component<ArtistsProps> {
...
-    openAuthModal(mediator, {
+    openAuthToFollowSave(mediator, {
       entity: artist,
       contextModule: Schema.ContextModule.ArtworkPage,
       intent: AuthModalIntent.FollowArtist,
     })
   }
 
-  private renderSingleArtist(artist: Artist, user, mediator) {
+  renderSingleArtist = (artist: Artist, user, mediator) => {
```
xtina-starr: Do you know why this was initially a private method?

src/Components/NavBar/Menus/MobileNavMenu.tsx
```diff
@@ -66,12 +66,12 @@ export const MobileNavMenu: React.FC<MobileNavMenuProps> = props => {
       ) : (
         <>
           <MobileLink
-            href={`/log_in?intent=signup&signupIntent=signup&trigger=click&contextModule=Header`}
+            href={`/log_in?intent=signup&trigger=click&contextModule=Header`}
```
xtina-starr: 🎉

xtina-starr:
Looks like there are some failings specs.

docs/authentication.md
eessex: Yes, one of the goals here is to standardize how we're calling `mediator`, which will make it easier to deprecate -- this is the long goal.

src/Utils/openAuthModal.ts
eessex: I think this would be helpful as a next step, we are calling on `mediator` via reaction with a variety of strategies. Once we've got the helper in place it will be easier to manage this change as a separate work item.

eessex:
The failing specs are strange to me, these all pass locally. Will need to pair with someone more familiar with the relay/test setup to get these 🍏

damassi:
Yayy 🍏

artsyit:
:rocket: PR was released in `v25.31.1` :rocket:

PR #3265, opened 2020-03-13
Attempt to fix Flickity error

damassi:
Still trying to fix the unreproducible  
```
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node
```
error around flickity as seen in this [sentry](https://sentry.io/organizations/artsynet/issues/1497212332/?project=28316&query=removeChild&statsPeriod=24h) ticket. 

This [stack overflow](https://stackoverflow.com/a/54882016/1038901) comment made me think we might be using a fragment somewhere in the carousel around images, and indeed we were. 

🤞


artsyit:
:rocket: PR was released in `v25.27.17` :rocket:

PR #3266, opened 2020-03-13
[Clientside Nav] Remove mobile loader test

damassi:
We have enough data now where we don't need to continue showing the loader in the middle of the screen on mobile. This removes that weirdness.

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -24,7 +24,6 @@ import { Footer } from "Components/v2/Footer"
 import { RecentlyViewedQueryRenderer as RecentlyViewed } from "Components/v2/RecentlyViewed"
 import { RouterContext } from "found"
 import { TrackingProp } from "react-tracking"
-import { get } from "Utils/get"
 import { Media } from "Utils/Responsive"
 
 export interface Props {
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -122,9 +121,9 @@ export class ArtworkApp extends React.Component<Props> {
   }
 
   renderArtists() {
-    const artists = get(this.props, p => p.artwork.artists)
+    const artists = this.props.artwork?.artists
 
-    if (!artists.length) {
+    if (!artists?.length) {
```
damassi: Fixes https://sentry.io/organizations/artsynet/issues/1529777601/?project=28316&query=is%3Aunresolved&statsPeriod=24h

zephraph: Nice. I was unsure of what `artist?.length` would be if `artist` was undefined and it looks like the answer to that is... well.. `undefined`.

👍 Love that syntax though.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -122,9 +121,9 @@ export class ArtworkApp extends React.Component<Props> {
   }
 
   renderArtists() {
-    const artists = get(this.props, p => p.artwork.artists)
+    const artists = this.props.artwork?.artists
```
zephraph: 🙌

zephraph:
Looks good! Does this mean that the a/b test was a success?

damassi:
The A/B test data is all over the place unfortunately. It's hard to know, because there's been so many pushes to master that have fixed little things here and there.

artsyit:
:rocket: PR was released in `v25.29.1` :rocket:

PR #3268, opened 2020-03-13
Sync schema

zephraph:
Do the thing!

I ran `yarn sync-schema` which is needed after an update to metaphysics has been  shipped.


artsyit:
:rocket: PR was released in `v25.29.3` :rocket:

PR #3269, opened 2020-03-14
[Carousel] Fix alignment in image

damassi:
Fixes an alignment bug that [originated here](https://github.com/artsy/reaction/pull/3265) that only appears during SSR render / when JS is disabled.


artsyit:
:rocket: PR was released in `v25.29.4` :rocket:

PR #3270, opened 2020-03-14
Add initial UI for years created filters on auction results

zephraph:
Fixes [PURCHASE-1810](https://artsyproduct.atlassian.net/browse/PURCHASE-1810)

Adds the **Year Created** filter to the auction results tab on the artist page. 

image

## TODO

- [x] Actually pull the date range from metaphysics
- [x] Wire up new filter to the refetch mutation
- [x] Write tests

<details>
  <summary>📦 Published PR as canary version: <code>25.33.0-canary.3270.54783.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.33.0-canary.3270.54783.0
  # or 
  yarn add @artsy/reaction@25.33.0-canary.3270.54783.0
  ```
</details>


zephraph:
Still need to add some tests, but if folks want to start taking a look I'd appreciate it 🙏

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -200,8 +201,21 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
...
+    const birthYear = DateTime.fromJSDate(new Date(birthday)).year
     return (
-      <AuctionResultsFilterContextProvider>
+      <AuctionResultsFilterContextProvider
+        filters={{
+          earliestCreatedYear:
+            birthday && birthYear > startAt ? birthYear : startAt,
+          latestCreatedYear: endAt,
+        }}
+      >
```
zephraph: @jpotts244, I'm doing a _bit_ of massaging here just to ensure we're not showing a work with a year created that's earlier than the artist's birth year. Once we're able to clean up the data a bit in diffusion to ensure this isn't really much of an issue we can nix this.

src/Apps/Artist/Routes/AuctionResults/AuctionResultsFilterContext.tsx
```diff
@@ -6,6 +6,13 @@ export interface AuctionResultsFilters {
   sizes?: string[]
   page?: number
   sort?: string
+  createdAfterYear?: string
+  createdBeforeYear?: string
+
+  /** Used to get the overall earliest created year for all lots of given artist */
+  readonly earliestCreatedYear?: number
+  /** Used to get the overall latest created year for all lots of given artist */
+  readonly latestCreatedYear?: number
```
zephraph: These are only set initially via the context provider and aren't actually intended to be changed later. They're just holding the _total_ range of years created. Needed to populate the selector w/ dates.

src/Apps/Artist/Routes/AuctionResults/AuctionResultsFilterContext.tsx
```diff
@@ -144,13 +166,31 @@ const AuctionResultsFilterReducer = (
...
+        }
+      }
+
+      if (name === "createdAfterYear" && value) {
+        if (!state.createdBeforeYear) {
+          filterState.createdBeforeYear = `${state.latestCreatedYear}`
+        } else if (parseInt(state.createdBeforeYear) < parseInt(value)) {
+          filterState.createdBeforeYear = value
+        }
+      }
```
zephraph: I added some extra filtering logic to test the cases when the filters are set in a weird way. I.e. when the earliest year is set to a time later than the latest year.

src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/YearCreated.tsx
```diff
@@ -0,0 +1,57 @@
...
+    createdAfterYear,
+    createdBeforeYear,
+  } = filterContext?.filters
+  if (
+    typeof earliestCreatedYear === undefined ||
+    typeof latestCreatedYear === undefined
+  ) {
+    console.error("Couldn't display year created filter due to missing data")
+    return null
+  }
```
zephraph: If, for whatever reason, metaphysics isn't giving us back either of the created year bounds, just don't render the filter.

src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/YearCreated.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const fullDateRange = useMemo(
+    () => buildDateRange(earliestCreatedYear, latestCreatedYear),
+    [earliestCreatedYear, latestCreatedYear]
+  )
```
zephraph: A little optimization to ensure we only calculate this once for each unique value, despite how many times it re-renders.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
jpotts244: 🙏 thanks for the coverage here! hoping the additional constraints in the new ArtistMatch job will help with cases like these

zephraph: It may be worth creating a task to actually do this logic in diffusion (on route that actually returns the range).

src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/YearCreated.tsx
jpotts244: Good call!

src/Apps/Artist/Routes/AuctionResults/__tests__/AuctionResults.test.tsx
```diff
@@ -255,6 +259,10 @@ describe("AuctionResults", () => {
                   sort: "DATE_DESC",
                   organizations: [],
                   categories: [],
+                  createdAfterYear: "1881",
```
jpotts244: The different data types for years (string vs integer) is a little challenging for my brain to parse.

Wondering if it is worth discussing modifying the response from diffusion to return a string? I think gravity uses strings to persist years as well.

jpotts244:
the UI looks really sharp! 🎉

src/Apps/Artist/Routes/AuctionResults/__tests__/AuctionResults.test.tsx
zephraph: I agree that it's a bit weird. I'd much prefer to lean into having the types really match what the value is. So in this case, we know that year should always be an integer. For all the things that use year, I'd much prefer it to be a number/int/etc. 

I could probably look to consolidate this to numbers across the different usages. Let me take a peek at that.

zephraph:
I found a bug where when you tap reset on mobile it throws and error. Looking into that.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
jpotts244: I dig it -- might be fun to write it together if you have some time tomorrow. Would like to get your ideas on endpoint design. We could probably add some fun caching to that endpoint too!

zephraph:
Hi friends!

I'm gonna ship this 🚀

artsyit:
:rocket: PR was released in `v25.33.0` :rocket:

PR #3271, opened 2020-03-15
[Clientside Nav] Add more safety around network failures with fallback UI

damassi:
While browsing Sentry I've noticed [occasional errors](https://sentry.io/organizations/artsynet/issues/1564753432/?project=28316&query=is%3Aunresolved+chunk&statsPeriod=24h) appear that look like 

> Loading chunk 24 failed. (error: https://d1s2w0upia4e9w.cloudfront.net/assets/Routes-Works.22820.adcce79ed4f87e3fc6ed.js)

This typically happens when a user (often on a mobile phone) is navigating between routes and their network connection goes out, which leads to a split JS bundle failing to load. 

We've already [added retry support](https://github.com/artsy/force/pull/5186) on the Force side of things, which will re-try the request up to five times. However, that doesn't handle absolute failures, nor does it gracefully degrade with UI that can lead to a recovery.

According to React's [code splitting docs](https://reactjs.org/docs/code-splitting.html), the best way to handle this is within an Error boundary: 

> If the other module fails to load (for example, due to network failure), it will trigger an error. You can handle these errors to show a nice user experience and manage recovery with Error Boundaries. Once you’ve created your Error Boundary, you can use it anywhere above your lazy components to display an error state when there’s a network error.

Since we already have an error boundary at the root of our app we can listen for specific errors here and handle gracefully, prompting the user to reload the page via our standard `ErrorModal`. (In the future we can further flesh this out for other kinds of global errors.)

retry

### DIFF
--- a/src/Artsy/Router/Boot.tsx
+++ b/src/Artsy/Router/Boot.tsx
@@ -57,18 +57,18 @@ export class Boot extends React.Component<BootProps> {
     }
 
     return (
-      <ErrorBoundary>
+      <Theme>
         <HeadProvider headTags={headTags}>
           <StateProvider>
             <Artsy.SystemContextProvider {...contextProps}>
-              <MediaContextProvider onlyMatch={props.onlyMatchMediaQueries}>
-                <ResponsiveProvider
-                  mediaQueries={themeProps.mediaQueries}
-                  initialMatchingMediaQueries={
-                    props.onlyMatchMediaQueries as any
-                  }
-                >
-                  <Theme>
+              <ErrorBoundary>
+                <MediaContextProvider onlyMatch={props.onlyMatchMediaQueries}>
+                  <ResponsiveProvider
+                    mediaQueries={themeProps.mediaQueries}
+                    initialMatchingMediaQueries={
+                      props.onlyMatchMediaQueries as any
+                    }
+                  >
                     <Grid fluid>
                       <GlobalStyles suppressMultiMountWarning />
                       {children}
@@ -76,13 +76,13 @@ export class Boot extends React.Component<BootProps> {
                         <BreakpointVisualizer />
                       )}
                     </Grid>
-                  </Theme>
-                </ResponsiveProvider>
-              </MediaContextProvider>
+                  </ResponsiveProvider>
+                </MediaContextProvider>
+              </ErrorBoundary>
             </Artsy.SystemContextProvider>
           </StateProvider>
         </HeadProvider>
-      </ErrorBoundary>
+      </Theme>
     )
   }
 }
--- a/src/Artsy/Router/__tests__/Boot.test.tsx
+++ b/src/Artsy/Router/__tests__/Boot.test.tsx
@@ -1,6 +1,5 @@
 import { SystemContextConsumer } from "Artsy"
 import { Boot } from "Artsy/Router"
-import { ErrorBoundary } from "Components/ErrorBoundary"
 import { mount } from "enzyme"
 import React from "react"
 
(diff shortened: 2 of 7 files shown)


src/Artsy/Router/Boot.tsx
```diff
@@ -57,32 +57,30 @@ export class Boot extends React.Component<BootProps> {
...
+        <StateProvider>
+          <Artsy.SystemContextProvider {...contextProps}>
+            <MediaContextProvider onlyMatch={props.onlyMatchMediaQueries}>
+              <ResponsiveProvider
+                mediaQueries={themeProps.mediaQueries}
+                initialMatchingMediaQueries={props.onlyMatchMediaQueries as any}
+              >
+                <Theme>
+                  <Grid fluid>
+                    <ErrorBoundary>
```
damassi: Moved the `ErrorBoundary` down so that it starts at our component tree and can inherit styles

src/Artsy/Router/__tests__/Boot.test.tsx
```diff
@@ -56,20 +55,4 @@ describe("Boot", () => {
   it("injects Grid", () => {
     expect(mount(<Boot {...bootProps} />).find("Grid").length).toEqual(1)
   })
-
-  it("catches errors with componentDidCatch", () => {
```
damassi: Moved this test to `ErrorBoundary.test`

src/Artsy/Router/__tests__/buildClientApp.test.tsx
```diff
@@ -6,6 +6,16 @@ import { mount } from "enzyme"
...
 
+jest.mock("Components/NavBar", () => ({
+  NavBar: () => <div />,
+}))
+
+jest.mock("react-relay", () => ({
+  ReactRelayContext: {
+    Provider: ({ children }) => children,
+  },
+}))
```
damassi: The use of `getDerivedStateFromError` revealed that we were masking a test down below with an unrelated error

src/Components/Payment/__tests__/PaymentForm.test.tsx
```diff
@@ -8,6 +8,7 @@ import { creatingCreditCardFailed } from "Apps/Order/Routes/__fixtures__/Mutatio
 jest.mock("react-relay", () => ({
   commitMutation: jest.fn(),
   createFragmentContainer: component => component,
+  createRefetchContainer: component => component,
```
damassi: Another subtle test bug that the error boundary caught

src/Artsy/Router/Boot.tsx
```diff
@@ -57,32 +57,32 @@ export class Boot extends React.Component<BootProps> {
     }
 
     return (
-      <ErrorBoundary>
+      <Theme>
```
zephraph: Was it necessary to move the existing error boundary in order to get this to work? Are there any tradeoffs in having it as the outermost wrapper vs having it inside the system context provider?

src/Components/ErrorBoundary.tsx
```diff
@@ -18,7 +28,45 @@ export class ErrorBoundary extends React.Component<Props> {
...
+      return {
+        asyncChunkLoadError: true,
+      }
+    }
+  }
+
   render() {
+    if (this.state.asyncChunkLoadError) {
+      return <ErrorModalWithReload show={this.state.asyncChunkLoadError} />
+    }
```
zephraph: Are there other cases we should be checking for? It's not clear to me what this component was really doing previously aside from logging things out.

zephraph:
Oh, and one other note. We should clue in design on some of these edge case experiences.

src/Artsy/Router/Boot.tsx
damassi: This was just so that I could use Theme file inside the error boundary, which shows the nav. 

> Are there any tradeoffs in having it as the outermost wrapper vs having it inside the system context provider

The only trade off i can see is that an error thrown from within the render function of the context provider proper -- not its children -- wouldn't get caught by it, but there isn't too much logic there. We might want to think about it in the future if we decide to expand functionality at that level.

src/Components/ErrorBoundary.tsx
damassi: Yup, was just used for logging for the time being. I think we can experiment a bit here!

damassi:
> We should clue in design on some of these edge case experiences

Yup, I definitely planned on doing this once we were done with the A/B test.

artsyit:
:rocket: PR was released in `v25.30.0` :rocket:

PR #3272, opened 2020-03-17
[ErrorBoundary] Add generic error handler

damassi:
This adds a generic error handler for catch-alls runtime errors. I realized we needed this when working in storybooks and hitting a syntax error and thinking things through: if `getDerivedStateFromError` is present, then it will rerender the component for a fallback, but if theres an error, it will call `getDerivedStateFromError`, which will rerender -- etc. So this adds more safety for all other cases, which should be in place anyway.


artsyit:
:rocket: PR was released in `v25.31.0` :rocket:

PR #3275, opened 2020-03-18
Fix admin check on staging

damassi:
Noticed that there was some weirdness around isAdmin on staging, so moving the check into the useEffect hook so only evaluated on client.


artsyit:
:rocket: PR was released in `v25.32.2` :rocket:

PR #3279, opened 2020-03-18
[Artist/Consign] Add tests 

damassi:
Adds some tests and completes functionality for the new Artist/Consign pages. 

Still waiting on:
- Individual artwork stats to add to bricks
- Need to discuss FAQ page with Nicole, and update link.


src/Apps/Artist/Routes/Consign/__tests__/ConsignRoute.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+              router={router}
+            />
+          </SystemContextProvider>
+        )
+      },
+      query: graphql`
+        query ConsignRoute_Test_Query(
+          $artistID: String!
+          $recentlySoldArtworkIDs: [String]!
+        ) @raw_response_type {
```
damassi: See [@raw_response_type](https://mrtnzlml.com/docs/relay/directives#raw_response_type) for full explanation, but the TLDR is: 

If you have GraphQL query `A` that contains a nested fragment query `B`, importing the generated type for `A` would only contain a `$ref` to `B` due to how relay implements data masking. By using `@raw_response_type` it will spit out the whole deal, which makes it ideal for typing raw JSON data output.

src/Apps/__tests__/Fixtures/Artist/Routes/ConsignRouteFixture.ts
```diff
@@ -0,0 +1,331 @@
+import { ConsignRoute_Test_QueryRawResponse } from "__generated__/ConsignRoute_Test_Query.graphql"
+import { ArtistConsignment } from "Apps/Artist/Routes/Consign/Utils/getConsignmentData"
+
+export const ConsignRouteFixture: ConsignRoute_Test_QueryRawResponse = {
```
damassi: Where did this data come from? 

data

jonallured: Whoa, cool technique!!! 😍

src/Apps/Artist/Routes/Consign/index.tsx
```diff
@@ -25,11 +25,11 @@ interface ConsignRouteProps {
   router: Router
 }
 
-const ConsignRoute: React.FC<ConsignRouteProps> = props => {
-  const { artist } = props
-  const { user } = useSystemContext()
-  const pathname = props.match.location.pathname.replace("/consign", "")
+export const ConsignRoute: React.FC<ConsignRouteProps> = props => {
+  const { artist, artworksByInternalID, match, router } = props
+  const pathname = match.location.pathname.replace("/consign", "")
```
jonallured:
```suggestion
  const artistPagePath = match.location.pathname.replace("/consign", "")
```

src/Apps/Artist/Routes/Consign/index.tsx
```diff
@@ -38,7 +38,7 @@ const ConsignRoute: React.FC<ConsignRouteProps> = props => {
     const isAuthorizedToView = Boolean(userIsAdmin(user) && artistConsignment)
 
     if (!isAuthorizedToView) {
-      props.router.replace(pathname)
+      router.replace(pathname)
```
jonallured:
```suggestion
      router.replace(artistPagePath)
```

src/Apps/Artist/Routes/Consign/Utils/__tests__/getConsignmentData.test.tsx
```diff
@@ -0,0 +1,14 @@
+import { artistConsignmentFixture } from "Apps/__tests__/Fixtures/Artist/Routes/ConsignRouteFixture"
+import { getConsignmentData } from "../getConsignmentData"
+
+describe("getConsignmentData", () => {
```
jonallured: Good call to get some coverage here. 👍

src/Apps/Artist/Routes/Consign/__tests__/ConsignRoute.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    const {
+      artistConsignmentFixture,
+    } = require("Apps/__tests__/Fixtures/Artist/Routes/ConsignRouteFixture")
+
+    return artistConsignmentFixture
+  },
+}))
+
+describe("ConsignRoute", () => {
+  const getWrapper = async (
```
jonallured: This thing breaks my brain, haha. I think what I'm reading here is a mocking of relay? It allows us to get a wrapper in the enzyme sense, right? But it's not really an enzyme wrapper? Sorta confusing and lots of abstraction that I'm trying to understand but I guess in the end if you can use it like the wrappers I'm used to, then the details probably don't matter.

Which I guess makes me wonder if this construct is more of a distraction in this test file than it's worth. Have you considered pulling something like this out into a separate location? Just a thought!

damassi: This is the typical Relay pattern that we use all over Reaction, and it does return an enzyme wrapper! Lots of goodies in here: https://github.com/artsy/reaction/tree/master/src/DevTools

artsyit:
:rocket: PR was released in `v25.34.1` :rocket:

PR #3283, opened 2020-03-18
Update remaining mediator calls

eessex:
Moves all remaining auth modal triggers to use `openAuthModal` helper. Follow up to #3264


artsyit:
:rocket: PR was released in `v25.32.8` :rocket:

PR #3238, opened 2020-03-05
[FX-1702] new mobile nav menu

xtina-starr:
Addresses: [FX-1702](https://artsyproduct.atlassian.net/browse/FX-1702)

This adds the new mobile nav menu that introduces nested sub menus. 

mobile nav menu

PR #3289, opened 2020-03-20
Mobile nav menu analytics

xtina-starr:
Addresses: [FX-1829](https://artsyproduct.atlassian.net/browse/FX-1829)

This PR adds analytics tracking to mobile nav menu.

Screen Shot 2020-03-19 at 10 02 12 PM

Screen Shot 2020-03-19 at 10 02 34 PM

Screen Shot 2020-03-19 at 10 04 02 PM

### DIFF
--- a/src/Components/NavBar/Menus/MobileLink.tsx
+++ b/src/Components/NavBar/Menus/MobileLink.tsx
@@ -1,9 +1,12 @@
+import { AnalyticsSchema } from "Artsy"
+import { useTracking } from "Artsy/Analytics"
 import React, { useState } from "react"
 import styled from "styled-components"
 
 import { Box, color, Link, Sans } from "@artsy/palette"
 
 interface MobileLinkProps {
+  contextModule?: string
   children: React.ReactNode
   href?: string
   onClick?: (event?: React.MouseEvent<HTMLElement>) => void
@@ -12,10 +15,22 @@ interface MobileLinkProps {
 export const MobileLink: React.FC<MobileLinkProps> = ({
   href,
   children,
+  contextModule,
   ...props
 }) => {
   const [isPressed, setPressed] = useState(false)
   const bg = isPressed ? "black5" : "white100"
+  const { trackEvent } = useTracking()
+
+  const handleClickTracking = (linkHref: string) => {
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.Click,
+      context_module: contextModule,
+      flow: "Header",
+      subject: children.toString(),
+      destination_path: linkHref,
+    })
+  }
 
   return (
     <MobileLinkContainer
@@ -28,7 +43,11 @@ export const MobileLink: React.FC<MobileLinkProps> = ({
     >
       <Box>
         {href ? (
-          <Link href={href} underlineBehavior="none">
+          <Link
+            href={href}
+            underlineBehavior="none"
+            onClick={() => handleClickTracking(href)}
+          >
             <Sans size={["5t", "6"]} color={color("black60")}>
               {children}
             </Sans>
--- a/src/Components/NavBar/Menus/MobileLink.tsx
+++ b/src/Components/NavBar/Menus/MobileLink.tsx
@@ -6,7 +6,7 @@ import styled from "styled-components"
 import { Box, color, Link, Sans } from "@artsy/palette"
 
 interface MobileLinkProps {
-  contextModule?: string
+  contextModule?: any
   children: React.ReactNode
   href?: string
   onClick?: (event?: React.MouseEvent<HTMLElement>) => void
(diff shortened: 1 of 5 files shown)


src/Components/NavBar/Menus/__tests__/NewMobileNavMenu.test.tsx
```diff
@@ -56,4 +70,66 @@ describe("MobileNavMenu", () => {
       expect(linkText).not.toContain("Works for you")
     })
   })
+
+  describe("Analytics tracking", () => {
```
xtina-starr: I'd like to add a few more tests for when a user clicks items from the Artworks or Artists dropdown, but I'm finding  it difficult to setup tests for two or more layers deep. I'm open to suggestions and would love to pair on this.

src/Components/NavBar/Menus/NewMobileNavMenu.tsx
```diff
@@ -93,32 +155,56 @@ const Menu = ({ isOpen, title, links, showBacknav = true }) => {
...
+        height="10px"
+        width="10px"
+        top="7px"
+        left="5px"
+      />
     </Box>
   )
 }
 
+const getContextModule = () => {
```
sweir27: Instead of basing the `contextModule` off of the path, could we instead pass it down through the `MobileSubmenuLink` and access it directly?

src/Components/NavBar/Menus/NewMobileNavMenu.tsx
```diff
@@ -33,13 +45,63 @@ export const NewMobileNavMenu: React.FC<Props> = props => {
...
+            </MobileLink>
+            <MobileLink
+              href="/institutions"
+              onClick={() => {
+                handleClickTracking("/institutions", "Museums")
+              }}
+            >
+              Museums
+            </MobileLink>
+            <MobileLink
```
sweir27: Instead of having an explicit `onClick` call on each of these instances of `<MobileLink>`, could we try putting the click/tracking handler on the `<MobileLink>` component directly?

It looks like most of the custom properties (`destination_path`, `subject`) should be present there anyways, and where not perhaps we can pass that through from a parent component?

If possible to do so, it could make this code a bit more concise (and also make it less likely for someone adding a new link to forget to add tracking in the future!).

src/Components/NavBar/Menus/__tests__/NewMobileNavMenu.test.tsx
pepopowitz: My personal feeling is that testing the analytics calls of these links could go one of two ways: 

* Verify that links are _in general_ hooked up to analytics, which the tests here cover. 
* Verify that _all_ links are comprehensively hooked up to the correct analytics. 

I don't think adding tests of second-level links is necessary unless you're looking to go comprehensive. 

If you _are_ looking to be comprehensive, I think there are some things you could do to these tests to make it easier to write them all:

* Add a helper for finding a specific link by text. This would make it easier to target each individual link/sub-link without being tied to order/sequence of elements.
* Use [`jest.each`](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout) to write one test to cover lots of similar test cases. It would save a good amount of copy-paste.

I'd love to pair on it! Feel free to put some time on my calendar.

src/Components/NavBar/Menus/NewMobileNavMenu.tsx
xtina-starr: We talked about this IRL. This is related to the [refactor](https://github.com/artsy/reaction/pull/3238#discussion_r388608576) I made that switched  from passing down `path`, `pop`, and `push` props down through several children props to accesses it where is was to be used to accessing it from the `NavigatorContextProvider` via a local hook. Determining what the context module should be based on the `path` is consistent with that pattern so the conclusion was to keep this as is.

src/Components/NavBar/Menus/MobileLink.tsx
```diff
@@ -12,10 +15,23 @@ interface MobileLinkProps {
...
 }) => {
   const [isPressed, setPressed] = useState(false)
   const bg = isPressed ? "black5" : "white100"
+  const { trackEvent } = useTracking()
+  const text = children.toString()
+
+  const handleClickTracking = (linkHref: string) => {
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.Click,
+      context_module: contextModule,
```
sweir27: This is great! Nice refactor.

One last question. It _looks like_ the vast majority of the `contextModule`s that are passed in have the same value of `AnalyticsSchema.ContextModule.Header`. If that's the case, can we give this a default value and make the prop optional?

src/Components/NavBar/Menus/MobileLink.tsx
```diff
@@ -15,7 +15,7 @@ interface MobileLinkProps {
 export const MobileLink: React.FC<MobileLinkProps> = ({
   href,
   children,
-  contextModule,
+  contextModule = AnalyticsSchema.ContextModule.Header,
```
xtina-starr: Set a default value for `contextModule`

artsyit:
:rocket: PR was released in `v25.36.5` :rocket:

PR #3291, opened 2020-03-20
Prepopulate email input from URL query string on forgot password page

dblandin:
Links to the reset password page will often include a query param including the target user's email address. This commit uses that value to prepopulate the reset password form, simplifying the experience.

Co-authored-by: @anandaroop 

jira :lock: : https://artsyproduct.atlassian.net/browse/TRUST-92

### DIFF
--- a/src/Components/Authentication/FormSwitcher.tsx
+++ b/src/Components/Authentication/FormSwitcher.tsx
@@ -132,6 +132,19 @@ export class FormSwitcher extends React.Component<FormSwitcherProps, State> {
     )
   }
 
+  getEmailValue = (): string => {
+    const { values } = this.props
+    const isClient = typeof window !== "undefined"
+    let email
+
+    if (isClient) {
+      const searchQuery = window.location.search.slice(1)
+      email = qs.parse(searchQuery).email as string
+    }
+
+    return email || values.email || ""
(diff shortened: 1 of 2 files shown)


src/Components/Authentication/FormSwitcher.tsx
```diff
@@ -132,6 +132,19 @@ export class FormSwitcher extends React.Component<FormSwitcherProps, State> {
...
   }
 
+  getEmailValue = (): string => {
+    const { values } = this.props
+    const isClient = typeof window !== "undefined"
+    let email
+
+    if (isClient) {
+      const searchQuery = window.location.search.slice(1)
+      email = qs.parse(searchQuery).email as string
```
joeyAghion: Presumably this does proper URL-decoding?

artsyit:
:rocket: PR was released in `v25.34.0` :rocket:

src/Components/Authentication/FormSwitcher.tsx
dblandin: Confirmed that this works with an encoded email string like `user+with-extra@example.com` / `user%2Bwith-extra%40example.com`

Screenshot from 2020-03-20 14-32-01

PR #3292, opened 2020-03-20
[Consignments] Add Artist/Consign SEO/meta tags

damassi:
Does a couple things: 
- Adds some meta tags to the new `artist/consign` page
- Updates titles and subtitles to `h1` / `h2` for SEO

### DIFF
--- a/src/Apps/Artist/Routes/Consign/Components/__tests__/ArtistConsignMeta.test.tsx
+++ b/src/Apps/Artist/Routes/Consign/Components/__tests__/ArtistConsignMeta.test.tsx
@@ -1,0 +1,99 @@
+import { MockBoot } from "DevTools"
+import { mount } from "enzyme"
+import React from "react"
+import { ArtistConsignMeta } from "../ArtistConsignMeta"
+
+jest.mock("Utils/getENV", () => ({
+  getENV: () => "https://artsy.net",
+}))
+
+describe("ArtistConsignMeta", () => {
+  const props = {
+    artistName: "Alex Katz",
+    artistHref: "/artist/alex-katz",
+    imageURL: "path/to/image.jpg",
+  }
+  const getWrapper = (passedProps = {}) => {
+    return mount(
+      <MockBoot>
+        <ArtistConsignMeta {...props} {...passedProps} />
+      </MockBoot>
+    )
+  }
+
+  it("outputs correct title tags", () => {
+    const wrapper = getWrapper()
+    expect(wrapper.find("Title").debug()).toContain(props.artistName) // use `debug` to assert on component tree, vs html tree
+    expect(
+      wrapper
+        .find("Meta")
+        .findWhere(c => c.props().property === "og:title")
+        .debug()
+    ).toContain(props.artistName)
+    expect(
+      wrapper
+        .find("Meta")
+        .findWhere(c => c.props().property === "og:title")
+        .debug()
+    ).toContain(props.artistName)
+  })
+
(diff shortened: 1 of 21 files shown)


src/Apps/Artist/Routes/Consign/Components/__tests__/ArtistConsignMeta.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return mount(
+      <MockBoot>
+        <ArtistConsignMeta {...props} {...passedProps} />
+      </MockBoot>
+    )
+  }
+
+  it("outputs correct title tags", () => {
+    const wrapper = getWrapper()
+    expect(wrapper.find("Title").debug()).toContain(props.artistName) // use `debug` to assert on component tree, vs html tree
```
pepopowitz: I don't think this is super important, but since it's a text element, you could do `wrapper.find("Title").text()` to assert against a more direct/less noisy value.

src/Utils/getENV.tsx
```diff
@@ -1,6 +1,6 @@
-import { data as sd } from "sharify"
+import { data as sd, GlobalData } from "sharify"
 
-export function getENV(ENV_VAR) {
+export function getENV(ENV_VAR: keyof GlobalData) {
```
pepopowitz: 💪

typings/sharify.d.ts
```diff
@@ -22,6 +22,7 @@ declare module "sharify" {
       readonly ARTIST_COLLECTIONS_RAIL_IDS: string[]
       readonly CLIENT_NAVIGATION_V4: "experiment" | "control" // TODO: Remove after A/B test.
       readonly CMS_URL: string
+      readonly CDN_URL: string
```
pepopowitz: Is this related to this work, or did it come from somewhere else?

src/Apps/Artist/Routes/Consign/Components/__tests__/ArtistConsignMeta.test.tsx
damassi: So with the title tag, it actually does some magic in the background to inject a `<title>` meta tag into the header. Since we're not rendering the full page, and thus don't have access to stuff in the `<header>`, we can use `debug` to assert on the react tree itself. This provides sufficient assurances.

typings/sharify.d.ts
damassi: I added `keyof GlobalData` to our `getENV` helper for better autocomplete. (`getENV` is a util for plucking isomorphic ENV vars -- `process.env` on the server, and `sd` on the client.) After doing that it revealed a type error which pointed to this missing value.

artsyit:
:rocket: PR was released in `v25.34.4` :rocket:

src/Apps/Artist/Routes/Consign/Components/__tests__/ArtistConsignMeta.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+describe("ArtistConsignMeta", () => {
+  const props = {
+    artistName: "Alex Katz",
+    artistHref: "/artist/alex-katz",
+    imageURL: "path/to/image.jpg",
+  }
+  const getWrapper = (passedProps = {}) => {
+    return mount(
+      <MockBoot>
```
damassi: Useful for testing stuff that normally boots inside of apps (all React context-related stuff) -- also needed to test responsive layouts around the `<Media>` component.

PR #3282, opened 2020-03-18
User can navigate to IDV FAQ from the IDV landing page

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-930

## Todos

 * [x] Merge and deploy https://github.com/artsy/force/pull/5240

### Desktop

AUCT-911-desktop

### Mobile

AUCT-911-mobile

PR #3294, opened 2020-03-21
Show a pass/fail message for a completed identity verification

erikdstock:
This PR changes the IdentityVerificationApp (at `/identity-verification/:id`) to show different pages if the IDV is complete (that is has a state of 'passed' or 'failed'), blocking someone from beginning new flows at this point.


src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box, Button, Link, Sans, Serif } from "@artsy/palette"
+import React from "react"
+
+export const CompleteFailed = () => {
```
damassi: `React.FC`

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  return (
+    <AppContainer>
+      <HeadTitle>Artsy | ID Verification</HeadTitle>
+      <Meta
+        name="viewport"
+        content="width=device-width, initial-scale=1, maximum-scale=5 viewport-fit=cover"
+      />
+      <InnerComponent />}
```
damassi: Stray `}`

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box, Button, Link, Sans, Serif } from "@artsy/palette"
+import React from "react"
+
+export const CompleteFailed: React.FC = () => {
+  return (
+    <Box px={[2, 3]} mb={6} mt={4}>
+      <Box
+        mx={["auto"]}
```
yuki24:
```suggestion
        mx="auto"
```

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
```diff
@@ -0,0 +1,43 @@
...
+        width={["100%", "80%"]}
+        maxWidth={"400px"}
```
yuki24:
```suggestion
        maxWidth="400px"
```

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
```diff
@@ -0,0 +1,43 @@
...
+        textAlign="center"
+      >
+        <Serif size="6" color="black100">
```
yuki24:
```suggestion
        <Serif size="6">
```

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box, Button, Link, Sans, Serif } from "@artsy/palette"
+import React from "react"
+
+export const CompleteFailed: React.FC = () => {
+  return (
+    <Box px={[2, 3]} mb={6} mt={4}>
+      <Box
```
yuki24: I'm not sure if you have to nest `<Box>`es here. You could find [a refactoring example here](https://github.com/artsy/reaction/pull/3282/files#diff-b6e1258870cda8e1ddc88e37a1932561L129-R123).

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const trackClickedContinueToVerification = () => {
+    trackEvent({
+      context_page_owner_id: identityVerification.internalID,
+      action_type: Schema.ActionType.ClickedContinueToIdVerification,
+      context_page: Schema.PageName.IdentityVerificationPage,
+    })
+  }
+
+  async function startIdentityVerification() {
```
yuki24: This function doesn't seem to use `await`. We should stick with using `async`/`await` together or not using them at all.

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    )
+  }
+
+  return (
+    <AppContainer>
+      <HeadTitle>Artsy | ID Verification</HeadTitle>
+      <Meta
+        name="viewport"
+        content="width=device-width, initial-scale=1, maximum-scale=5 viewport-fit=cover"
+      />
```
yuki24: I don't think we need this `<Meta>` tag.

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { RelayProp } from "react-relay"
+import { useTracking } from "react-tracking"
+import { commitMutation, graphql } from "relay-runtime"
+import createLogger from "Utils/logger"
+
+const logger = createLogger("StartIdentityVerification.tsx")
+
+interface Props {
+  relay: RelayProp
+  identityVerification: {
```
yuki24:
```suggestion
  identityVerification: IdentityVerificationApp_me['identityVerification']
```

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
```diff
@@ -0,0 +1,168 @@
...
+    mutation
+      .then(handleMutationSuccess)
+      .catch(handleMutationError)
+      .finally(() => {
+        setRequesting(false)
+      })
+  }
+
+  const handleMutationSuccess = (identityVerificationFlowUrl: string) => {
+    location.assign(identityVerificationFlowUrl)
```
yuki24: We should probably use `props.router` or `props.location` rather than relying on the global one.

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
dleve123: When are the `{}` needed? When this is evaluating an expression?

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
```diff
@@ -0,0 +1,43 @@
...
+            .
+          </Sans>
+        </Box>
+        <a href="mailto:verification@artsy.net">
+          <Button
+            block
+            width={["100%", 335]}
+            mt={4}
+            onClick={() => {
+              // trackClickedContinueToVerification()
```
dleve123: Should we delete this commented out code?

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <ErrorModal
+        show={showErrorModal}
+        contactEmail="verification@artsy.net"
+        onClose={() => {
+          setShowErrorModal(false)
+        }}
+      />
+      <Box px={[2, 3]} mb={6} mt={4}>
+        <Box
+          mx={["auto"]}
```
dleve123: Similar to @yuki24's [comment above](https://github.com/artsy/reaction/pull/3294/files#r396483819), I wonder if `mx="auto"` is sufficient here.

src/Apps/IdentityVerification/__tests__/IdentityVerificationApp.test.tsx
```diff
@@ -134,13 +173,13 @@ describe("IdentityVerification route", () => {
...
         env.mutations.mockNetworkFailureOnce()
 
-        await page.clickStartVerification()
-
-        const errorModal = expectOne(page.find(ErrorModal))
-        expect(errorModal.props().show).toBe(true)
-        expect(page.text()).toContain(
-          "Something went wrong. Please try again or contact verification@artsy.net."
-        )
+        page.clickStartVerification().then(() => {
```
dleve123: For my own education – why the change the `await` to `then` here? Does this have to do with the Jest error that you tracked down? Is it just for consistency purposes?

yuki24:
I think this PR needs a bit more work to be ready as it requires a sizable change (global variable leak, typing, nested components, etc). I'd be also happy to pair on this if the comments here are overwhelming.

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
erikdstock: Strings work by default, but for other js types including numbers you need `{}`. In some cases they get left behind when a responsive value for palette/styled-system (in the form of an array) gets removed - with the single string they are no longer necessary.

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
erikdstock: stylistic choice - I'm fine with doing that but it felt more react-y to just pass the props i need since this component is not a fragment container and knows nothing about relay (except for having a prop named `relay` of course). I'm fine either way, happy to move it.

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
damassi: We shouldn't be using `{}` unless interpolatred JS is actually being used.

src/Apps/IdentityVerification/IdentityVerificationApp/CompleteFailed.tsx
damassi: Saw that this was resolvrd, but @yuki24's requested changes wern't applied.

erikdstock: yes, i have them all in a working branch, sorry - will change pr to WIP

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <InnerComponent />
+    </AppContainer>
+  )
+}
+
+export const IdentityVerificationAppFragmentContainer = createFragmentContainer(
```
yuki24: Now that I look at this component again I think we should co-locate this fragment and `StartIdentityVerification` since this `IdentityVerificationApp` component doesn't directly depend on any GraphQL data.

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
erikdstock: I updated the props to be only `relayEnvironment` and `identityVerificationId`.

src/Apps/IdentityVerification/__tests__/IdentityVerificationApp.test.tsx
erikdstock: Yes, i think this was while i was trying to debug errors in our tests. As it turns out this didn't fix the issue so i'll fix now.

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
erikdstock: It actually does depend on the state and that the `me` owns the `identityVerification` (their ids match). So i could make another fragment container but the amount of threading seems relatively small for the added complexity there.

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
erikdstock: (and now `router`)

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
```diff
@@ -0,0 +1,160 @@
...
+            onClick={() => {
+              setRequesting(true)
+              trackClickedContinueToVerification()
+              startIdentityVerification()
+            }}
+          >
+            Continue to verification
+          </Button>
+        </Box>
+      </Box>
```
yuki24: I think we should bring back what is currently in the `<IdentityVerificationApp>` component. It seems like the old code was copy-&-pasted here, and it isn't just un-refactored code but also missing a link to FAQ.

https://github.com/yuki24/reaction/blob/bfffe3297b704f10378fa7d7831bac05427659da/src/Apps/IdentityVerification/IdentityVerificationApp.tsx#L117-L169

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+interface Props {
+  me: IdentityVerificationApp_me
+  relay: RelayProp
+  router: Router
+}
+
+const IdentityVerificationApp: React.FC<Props> = ({ me, relay, router }) => {
+  const { identityVerification } = me
+
+  if (!identityVerification || identityVerification.userID !== me.internalID) {
```
yuki24: You don't have to change this as part of this PR, but It's odd that you have to do this on the front-end when the `identityVerification` field is a nested field of `me`. The front-end should only check the presence of `identityVerification` and the back-end should take care of hiding `identityVerification`s that belongs to someone else.

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const IdentityVerificationAppFragmentContainer = createFragmentContainer(
+  IdentityVerificationApp,
+  {
+    me: graphql`
+      fragment IdentityVerificationApp_me on Me
+        @argumentDefinitions(id: { type: "String!" }) {
+        internalID
+        name
```
yuki24: We don't seems to be using it:

```suggestion
```

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
erikdstock: I agree with you technically - but we had many long conversations and about this approach last month and I don't think it is going to change; a `me.identityVerification` is an idv that the user has access to, not necessarily that _belongs to_ the user, so we do need this check on the client.

src/Apps/IdentityVerification/IdentityVerificationApp/StartIdentityVerification.tsx
erikdstock: Good catch! probably lost on rebase. I'll try restoring this version but personally don't like the mixture of inner components.

erikdstock: replaced :+1:

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -113,60 +119,66 @@ const IdentityVerificationApp: React.FC<Props> = ({ me, relay }) => {
...
+              Artsy identity verification
+            </Serif>
+
+            <Sans size="4" mt="2" weight="medium">
+              You’ll need
+            </Sans>
+            <Sans size="4">
+              • A camera on your phone or computer
+              <br />
+              respont • Your government ID
```
yuki24: Maybe this might have slipped in by accident?
```suggestion
             • Your government ID
```

src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
dleve123: If I am remembering context correctly, it's worth noting that this is _just_ to protect against the case of an admin accessing this page for an IDV that isn't theirs. At an API layer, we don't allow regular user A to query for the identity verifications for regular user B. In that case, the identity_verification would be `null` (the LHS of this expression).

cc @yuki24

artsyit:
:rocket: PR was released in `v25.35.3` :rocket:

PR #3297, opened 2020-03-22
Update NetworkTimeout language

damassi:
Updates the language on the network timeout to be something a bit more specific:
"An error occurred" => "Network timeout"

Also removes the faux nav from the ErrorBoundary state as it conflicts with a few different scenarios (app sections that don't have a nav, etc).


artsyit:
:rocket: PR was released in `v25.34.6` :rocket:

PR #3301, opened 2020-03-23
Adds social_embed section type to ad insertion logic

ashleyjelks:
Social Embed sections types were missing from the logic that determines if an ad should be rendered in a standard article section, this PR addresses that.

After fix:
Screen Shot 2020-03-23 at 12 05 21 PM

Before fix:
Screen Shot 2020-03-23 at 12 07 45 PM


sweir27:
👍 looks good, thanks for the screenshots!

artsyit:
:rocket: PR was released in `v25.35.1` :rocket:

PR #3307, opened 2020-03-23
Redirect logged in users to the works for sale tab

zephraph:
Fixes [PURCHASE-1689](https://artsyproduct.atlassian.net/browse/PURCHASE-1689).

When a logged in user navigates to the artist page we want it to default to the `/works-for-sale` tab.

The logic is as follows

1. If a user is logged in `/` redirects for `/works-for-sale`
2. If a user is logged in `/overview` opens the overview tab (which also links off to `/overview`)
3. If a user is **not** logged in `/` leads to the overview tab
4. If a user is **not** logged in `/overview` redirects to `/` and therefore the overview tab

The not logged in case has the _exact_ same behavior as production so SEO won't be impacted. 

<details>
  <summary>📦 Published PR as canary version: <code>25.42.0-canary.3307.56198.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.42.0-canary.3307.56198.0
  # or 
  yarn add @artsy/reaction@25.42.0-canary.3307.56198.0
  ```
</details>


src/Apps/Artist/routes.tsx
```diff
@@ -75,13 +75,40 @@ if (typeof window !== "undefined") {
   ArtistApp.preload()
 }
 
+const routesMiddleware = (originalRoutes, user): RouteConfig[] => {
+  if (user) {
```
damassi: Can we add a comment describing whats happening here and why?

src/Apps/Artist/routes.tsx
```diff
@@ -75,13 +75,40 @@ if (typeof window !== "undefined") {
...
+  return originalRoutes
+}
 // FIXME:
 // * `render` functions requires casting
 // * `Redirect` needs to be casted, as it’s not compatible with `RouteConfig`
 export const routes: RouteConfig[] = [
   {
     path: "/artist/:artistID",
     getComponent: () => ArtistApp,
+    routesMiddleware,
```
damassi: Maybe `redirectIfLoggedIn`?

src/Apps/Artist/routes.tsx
```diff
@@ -122,8 +123,27 @@ export const routes: RouteConfig[] = [
     children: [
       // Routes in tabs
       {
-        path: "/",
+        path: ":regexParam(\\overview)?",
         getComponent: () => OverviewRoute,
+        render: ({ Component, props, match }) => {
```
damassi: @zephraph - one other solution would be to do 

```
me { 
  id
}
```
in the query itself, as non-null indicates logged-in-ness

zephraph: That's true, yeah. 

I discovered this when looking through `buildClientApp`

https://github.com/artsy/reaction/blob/0535652b04b727f7149c6f24e6b1963097f30e1b/src/Artsy/Router/buildClientApp.tsx#L86

It is, so far as I can tell, driven by the system context. It's just passed through a different level. Is it worth adding more to get the user if it's already available otherwise?

damassi: I _think_ that user is driven by headers, so its zero cost from a MP query perspective, and would be a dead-simple data-driven fix, versus leveraging router-specific stuff. That said i'm not opposed either way, we just need to be sure to pass it in on `buildServerApp` as well.

zephraph: It's already in `buildServerApp`

https://github.com/artsy/reaction/blob/0535652b04b727f7149c6f24e6b1963097f30e1b/src/Artsy/Router/buildServerApp.tsx#L82

I was able to trace it down to this commit: https://github.com/artsy/reaction/commit/36d07a381710d7ab7e0493b7f803ad935a6dc792

zephraph: I'm looking for other usages, but I'm not seeing anything. If that was the only thing that used it maybe we should just remove it and fallback to the query?

zephraph: Yeah, confirmed no other usages. Maybe we can just nix it and move to the query based approach.

damassi: Its nice to query for what one wants :)

zephraph:
I'm going to follow up with a test before merging

src/Apps/Artist/routes.tsx
zephraph: Done

zephraph:
Okay @damassi, I think this is good now. We consolidated the routing logic at the top level (where the other route logic already was). Makes us only need to go to one place for it.

src/Apps/Artist/routes.tsx
```diff
@@ -87,42 +87,56 @@ export const routes: RouteConfig[] = [
...
+        `${artist.slug}/works-for-sale`
+      )
 
-        if (!canShowOverview && !alreadyAtWorksForSalePath) {
-          throw new RedirectException(`/artist/${artist.slug}/works-for-sale`)
-        }
+      const canShowOverview = showArtistInsights || hasArtistContent
 
-        return <Component {...props} />
+      if (user && !pathname.includes("/overview")) {
```
damassi, line 122: It would be nice if this was added verbatim as a multiline comment just above these conditionals: 

Screen Shot 2020-04-01 at 11 39 53 AM

src/Apps/Artist/routes.tsx
```diff
@@ -87,42 +87,56 @@ export const routes: RouteConfig[] = [
...
+        throw new RedirectException(`/artist/${artist.slug}/works-for-sale`)
+      }
+
+      return <Component {...props} />
     },
     children: [
       // Routes in tabs
       {
-        path: "/",
+        path: ":regexParam(\\overview)?",
```
damassi, line 139: is `:regexParam` a special flag in found?

zephraph: It is, yeah. This is required to ensure that the overview route will correctly render for both `/` and `/overview` depending on the case.

src/Apps/Artist/routes.tsx
zephraph: Done 👍

src/Apps/Artist/routes.tsx
damassi: That's cool, i was wondering if Found supported this kind of thing

artsyit:

PR #3308, opened 2020-03-24
[Consignments] Artist/Consign QA

damassi:
Addresses feedback in https://artsyproduct.atlassian.net/browse/CSGN-93

A couple post-launch follow-up items as noted in JIRA comments:
- Ensure all Recently Sold artworks are public in CMS
- Move Recently Sold items into a proper carousel

### DIFF
--- a/src/Apps/Artist/Routes/Consign/index.tsx
+++ b/src/Apps/Artist/Routes/Consign/index.tsx
@@ -15,19 +15,18 @@ import { ArtistConsignPageViews } from "./Components/ArtistConsignPageViews"
 import { ArtistConsignRecentlySold } from "./Components/ArtistConsignRecentlySold"
 import { ArtistConsignSellArt } from "./Components/ArtistConsignSellArt"
 
-import { getConsignmentData } from "./Utils/getConsignmentData"
+import { ArtistConsignment } from "./Utils/getConsignmentData"
 
 interface ConsignRouteProps {
   artist: Consign_artist
   artworksByInternalID: Consign_artworksByInternalID
+  artistConsignment: ArtistConsignment
   match: Match
   router: Router
 }
 
 export const ConsignRoute: React.FC<ConsignRouteProps> = props => {
-  const { artist, artworksByInternalID, match } = props
-  const artistPathName = match.location.pathname.replace("/consign", "")
-  const artistConsignment = getConsignmentData(artistPathName)
+  const { artist, artistConsignment, artworksByInternalID, match } = props
   const imageURL = artworksByInternalID[0]?.image.imageURL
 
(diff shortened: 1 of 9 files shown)


src/Apps/Artist/routes.tsx
```diff
@@ -240,6 +240,22 @@ export const routes: RouteConfig[] = [
...
+          const hasConsignPage = !!getConsignmentData(artistPathName)
+
+          // Check to see if artist is within top 20 consignable artists, and if
+          // not, redirect to home route.
+          if (!hasConsignPage) {
+            throw new RedirectException(artistPathName)
+          } else {
+            return <Component {...props} match={match} />
+          }
+        },
```
jonallured: What about something like this:

```ts
const noConsignmentData = !getConsignmentData(artistPathName)

if (noConsignmentData) {
  throw new RedirectException(artistPathName)
} else {
  return <Component {...props} match={match} />
}
```

Not a big deal, just suggesting that the name of the boolean condition reflect the case where we throw so that we don't have to recognize the exclamation points as much.

On the other hand, if we're grabbing this data here, we might as well pass it down to the component, no?

```ts
const consignmentData = getConsignmentData(artistPathName)

if (!consignmentData) {
  throw new RedirectException(artistPathName)
} else {
  return <Component {...props} match={match} consignmentData={consignmentData} />
}
```

This feels a little more clear to me, what do you think? If we have to check this data at the router level anyway seems like we should just pass it down rather than having it computed again.

src/Apps/Artist/Routes/Consign/index.tsx
```diff
@@ -27,22 +25,9 @@ interface ConsignRouteProps {
...
-  // FIXME: Move this check to the router level when we're ready launch
-  useEffect(() => {
-    // FIXME: Ungate admin-only feature when ready to launch
-    const isAuthorizedToView = Boolean(userIsAdmin(user) && artistConsignment)
-
-    if (!isAuthorizedToView) {
-      router.replace(artistPathName)
-    }
-  }, [])
-
```
jonallured: Love to delete code. 🤗

src/Apps/Artist/routes.tsx
pepopowitz: :+1: to the idea of passing the data into the component from here.

damassi: Yup! both good suggestions, will do

damassi:
@jonallured  

> And just so I'm clear, merging this PR removes the admin guard and releases these pages, right?

Yup

artsyit:
:rocket: PR was released in `v25.36.1` :rocket:

PR #3312, opened 2020-03-25
Switch props.router.push back to window.locaion.assign for idv flow

erikdstock:
This is due to a failure to direct the user to the identity verification
flow, probably because that `push()` function is intended for navigating
programatically within the found router on the same domain.

[Bug report on slack](https://artsy.slack.com/archives/CA3D02SDN/p1585152180005500) 🔒

### DIFF
--- a/src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
+++ b/src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
@@ -5,7 +5,6 @@ import { AppContainer } from "Apps/Components/AppContainer"
 import * as Schema from "Artsy/Analytics/Schema"
 import { ErrorPage } from "Components/ErrorPage"
 import { ErrorModal } from "Components/Modal/ErrorModal"
-import { Router } from "found"
 import React, { useState } from "react"
 import { Title as HeadTitle } from "react-head"
 import {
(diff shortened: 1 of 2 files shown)


src/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
```diff
@@ -108,7 +106,7 @@ const IdentityVerificationApp: React.FC<Props> = ({ me, relay, router }) => {
   }
 
   const handleMutationSuccess = (identityVerificationFlowUrl: string) => {
-    router.push(identityVerificationFlowUrl)
+    window.location.assign(identityVerificationFlowUrl)
```
damassi: What is `identityVerificationFlowUrl`?

dleve123: it's an external URL to jumio- https://artsy.netverify.com/adsfasdf

dleve123:
Is there a way that we can better test this flow within reaction? Right now, we're replacing one mock expectation with another.

erikdstock:
>Is there a way that we can better test this flow within reaction? Right now, we're replacing one mock expectation with another.

I don't think so, but in this case we are now using a browser api on `window` that is as good as i'm aware of for being able to programmatically assert about navigation outside of a real networked browser environment. I feel like this is good and have now verified it on my machine, good to merge @dleve123 ?

artsyit:
:rocket: PR was released in `v25.36.6` :rocket:

PR #3313, opened 2020-03-25
the Address Form component should respect country selection for Euro only shipping

oxaudo:
The issue here is that when the Form loads technically there is nothing selected. But once the user starts interacting with country select - we need to let this interaction manage the state. The fact that we just set selected to whatever was passes as origin country broke this state interaction.

Hope that fixes this issue: https://artsyproduct.atlassian.net/browse/GALL-2703

Also @lilyfromseattle I think moved the AddressForm to the general components from Order. So I reshuffled the story a bit as well.

### DIFF
--- a/src/Components/AddressForm.tsx
+++ b/src/Components/AddressForm.tsx
@@ -120,7 +120,8 @@ export class AddressForm extends React.Component<
             </Serif>
             <CountrySelect
               selected={
-                lockCountryToOrigin || lockCountriesToEU
+                lockCountryToOrigin ||
+                (lockCountriesToEU && !this.state.address.country)
                   ? this.props.shippingCountry
                   : this.state.address.country
               }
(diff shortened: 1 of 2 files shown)


src/Components/AddressForm.tsx
```diff
@@ -120,19 +120,22 @@ export class AddressForm extends React.Component<
             </Serif>
             <CountrySelect
               selected={
-                lockCountryToOrigin || lockCountriesToEU
+                lockCountryToOrigin ||
+                (lockCountriesToEU && !this.state.address.country)
```
ansor4: #minor Would this be cleaner if we pulled it out into another `const` above? Something like `useCountryFromState` vs `useCountryFromProps`?

artsyit:
:rocket: PR was released in `v25.36.7` :rocket:

PR #3314, opened 2020-03-25
Video meta bug

ashleyjelks:
This PR fixes a bug in the article meta tags causing a the text 0 to render at the top of video editorial series pages. 

https://artsyproduct.atlassian.net/browse/FX-1875

The Bug:
Screen Shot 2020-03-25 at 12 27 43 PM

The Fix:
Screen Shot 2020-03-25 at 5 15 51 PM

### DIFF
--- a/src/Components/Publishing/ArticleMeta.tsx
+++ b/src/Components/Publishing/ArticleMeta.tsx
@@ -95,7 +95,7 @@ export const ArticleMeta: React.SFC<{
         </>
       )}
 
-      {keywords.length && (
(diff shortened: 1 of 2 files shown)


eessex:
This fix looks good to me but I'm also surprised by the bug because we automatically add keywords to the article on publish here - https://github.com/artsy/positron/blob/master/src/api/apps/articles/model/save.coffee#L93

artsyit:
:rocket: PR was released in `v25.36.8` :rocket:

PR #3316, opened 2020-03-25
Ensure year created check doesn't error on nulls

zephraph:
The original logic check here was checking for an `undefined` type but those results could be `null` too. 

I updated the logic to check for a `number` which is the _exact_ type we're expecting.


jpotts244:
🔥 🔥

artsyit:
:rocket: PR was released in `v25.36.10` :rocket:

PR #3317, opened 2020-03-26
[Consignments] Add ArtistConsignButton

damassi:
Addresses 
- https://artsyproduct.atlassian.net/browse/CSGN-90
- https://artsyproduct.atlassian.net/browse/CSGN-92

consignbutton

Adds a new responsive `ArtistConsignButton` that covers two cases: "Top 20", and everything else.

We're not implementing "Target Supply" version yet because we don't yet know how to get data from RedShift over to Metaphysics.

### DIFF
--- a/src/Apps/Artist/Components/ArtistConsignButton.tsx
+++ b/src/Apps/Artist/Components/ArtistConsignButton.tsx
@@ -15,12 +15,31 @@ import {
   Image,
   Sans,
 } from "@artsy/palette"
+import { AnalyticsSchema, useTracking } from "Artsy"
 
 export interface ArtistConsignButtonProps {
   artist: ArtistConsignButton_artist
 }
 
-export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props => {
+export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = ({
+  artist,
+}) => {
+  const tracking = useTracking()
+
+  const trackGetStartedClick = ({ destinationPath }) => {
+    tracking.trackEvent({
+      context_page: AnalyticsSchema.PageName.ArtistPage,
+      context_page_owner_id: artist.internalID,
+      context_page_owner_slug: artist.slug,
+      context_page_owner_type: AnalyticsSchema.OwnerType.Artist,
+      context_module: AnalyticsSchema.ContextModule.ArtistConsignment,
+      subject: AnalyticsSchema.Subject.GetStarted,
+      destination_path: destinationPath,
+    })
+  }
+
+  const props = { artist, trackGetStartedClick }
+
   return (
     <>
       <Media at="xs">
@@ -33,8 +52,13 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props => 
   )
 }
 
-export const ArtistConsignButtonLarge: React.FC<ArtistConsignButtonProps> = props => {
-  const { artistHasOwnConsignRoute, imageURL, headline, consignLink } = getData(
+interface Tracking {
+  trackGetStartedClick: (props: { destinationPath: string }) => void
+}
+
+export const ArtistConsignButtonLarge: React.FC<ArtistConsignButtonProps &
+  Tracking> = props => {
+  const { artistHasOwnConsignRoute, imageURL, headline, consignURL } = getData(
     props
   )
 
@@ -57,7 +81,14 @@ export const ArtistConsignButtonLarge: React.FC<ArtistConsignButtonProps> = prop
           </Flex>
         </Flex>
         <Box>
-          <RouterLink to={consignLink}>
+          <RouterLink
+            to={consignURL}
+            onClick={() => {
+              props.trackGetStartedClick({
+                destinationPath: consignURL,
+              })
+            }}
+          >
             <Button variant="secondaryGray">Get started</Button>
           </RouterLink>
         </Box>
@@ -66,8 +97,9 @@ export const ArtistConsignButtonLarge: React.FC<ArtistConsignButtonProps> = prop
   )
 }
 
-export const ArtistConsignButtonSmall: React.FC<ArtistConsignButtonProps> = props => {
-  const { artistHasOwnConsignRoute, imageURL, headline, consignLink } = getData(
+export const ArtistConsignButtonSmall: React.FC<ArtistConsignButtonProps &
+  Tracking> = props => {
+  const { artistHasOwnConsignRoute, imageURL, headline, consignURL } = getData(
     props
   )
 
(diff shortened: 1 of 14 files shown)


src/Apps/Artist/Components/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props => {
+  return (
+    <>
+      <Media at="xs">
+        <ArtistConsignButtonSmall {...props} />
+      </Media>
+      <Media greaterThan="xs">
+        <ArtistConsignButtonLarge {...props} />
+      </Media>
```
damassi: The layouts diverge subtly, enough to warrant clarity over duplication concerns.

src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -412,8 +412,8 @@ const renderAuctionHighlight = artist => {
 }
 
 const renderRepresentationStatus = artist => {
-  const { highlights } = artist
-  const { partnersConnection } = highlights
+  const { artistHightlights } = artist
```
damassi: These `artistHighlight` changes get our VSCode error panel back to a healthy state /cc @sweir27

src/Apps/Artist/Components/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </Flex>
+    </BorderBox>
+  )
+}
+
+function getData(props) {
+  const {
+    artist: { href, name, image },
+  } = props
+  const isTop20 = Boolean(getConsignmentData(href))
```
damassi: Around the time we we address the Redshift data issue I'll move this CSV json functionality into Convection or MP GraphQL layer where we can retrieve it via API request.

sweir27: May I suggest slightly different naming here? The `top20` seems like it will become outdated/confusing soon (I've already heard talks of how these artists aren't really "top"). Can we refer more to the feature at-hand? `hasRichConsignContent` (that's not great, but hopefully you get the idea 😅)

src/Apps/Artist/Components/ArtistHeader.tsx
sweir27: Thank you! Sorry for dropping the ball on that one.

src/Apps/Artist/Components/ArtistConsignButton.tsx
damassi: Sure thing - this was referenced in design docs so I'll raise in consign channel 👍

src/Apps/Artist/Components/__tests__/ArtistConsignButton.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+  // TODO: Functionality doesn't yet exist
+  xdescribe("Target Supply Button", () => {
+    it("renders desktop version", async () => {
+      //
+    })
+
+    it("renders mobile version", async () => {
+      //
+    })
+  })
```
jonallured:
```suggestion
```

This seems unnecessary to me - what if we change our minds and don't do this button?

damassi: That makes sense 👍

src/Apps/Artist/Components/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+    <BorderBox p={1} width="100%">
+      <Flex alignItems="center" width="100%" justifyContent="space-between">
+        <Flex>
+          {isTop20 && <Image src={imageURL} width={50} height={50} />}
+          <Flex flexDirection="column" justifyContent="center" pl={1}>
+            <Sans size="3t" weight="medium">
+              {headline}
+            </Sans>
+            <Box position="relative">
+              <Sans size="3t" color={color("black60")}>
```
pepopowitz: Something I've always wondered - some reaction components use `color={color("black60")}`, and some use `color="black60"`. Is there a reason to favor one over the other?

src/Apps/Artist/Components/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+}
+
+export const ArtistConsignButtonLarge: React.FC<ArtistConsignButtonProps> = props => {
+  const { isTop20, imageURL, headline, consignLink } = getData(props)
+
+  return (
+    <BorderBox p={1} width="100%">
+      <Flex alignItems="center" width="100%" justifyContent="space-between">
+        <Flex>
+          {isTop20 && <Image src={imageURL} width={50} height={50} />}
```
pepopowitz:
```suggestion
          {isTop20 && imageURL && <Image src={imageURL} width={50} height={50} />}
```

The value of `imageURL` is obtained in `getData` with optional chaining - should this also guard against non-existent images?

src/Apps/Artist/Components/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+  )
+}
+
+export const ArtistConsignButtonSmall: React.FC<ArtistConsignButtonProps> = props => {
+  const { isTop20, imageURL, headline, consignLink } = getData(props)
+
+  return (
+    <BorderBox maxWidth={335} p={1}>
+      <Flex alignItems="center">
+        {isTop20 && <Image src={imageURL} width={75} height={66} />}
```
pepopowitz:
```suggestion
        {isTop20 && imageURL && <Image src={imageURL} width={75} height={66} />}
```

Same comment as above, regarding potentially non-existent imageURLs.

src/Apps/Artist/Components/__tests__/ArtistConsignButton.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      )
+    })
+
+    it("renders mobile version", async () => {
+      const wrapper = await getWrapper({ breakpoint: "xs", response })
+      expect(wrapper.find("Image").length).toEqual(1)
+      expect(wrapper.text()).toContain("Sell your Alex Katz")
+      expect(wrapper.find("RouterLink").html()).toContain(
+        `href="/artist/alex-katz/consign"`
+      )
```
pepopowitz: These tests are rendering with different breakpoints, but because the assertions are identical, there's not much safety that we're rendering `xs` differently from `>xs`. I think this is because there are *aren't* any significant differences between the breakpoints, other than layout implementation details. Am I correct in my understanding?

damassi: There's two different components triggered by breakpoints, so we want to ensure that both work as expected.

src/Apps/Artist/Components/ArtistConsignButton.tsx
damassi: Nice catch 👍

src/Apps/Artist/Components/ArtistConsignButton.tsx
damassi: I guess in this case we don't _need_ `color` because this component will only ever live inside of our app, which is contained by boot, but color makes it failproof should it be moved, etc in the future, while also giving us intellisense.

artsyit:
:rocket: PR was released in `v25.36.13` :rocket:

PR #3321, opened 2020-03-26
[Consignments] Add "Sold for" price data

damassi:
This PR adds new 'Sold for $' data: 

Screen Shot 2020-03-26 at 3 55 21 PM


src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
```diff
@@ -7,7 +7,11 @@ import { flatten, groupBy } from "lodash"
 import { staticCSVToJSONData } from "./staticCSVToJSONData"
 
 export interface ArtistConsignment {
-  artworks: object[]
+  artworks: Array<{
+    internalID: string
+    realizedPrice: string
```
damassi: This is purely for code intellisense, in the sense that we reference aliased CSV column fields (which aren't too pretty)

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
```diff
@@ -37,8 +41,9 @@ export function getConsignmentData(pathname: string): ArtistConsignment {
       const recentlySoldArtworkIDs = flatten(
         artworks.map((artwork: any) => {
           const id = artwork["Artwork ids (recently sold) (comma separated)"]
-          // FIXME: Add proper type
-          artwork.artworkID = id
+          const realizedPrice = artwork["Realized Price (in dollars)"]
+          artwork.internalID = id
+          artwork.realizedPrice = realizedPrice
```
damassi: Related to comment above

src/Components/Artwork/FillwidthItem.tsx
```diff
@@ -141,7 +141,7 @@ export class FillwidthItemContainer extends React.Component<
             />
           </ImageLink>
 
-          <Badge artwork={artwork} width={this.imageWidth} />
+          {showExtended && <Badge artwork={artwork} width={this.imageWidth} />}
```
damassi: Don't show the weird buy now badges on artwork bricks:

Screen Shot 2020-03-26 at 3 59 02 PM

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
jonallured: I'm curious to know how you were thinking about types here. I see a note about a proper type, but then it looks like you opted for an unnamed type - any particular reason for that? No change necessary, just wondering!

src/Apps/Artist/Routes/Consign/Components/ArtistConsignRecentlySold.tsx
```diff
@@ -33,15 +33,23 @@ export const ArtistConsignRecentlySold: React.FC<ArtistConsignRecentlySoldProps>
...
+                <Box p={2} key={key} textAlign="left">
                   <FillwidthItem
                     artwork={artwork}
                     targetHeight={150}
                     imageHeight={150}
                     showExtended={false}
                     width={150 * artwork.image.aspectRatio}
                   />
+                  <Sans size="2" weight="medium">
+                    Sold for {artworkData.realizedPrice}
```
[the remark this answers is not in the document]
damassi: This is curated spreadsheet data so there should always be data for this field.

src/Apps/Artist/Routes/Consign/Utils/getConsignmentData.ts
damassi: The idea here is to not mess with the original data too much, but also dont want to access data with these really long keys, so I added an alias on the object and then added some types to indicate useage. The unnamed types permit everything to remain how it is and not clutter up the interface with noise, and the additions that I made were made because they're used in code, and we want intellisense. 

Basically, just striking a balance between the CSV / excel table and the minimum amount of code required to parse / use it in a friendly way.

artsyit:
:rocket: PR was released in `v25.36.14` :rocket:

PR #3322, opened 2020-03-27
[Consignments] Add images to Artist/Consign header

damassi:
Rebased from https://github.com/artsy/reaction/pull/3321. That needs to be merged first. 

Addresses https://artsyproduct.atlassian.net/browse/CSGN-79 

This adds images to the main header. There was a lot of nuance to this, and its not 100% [to the comp](https://www.figma.com/file/MFZBkqq9A5E7msdDYUkyJK/Consign-Top-20-Landing-Page?node-id=390%3A1214&viewport=-9442%2C2458%2C1) but its about as close as its going to be given

1) Different aspect ratios across `n` number of artist images
1) Responsiveness, including the inability to create a "normal" border
1) Server-side rendering, where we don't have the luxury or performing client-side math

I added a bit of a nice touch by randomly selecting images based on the recently sold, so users returning to the page will see something different each time. 

images

Also randomizes the Recently sold brick order, giving it a nice living feel.


src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -1,19 +1,41 @@
...
+  artworksByInternalID: Consign_artworksByInternalID
 }
 
 export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props => {
   const tracking = useTracking()
 
   return (
     <SectionContainer background={LightPurpleColor}>
+      <Media greaterThan="xs">
+        {classNames => {
```
damassi: Media component creates its own wrapping component. Since we're doing some custom layout stuff, one can use the renderProps pattern to apply classes to your own component.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -46,3 +68,122 @@ export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props =>
...
+    // Layout requires that we only use vertically oriented images
+    const widthExceedsHeight = ({ image }) => {
+      return image.resized.width > image.resized.height
+    }
+
+    if (widthExceedsHeight(left)) {
+      return getRandomImages()
+    }
+    if (widthExceedsHeight(right)) {
+      return getRandomImages()
```
damassi: We're doing some recursion here, but the array size is small. I'm not worried.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
jonallured: I see what's happening here, that class names are being passed from the generated wrapper down into the box below, but I can't quite figure out why? Is it because you want some styling on both containers? 🤔

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -46,3 +68,137 @@ export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props =>
...
+ *    container and then applying a "fit" to a background image.
+ * 4) Therefore, conventional borders cannot be applied, and a CSS filter is
+ *    required. This is based upon pixel data, and border faked.
+ * 5) The pixel data comes from a second image, sitting below the first, with
+ *    its brightness turned to -100.
+ * 5) It's not perfect, and the layout needs to also account for variable
+ *    aspect ratios.
+ * 6) The positioning CSS is also tricky, so please forgive...
+ */
+const HeaderImages = (props: {
```
jonallured: Nice docs! Given the complexity here, I might suggest extracting this to a separate file.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -46,3 +68,137 @@ export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props =>
...
+  artworksByInternalID: Consign_artworksByInternalID
+}) => {
+  const getRandomImages = () => {
```
jonallured: This function seems worth getting some specs around if for no other reason than to help future us remember what it's doing in case we have to change it! 😝

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -46,3 +68,137 @@ export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props =>
...
+      return getRandomImages()
+    }
+
+    return [left, right]
+  }
+
+  const [leftImage, rightImage] = getRandomImages()
+
+  return (
+    <HeaderImageContainer>
```
jonallured: I think it would be cool to extract and name some concepts here. What about something like this:

```ts
const HeaderImages = props => {
  // .. stuff ..
  return (
    <HeaderImageContainer>
      <Flex width="100%" justifyContent="space-between">
        <LeftImage image={leftImage} />
        <RightImage image={leftImage} />
      </Flex>
    </HeaderImageContainer>
  )
}

const LeftImage = props => {
  // .. stuff ..
  return (
    <LeftImageWrapper>
      <BorderImage image={image} />
      <ShrunkImage image={image} />
    </LeftImageWrapper>
  )
}

const RightImage = props => {
  // .. stuff ..
  return (
    <RightImageWrapper>
      <BorderImage image={image} />
      <ShrunkImage image={image} />
    </RightImageWrapper>
  )
}

const BorderImage = props => {
  // .. stuff ..
  return (
    <ResponsiveImage
      src={src}
      width={width}
      style={{
        transformOrigin: "center",
        filter: "invert(100%) brightness(-100%)",
        backgroundPosition: "left",
      }}
    />
  )
}

const ShrunkImage = props => {
  // .. stuff ..
  return (
    <ResponsiveImage
      src={src}
      width={width}
      style={{
        position: "absolute",
        top: 0,
        left: -4,
        transformOrigin: "center",
        backgroundPosition: "left",
        transform: "scaleX(.97) scaleY(.96)",
      }}
    />
  )
}
```

Hopefully I copy/pasted that mostly correct! 😝 So yeah, my approach here is to pull out things and give them names so that I can see the layers and how they all compose together. I think something like this would make this code easier to change. What do you think?

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
pepopowitz: If I'm understanding correctly, it will recursively call itself until it chooses two images randomly that are both vertically oriented. Is it possible that we could have an artist who has no vertically oriented images (or only one), in which case this would never actually find two usable images?

I also think iterating through a randomly sorted collection of the images until two good ones are found would be easier to understand and safer than using recursion.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -46,3 +68,137 @@ export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props =>
...
+          <ResponsiveImage
+            src={leftImage.image.resized.url}
+            width={leftImage.image.resized.width}
+            style={{
+              position: "absolute",
+              top: 0,
+              left: -4,
+              transformOrigin: "center",
+              backgroundPosition: "left",
+              transform: "scaleX(.97) scaleY(.96",
```
pepopowitz:
```suggestion
              transform: "scaleX(.97) scaleY(.96)",
```

I'm not sure what happens without a terminating parentheses, but I noticed one missing

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
```diff
@@ -46,3 +68,137 @@ export const ArtistConsignHeader: React.FC<ArtistConsignHeaderProps> = props =>
...
+            And the image
+          */}
+          <ResponsiveImage
+            src={rightImage.image.resized.url}
+            width={rightImage.image.resized.width}
+            style={{
+              position: "absolute",
+              top: 0,
+              transformOrigin: "center",
+              transform: "scaleX(.95) scaleY(.96",
```
pepopowitz:
```suggestion
              transform: "scaleX(.95) scaleY(.96)",
```

src/Apps/Artist/Routes/Consign/__tests__/ConsignRoute.test.tsx
```diff
@@ -93,6 +96,13 @@ describe("ConsignRoute", () => {
       expect(wrapper.find("ArtistConsignHeader").text()).toContain("Alex Katz")
     })
 
+    it("displays two images in header", async () => {
+      const wrapper = await getWrapper()
+      expect(
+        wrapper.find("ArtistConsignHeader").find("ResponsiveImage").length
+      ).toEqual(4) // actually 2, but second set of images creates border
```
pepopowitz: 👍 appreciate the comment

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
damassi: the `<Media>` container doesn't accept a styles or attributes, but has certain layout properties. I need a subset of those layout properties for the media query aspect, but need to change others because we're doing some position absolute magic. So I grab the classNames, apply them to my own container, then augment my container with desired props.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
damassi: I can add an iteration guard, but I think for the 5 or so lines here, recursion is more readable and much simpler, so gonna stick with this.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
damassi: Ah, good idea 👍

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
damassi: Great suggestion 👍 Didn't follow _exactly_, but kept the spirit of it. Much easier to read now.

src/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader.tsx
damassi: I've added an iteration cap of 100 while also try / catching the entire function for further safety should any other problems occur. If caught, the images don't render.

damassi:
I've addressed all of the comments here so going to go ahead and merge on green to get this out. We can follow up if there are any other comments. 

Thanks for the great feedback @jonallured and @pepopowitz 👍

artsyit:
:rocket: PR was released in `v25.38.1` :rocket:

PR #3323, opened 2020-03-27
gitignore node diagnostic reports

erikdstock:
#minor

This PR adds a line to gitignore node diagnostic reports. It is copied from https://gitignore.io/api/node.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ yalc.lock
 coverage*
 reports*
 2
+
+# Diagnostic reports (https://nodejs.org/api/report.html)
+report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -26,3 +26,6 @@ yalc.lock
 coverage*
 reports*
 2
+
+# Diagnostic reports (https://nodejs.org/api/report.html)
+report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
```
zephraph: Could this just be `report.*.json`?

erikdstock: I used this because it was the 'official' line from gitignore.io. Don't have a strong opinion either way if you think that is better.

erikdstock:
```suggestion
report.*.json
```

artsyit:
:rocket: PR was released in `v25.38.0` :rocket:

PR #3326, opened 2020-03-27
[Consignments] Remove shuffle on recently sold

damassi:
Little SSR edgecase -- if `shuffle` is used, it will execute once on the server, and then once more on the client, leading to dom mismatch. Need to think this through a bit better.


artsyit:
:rocket: PR was released in `v25.39.1` :rocket:

jonallured:
Oh weird, how did you notice this??

pepopowitz:
Good catch!

damassi:
The text would shuffle, but the image wouldn't. I was only deving in Storybooks, and since the page isn't complicated just pushed things along. When SSR kicked in it became clear. I imagine this is fixible tho, just not worth digging right now.

PR #3329, opened 2020-03-30
[Artist] Add ConsignButton to Artist/Overview page

damassi:
Addresses: https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=86&projectKey=CSGN&modal=detail&selectedIssue=CSGN-109

This adds the new consign  button to the Artist/Overview page: 

#### Top 20: 

Screen Shot 2020-03-30 at 1 33 25 PM

#### General Artist: 

Screen Shot 2020-03-30 at 1 31 15 PM

### DIFF
--- a/src/Apps/Artist/Routes/Overview/__tests__/index.test.tsx
+++ b/src/Apps/Artist/Routes/Overview/__tests__/index.test.tsx
@@ -210,7 +210,7 @@ describe("OverviewRoute", () => {
   })
 
   describe("ConsignButton", () => {
-    it("shows a consign button", () => {
+    it("shows a default consign button", () => {
       const wrapper = getWrapper({
         ...defaultArtist,
       })
--- a/src/Apps/Artist/Routes/Overview/__tests__/index.test.tsx
+++ b/src/Apps/Artist/Routes/Overview/__tests__/index.test.tsx
@@ -209,6 +209,15 @@ describe("OverviewRoute", () => {
     })
   })
 
+  describe("ConsignButton", () => {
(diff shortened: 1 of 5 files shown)


src/Apps/Artist/Routes/Overview/__tests__/index.test.tsx
```diff
@@ -209,6 +209,15 @@ describe("OverviewRoute", () => {
     })
   })
 
+  describe("ConsignButton", () => {
+    it("shows a consign button", () => {
+      const wrapper = getWrapper({
+        ...defaultArtist,
+      })
+      expect(wrapper.find("ArtistConsignButton").length).toEqual(1)
```
pepopowitz: This assertion is pretty tied to our specific implementation. I know we are using enzyme and not testing-library, but I like to use the [testing-library guiding principles](https://testing-library.com/docs/guiding-principles) because I think their approach results in more robust tests & easier refactoring. In this case, it would mean asserting that specific text or a button with specific text appeared on the page, instead of a specific component type.

No changes expected here - It's opinionated feedback.

damassi: Will update 👍

damassi: These assertions are encapsulated / tested in the button itself, up a level; since this is the container we're just checking for the existence of the button on the page.

artsyit:
:rocket: PR was released in `v25.40.0` :rocket:

PR #3333, opened 2020-03-31
[Consignments] Guard against missing data

damassi:
Noticed that `artsy.net/artist/genieve-figgis/consign` has no recently sold artworks, but is still in the microfunnel, and thus breaks as the assumption was that only artists with sold items were in this list. This guards against this case in the future by hiding the section entirely. 

#trivial


artsyit:
:rocket: PR was released in `v25.41.1` :rocket:

src/Apps/Artist/routes.tsx
```diff
@@ -97,7 +98,7 @@ export const routes: RouteConfig[] = [
         const { artist } = props as any
 
         if (!artist) {
-          return null
+          return undefined
```
damassi, line 101: Need to figure out how to address this more globally, but in order to trigger the top loading bar we need to return `undefined` rather than `null`

PR #3334, opened 2020-04-01
Display information about identity verification on credit card form

bhoggard:
completes https://artsyproduct.atlassian.net/browse/AUCT-949

Screen Shot 2020-03-31 at 6 35 15 PM

<details>
  <summary>📦 Published PR as canary version: <code>25.41.2-canary.3334.56177.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.41.2-canary.3334.56177.0
  # or 
  yarn add @artsy/reaction@25.41.2-canary.3334.56177.0
  ```
</details>

### DIFF
--- a/src/Apps/Auction/Routes/Register/index.tsx
+++ b/src/Apps/Auction/Routes/Register/index.tsx
@@ -99,6 +99,8 @@ export function createCreditCardAndUpdatePhone(relayEnvironment, phone, token) {
 
 export const RegisterRoute: React.FC<RegisterProps> = props => {
   const { me, relay, sale, tracking } = props
+  const needsIdentityVerification =
+    sale.requireIdentityVerification && !me.identityVerified
 
   const commonProperties = {
     auction_slug: sale.slug,
@@ -201,6 +203,7 @@ export const RegisterRoute: React.FC<RegisterProps> = props => {
         <StripeWrappedRegistrationForm
           onSubmit={handleSubmit}
           trackSubmissionErrors={trackRegistrationFailed}
+          needsIdentityVerification={needsIdentityVerification}
         />
       </Box>
     </AppContainer>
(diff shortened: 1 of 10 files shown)


artsyit:

src/Apps/Auction/Routes/Register/index.tsx
```diff
@@ -99,6 +99,8 @@ export function createCreditCardAndUpdatePhone(relayEnvironment, phone, token) {
 
 export const RegisterRoute: React.FC<RegisterProps> = props => {
   const { me, relay, sale, tracking } = props
+  const needsIdentityVerification =
```
dleve123, line 102: @bhoggard @yuki24 I'm working on a related story and realized this is missing a conditional. I believe this should really be the actual definition of "user needs identity verification" (`needsIdentityVerification` on the commented line; `userNeedsIdentityVerification` in the codeblock below):

```ts
/*
* Note: If the user is already registered to bid, then the sale doesn't
* requireIdentityVerification OR the user was manually approved by an
* admin. In either case, they don't need identity verification at this
* time.
*/
const userNeedsIdentityVerification: boolean =
  !registeredToBid &&
  artwork.sale.requireIdentityVerification &&
  !me?.identityVerified
```

Note the first condition – we don't "need" identity verification if the user is approved for bidding anyways (aka they were manually approved by an admin).

Let's align on Monday -- assuming we're on the same page, let's update this call-site to reflect this predicate. If we're aligned here, then we're going to need to update this [Eigen call-site as well](https://github.com/artsy/eigen/pull/3138/files#diff-e5bbaad5f015fcc332d39de302973805R55) (cc @yuki24).

---

The evolving complexity of this conditional here makes me want to [re-visit this conversation](https://artsy.slack.com/archives/CA3D02SDN/p1584993767026400) we had about whether we should place this logic in MP directly (or Gravity, some common service). Seems high risk to have to write out this conditional in many places.

---

Sorry for the weekend ping! Just didn't want to forget about this.

yuki24: I don't think it is extremely necessary to change it. The whole point of the registration screen is to allow unregistered users to register, and users are always unregistered unless deliberately attempted. I don't think it makes sense to add a check that always evaluates to one condition.

PR #3335, opened 2020-04-01
Organize MobileNav into folder

damassi:
Noticed that the new MobileNav and associated files were living freely inside of the `Menus` folder. This just moves things into a `Menus/MobileNavMenu` folder for clarity.


src/Components/NavBar/Menus/MobileNavMenu/mobileNavMenuData.ts
```diff
@@ -19,7 +19,7 @@ export interface MenuLinkData {
   dividerBelow?: boolean
 }
 
-export const menuData: MenuData = {
```
anandaroop, line 22: This is a data file meant to power both the new mobile menu (the first one we tackled) as well as the new desktop (currently in the works).

So I'd favor giving this a more generic name, and maybe even moving it a folder up so that it does not appear coupled solely with the mobile nav.

(At least that was my understanding when I was doing FX work actively, @xtina-starr feel free to correct me 😄)

damassi: We should rename it when needed, unless that's currently in progress.

xtina-starr: Yeah it is currently in progress. I think the `menuData` should live outside of the MobileNavMenu folder but everything else makes perfect sense.

damassi: 👍 i'll move it back down to the root

artsyit:

PR #3337, opened 2020-04-02
Revert "Redirect logged in users to the works for sale tab"

zephraph:
Temporarily reverts artsy/reaction#3307 until we can fix a redirect loop that's causing the artist page to fail to load on staging.


artsyit:
:rocket: PR was released in `v25.42.1` :rocket:

PR #3338, opened 2020-04-02
Fix redirect loop when user logged in on artist page

zephraph:
This fixes the redirect loop we encountered on the artist page. 

Most of the changes here are the actual revert of the former revert. The fix can be found here: https://github.com/artsy/reaction/commit/478389537863c91aa77d500469207ab97be57a6a#diff-53ea779b20e5bd37af12c435a46415edR135

<details>
  <summary>📦 Published PR as canary version: <code>25.42.3-canary.3338.56288.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@25.42.3-canary.3338.56288.0
  # or 
  yarn add @artsy/reaction@25.42.3-canary.3338.56288.0
  ```
</details>

### DIFF
--- a/src/Apps/Artist/Components/NavigationTabs.tsx
+++ b/src/Apps/Artist/Components/NavigationTabs.tsx
@@ -51,6 +51,7 @@ export class NavigationTabs extends React.Component<Props> {
   renderTabs() {
     const {
       artist: { slug, statuses, counts },
+      user,
     } = this.props
 
     const route = path => `/artist/${slug}${path}`
@@ -62,7 +63,7 @@ export class NavigationTabs extends React.Component<Props> {
 
     return (
       <>
-        {this.renderTab("Overview", route(""), {
+        {this.renderTab("Overview", route(user ? "/overview" : ""), {
           exact: true,
         })}
         {statuses.artworks &&
--- a/src/Apps/Artist/routes.tsx
+++ b/src/Apps/Artist/routes.tsx
@@ -121,14 +121,18 @@ export const routes: RouteConfig[] = [
       /**
        * The logic is as follows
        *
-       * 1. If a user is logged in / redirects for /works-for-sale
-       * 2. If a user is logged in /overview opens the overview tab (which also links off to /overview)
-       * 3. If a user is not logged in / leads to the overview tab
-       * 4. If a user is not logged in /overview redirects to / and therefore the overview tab
-       * 5. If there's insufficient data, all tabs redirect to /works-for-sale
+       * 1. If a user somehow ends up at /artist/<slug>/ redirect to /artist/<slug> because the former causes weird issues
+       * 2. If a user is logged in / redirects for /works-for-sale
+       * 3. If a user is logged in /overview opens the overview tab (which also links off to /overview)
+       * 4. If a user is not logged in / leads to the overview tab
+       * 5. If a user is not logged in /overview redirects to / and therefore the overview tab
+       * 6. If there's insufficient data, all tabs redirect to /works-for-sale
        */
+      if (pathname === `/artist/${artist.slug}/`) {
+        throw new RedirectException(`/artist/${artist.slug}`)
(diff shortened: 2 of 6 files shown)


src/Apps/Artist/routes.tsx
```diff
@@ -87,42 +87,69 @@ export const routes: RouteConfig[] = [
...
+       * The logic is as follows
+       *
+       * 1. If a user somehow ends up at /artist/<slug>/ redirect to /artist/<slug> because the former causes weird issues
+       * 2. If a user is logged in / redirects for /works-for-sale
+       * 3. If a user is logged in /overview opens the overview tab (which also links off to /overview)
+       * 4. If a user is not logged in / leads to the overview tab
+       * 5. If a user is not logged in /overview redirects to / and therefore the overview tab
+       * 6. If there's insufficient data, all tabs redirect to /works-for-sale
+       */
+      if (pathname === `/artist/${artist.slug}/`) {
```
sweir27, line 131: Interesting 😅

zephraph: Yeaaahh. It's almost considered a separate route, but it's not. You can navigate to it and it won't do the redirects and it also breaks the navigation tabs highlighting. 🤷‍♂

src/Apps/Artist/routes.tsx
```diff
@@ -87,42 +87,69 @@ export const routes: RouteConfig[] = [
...
+        throw new RedirectException(`/artist/${artist.slug}`)
+      }
 
-        return <Component {...props} />
+      if (user && pathname === `/artist/${artist.slug}`) {
+        throw new RedirectException(`/artist/${artist.slug}/works-for-sale`)
       }
+
+      if (!user && pathname.includes("/overview") && canShowOverview) {
+        throw new RedirectException(`/artist/${artist.slug}`)
```
sweir27, line 140: Do we need this? It seems like this is saying if a non-logged-in user tries to visit `/overview`, we redirect to `/`. Do we need to?

zephraph: Technically, no. The only way a user could get to this would be either by being on the overview tab when they logged out or by going to the URL directly.

Are you in favor of removing it?

sweir27: I'd remove it unless there's a reason to keep (just in the interest of keeping the logic here minimal).

zephraph: Done

src/Apps/Artist/Components/NavigationTabs.tsx
```diff
@@ -62,7 +63,7 @@ export class NavigationTabs extends React.Component<Props> {
 
     return (
       <>
-        {this.renderTab("Overview", route(""), {
+        {this.renderTab("Overview", route(user ? "/overview" : ""), {
```
sweir27, line 66: Ah-- will this break now that you made the other change?

zephraph: Nope, this will continue to function fine. Both `/` and `/overview` will lead to the overview route.

zephraph: The code to support that is here: https://github.com/artsy/reaction/pull/3338/files#diff-53ea779b20e5bd37af12c435a46415edR145

sweir27: Ahhh I was misunderstanding what this option was even for 😅 . Assumed it was about matching/highlighting.

zephraph: Actually, @sweir27 and I talked about this a bit offline. After doing some testing I saw the above change actually _does_ break tab highlighting which I think was the original reason for adding it. Gonna re-add.

artsyit:
:rocket: PR was released in `v25.42.5` :rocket:

PR #3339, opened 2020-04-02
Use "Artsy for Galleries" instead of "Partner with Artsy" naming in mobile nav

sweir27:
Addresses: https://artsyproduct.atlassian.net/browse/FX-1880

image

### DIFF
--- a/src/Components/NavBar/menuData.ts
+++ b/src/Components/NavBar/menuData.ts
@@ -440,7 +440,7 @@ export const menuData: MenuData = {
       href: "/institutions",
     },
     {
-      text: "Partner with Artsy",
+      text: "Artsy for Galleries",
       href: "/gallery-partnerships",
     },
     // TODO: links here for logged-in vs. logged-out state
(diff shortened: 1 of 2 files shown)


src/Components/NavBar/menuData.ts
```diff
@@ -440,7 +440,7 @@ export const menuData: MenuData = {
       href: "/institutions",
     },
     {
-      text: "Partner with Artsy",
+      text: "Artsy for Galleries",
```
sweir27, line 443: @xtina-starr Given we hard-code this in the file above, are we using this data anywhere? Wondering if we should remove it from this file or update the logic above-- what do you think?

xtina-starr: Yeah I think we can remove it from this file.

sweir27:
Alright @xtina-starr updated the screenshot! I think this is good to go.

src/Components/NavBar/menuData.ts
sweir27: It looks like there are a few entries that are unused. Would you mind giving it a once-over and removing the ones that aren't necessary as you're working on the desktop-side?

xtina-starr: Yeah! Definitely.

artsyit:
:rocket: PR was released in `v25.42.4` :rocket:

PR #3340, opened 2020-04-02
Ensure consign button shows for all artists

damassi:
Noticed that we were only showing the consign button if the artist had `is_consignable` on the model. Also noticed that it didn't appear in the mobile layout.


artsyit:
:rocket: PR was released in `v25.42.3` :rocket:

PR #3341, opened 2020-04-02
[Bugfix] Add missing `action` call to ArtworkConsignButton

damassi:
Was flagged that tracking was missing from the button click and noticed that `action` was missing from the schema props. 

Raised the issue here in an effort to make things a little more frictionless: https://artsy.slack.com/archives/C0KEQD4B0/p1585865660062300

#trivial


artsyit:
:rocket: PR was released in `v25.42.6` :rocket:

PR #3342, opened 2020-04-03
[Clientside Nav] Use app shell top-loader by default

damassi:
We have the loader split between a few different styles, from the spinner to the top purple bar. Rather than default to spinner on route transition, which was in place before the router work, lets use the purple bar by default unless we explicitly want the spinner. (Spinner is good for nested content like artist tabs)

Also cleans up error modal stack trace a bit by making it wider in development and respecting `\n` breaks: 

Screen Shot 2020-04-02 at 10 13 30 PM

### DIFF
--- a/src/Artsy/Router/Route.tsx
+++ b/src/Artsy/Router/Route.tsx
@@ -4,7 +4,6 @@
  */
 
 import { RouteSpinner } from "Artsy/Relay/renderWithLoadProgress"
-import { HttpError } from "found"
 import BaseRoute from "found/lib/Route"
 import React from "react"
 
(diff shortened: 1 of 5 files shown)


src/Artsy/Router/Route.tsx
```diff
@@ -22,12 +21,12 @@ interface RenderArgProps {
 }
 
 function createRender({
-  fetchIndicator = "spinner",
+  fetchIndicator = "overlay",
   render,
 }: CreateRenderProps) {
   return (renderArgs: RenderArgProps) => {
     const { Component, props, error } = renderArgs
-    if (error && error instanceof HttpError) {
```
damassi, line 30: Realized this was actually masking errors in the stack

damassi:
@zephraph - going to go ahead and merge this so that I can rebase it it into the new app shell which is going to launch in new A/B test tomorrow. Need to get it into review app upload tonight before bed.

artsyit:
:rocket: PR was released in `v26.0.1` :rocket:

PR #3344, opened 2020-04-03
[Route] Log generic errors but throw on HttpError

damassi:
This reverts a change in https://github.com/artsy/reaction/pull/3342.

In short, I noticed that we were masking errors in our routing layer by only throwing if it was an HTTP error. If a non-http error throws in the stack during a fetch it would get quietly consumed. In most cases this is fine as we surface these errors in other ways (such as our ErrorBoundary) but if something threw inside of our graphql stack [as this test shows](https://github.com/damassi/reaction/blob/914699fbe737dba9ece6bc9ff8e48ef6bdd3c7e1/src/Artsy/Router/__tests__/buildServerApp.test.tsx#L233-L253) we would never know about it. 

This ended up failing our acceptance test because it surfaced a masked error related to fixture data (or something):

Screen Shot 2020-04-03 at 4 20 46 PM

In any case, reverting that change and simply logging the error so that it's traceable if need be.

### DIFF
--- a/src/Artsy/Router/Route.tsx
+++ b/src/Artsy/Router/Route.tsx
@@ -4,6 +4,7 @@
  */
 
 import { RouteSpinner } from "Artsy/Relay/renderWithLoadProgress"
+import { HttpError } from "found"
 import BaseRoute from "found/lib/Route"
 import React from "react"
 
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in `v26.0.3` :rocket:

PR #3348, opened 2020-04-06
Adds the Art Keeps Going feature app + content

sweir27:
This adds the "Art Keeps Going" campaign landing page.

Blocked by https://github.com/artsy/metaphysics/pull/2274

The page itself is a one-off because we don't yet have a flexible-enough solution in place for pages like this (FX team is going to look into it 💪 🎉 ), and we want to ship it by Wednesday.

Some goals for this effort:
- I wanted to use an existing solution for static data, and settled on force's `JSONPage` as it's infinitely flexible and already comes with a handy UI. It also makes it clear that this is a one-off. 😄
- I wanted to use reaction for the UI to take advantage of existing components, etc.
- I had hoped to make it _somewhat_ future-proof (even just at the UI layer). I didn't get very far with that mostly because of limitations around relay/data-fetching for this specific use-case.

When we design a more flexible marketing solution (likely something akin to gravity's `Feature` page), we'll be able to have metaphysics serve us the exact shape of data we need to show all of the sections/content that exist. Unfortunately, for this we instead have to explicitly query for every piece of data we need. See the corresponding metaphysics PR for more info on limitations I found while doing this.

Major 👏 👏 for all of the work that went into making it easy to craft responsive layouts. I thought it would be frustrating but was one of the more straightforward things!

Follow-ups:
- [ ] Analytics
- [ ] Video/GIF header/sections

### Admin interface:
image

### Large breakpoint:
akg-desktop

### XS breakpoint:
akg-mobile


src/Apps/FeatureAKG/__tests__/FeatureAKGRoute.test.tsx
```diff
@@ -0,0 +1,815 @@
...
+      const wrapper = await getWrapper(
+        ArtKeepsGoingFixture,
+        defaultVariables,
+        injectedData
+      )
+
+      expect(wrapper.find("FeaturedFairsRail").length).toEqual(0)
+    })
+
+    it.todo("tracks clicking on a collection")
```
sweir27, line 384: Leaving these around intentionally. 😄

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -0,0 +1,227 @@
...
+import { crop, resize } from "Utils/resizer"
+
+interface FeatureProps {
+  viewer: Feature_viewer
+}
+
+const Feature: React.FC<FeatureProps> = props => {
+  const { injectedData } = useSystemContext()
+
+  if (injectedData === undefined) {
```
damassi, line 22: Is it for intentional readability to call out `undefined`? If so nothing to change, but around reaction we mostly see 
```js
if (!data) {
  return null
}
```

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -0,0 +1,227 @@
...
+    return null
+  }
+
+  const resizedUrl = resize(injectedData.hero_video_src, {
```
damassi, line 26: `hero_video_src` - any reason this isn't `camelCase`?  (Haven't got to where the data comes from yet.)

sweir27: Oh yeah, all the stuff that's stored in s3 (via `JSONPage`) is snake_case-- only because the admin UI makes everything uppercase so without the `_` it gets confusing.

(It also makes it clear what's injected via force vs. fetched from metaphysics, which while developing was helpful).

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -0,0 +1,227 @@
...
+    featuredRails?.collections_rail?.items?.length ||
+    featuredRails?.auctions_rail?.items?.length ||
+    featuredRails?.fairs_rail?.items?.length
+
+  return (
+    <>
+      <Box width="100%">
+        <Image src={resizedUrl} />
+      </Box>
+      <Box pt="3" maxWidth="475px" style={{ margin: "0 auto" }}>
```
damassi, line 48: Should also be able to do `m="0 auto"`

src/Apps/FeatureAKG/Components/FeaturedArticles.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const FeaturedArticles: React.FC<FeaturedArticlesProps> = props => {
+  const articles = props.articles
+
+  if (!articles?.length) {
+    return null
+  }
+
+  // @ts-ignore
+  const firstArticle = articles.shift()
```
damassi, line 28: Is this some kind of `no-mutation` lint error? If so can also do:
```js
const [firstArticle] = articles
```

src/Apps/FeatureAKG/Components/FeaturedArticles.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Grid fluid>
+      <Row>
+        <Col md={6}>
+          <Box pr={[0, 0, 1]}>
```
damassi, line 34: Dev time is so fast with these responsive props, its kinda crazy

src/Apps/FeatureAKG/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const FeaturedArtists: React.FC<FeaturedArtistsProps> = props => {
+  return (
+    <Flex
+      flexDirection="row"
+      justifyContent={["center", "space-between"]}
+      flexWrap="wrap"
+      maxWidth="720px"
+      style={{ margin: "0 auto" }}
```
damassi, line 16: `m="0 auto"`

src/Apps/FeatureAKG/Components/FeaturedRails/FeaturedAuctions.tsx
```diff
@@ -0,0 +1,65 @@
...
+  auctions: FeaturedAuctions_auctions
+  railMetadata: RailMetadata
+}
+
+const FeaturedAuctionsRail: React.FC<FeaturedAuctionsRailProps> = props => {
+  const { title, subtitle, items } = props.railMetadata
+  const { auctions } = props
+
+  const itemsForCarousel = auctions.edges.map(auction => {
+    const matchingAuctionFromSpreadsheet = find(
```
damassi, line 22: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/Apps/FeatureAKG/Components/FeaturedRails/FeaturedCollections.tsx
```diff
@@ -0,0 +1,62 @@
...
+  collections: FeaturedCollections_collections
+  railMetadata: RailMetadata
+}
+
+const FeaturedCollectionsRail: React.FC<FeaturedCollectionsRailProps> = props => {
+  const { title, subtitle, items } = props.railMetadata
+  const { collections } = props
+
+  const itemsForCarousel = collections.map(collection => {
+    const matchingCollectionFromSpreadsheet = find(
```
damassi, line 22: Ditto

src/Apps/FeatureAKG/__tests__/FeatureAKGRoute.test.tsx
```diff
@@ -0,0 +1,815 @@
...
+    it("displays a featured article and 3 other articles", () => {
+      expect(defaultWrapper.html()).toContain("Editorial")
+      expect(
+        defaultWrapper.find("FeaturedArticles").find("StyledLink").length
+      ).toEqual(4)
+      expect(defaultWrapper.find("FeaturedArticles").text()).toContain(
+        "View more"
+      )
+    })
+    it.todo("tracks clicking the featured article")
```
damassi, line 123: Nice, did not know about `todo`

src/Apps/FeatureAKG/routes.tsx
```diff
@@ -0,0 +1,46 @@
+import loadable from "@loadable/component"
+import { RouteConfig } from "found"
+import { graphql } from "react-relay"
+
+export const routes: RouteConfig[] = [
+  {
+    path: "/campaign/art-keeps-going",
+    getComponent: () => loadable(() => import("./FeatureApp")),
```
damassi, line 8: This only applies in the new app shell but we'll want to preload this JS bundle in parallel with the relay req. This prevents the slight flash post route transition as the bundle split loads:

```jsx
const FeatureApp = loadable(() => import("./FeatureApp"))
...
prepare: () => {
  FeatureApp.preload()
},
```

Here's where [it's called](https://github.com/artsy/reaction/blob/master/src/Apps/Components/AppShell.tsx#L21-L34).

src/Apps/FeatureAKG/routes.tsx
```diff
@@ -0,0 +1,46 @@
...
+      const data = props.context.injectedData
+      const collectionIDs = data?.browse?.collections_rail?.items
+      const fairIDs = data?.browse?.fairs_rail?.items
+      const auctionIDs = data?.browse?.auctions_rail?.items
+
+      return {
+        articleIDs: data?.editorial?.article_ids,
+        selectedWorksSetID: data?.selected_works?.set_id,
+        collectionRailItemIDs:
+          collectionIDs.length && collectionIDs.map(item => item.id),
```
damassi, line 39: Ok that this as well as the other `IDs` props could be falsy? Better to set at the definition and use the new `??` [nullish coalescing](https://levelup.gitconnected.com/understand-the-nullish-coalescing-operator-coming-in-es2020-2d1c6df1765f) operator: 

```ts
const collectionIDs = data?.browse?.collections_rail?.items.map(item => item.id) ?? []
const fairIDs = ...
const auctionIDs = ...
```

src/Artsy/Router/buildClientApp.tsx
```diff
@@ -83,6 +83,7 @@ export function buildClientApp(config: RouterConfig): Promise<Resolve> {
         historyMiddlewares,
         historyOptions: history.options,
         routeConfig: createRouteConfig(routes),
+        matchContext: context,
```
damassi, line 86: Yayy 😄

artsyit:
:rocket: PR was released in `v26.2.0` :rocket:

PR #3351, opened 2020-04-06
Search bar artist results link to works-for-sale page

lilyfromseattle:
This was meant to be completed as part of [PURCHASE-1787]

[PURCHASE-1787]: https://artsyproduct.atlassian.net/browse/PURCHASE-1787

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -30,6 +30,7 @@ import { Media } from "Utils/Responsive"
 import { SearchInputContainer } from "./SearchInputContainer"
 
 const logger = createLogger("Components/Search/SearchBar")
+const regex = new RegExp(/\/artist\/*/, "i")
 
 export interface Props extends SystemContextProps {
   relay: RelayRefetchProp
@@ -244,7 +245,7 @@ export class SearchBar extends Component<Props, State> {
     ) => ({
       action_type: Schema.ActionType.SelectedItemFromSearch,
       query: state.term,
-      destination_path: href,
+      destination_path: regex.test(href) ? href + "/works-for-sale" : href,
       item_type: displayType,
       item_id: id,
       item_number: suggestionIndex,
@@ -256,6 +257,7 @@ export class SearchBar extends Component<Props, State> {
     },
   }) {
     this.userClickedOnDescendant = true
+    const newHref = regex.test(href) ? href + "/works-for-sale" : href
 
     if (this.enableExperimentalAppShell) {
       if (this.props.router) {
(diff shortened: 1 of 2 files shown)


src/Components/Search/SearchBar.tsx
```diff
@@ -244,7 +245,7 @@ export class SearchBar extends Component<Props, State> {
     ) => ({
       action_type: Schema.ActionType.SelectedItemFromSearch,
       query: state.term,
-      destination_path: href,
+      destination_path: regex.test(href) ? href + "/works-for-sale" : href,
```
zephraph, line 248: Hmmm.... what is the value of `href` here? 

Is it `/artist/foo`, `/artwork/bar`, etc? Or is it a full url starting with `http`?

mzikherman: 🐈 Can use `displayType` prop in place of the regex, that should be "Artist" for all artists in the list.

That propagates from MP here: https://github.com/artsy/metaphysics/blob/e0a2c0e81953571a80eeba405118ca688170864a/src/schema/v2/SearchableItem/index.ts#L42. In the case of artists, that falls thru that helper to here: https://github.com/artsy/metaphysics/blob/e0a2c0e81953571a80eeba405118ca688170864a/src/schema/v2/SearchableItem/SearchableItemPresenter.ts#L93

So, that corresponds to the `label` property of the upstream response.

Screen Shot 2020-04-07 at 9 22 34 AM

Looks promising! But what's `label`?

That is in Gravity: https://github.com/artsy/gravity/blob/09ef21978f278cca4184e2bd1db03ade6039712d/app/api/util/search_helpers.rb#L54

The `_type` field comes from our ES index, and corresponds to the `document_type` of a model.

In the version of ES and associated gems we're using, that's defined as: https://github.com/elastic/elasticsearch-rails/blob/5251a5201f0b3fa1278e4366e364f9aaded81be6/elasticsearch-model/lib/elasticsearch/model/naming.rb#L93 (which uses some Rails helpers to get the model name), unless it's overridden (which we do for partner sub-classes): https://github.com/artsy/gravity/blob/09ef21978f278cca4184e2bd1db03ade6039712d/app/models/domain/partner_gallery.rb#L3

I apologize for this insanely pedantic comment. It's just, regexes.... 😅

zephraph: So per @mzikherman's suggestion...

```suggestion
      destination_path: displayType === "Artist" ? `${href}/works-for-sale` : href,
```

src/Components/Search/SearchBar.tsx
```diff
@@ -244,7 +245,7 @@ export class SearchBar extends Component<Props, State> {
     ) => ({
       action_type: Schema.ActionType.SelectedItemFromSearch,
       query: state.term,
-      destination_path: href,
+      destination_path: displayType === "Artist" ? `${href}/works-for-sale` : href,
```
zephraph, line 248: Sorry, I didn't mean to just update this. The suggestion made that a little unclear, my bad!

We can use `displayType` to build the url _instead_ of using `regex`. So we'd want to update [line 33](https://github.com/artsy/reaction/pull/3351/files#diff-195cf326e4088c5f9ff2c6d47cc4248cR33) and [line 337](https://github.com/artsy/reaction/pull/3351/files#diff-195cf326e4088c5f9ff2c6d47cc4248cR337)

lilyfromseattle: My bad! I used that github automatic "commit this suggestion" thing

zephraph: You did, I just shouldn't have used GitHub's suggestion for that because it wasn't actually the right place to make a code change 😅

artsyit:
:rocket: PR was released in `v26.7.0` :rocket:

PR #3352, opened 2020-04-06
[Clientside Nav] Clear cache on mutation

damassi:
Noticed an issue where clicking an artist, clicking follow, then clicking an artwork, then clicking back to an artist the follow button wasn't selected because we had a cache hit on artist view. 

There's a config setting in our network layer to bust cache on mutations. Setting this to true. 

cache


mzikherman:
Good call.

artsyit:
:rocket: PR was released in `v26.1.3` :rocket:

PR #3353, opened 2020-04-06
Update `back to <artist>` links to point directly to the overview page

lilyfromseattle:
[PURCHASE-1853]
Currently on the cv, shows, and article pages there's a Back to <artist> link that the top which points to the root artist page (/). Given that we're doing redirects now when a user is logged in to the works-for-sale tab, it won't take logged in users who navigated to that page explicitly back to the overview which is a confusing experience. 

We should change all of those links to go specifically to /overview

Here's an example of what the link on https://www.artsy.net/artist/andy-warhol/cv

[PURCHASE-1853]: https://artsyproduct.atlassian.net/browse/PURCHASE-1853


artsyit:
:rocket: PR was released in `v26.2.2` :rocket:

PR #3354, opened 2020-04-06
Adds video support for the Art Keeps Going campaign

sweir27:
We have 3 different videos (one for the top and 2 sections in the middle), with two different assets depending on the breakpoint.

akg-vids


artsyit:
:rocket: PR was released in `v26.2.1` :rocket:

PR #3355, opened 2020-04-07
[Experiment] SSR Render Collections rail on Apps/Artist

damassi:
Depends on https://github.com/artsy/metaphysics/pull/2279

Seeing what it looks like to SSR render the collections rail. There are concerns about performance, but that might be offset by SEO impact since these rails currently only render on the client. (We're pushing collections content a lot lately.) Wanted to get y'alls thoughts.

#### Before: 

Screen Shot 2020-04-06 at 11 23 12 PM

#### After: 

Screen Shot 2020-04-06 at 10 55 23 PM

The difference is ~900ms (pointing at staging MP) for SSR. Seems maybe worth it since these pages are going to be in the top search results. It also fixes the poor UX problem around above-the-fold loading spinner > show rail > shift entire layout down.

#### Fix top hz border on artwork grid

Also adds a new prop `showTopBorder` to the `ArtworkFilter` component -- its been driving me crazy that its always there even if there's not a top rail sitting above it. The problem is we can only compute showing / hiding based on collect rail data. If it only renders on the client then it's impossible to check it before the page renders, throwing the layout off. 

#### Before: 

Screen Shot 2020-04-06 at 11 03 41 PM

#### After: 

Screen Shot 2020-04-06 at 10 46 32 PM

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -750,7 +750,14 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
   statuses: ArtistStatuses
   targetSupply: ArtistTargetSupply
   years: String
-  marketingCollections(size: Int): [MarketingCollection]
+  marketingCollections(
+    slugs: [String!]
+    category: String
+    randomizationSeed: String
+    size: Int
+    isFeaturedArtistContent: Boolean
+    showOnEditorial: Boolean
+  ): [MarketingCollection]
 }
 
 type ArtistArtworkGrid implements ArtworkContextGrid {
@@ -4059,6 +4066,19 @@ type FairSponsoredContent {
   pressReleaseUrl: String
 }
 
+type FeaturedLink {
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: String
+  href: String
+  image: Image
+  initials(length: Int = 3): String
+  subtitle: String
+  title: String
+}
+
 type Feedback {
   # A globally unique ID.
   id: ID!
@@ -5594,6 +5614,30 @@ type OpeningHoursText {
 
 union OpeningHoursUnion = OpeningHoursArray | OpeningHoursText
 
+type OrderedSet {
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID.
+  internalID: ID!
+  cached: Int
+  description: String
+  key: String
+  itemType: String
+  items: [OrderedSetItem]
+
+  # Returns a connection of the items. Only Artwork supported right now.
+  itemsConnection(
+    after: String
+    first: Int
+    before: String
+    last: Int
+  ): ArtworkConnection
+  name: String
+}
+
+union OrderedSetItem = Artist | Artwork | FeaturedLink | Gene
+
 union OrderParty = Partner | User
 
 type organizer {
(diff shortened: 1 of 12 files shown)


damassi:
> Is this content added to the single SSR fetch.

Yup! Everything is being grabbed at once. 

>  so I generally just do a sanity check, and use the Metaphysics logging middleware in Reaction/storybooks to look at the number of backend requests that needed to happen

There are calls to the artwork grid on the `works-for-sale` page, and a single stitched req to Kaws to grab the rail.

> I don't know we have any good rule on how to judge these kinds of tradeoffs (performance vs. SEO or UX concerns). Depending on the level of concern, we should isolate this change in a single Force deploy, and monitor. If there's an immediate issue can always rollback. Otherwise, it's probably a worthy followup later in the sprint/next sprint to review and see if we're satisfied or not.

This sounds good to me. Thoughts @sweir27? From a UX perspective it's hardly noticable between the route loading indicator and so on, but I can only speak to the UX. We SSR collections stuff all over the site via `/collection/id` so its not an uncommon thing. From an SEO perspective it would be nice to concretely associate results on `/artist/id` back to those pages.

mzikherman:
> single stitched req to Kaws to grab the rail.

That's totally fine then, IMO. Are you sure though? I think the main issue with the rails is that there's a single KAWS call to get related collections, but then for each collection, there's a separate ES-backed call to get the `artworksConnection` (to display those first 3 artwork thumbnails). Almost an N+1 type situation, and with ES thrown in the mix (hence the caution). So there might be a lot of `filter/artworks` calls in the logs in addition to the existing grid one.

damassi:
Ah, I see -- in the inspector there were two fields, `gravity` and `stitched`, which pointed to kaws. So I guess the kaws request is rerouted through the `gravity` namespace? 

On the page there's the one rail and the grid:

Screen Shot 2020-04-07 at 9 34 20 AM

So what you're saying is that for the artwork grid there's one request to ES, but for each of the items in the rail there's a cascade of requests.

One thing i'm noticing is that the results of each of these are cached; maybe that's why the response time is relatively low for whats happening under the hood:

Screen Shot 2020-04-07 at 9 32 47 AM

damassi:
Another thought is that rather than [hard-coding a size of 16](https://github.com/damassi/reaction/blob/634e5ab6e20d21cad632c8713f8f07cf907ff973/src/Apps/Artist/Routes/Works/index.tsx#L86) we lower this number.

damassi:
Here's how it feels in the new app shell (noting that perf should be improved in prod). It does feel slower, but not prohibitively so, and there's the loading indicator at the top:  

no-flash-app-shell

mzikherman:
We do cache filter results that are queried for by Kaws in the extended schema: https://github.com/artsy/metaphysics/blob/4024644d124818da315dfa7461ffce0267bb820c/src/lib/stitching/kaws/stitching.ts#L74-L76

But that screenshot you posted is what I was afraid of. Uncached, it will now take 16+ ES queries to render the artist page? (One for each related collection w/ a max of 16, and once for the main artwork grid). I mean....that's just setting us up for pain, and not success IMO. I don't think it's a great idea to build something that will rely on a warm and functioning cache w/o considering alternatives.

One idea:
  - why fetch artworks for the rail items that aren't visible? Shouldn't the first 3 rail items be fully fetched (with ES-enriched artworks), and then only as you scroll to the next 3, can it kick off those fetches?

Subsequent follow-up:
  Unfortunately the architecture of KAWS (a separate repo that contains metadata about collections of artworks), IMO, lends itself to needing many API calls to fetch ES data all the time, in a somewhat roundabout way. Really, collections should be a first-class gravity citizen, and we should build it there to be _fast_ to return artworks that are members. Obviously that's out of scope here! But esp. as we push collections more I would hope that's considered.

It still might be ok to merge this as-is and monitor closely, but if there were lazy loading implemented on the rails that might be a worthy improvement.

damassi:
Thanks @mzikherman for all of the feedback. Yeah, learning about all of those requests definitely dampened things a bit :/. 

@sweir27 - I'll close this but we can refer back to it at a later point. 

(The [associated MP PR](https://github.com/artsy/metaphysics/pull/2279) should still be merged tho as that was a v1 -> v2 migration bug that was never caught.)

PR #3356, opened 2020-04-07
Visual fixes based on QA

sweir27:
This PR updates the art keeps going layout based on some visual QA by Nicole and Megan.

I've left some comments/questions inline!

### DIFF
--- a/src/Apps/FeatureAKG/Components/Feature.tsx
+++ b/src/Apps/FeatureAKG/Components/Feature.tsx
@@ -1,4 +1,4 @@
-import { Box, Image, Sans, Serif } from "@artsy/palette"
+import { Box, ResponsiveImage, Sans, Serif } from "@artsy/palette"
 import { Feature_viewer } from "__generated__/Feature_viewer.graphql"
 import { FeaturedArticlesFragmentContainer as FeaturedArticles } from "Apps/FeatureAKG/Components/FeaturedArticles"
 import { FeaturedArtists } from "Apps/FeatureAKG/Components/FeaturedArtists"
@@ -34,22 +34,22 @@ const Feature: React.FC<FeatureProps> = props => {
   const video2 = injectedData.video_2
 
   const showRails =
-    featuredRails?.collections_rail?.items?.length ||
-    featuredRails?.auctions_rail?.items?.length ||
-    featuredRails?.fairs_rail?.items?.length
+    !!featuredRails?.collections_rail?.items?.length ||
+    !!featuredRails?.auctions_rail?.items?.length ||
+    !!featuredRails?.fairs_rail?.items?.length
 
   return (
     <>
-      <Media greaterThanOrEqual="md">
+      <Media greaterThanOrEqual="sm">
         {heroVideo?.large_src && (
-          <Box textAlign="center" mb="-10px">
+          <Box textAlign="center" mb={-1}>
             <Video src={heroVideo.large_src} />
           </Box>
         )}
       </Media>
-      <Media lessThan="md">
+      <Media at="xs">
         {heroVideo?.small_src && (
-          <Box textAlign="center" mb="-10px">
+          <Box textAlign="center" mb={-1}>
             <Video src={heroVideo.small_src} />
           </Box>
         )}
@@ -77,11 +77,21 @@ const Feature: React.FC<FeatureProps> = props => {
       )}
 
       {/* Video 1 */}
-      <Media greaterThanOrEqual="md">
-        {video1?.large_src && <ImageSection src={video1.large_src} />}
+      <Media greaterThanOrEqual="sm">
+        {video1?.large_src && (
+          <ImageSection
+            src={video1.large_src}
+            aspectRatio={video1.large_aspect_ratio}
+          />
+        )}
       </Media>
-      <Media lessThan="md">
-        {video1?.small_src && <ImageSection src={video1.small_src} />}
+      <Media at="xs">
+        {video1?.small_src && (
+          <ImageSection
+            src={video1.small_src}
+            aspectRatio={video1.small_aspect_ratio}
+          />
+        )}
       </Media>
 
       {/* Selected works */}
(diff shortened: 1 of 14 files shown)


src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -77,10 +78,10 @@ const Feature: React.FC<FeatureProps> = props => {
 
       {/* Video 1 */}
       <Media greaterThanOrEqual="md">
-        {video1?.large_src && <VideoSection src={video1.large_src} />}
+        {video1?.large_src && <ImageSection src={video1.large_src} />}
```
sweir27, line 81: These are going to be black and white gifs, so they should be small.

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -173,24 +176,38 @@ const Section: React.FC<SectionProps> = props => {
...
-  // Constrain the height, add 1px for the border
+const Video: React.FC<{ src: string }> = props => {
   return (
-    <Box mb="-40px" height="591px">
-      <SectionSeparator mt="4" />
-      <Box textAlign="center">
-        <Video src={props.src} />
-      </Box>
-    </Box>
+    <StyledVideo
```
sweir27, line 181: The header is a big mp4 video that they're going to compress as much as possible.

I'm worried about performance here-- do you have any suggestions?

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -173,24 +176,38 @@ const Section: React.FC<SectionProps> = props => {
...
+const StyledVideo = styled("video")`
+  width: 100%;
+  height: auto;
+`
+
+const ImageSection: React.FC<{ src: string }> = props => {
   return (
-    <video autoPlay loop muted playsInline controls={false} src={props.src} />
+    <BorderedSection>
+      <Image src={props.src} width="100%" height="auto" />
```
sweir27, line 200: These gifs are well below the fold, but they are meant to be responsive. Do you know of any techniques for lazyloading images that don't have a fixed height?

src/Apps/FeatureAKG/Components/Feature.tsx
zephraph: You may want check what happens when this renders. Given that you're using `Media`, it's rendering both and hiding one. It's very likely both are being downloaded despite the breakpoint. If they're both small, maybe it's a non-issue. 

Might be time that we investigate adding support for the [picture element](https://www.w3schools.com/tags/att_source_srcset.asp)

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -93,24 +94,26 @@ const Feature: React.FC<FeatureProps> = props => {
...
-      </Media>
-
       {/* Featured Artists */}
       {featuredArtists?.artists?.length && (
         <Section
           title={featuredArtists.title}
           subtitle={featuredArtists.subtitle}
         >
-          <FeaturedArtists {...featuredArtists} />
+          <Box mb="-30px">
```
zephraph, line 103: Does `mb={-3}` work?

src/Apps/FeatureAKG/Components/Feature.tsx
sweir27: 😮Gah.

I can try giving `picture` a spin tomorrow here. (This project has a tight deadline-- Wednesday, so I likely won't get to supporting it in palette by then).

src/Apps/FeatureAKG/Components/Feature.tsx
zephraph: Generally we have three levers here:

1. Make it smaller
2. Transfer it faster
3. Load it later

I don't have a ton of recommendations. So long as it's being sent from a CDN, the only other thing I can really thing of is to defer its download until later. The loading experience might be hard to get right, but that's worth thinking about anyway.

src/Apps/FeatureAKG/Components/Feature.tsx
zephraph: Build it in place and promote it once you've proved it out.

src/Apps/FeatureAKG/Components/Feature.tsx
zephraph: Hmmm... So you tried the `ResponsiveImage` component and that didn't work?

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -34,27 +34,28 @@ const Feature: React.FC<FeatureProps> = props => {
   const video2 = injectedData.video_2
 
   const showRails =
-    featuredRails?.collections_rail?.items?.length ||
-    featuredRails?.auctions_rail?.items?.length ||
-    featuredRails?.fairs_rail?.items?.length
+    !!featuredRails?.collections_rail?.items?.length ||
```
sweir27, line 37: These (and similar changes) are to enable the edge case of us not having content in the JSON file for one or more rails.

There's also a slight performance win in that we're only fetching for the rails data at all if we have data for it (via an `@include(if:...` directive.

src/Apps/FeatureAKG/Components/Feature.tsx
```diff
@@ -173,24 +202,38 @@ const Section: React.FC<SectionProps> = props => {
...
+const StyledVideo = styled("video")`
+  width: 100%;
+  height: auto;
+`
+
+const ImageSection: React.FC<{ src: string; aspectRatio: number }> = props => {
   return (
-    <video autoPlay loop muted playsInline controls={false} src={props.src} />
+    <BorderedSection>
+      <ResponsiveImage lazyLoad src={props.src} ratio={props.aspectRatio} />
```
sweir27, line 226: This is where we handle the gif-- lazy loading works but we're still getting the accordion effect. I'll continue looking but it can be a follow up.

sweir27:
@zephraph I didn't get a chance to investigate using `picture` (instead attempted to get lazyloading to work).

The bigger issue with loading multiple files is for the video, but unfortunately the `media` attr on `source` within and html5 `video` component has been deprecated 😢.

artsyit:
:rocket: PR was released in `v26.3.1` :rocket:

PR #3365, opened 2020-04-08
Convert images to JPG and add poster for video

sweir27:
Couple of performance wins:
- Images were being uploaded a `png`s-- this explicitly converts them to jpeg.
- Added a `poster` attribute for the video so if someone has a setting or otherwise that doesn't let it autoplay, they get the first contentful frame vs. a white screen.


artsyit:
:rocket: PR was released in `v26.6.0` :rocket:

PR #3366, opened 2020-04-08
Use auth helpers for minimalCTABanner tracking args

eessex:
The tracking on the `MinimalCtaBanner` component was double-firing, because we get an `authImpression` event once we hit `/sign_up`.

This updates the link href for the banner to use updated auth tracking args and helpers, and removes the extraneous impression.


artsyit:
:rocket: PR was released in `v26.8.2` :rocket:

PR #3369, opened 2020-04-08
Update desktop nav menu

xtina-starr:
Addresses: [FX-1887](https://artsyproduct.atlassian.net/browse/FX-1887)

This is a WIP PR that introduces a new DropDownMenu component. It's blocked by some work on [palette Menu component](https://github.com/artsy/palette/pull/664). This PR includes most of the styling and layout structure in addition to hiding the new drop down behind a feature flag.

Missing the following ToDos:
- [x] Fix the broken tests + add new ones.
- [ ] Add chevron + underline to "Artworks" link.
- [x] Create a util to use for both Artworks & Artists data (might do this in a separate PR).

nav menu drop down

Screen Shot 2020-04-08 at 5 10 37 PM

### DIFF
--- a/src/Components/NavBar/Menus/DropDownMenu.tsx
+++ b/src/Components/NavBar/Menus/DropDownMenu.tsx
@@ -67,8 +67,8 @@ export const DropDownNavMenu: React.FC<DropDownNavMenuProps> = ({
                 return (
                   <MenuItemContainer mb={1}>
                     <MenuItem
-                      paddingX={0}
-                      paddingY={0}
+                      px={0}
+                      py={0}
                       href={menuItem.href}
                       textColor={color("black60")}
                       textWeight="regular"
--- a/src/Components/NavBar/Menus/DropDownMenu.tsx
+++ b/src/Components/NavBar/Menus/DropDownMenu.tsx
@@ -60,8 +60,8 @@ export const DropDownNavMenu: React.FC<DropDownNavMenuProps> = ({
   return (
     <Menu onClick={trackClick} width={width}>
       <ItemsContainer>
-        <OuterLinkContainer py={4}>
-          <Box mr={3} width={150}>
+        <OuterLinkContainer py={4} mr={[4, 4, 4, 6]}>
+          <Box mr={[2, 2, 3, 3]} width={[110, 110, 110, 135, 150]}>
             {menu.links.map(menuItem => {
               if (!menuItem.menu) {
                 return (
@@ -100,7 +100,6 @@ const MenuItemContainer = styled(Box)``
 
 const OuterLinkContainer = styled(Box)`
   border-right: 1px solid ${color("black10")};
-  margin-right: 60px;
   ${MenuItemContainer} {
     &:last-child {
       margin-top: 130px;
--- a/src/Components/NavBar/Menus/DropDownMenu.tsx
+++ b/src/Components/NavBar/Menus/DropDownMenu.tsx
@@ -65,7 +65,7 @@ export const DropDownNavMenu: React.FC<DropDownNavMenuProps> = ({
             {menu.links.map(menuItem => {
               if (!menuItem.menu) {
                 return (
-                  <MenuItemContainer mb={1}>
+                  <MenuItemContainer mb={1} key={menuItem.text}>
                     <MenuItem
                       px={0}
                       py={0}
@@ -96,7 +96,7 @@ const ItemsContainer = styled(Flex)`
   margin: auto auto;
 `
 
-const MenuItemContainer = styled(Box)``
+export const MenuItemContainer = styled(Box)``
 
 const OuterLinkContainer = styled(Box)`
   border-right: 1px solid ${color("black10")};
--- a/src/Components/NavBar/Menus/DropDownMenu.tsx
+++ b/src/Components/NavBar/Menus/DropDownMenu.tsx
@@ -1,0 +1,102 @@
+import { Box, Flex, Menu, MenuItem } from "@artsy/palette"
+import { AnalyticsSchema } from "Artsy"
+import { useTracking } from "Artsy/Analytics/useTracking"
+import React from "react"
+import styled from "styled-components"
+
+interface DropDownNavMenuProps {
+  width?: string
+  paddingTop?: number
+  paddingBottom?: number
+}
+
+export const DropDownNavMenu: React.FC<DropDownNavMenuProps> = ({
+  width = "100%",
+}) => {
+  const { trackEvent } = useTracking()
+
+  const trackClick = event => {
+    const link = event.target
+    const text = link.innerText
+    const href = link.parentNode.parentNode.getAttribute("href")
+
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.Click,
+      context_module: AnalyticsSchema.ContextModule.HeaderMoreDropdown,
+      subject: text,
+      destination_path: href,
+    })
+  }
+
+  return (
+    <Menu onClick={trackClick} width={width}>
+      <ItemsContainer>
+        <Box width={150}>
+          <MenuItem href="/galleries">Galleries</MenuItem>
+          <MenuItem href="/fairs">Fairs</MenuItem>
+          <MenuItem href="/shows">Shows</MenuItem>
+          <MenuItem href="/institutions">Museums</MenuItem>
+          <MenuItem href="/consign">Consign</MenuItem>
+          <MenuItem href="https://partners.artsy.net">
+            Artsy for Galleries
+          </MenuItem>
+        </Box>
+        <Box width={150}>
+          <MenuItem href="/galleries">Galleries</MenuItem>
+          <MenuItem href="/fairs">Fairs</MenuItem>
+          <MenuItem href="/shows">Shows</MenuItem>
+          <MenuItem href="/institutions">Museums</MenuItem>
+          <MenuItem href="/consign">Consign</MenuItem>
+          <MenuItem href="https://partners.artsy.net">
+            Artsy for Galleries
+          </MenuItem>
+        </Box>
+        <Box width={150}>
+          <MenuItem href="/galleries">Galleries</MenuItem>
+          <MenuItem href="/fairs">Fairs</MenuItem>
+          <MenuItem href="/shows">Shows</MenuItem>
+          <MenuItem href="/institutions">Museums</MenuItem>
+          <MenuItem href="/consign">Consign</MenuItem>
+          <MenuItem href="https://partners.artsy.net">
+            Artsy for Galleries
+          </MenuItem>
+        </Box>
+        <Box width={150}>
+          <MenuItem href="/galleries">Galleries</MenuItem>
+          <MenuItem href="/fairs">Fairs</MenuItem>
+          <MenuItem href="/shows">Shows</MenuItem>
+          <MenuItem href="/institutions">Museums</MenuItem>
+          <MenuItem href="/consign">Consign</MenuItem>
+          <MenuItem href="https://partners.artsy.net">
+            Artsy for Galleries
+          </MenuItem>
+        </Box>
+        <Box width={150}>
+          <MenuItem href="/galleries">Galleries</MenuItem>
+          <MenuItem href="/fairs">Fairs</MenuItem>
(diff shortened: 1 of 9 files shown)


src/Components/NavBar/NavItem.tsx
```diff
@@ -36,6 +40,8 @@ export const NavItem: React.FC<NavItemProps> = ({
   const getAnimation = h => ({
     opacity: h ? 0 : 1,
     transform: `translate3d(0, ${h ? -90 : -65}px, 0)`,
+    position: isFullScreenDropDown ? ("absolute" as Position) : "static",
```
damassi, line 43: Nice, so this is how it was done 👍

src/Components/NavBar/Menus/DropDownMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  // const getSubMenu = text => {
+  //   menu.links.filter(item => {
+  //     return item.text === text
+  //   })[0]
+  // }
+
+  const mediumLinks = menu.links.filter(item => {
+    return item.text === "Medium"
+  })[0]
```
damassi, line 42: If `filter` returns undefined, then `[0]` access could result in an error. Since this invocation is repeated a few times we could make things safer in the commented out helper function just above.

src/Components/NavBar/Menus/DropDownMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <DropDownSection section={priceLinks} />
+        <DropDownSection section={sellerLocationLinks} />
+      </ItemsContainer>
+    </Menu>
+  )
+}
+
+const ItemsContainer = styled(Flex)`
+  margin: auto auto;
+`
```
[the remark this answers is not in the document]
damassi: While less descriptive, note that styled-system's margin and padding (and so on) takes the full suite of CSS props, so something like `<Flex m="auto auto">` is valid.

src/Components/NavBar/NavItem.tsx
```diff
@@ -15,8 +15,11 @@ interface NavItemProps extends BoxProps {
   className?: string
   href?: string
   onClick?: () => void
+  isFullScreenDropDown?: boolean
 }
 
+type Position = React.CSSProperties["position"]
```
damassi, line 21: Can also use styled-systems's `PositionProps` interface here.

src/Components/NavBar/NavItem.tsx
damassi: Is `"absolute" as Position` needed?

xtina-starr:
Thanks for the awesome feedback @damassi @anandaroop!

src/Components/NavBar/Menus/DropDownMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { AnalyticsSchema } from "Artsy"
+import { useTracking } from "Artsy/Analytics/useTracking"
+import React from "react"
+import styled from "styled-components"
+import { DropDownSection } from "./DropDownSection"
+
+interface DropDownNavMenuProps {
+  width?: string
+  paddingTop?: number
+  paddingBottom?: number
```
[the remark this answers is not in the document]
xtina-starr: You are right it's not needed.

src/Components/NavBar/NavItem.tsx
xtina-starr: I get the following type errors without it: `Types of property 'position' are incompatible. Type 'string' is not assignable to type 'PositionProperty'.`

src/Components/NavBar/Menus/DropDownMenu.tsx
xtina-starr: In theory they should always return something because the data is hardcoded. However, that might not be the case forever. I could conditional render the `DropDownSection` component inside of `DropDownMenu` (i.e.):

```tsx
  {mediumLinks.length > 0 && <DropDownSection section={mediumLinks} />
```

or handle this inside of `DropDownSection` by returning `null` when the section prop is undefined.

Thoughts??

damassi: If it's hardcoded I think its fine for now! We'd want this to blow up early in that case, vs mask an error (like missing data for a menu item). When its dynamic and controlled from elsewhere (like say MP) then it would be good to add guards since upstream issues could involve any number of things.

src/Components/NavBar/NavItem.tsx
```diff
@@ -75,7 +81,6 @@ export const NavItem: React.FC<NavItemProps> = ({
 
   return (
     <Box
-      position="relative"
       onMouseEnter={() => toggleHover(true)}
       onMouseLeave={() => toggleHover(false)}
```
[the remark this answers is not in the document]
damassi: We should be careful here in terms of fast UI responsiveness since most nav items are smaller and vertically oriented. That said, this addition might improve _all_ of the nav items, but we also might want to make this timeout configurable from the parent and only apply it to the new `Artworks` mega nav. I've noticed in the past having trouble "catching" even the small navs, so a timeout might be just the solution.

xtina-starr:
FYI I'm planning to address [disappearing menu problem](https://github.com/artsy/reaction/pull/3369#discussion_r406590521) in a separate PR.

src/Components/NavBar/NavItem.tsx
xtina-starr: I tried pull `PositionProps` but was getting the same type error.

src/Components/NavBar/Menus/DropDownSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export const DropDownSection: React.FC<DropDownSectionProps> = ({
+  section,
+}) => {
+  if (!section) return null
+
+  return (
+    <>
+      {section.text === "Artist Nationality & Region" ? (
```
xtina-starr, line 16: This bit of logic here is to achieve the layout outlined in the spec.

Screen Shot 2020-04-14 at 12 39 04 PM

src/Components/NavBar/Menus/DropDownSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <Flex>
+          <Box width={[110, 110, 110, 135, 150]} py={4} mr={[2, 2, 3, 3]}>
+            <Sans size="2" mb={1}>
+              {section.text}
+            </Sans>
+            {section.menu &&
+              section.menu.links.slice(0, 8).map((menuItem: SimpleLinkData) => {
```
sweir27, line 23: Instead of explicitly breaking this into two separate columns, could we leverage flexbox to generate the layout based on our list of items? I know it's hard-coded data right now, but I think this'll be easier to maintain if the layout can adjust based on the number of items.

src/Components/NavBar/Menus/DropDownSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    fontSize="3t"
+                  >
+                    {menuItem.text}
+                  </MenuItem>
+                )
+              })}
+          </Box>
+        </Flex>
+      ) : (
+        <Box width={[110, 110, 110, 135, 150]} py={4} mr={[2, 2, 3, 3]}>
```
sweir27, line 64: If the logic above can be made simpler-- it's not clear to me which part of this display is different from the above section. For the most part they look similar.

Is it possible to isolate the difference and only add logic for that piece? (i.e. if it's a single padding value that's different, we can add logic around that value vs. replicating the entire component).

src/Components/NavBar/Menus/__tests__/DropDownMenu.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+    ;(useTracking as jest.Mock).mockImplementation(() => {
+      return { trackEvent }
+    })
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it("renders simple links", () => {
```
sweir27, line 28: Mind adding tests to ensure that we're tracking events correctly?

src/Components/NavBar/Menus/DropDownMenu.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { trackEvent } = useTracking()
+
+  const trackClick = event => {
+    const link = event.target
+    const text = link.innerText
+    const href = link.parentNode.parentNode.getAttribute("href")
+
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.Click,
+      context_module: AnalyticsSchema.ContextModule.HeaderArtworksDropdown,
```
sweir27, line 26: This should be updated to handle both artists and artworks, right?

src/Components/NavBar/Menus/DropDownMenu.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    </MenuItem>
+                  </MenuItemContainer>
+                )
+              }
+            })}
+          </Box>
+        </SimpleLinksContainer>
+
+        {/* We use splice here to ignore the first 4 links which are simple links and handled above */}
+        {menu.links.slice(4).map(subMenu => {
```
sweir27, line 59: Instead of slicing here, could we dynamically determine whether or not to show a `<DropDownSection` based on whether the `menu` key exists?

Even like (very rough pseudocode):
```
if (menuItem.menu) {
  <DropDownSection...
} else {
  ...

```

src/Components/NavBar/Menus/DropDownSection.tsx
xtina-starr: Screen Shot 2020-04-14 at 12 39 04 PM

@sweir27 I thought about that and explored it for a bit. All of the `MenuItem` links fit in a container with a specified width. What makes this UI somewhat tricky is that ['Artists Nationality & Region'](https://github.com/artsy/reaction/blob/7d9348252dce9d66dc30a820c2b86143547115aa/src/Components/NavBar/menuData.ts#L302-L367) is one long array of links that needs to break off (halfway) and fit in a new container with the specified width. Do you maybe have some time to chat about the implementation/pair?

src/Components/NavBar/Menus/__tests__/DropDownMenu.test.tsx
```diff
@@ -46,4 +47,19 @@ describe("DropDownMenu", () => {
...
+      .first()
+      .simulate("click")
+
+    expect(trackEvent).toHaveBeenCalledWith({
+      action_type: "Click",
+      context_module: "HeaderArtworksDropdown",
+      subject: "New this Week",
+      destination_path: "/collection/new-this-week",
+    })
+  })
```
xtina-starr, line 64: This seem pretty legit but I can't seem to get this spec to pass. The analytics event also works in the browser.

Screen Shot 2020-04-15 at 2 14 40 PM

src/Components/NavBar/Menus/TwoColumnDropDownSection.tsx
```diff
@@ -0,0 +1,55 @@
...
+  section,
+}) => {
+  return (
+    <Flex>
+      <Box width={[110, 110, 110, 135, 150]} py={4} mr={[2, 2, 3, 3]}>
+        <Sans size="2" mb={1}>
+          {section.text}
+        </Sans>
+        {section.menu &&
+          section.menu.links.slice(0, 8).map((menuItem: SimpleLinkData) => {
```
sweir27, line 19: So this wasn't possible to do with fixed widths/flexbox after all? Bummer!

Would you mind adding a comment to the code explaining why this logic is necessary (for our future selves 😄)?

damassi: Would also be good to assign `section.menu.links.slice(0, 8)` to a good variable name, which is self-documenting

xtina-starr: That's a good call!

sweir27:
Nice work!!! Excited to see this out in the wild 🎉 🙌 !

artsyit:
:rocket: PR was released in [`v26.21.0`](https://github.com/artsy/reaction/releases/tag/v26.21.0) :rocket:

PR #3370, opened 2020-04-09
Sync schema

eessex:
`yarn sync-schema`


damassi:
@eessex - there are two MP prs cycling through right now that need to make it in here, one sec...

damassi:
Closing in favor of https://github.com/artsy/reaction/pull/3371

PR #3373, opened 2020-04-09
[AUCT-952] Bidder sees copy in "about to register" and "registration pending" modals with information about identity verification

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-952

This PR is coupled to https://github.com/artsy/force/pull/5392. Please see the other PR for more details.

### DIFF
--- a/src/Components/Auction/PostRegistrationModal.tsx
+++ b/src/Components/Auction/PostRegistrationModal.tsx
@@ -4,6 +4,7 @@ import React, { useEffect, useState } from "react"
 export type ContentKey =
   | "registrationConfirmed"
   | "registrationPending"
+  | "registrationPendingUnverified"
   | "bidPending"
 
 interface Props {
@@ -35,6 +36,32 @@ const RegistrationPending: ModalContent = ({ onClick }) => {
     </>
   )
 }
+
+const RegistrationPendingUnverified: ModalContent = ({ onClick }) => {
+  return (
+    <>
+      <RegistrationPendingHeader />
+      <Serif my={3} size="3t">
+        This auction requires Artsy to verify your identity before bidding.
+        <br />
+        <br />
+        For details about identity verification, please see the
+        <a
+          target="_blank"
+          href="https://www.artsy.net/identity-verification-faq"
+        >
+          FAQ
+        </a>
+        or contact verification@artsy.net.
+        <br />
+        <br />A link to complete identity verification has been sent to your
+        email.
+      </Serif>
+      <ViewWorksButton onClick={onClick} />
+    </>
+  )
+}
+
 const RegistrationComplete: ModalContent = ({ onClick }) => {
   return (
     <>
@@ -55,6 +82,7 @@ const RegistrationComplete: ModalContent = ({ onClick }) => {
 const contentFor: { [key in ContentKey]: ModalContent } = {
   registrationConfirmed: RegistrationComplete,
   registrationPending: RegistrationPending,
+  registrationPendingUnverified: RegistrationPendingUnverified,
   bidPending: BidPending,
 }
 
--- a/src/Components/Auction/PostRegistrationModal.tsx
+++ b/src/Components/Auction/PostRegistrationModal.tsx
@@ -48,7 +48,7 @@ const RegistrationPendingUnverified: ModalContent = ({ onClick }) => {
         For details about identity verification, please see the
         <a
           target="_blank"
-          href="https://www.artsy.net/identity-verification-faq"
+          href="/identity-verification-faq"
         >
           FAQ
         </a>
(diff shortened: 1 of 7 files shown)


src/Components/Auction/PostRegistrationModal.tsx
```diff
@@ -35,6 +36,26 @@ const RegistrationPending: ModalContent = ({ onClick }) => {
...
+
+const RegistrationPendingUnverified: ModalContent = ({ onClick }) => {
+  return (
+    <>
+      <RegistrationPendingHeader />
+      <Serif my={3} size="3t">
+        This auction requires Artsy to verify your identity before bidding.
+        <br />
+        <br />
+        For details about identity verification, please see the FAQ or contact
```
bhoggard, line 48: Is `FAQ` supposed to be a link here?

yuki24: It should be a link, added a todo to https://github.com/artsy/force/pull/5392.

src/Components/Auction/PostRegistrationModal.tsx
```diff
@@ -35,6 +36,32 @@ const RegistrationPending: ModalContent = ({ onClick }) => {
...
+    <>
+      <RegistrationPendingHeader />
+      <Serif my={3} size="3t">
+        This auction requires Artsy to verify your identity before bidding.
+        <br />
+        <br />
+        For details about identity verification, please see the
+        <a
+          target="_blank"
+          href="https://www.artsy.net/identity-verification-faq"
```
dleve123, line 51: can we scope this to the environment that we're in?  I believe we can use `getENV('APP_URL')`, similar to what we do here:

https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx#L103

src/Components/Auction/__tests__/AuctionRegistrationModal.test.tsx
```diff
@@ -11,31 +11,82 @@ const closeMock = jest.fn()
...
+      },
+    })
+
+    expect(wrapper.find(Modal).text()).toMatch("Register for IDV Sale")
+    expect(wrapper.find(Modal).text()).toMatch(
+      "Welcome back. To complete your registration, please confirm that you agree to the Conditions of Sale."
+    )
+  })
+
+  it("shows an IDV registration message if the sale requires IDV and the user is verified", async () => {
```
dleve123, line 62:
```suggestion
  it("shows an IDV registration message if the sale requires IDV and the user is not verified", async () => {
```

src/Utils/identityVerificationRequirements.ts
```diff
@@ -25,8 +25,8 @@ export const bidderNeedsIdentityVerification = ({
   bidder,
 }: {
   sale: IdentityVerificationRequireable
-  user: IdentityVerifiable | null
-  bidder: Bidder | null
+  user?: IdentityVerifiable
```
dleve123, line 28: What motivates this diff? This makes sense for bidder, as sometimes there is no bidder object within scope, but there's always a `me` for us to consider (even if that me is `null`). I'm probably just not understanding some nuance of TS here, but thanks for clarifying!

yuki24: Having both `value: Type | null` and `value?: Type` next to each other makes understanding the signature quite tricky. Also the `?` symbol aligns more with TS's new optional chaining `value?.attr` both semantically and visually.

src/Components/Auction/PostRegistrationModal.tsx
yuki24, line 51:
```suggestion
          href="/identity-verification-faq"
```

yuki24: just updated to use a relative path.

src/Utils/identityVerificationRequirements.ts
```diff
@@ -25,8 +25,8 @@ export const bidderNeedsIdentityVerification = ({
...
+  bidder?: Bidder
 }) => {
   return (
     !bidder?.qualified_for_bidding &&
```
erikdstock, line 32: I realize this is not part of the diff but why is it necessary to check if the bidder is qualified for bidding? Wouldn't that be encapsulated by the checks below?

artsyit:

src/Utils/identityVerificationRequirements.ts
dleve123: @erikdstock Check out [this slack thread](https://artsy.slack.com/archives/CA3D02SDN/p1586546440286600) and [this bug](https://artsyproduct.atlassian.net/browse/AUCT-985) – both should help clarify why this is needed.

PR #3374, opened 2020-04-10
Bump timeout

damassi:
Looking at sentry we're getting a lot of timeout errors: 

Screen Shot 2020-04-09 at 10 19 15 PM

I bumped the timeout from 10s to 15s -- but maybe its better that the dialog with the reload button appears at 10, forcing an action in order to jog MP in some way?


artsyit:
:rocket: PR was released in `v26.8.5` :rocket:

PR #3377, opened 2020-04-10
[Clientside Nav] Move Pageloader from Palette

damassi:
Follow-up to https://github.com/artsy/palette/pull/665

Moving the top loading bar logic back over to reaction so that we can have more control. 

Eventually want to replace this with something like https://github.com/klendi/react-top-loading-bar, or at least match the functionality better.


artsyit:
:rocket: PR was released in `v26.8.8` :rocket:

PR #3384, opened 2020-04-13
Add data-test attribute on ForgotPassword forms

dblandin:
Similar to the `data-test` attribute on the login and signup forms, this allows us to query `ForgotPasswordForm` elements in the DOM during Cypress integration testing.

@eessex I followed the Login/SignUp examples found elsewhere. I wonder if we should at some point create a typed Cypress-specific schema which we can use in the artsy/integrity codebase similar to the analytics schema: https://github.com/artsy/reaction/blob/master/src/Artsy/Analytics/Schema/index.ts

Maybe we can also be more explicit in marking components as "queryable" within a Cypress test suite.

### DIFF
--- a/src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
+++ b/src/Components/Authentication/Desktop/ForgotPasswordForm.tsx
@@ -54,7 +54,11 @@ export class ForgotPasswordForm extends Component<
           }
 
           return (
-            <Form onSubmit={handleSubmit} height={180}>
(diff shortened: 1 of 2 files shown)


eessex:
Looks good! Regarding the schema for test components, I'm moving towards using the analytics schema v2 -- `contextModule` should map pretty accurately for most uses.

artsyit:
:rocket: PR was released in `v26.11.0` :rocket:

PR #3388, opened 2020-04-14
Upgrade storybook and related dependencies

dblandin:
`yarn upgrade @storybook/*`

PR #3391, opened 2020-04-14
Remove trailing separator char from Storybook story path

dblandin:
Follow-up to https://github.com/artsy/reaction/pull/3388

Starting at some version between v5.1.10 and v5.3.18 of Storybook, separator chars like `/` should only be used to separate path parts and not present at the end of a path string, otherwise the UI breaks. /shrug

```
Error: Invalid part '', leading to id === parentId
('publishing-tooltips'), inside kind 'Publishing/ToolTips/'

Did you create a path that uses the separator char accidentally, such as
'Vue <docs/>' where '/' is a separator char? See
https://github.com/storybookjs/storybook/issues/6128
```


artsyit:
:rocket: PR was released in `v26.14.0` :rocket:

PR #3394, opened 2020-04-14
[Clientside Nav] Slow down top bar a bit

damassi:
Its a bit too fast and frantic right now, slowing it down a bit...


artsyit:
:rocket: PR was released in `v26.14.2` :rocket:

PR #3407, opened 2020-04-15
[Clientside Nav] Fix referrer edgecase

damassi:
When looking closely at analytics data an edge-case was discovered around referrer being lost in events other than page views. 

Since we're now in a single-page-app context, we need a way to store referrers as the user progresses across clicks:
1) Compute the referrer in our analytics middleware as we have been
1) Store the refer on the global segment.js `window.analytics` object under `window.analaytics.__artsyReferrer`. 
1) In force, where we funnel analytics calls from Reaction, check to see if `__artsyReferrer` is populated and manually set the `referrer` before a `track` call is fired.

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -72,24 +72,28 @@ export class ArtworkApp extends React.Component<Props> {
       referrer,
     } = this.props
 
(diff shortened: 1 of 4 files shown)


artsyit:
:rocket: PR was released in `v26.17.2` :rocket:

PR #3402, opened 2020-04-15
Bump timeout of "ArtworkFilter/desktop/renders default UI items" spec

dblandin:
This spec is flaky on CI. We could skip it but might be better to give the spec a bit more time to complete and follow-up on why it's slow.

```
FAIL  src/Components/v2/ArtworkFilter/__tests__/ArtworkFilter.test.tsx (14.473s)
 ● ArtworkFilter › desktop › renders default UI items

   : Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.

     49 |
     50 |   describe("desktop", () => {
   > 51 |     it("renders default UI items", async () => {
        |     ^
     52 |       const wrapper = await getWrapper()
     53 |       expect(wrapper.find("ArtworkFilterArtworkGrid").length).toEqual(1)
     54 |       expect(wrapper.find("SortFilter").length).toEqual(1)

     at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
     at Suite.it (src/Components/v2/ArtworkFilter/__tests__/ArtworkFilter.test.tsx:51:5)
     at Suite.describe (src/Components/v2/ArtworkFilter/__tests__/ArtworkFilter.test.tsx:50:3)
     at Object.describe (src/Components/v2/ArtworkFilter/__tests__/ArtworkFilter.test.tsx:13:1)
```

PR #3400, opened 2020-04-15
Update Jest to use JSDOM v16

dblandin:
Update jest to use JSDOM v16 via the `jest-environment-jsdom-sixteen` package.

PR #3408, opened 2020-04-15
Rollback JSDOM v16 upgrade

dblandin:
We unfortunately discovered that some other specs are taking longer and hitting timeouts so we're going to postpone this library upgrade for now.

Partially revert https://github.com/artsy/reaction/pull/3400

Back out the library update but keep the `Object.defineProperty` mock updates which are backwards-compatible and should make a future upgrade attempt easier.

Also remove the timeout bump added in https://github.com/artsy/reaction/pull/3402.

Related issue thread on Jest 25 performance: https://github.com/facebook/jest/issues/9457

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,6 @@
     "babel-plugin-styled-components": "1.10.0",
     "babel-plugin-transform-decorators-legacy": "1.3.5",
     "cache-loader": "1.2.2",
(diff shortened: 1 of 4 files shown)


damassi:
LGTM 👍

artsyit:

PR #3410, opened 2020-04-15
[Clientside Nav] Add search to referrer

damassi:
We want to propagate queryParams across pages too, if they're there.


artsyit:

PR #3411, opened 2020-04-15
Use @artsy/cohesion

eessex:
Updates imports from `Artsy/Analytics/v2` to use `@artsy/cohesion` library.


src/Apps/Artist/Components/ArtistHeader.tsx
```diff
@@ -1,11 +1,11 @@
+import { AuthIntent, ContextModule } from "@artsy/cohesion/dist/Schema"
```
damassi, line 1: Left a comment in cohesion, but that should allow us to not need to drill into subdirectories of the package.

artsyit:
:rocket: PR was released in [`v26.18.0`](https://github.com/artsy/reaction/releases/tag/v26.18.0) :rocket:

PR #3431, opened 2020-04-21
Move cohesion to be a dev dep of reaction

zephraph:
Given that cohesion is a direct dependency of force, including it as a direct dependency of reaction can lead to cases where it's duplicated in Force's bundle. I'm moving it to a dev dependency in reaction so that it'll still be installed and available when developing locally, but won't be included in the production output (and therefore duplicated in force). 

I haven't included it in this PR, but if there are other services that depend on reaction which should also use cohesion then we should add cohesion as a peer dep too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   },
   "devDependencies": {
     "@artsy/auto-config": "1.0.2",
+    "@artsy/cohesion": "0.2.0",
     "@artsy/lint-changed": "3.0.4",
     "@artsy/palette": "8.1.3",
     "@babel/cli": "7.0.0",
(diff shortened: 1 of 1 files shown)


eessex:
@zephraph are you saying we'll need to install cohesion on all reaction-consuming apps to support this change? That would mean Positron and Rosalind for sure, not sure if there are more consumers, but I don't expect them to ever need Cohesion. 

I wonder if we can use resolvers or some other way to let force play nicer with the dependency? I fear that if we add cohesion to a reaction-consuming app that doesn't use it, it will get removed.

zephraph:
How it's set up now, no. We would _not_ need to install cohesion on other apps. 

The only reason we'd need to do that is if they needed cohesion.

eessex:
Ok sounds good! #mergeongreen

artsyit:
:rocket: PR was released in [`v26.22.3`](https://github.com/artsy/reaction/releases/tag/v26.22.3) :rocket:

PR #3433, opened 2020-04-21
Auction results auth intent conforms to schema

eessex:
Type checking is currently failing because we are using fields removed from cohesion in https://github.com/artsy/cohesion/pull/15

Updates references to use new naming pattern.

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
@@ -372,7 +372,7 @@ const renderPricing = (salePrice, saleDate, user, mediator, size) => {
               mode: ModalType.signup,
               copy: "Log in to see full auction records — for free",
               contextModule: ContextModule.auctionResults,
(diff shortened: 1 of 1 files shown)


eessex:
This change was recently requested by the data-team -- I had assumed it was already in reaction because the schema change was merged last week, but the references don't seem to have been updated. Context for the update to cohesion here: https://artsy.slack.com/archives/C9YNS4X32/p1587056175451400

dblandin:
> Originally these triggers had no intent

Gotcha. So passing in a blank intent for a period of time was probably fine.

I'm guessing that's what happened when we called `AuthIntent.loginToSeePrice` without that key/pair specific in the version reaction has been pinned to: https://github.com/artsy/cohesion/blob/f86d3248d31ce838b061a714ad98ff1a525b0bc9/src/Schema/Authentication.ts#L68-L90

The update looks good to me :+1:

artsyit:
:rocket: PR was released in [`v26.22.4`](https://github.com/artsy/reaction/releases/tag/v26.22.4) :rocket:

PR #3436, opened 2020-04-22
[Refactor] Modernize SystemContext, make app shell width configurable 

damassi:
This PR does a couple cleanup-y type things: 
- Converts `Boot` to an FC 
- Updates `SystemContext` to to be a bit more modern, and moves state updates to a reducer
- Adds a new action `setAppMaxWidth` which allows us to modify the outside container width to each of our breakpoints, or `100%`
- Moves our mechanism for catching a-tag links in the new app-shell context (similar to gatsby-catch-links) out of route construction and into its own file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "@artsy/auto-config": "1.0.2",
     "@artsy/cohesion": "0.2.0",
     "@artsy/lint-changed": "3.0.4",
-    "@artsy/palette": "8.1.3",
+    "@artsy/palette": "8.2.0",
     "@babel/cli": "7.0.0",
     "@babel/core": "7.7.2",
     "@babel/plugin-proposal-class-properties": "7.3.4",
--- a/package.json
+++ b/package.json
@@ -238,7 +238,8 @@
     "semi": false,
     "singleQuote": false,
     "trailingComma": "es5",
-    "bracketSpacing": true
+    "bracketSpacing": true,
+    "arrowParens": "avoid"
   },
   "jest": {
     "cacheDirectory": ".cache/jest",
--- a/src/Artsy/Router/Boot.tsx
+++ b/src/Artsy/Router/Boot.tsx
@@ -1,17 +1,18 @@
 import { Grid, injectGlobalStyles, Theme, themeProps } from "@artsy/palette"
 import * as Sentry from "@sentry/browser"
-import * as Artsy from "Artsy"
-import { track } from "Artsy/Analytics"
+import { SystemContextProvider, track } from "Artsy"
 import { RouteConfig } from "found"
-import React from "react"
+import React, { useEffect } from "react"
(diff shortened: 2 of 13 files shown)


src/Artsy/Router/Boot.tsx
```diff
@@ -19,74 +19,69 @@ import {
...
+      <HeadProvider headTags={headTags}>
+        <StateProvider>
+          <SystemContextProvider {...contextProps}>
+            <ErrorBoundary>
+              <MediaContextProvider onlyMatch={onlyMatchMediaQueries}>
+                <ResponsiveProvider
+                  mediaQueries={themeProps.mediaQueries}
+                  initialMatchingMediaQueries={onlyMatchMediaQueries as any}
+                >
+                  <Grid fluid>
```
zephraph, line 72: The grid component (in palette) has a max-width set on it.

damassi: Ahh -- so we had max-width set in a few areas. I can update palette with a new maxWidth prop and pass in systemContext maxwidth here. Will do a quick cycle over there and then rebase this PR.

src/Artsy/SystemContext.tsx
```diff
@@ -82,19 +69,20 @@ export const SystemContextProvider: SFC<SystemContextProps> = ({
...
   ...props
 }) => {
-  const [isFetching, setIsFetching] = useState(false)
-  const [router, setRouter] = useState(null)
-  const [user] = useState(getUser(props.user))
+  const [state, dispatch] = useReducer(systemContextReducer, {
+    isFetching: false,
+    router: null,
+    user: getUser(props.user),
+  })
```
zephraph, line 76: Nice 👌

package.json
```diff
@@ -238,7 +238,8 @@
     "semi": false,
     "singleQuote": false,
     "trailingComma": "es5",
-    "bracketSpacing": true
+    "bracketSpacing": true,
+    "arrowParens": "avoid"
```
damassi, line 242: Noticed that VSCode updated to 2, so setting this to v1 default until we want to update things codebase-wide

src/Artsy/Router/Boot.tsx
```diff
@@ -1,92 +1,89 @@
...
   }
-}
 
-// Tests
-GlobalStyles.displayName = "GlobalStyles"
-Grid.displayName = "Grid"
+  return (
+    <Theme>
+      <HeadProvider headTags={headTags}>
+        <StateProvider>
```
damassi, line 66: @dzucconi - this was pre hooks reaction, when we were using unstated. Theres just a few spots that still use the lib but most places have refactored it out

artsyit:
:rocket: PR was released in [`v26.24.0`](https://github.com/artsy/reaction/releases/tag/v26.24.0) :rocket:

PR #3438, opened 2020-04-23
Fix useSystemContext invocation

damassi:
Realized that `useSystemContext` was being called before we actually set up the context provider. Removed that and switched to `<SystemContextConsumer>` render props pattern in `Boot`.


artsyit:
:rocket: PR was released in [`v26.24.1`](https://github.com/artsy/reaction/releases/tag/v26.24.1) :rocket:

PR #3440, opened 2020-04-23
[Consignments] Add full-screen support to /consign

damassi:
This PR updates our microfunnel pages to be full-screen, and makes some adjustments elsewhere so full-screen support in the app shell is possible. 

Also removes randomization from the consign page due to SSR things. Normally a user navigates to that page from the button which works great, but direct links to the page will lead to random firing twice. Also removes image background shadow for now because its too tricky to get the border to match just right because of how the images are responsive. Sometimes its more px on the top, sometimes on the right, depending on the page. Looked sloppy. Also, many of the images already have a frame. Can look at that later. Nicole ok'd the change. Everything on that page is now simpler.

Also undoes changes from [this PR](https://github.com/artsy/reaction/pull/3436) as there was a weird edge-case discovered related to using reducer + react context leading to an infinite loop on state updates because of some kind of hook object equality somewhere. Very frustrating. 

Screen Shot 2020-04-23 at 2 11 44 PM


artsyit:
:rocket: PR was released in [`v26.25.1`](https://github.com/artsy/reaction/releases/tag/v26.25.1) :rocket:

PR #3445, opened 2020-04-27
[FX-1819] Links to consign page in mobile nav

dzucconi:
Re: [FX-1819](https://artsyproduct.atlassian.net/browse/FX-1819)

Same place as in the desktop hierarchy.

I don't really understand what this comment is specifically asking for — https://artsyproduct.atlassian.net/browse/FX-1819?focusedCommentId=27330 — should just be a separate issue.

### DIFF
--- a/src/Components/NavBar/menuData.ts
+++ b/src/Components/NavBar/menuData.ts
@@ -440,6 +440,10 @@ export const menuData: MenuData = {
       href: "/institutions",
     },
     {
+      text: "Consign",
+      href: "/consign",
+    },
+    {
       text: "Artsy for Galleries",
       href: "/gallery-partnerships",
     },
--- a/src/Components/NavBar/Menus/MobileNavMenu/MobileNavMenu.tsx
+++ b/src/Components/NavBar/Menus/MobileNavMenu/MobileNavMenu.tsx
@@ -55,6 +55,7 @@ export const MobileNavMenu: React.FC<Props> = props => {
             <MobileLink href="/fairs">Fairs</MobileLink>
             <MobileLink href="/shows">Shows</MobileLink>
             <MobileLink href="/institutions">Museums</MobileLink>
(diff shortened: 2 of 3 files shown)


xtina-starr:
@dzucconi agreed that should definitely be a separate ticket. It's asking for the additional links that are available in desktop but not mobile (see image below). I think that might require some pre-work though because those pages aren't optimized for mobile.

Screen Shot 2020-04-27 at 10 33 28 AM

src/Components/NavBar/menuData.ts
```diff
@@ -439,6 +439,10 @@ export const menuData: MenuData = {
       text: "Museums",
       href: "/institutions",
     },
+    {
+      text: "Consign",
+      href: "/consign",
+    },
```
dzucconi, line 445: @xtina-starr So, I actually don't understand this at all: I'm only adding it because there's a spec for it — Why does this data exist if it's not actually used? And what's with the type casting in the component spec (LinkData => SimpleLinkData)?

sweir27:
(We already have tickets for adding the /user/settings pages to the nav bar + making them more mobile-friendly!)

https://artsyproduct.atlassian.net/browse/FX-1848 and https://artsyproduct.atlassian.net/browse/FX-1903

src/Components/NavBar/menuData.ts
dzucconi: Just rebased and will merge on green. But if you know what's up with this — I'm interested. The duplication/spec doesn't make sense to me.

sweir27: I think the `menuData` file was a necessary part of an earlier iteration, but if it's not used then we should remove it + update the test!

artsyit:
:rocket: PR was released in [`v26.29.0`](https://github.com/artsy/reaction/releases/tag/v26.29.0) :rocket:

PR #3446, opened 2020-04-27
Make AKG page more resilient towards mismatched data

sweir27:
#minor but because we have to manually match the data from the JSON spreadsheet to the data coming back from metaphysics, it's possible that things are out-of-alignment.

This morning I noticed that the page was erroring out due to an instance of this ☝️. I fixed the data entry issue but this PR will make the page more resilient to issues of this type.

(Reminder that this entire page is still curated manually/is rather fragile. It's a one-off and we'll need to invest in a more sustainable solution before hosting other campaigns of this sort).

### DIFF
--- a/src/Apps/FeatureAKG/Components/FeaturedRails/FeaturedAuctions.tsx
+++ b/src/Apps/FeatureAKG/Components/FeaturedRails/FeaturedAuctions.tsx
@@ -45,7 +45,7 @@ const FeaturedAuctionsRail: React.FC<FeaturedAuctionsRailProps> = props => {
     return (
       <FeaturedRail title={title} subtitle={subtitle}>
         <FeaturedRailCarousel
-          itemsForCarousel={itemsForCarousel}
+          itemsForCarousel={compact(itemsForCarousel)}
           contextModule={AnalyticsSchema.ContextModule.BrowseAuctions}
         />
       </FeaturedRail>
--- a/src/Apps/FeatureAKG/Components/FeaturedRails/FeaturedCollections.tsx
+++ b/src/Apps/FeatureAKG/Components/FeaturedRails/FeaturedCollections.tsx
@@ -42,7 +42,7 @@ const FeaturedCollectionsRail: React.FC<FeaturedCollectionsRailProps> = props =>
     return (
       <FeaturedRail title={title} subtitle={subtitle}>
         <FeaturedRailCarousel
-          itemsForCarousel={itemsForCarousel}
+          itemsForCarousel={compact(itemsForCarousel)}
           contextModule={AnalyticsSchema.ContextModule.BrowseCollections}
         />
       </FeaturedRail>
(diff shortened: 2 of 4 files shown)


dzucconi:
This is fine, but seems weird to me — shouldn't the typing being generated accurately reflect this possibility or is part of this page just completely untyped?

sweir27:
Yeah, it's totally weird-- the architecture is like this:

(Using the "Auctions rail" as an example)

- Admins add static data via a `JSONPage`. In this case the data stored in s3 will look something like:
```json
"auctions_rail": {
    "items": [
      {
        "id": "moad-diaspora-unite-artists-of-african-descent-benefit-auction-2020",
        "image_src": "https://files.artsy.net/images/amoako-boafo-aurora-iradukunda-2020.jpg"
      },
      {
        "id": "food-bank-for-new-york-city-benefit-auction-2020",
        "image_src": "http://files.artsy.net/images/louis-fratino-nude-in-black-shorts-2020.jpg"
      }
    ],
    "subtitle": null,
    "title": "Benefit Auctions"
  },
```
The ID is used to fetch for more data, but they wanted to control the image that's shown, hence the `image_src`.
- We send that JSON blob over to our reaction app in the server context and use the data to construct graphQL queries that go through relay (in this case, we'll query for a list of auctions based on the IDs above).
- Then in the UI, we have to make sure we're pulling the correct manual image for each auction, so we have to stitch together both sources of data (JSON in s3 and response from graphQL query).

None of this is ideal. 😅

dzucconi:
Oof. Deeply troubling the JSONPage thing is still around. It'd be worth the (not even very large) amount of effort to make a real CMS for non-schema JSON and link it with MP.

sweir27:
@dzucconi Totally-- in this case, I'm OK with the hackiness because I don't want it to seem like this is even _potentially_ a long-term solution.

We have a desperate need for a proper data model for these marketing campaigns! Even the one-off data modeling is troubling in this case. Nicole and I are discussing that for this quarter on FX 😄.

damassi:
Consign team recently had to do a [marketing-ish page as well](https://www.artsy.net/artist/david-hockney/consign). It felt problematic not being near metaphysics / relay so we opted (for the time being) to use MP to serve up largely static data. So there's def an ongoing need for beter data management ATM. I believe the marketing team is just warming up, too. 

Right now there's:
- Torque (which is "deprecated", but is still actively being added to)
- JSONPage 
- And a new Rails-Admin system in Gravity (that's currently read only) 

Maybe others?

artsyit:
:rocket: PR was released in [`v26.27.5`](https://github.com/artsy/reaction/releases/tag/v26.27.5) :rocket:

sweir27:
Right now we have a few options for creating admin-managed "static pages":
1. JSONPage (data managed by bare-bones UI, stored in s3, requires a custom UI)
2. Static data stored in MP (data served my MP, no admin UI, requires a custom UI)
3. Feature pages (data served by gravity, has admin UI, does not require custom UI-- the page is dynamic based on the data)

The first two could be solved by the system that you + @dzucconi are mentioning-- some sort of combination of storing data in s3 + better admin interface + serving up data via metaphysics. It seems best-suited for a case where we truly _need_ a custom user interface (so probably best for something like the consign landing page, or other places where we need small bits of manually-entered data).

The last use-case is the one I'm looking into for this quarter-- the ability to create custom landing pages without needing a dev at all. 😻

PR #3450, opened 2020-04-27
Improves Carousel typings

dzucconi:
This lets us use the Carousel and access the type of the passed in object for each cell without it being an `any`.


src/Components/RecentlyViewed.tsx
```diff
@@ -59,6 +59,7 @@ export class RecentlyViewed extends React.Component<RecentlyViewedProps> {
                     return (
                       <FillwidthItem
                         lazyLoad={true}
+                        // @ts-ignore // TODO: Correct typing
```
dzucconi, line 62: Not 100% sure why it's not picking up the nested fragment type here. But better than the `any` for the time being.

damassi:
Nice, this came out simpler than it would have seemed!

artsyit:
:rocket: PR was released in [`v26.27.6`](https://github.com/artsy/reaction/releases/tag/v26.27.6) :rocket:

PR #3454, opened 2020-04-28
[NavBar] Add toggle menu support & fix mouse out hitarea on "Artworks" nav 

damassi:
Adds the ability to toggle closed `NavItem` components which render their own `Menu` via a `setIsVisible` prop. This is in support for the new app shell, where full page jumps don't occur. Now, when a user clicks, the menu toggles closed and the loading bar appears, making things feel nice and snappy. 

menu

(Hard to see loader in gif, but can try http://localhost:5000/?split_test[client_navigation_v5]=experiment to get into experiment group) 

Also fixed the issue where if the user mouses out of the "Artworks" nav item the full-screen menu goes away buy positioning an invisible box. Since users are going to want to get their cursor pointer to the far left of the menu in the most natural UX way this was necessary, otherwise they would have to go down then left, very exactly. The way I did this is very hacky, but not sure of a better way for the time being. It gets the job done: 

hover


src/Components/NavBar/NavItem.tsx
```diff
@@ -115,10 +117,31 @@ export const NavItem: React.FC<NavItemProps> = ({
...
+            zIndex: 1,
+            width: 300,
+            height: 50,
+            marginTop: -50, // using margin because we can't use relative positioning here
+            marginLeft: -220,
+            opacity: 0,
+          }}
+        />
+      )}
+
```
xtina-starr, line 140: I guess this wouldn't be necessary for the Artists drop down as well?

damassi: Nope, because it makes sense that if the user mouses over to the Artworks nav that a new nav would appear. Not so with angularly mousing over a wide search bar.

xtina-starr: Makes sense!

xtina-starr:
Thanks for this!

artsyit:
:rocket: PR was released in [`v26.27.11`](https://github.com/artsy/reaction/releases/tag/v26.27.11) :rocket:

src/Components/NavBar/NavItem.tsx
anandaroop: Crafty! 😄

PR #3457, opened 2020-04-28
[FX-1925] Fix some minor issues and launch nav drop down menus

xtina-starr:
Addresses: [FX-1925](https://artsyproduct.atlassian.net/browse/FX-1925)

- Fixes analytics hover tracking for Artworks & Artists drop downs.
- Removes the feature flag so the Artworks & Artists menus are visible to all users.
- Fixes a NavBar test.

**Analytics Before:**
Screen Shot 2020-04-28 at 4 38 03 PM

**Analytics After:**
Screen Shot 2020-04-28 at 4 37 34 PM

cc: @abhitip

### DIFF
--- a/src/Components/NavBar/NavItem.tsx
+++ b/src/Components/NavBar/NavItem.tsx
@@ -73,7 +73,7 @@ export const NavItem: React.FC<NavItemProps> = ({
     if (isString(navItemLabel) || label)
       trackEvent({
         action_type: AnalyticsSchema.ActionType.Hover,
-        subject: navItemLabel.toString() || label,
+        subject: label || navItemLabel.toString(),
       })
   }
 
--- a/src/Components/NavBar/__tests__/NavBar.test.tsx
+++ b/src/Components/NavBar/__tests__/NavBar.test.tsx
@@ -51,8 +51,8 @@ describe("NavBar", () => {
 
   describe("desktop", () => {
     const defaultLinks = [
-      ["/collect", "Artworks"],
-      ["/artists", "Artists"],
+      [undefined, "ArtworksReveal more"],
+      [undefined, "ArtistsReveal more"],
       ["/auctions", "Auctions"],
       ["/articles", "Editorial"],
     ]
--- a/src/Components/NavBar/NavBar.tsx
+++ b/src/Components/NavBar/NavBar.tsx
@@ -62,7 +62,6 @@ export const NavBar: React.FC = track(
     user,
     "User Conversations View"
   )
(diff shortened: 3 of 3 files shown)


src/Components/NavBar/__tests__/NavBar.test.tsx
```diff
@@ -51,8 +51,8 @@ describe("NavBar", () => {
 
   describe("desktop", () => {
     const defaultLinks = [
-      ["/collect", "Artworks"],
-      ["/artists", "Artists"],
+      [undefined, "ArtworksReveal more"],
+      [undefined, "ArtistsReveal more"],
```
xtina-starr, line 55: Two things regarding this test: 
1) For the urls, Artworks and Artists no longer have  a`href`. 
2) The "Reveal more" gets appended on because of the [`ChevronIcon` component's](https://github.com/artsy/palette/blob/d4acee6bbdf69269b0f6ec12702c40f4608345cb/packages/palette/src/svgs/ArrowDownIcon.tsx#L7) `<Title>` . I considered conditionally rendering that in palette but wondered if we had that there for SEO purposes. I'm open to a better way of handling that for this test.

src/Components/NavBar/NavItem.tsx
```diff
@@ -73,7 +73,7 @@ export const NavItem: React.FC<NavItemProps> = ({
     if (isString(navItemLabel) || label)
       trackEvent({
         action_type: AnalyticsSchema.ActionType.Hover,
-        subject: navItemLabel.toString() || label,
+        subject: label || navItemLabel.toString(),
```
xtina-starr, line 76: @damassi I remember why I initially checked for the label first in a previous PR. For the Artworks & Artists drop downs the child of `NavItem` is not a string. We pass `label` in as a prop in order to check the hover of the label.

src/Components/NavBar/__tests__/NavBar.test.tsx
sweir27: These changes make sense to me-- I think it's fine to keep the `Title` text there!

Eventually I think we'll add back the `href`s on those items, so this sets us up well for that future.

src/Components/NavBar/NavItem.tsx
damassi: Yah I noticed the "label" prop last night, that actually clears a lot of little things up versus having to check `parentNode` and so on. It might be good to dig through navbar code and see where we can leverage label instead.

artsyit:
:rocket: PR was released in [`v26.28.2`](https://github.com/artsy/reaction/releases/tag/v26.28.2) :rocket:

PR #3460, opened 2020-04-28
Improves server-side render of collect header

dzucconi:
Re: [FX-1842](https://artsyproduct.atlassian.net/browse/FX-1842)

This is an ugly diff for what it does — there's just a lot of moving around blocks.

* Removes the break-out-of-app-container CSS hack
* Renders all available artworks on initial pass rather than just 2
* Some light refactoring and removal of `any`s

I do think this could be a lot nicer in how it's structured/named. 🤷 

-----


src/Apps/Collect/Routes/Collection/Components/Header/DefaultHeader.tsx
```diff
@@ -1,181 +1,130 @@
...
       fragment DefaultHeader_headerArtworks on FilterArtworksConnection {
         edges {
           node {
-            href
-            slug
+            id
             image {
-              small: resized(height: 160) {
-                url
+              # TODO: Should accept arguments
```
dzucconi, line 114: I need to look over the docs on fragment args. In practice this isn't an issue since if the parent passes different args Relay errors, but...

src/Apps/Collect/Routes/Collection/index.tsx
```diff
@@ -82,84 +82,91 @@ export class CollectionApp extends Component<CollectionAppProps> {
...
+        {artworksConnection && (
+          <SeoProductsForArtworks artworks={artworksConnection} />
+        )}
+        {artworksConnection && (
+          <SeoProductsForCollections
+            descending_artworks={descending_artworks}
+            ascending_artworks={ascending_artworks}
+            collectionDescription={description}
+            collectionURL={collectionHref}
+            collectionName={title}
```
dzucconi, line 108: Preference — I like to move meta tags outside the app component into a fragment just so I *know* they aren't visual.

src/Apps/__stories__/Collect.story.tsx
```diff
@@ -32,9 +32,6 @@ storiesOf("Apps/Collect/Collection", module)
   })
   .add("Collection with Hub rails", () => {
     return (
-      <MockRouter
-        routes={collectRoutes}
-        initialRoute="/collection/contemporary"
```
dzucconi, line 37: Had to change the sample route cuz I couldn't look at the work of a certain someone. 😄

dzucconi:
Oh yeah: @damassi I wound up not using the Carousel because: well, there's no reason to if they aren't navigable, and there's some bugs that I didn't look into where when the amount of data changes it doesn't update. Didn't really look into why because no need.

src/Apps/Collect/Routes/Collection/Components/Header/DefaultHeaderArtwork.tsx
```diff
@@ -0,0 +1,87 @@
...
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.Click,
+      context_module: AnalyticsSchema.ContextModule.ArtworkBanner,
+      context_page_owner_type: AnalyticsSchema.OwnerType.Collection,
+      context_page: AnalyticsSchema.PageName.CollectionPage,
+      context_page_owner_id: collectionId,
+      context_page_owner_slug: collectionSlug,
+      destination_path: artwork.node.href,
+    })
+  }, [artwork, trackEvent])
```
damassi, line 42: @dzucconi - I've noticed this in a lot of your other work, but I was wondering if you could give a couple words on when to use `useCallback`. I've always thought about it as an opt-in for performance optimization of slow code, but I could also see it being used all the time, so that later optimizations don't need to take place. Wondering what yr thinking is.

src/Apps/Collect/Routes/Collection/Components/Header/__tests__/__snapshots__/DefaultHeader.test.tsx.snap
```diff
@@ -1,87 +1,180 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`default header component renders a snapshot 1`] = `
-<header>
+<div
```
damassi, line 4: Unless you feel strongly we should delete this snapshot in favor of assertions against actual content. Across Eigen and web there's been movement towards avoiding snapshot tests when possible.

src/Apps/Collect/Routes/Collection/Components/Header/index.tsx
```diff
@@ -22,14 +12,23 @@ import {
...
 } from "@artsy/palette"
-
-import { AuthIntent, ContextModule } from "@artsy/cohesion"
 import { Header_artworks } from "__generated__/Header_artworks.graphql"
 import { Header_collection } from "__generated__/Header_collection.graphql"
+import { CollectionDefaultHeaderFragmentContainer as CollectionDefaultHeader } from "Apps/Collect/Routes/Collection/Components/Header/DefaultHeader"
+import { HorizontalPadding } from "Apps/Components/HorizontalPadding"
 import { useSystemContext } from "Artsy"
+import { AnalyticsSchema } from "Artsy/Analytics"
+import { unica } from "Assets/Fonts"
```
damassi, line 21: If this is being used somewhere in the file it seems problematic. All type should flow through palette's typography components directly.

src/Apps/Collect/Routes/Collection/Components/Header/__tests__/__snapshots__/DefaultHeader.test.tsx.snap
dzucconi: Ah no yeah I never use them personally, just was updating what's here. 👍

src/Apps/Collect/Routes/Collection/Components/Header/index.tsx
dzucconi: Will do (these weren't actually added, just resorted)

src/Apps/Collect/Routes/Collection/Components/Header/__tests__/fixtures/artworks.ts
```diff
@@ -1,4 +1,4 @@
-import { DefaultHeader_headerArtworks } from "__generated__/DefaultHeader_headerArtworks.graphql"
+// import { DefaultHeader_headerArtworks } from "__generated__/DefaultHeader_headerArtworks.graphql"
```
dzucconi, line 1: Woops, yeah I could probably use a quick pair session on updating fragments and types for fixtures.

src/Apps/Collect/Routes/Collection/Components/Header/DefaultHeaderArtwork.tsx
dzucconi: I got in the habit of just using `useMemo` and `useCallback` everywhere just by default. But I think that's more to do with my mental model of hooks which might just be plain wrong. https://kentcdodds.com/blog/usememo-and-usecallback 🤔

damassi:
Just to confirm, this looks / functions good with JS disabled, and when enabled we no longer see the staggering in from left to right on load? 

Here's an illustration of the current problems on prod: 

#### JS Enabled: 
collection

#### Two image issue (looks like that's been addressed) 
ssr

#### SSR, JS disabled (only two images -- but that would be fixed with above I assume): 
ssr-2

dzucconi:
@damassi yeah indeed it does: basically just renders the full rail of 12 artworks on initial load and then trim or pad-out, depending, once we have the actual viewport width. I'll post a screen capture in a bit.

dzucconi:
JS Disabled:

damassi:
@dzucconi - i'm overjoyed ❤️ Collection pages are quite visible these days and will be even more so once the new full-screen nav launches so fix goes a long long way UI wise.

src/Apps/Collect/Routes/Collection/Components/Header/DefaultHeaderArtwork.tsx
damassi: This paragraph from that link is what I tend to think about until I've experientially encountered slow code:

> Specifically the cost for useCallback and useMemo are that you make the code more complex for your co-workers, you could make a mistake in the dependencies array, and you're potentially making performance worse by invoking the built-in hooks and preventing dependencies and memoized values from being garbage collected. Those are all fine costs to incur if you get the performance benefits necessary, but it's best to measure first.

ashleyjelks:
> Oh yeah: @damassi I wound up not using the Carousel because: well, there's no reason to if they aren't navigable, and there's some bugs that I didn't look into where when the amount of data changes it doesn't update. Didn't really look into why because no need.

This looks great! On this, I think there were plans to discuss with the design team about making the works clickable through to the artworks page. But I think it would be pretty easy to link the default header artworks to their respective artworks page if the designers decide they want that change implemented.

dzucconi:
> But I think it would be pretty easy to link the default header artworks to their respective artworks page if the designers decide they want that change implemented.

Ah, they still click thru to the artwork page, just not navigable in the pagination, left/right sense.

src/Apps/Collect/Routes/Collection/Components/Header/DefaultHeaderArtwork.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  return (
+    <Link
```
damassi, line 45: This should use a [`<RouterLink to="...">`](https://github.com/artsy/reaction/blob/master/src/Artsy/Router/RouterLink.tsx) so that we can tap into the new app shell.

src/Apps/Collect/Routes/Collection/Components/Header/__tests__/fixtures/artworks.ts
```diff
@@ -64,14 +65,13 @@ export const collectionHeaderArtworks: Header_artworks = {
   ],
 }
 
-export const defaultCollectionHeaderArtworks: Header_artworks &
-  DefaultHeader_headerArtworks = {
-  " $refType": null as never,
-  " $fragmentRefs": null as never,
-  merchandisableArtists: null,
+export const defaultCollectionHeaderArtworks: {
+  edges: Array<DeFraged<DefaultHeaderArtwork_artwork>>
```
dzucconi, line 69: This `DeFraged` should be a DeepOmit. We kinda should just install ts-essentials me-thinks.

artsyit:
:rocket: PR was released in [`v26.29.1`](https://github.com/artsy/reaction/releases/tag/v26.29.1) :rocket:

PR #3469, opened 2020-04-30
Correct findCurrentRoute logic to unbreak artist page

zephraph:
The acceptance tests started failing for force on a reaction renovate PR. @xtina-starr dug in and found out that the artist page was broken because the header wasn't being rendered. She dug in and noticed that it was because     `route.displayNavigationTabs` wasn't being set. 

When I started investigating, it turns out that when I updated the logic for `findCurrentRoute` I broke things. I updated it to just use `match.route`, but that's always just the top level route it seems. I updated it logic of that function to actually walk through the route indices and return the properly nested route. I also wrote a test. 

<details>
  <summary>📦 Published PR as canary version: <code>26.30.2-canary.3469.59299.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@26.30.2-canary.3469.59299.0
  # or 
  yarn add @artsy/reaction@26.30.2-canary.3469.59299.0
  ```
</details>

### DIFF
--- a/src/Artsy/Router/Utils/__tests__/findCurrentRoute.test.ts
+++ b/src/Artsy/Router/Utils/__tests__/findCurrentRoute.test.ts
@@ -1,0 +1,54 @@
+import { findCurrentRoute } from "../findCurrentRoute"
+import { Match, Router, Location } from "found"
+
+describe("findCurrentRoute", () => {
+  const getMatch = (config: Partial<Match>): Match => ({
+    routes: [],
+    router: ({} as unknown) as Router,
+    context: {},
+    routeIndices: [0],
+    location: ({} as unknown) as Location,
+    params: {},
+    ...config,
+  })
+
+  it("should return a base nested route", () => {
+    const match = getMatch({
+      routeIndices: [0, 0],
+      routes: [
+        {
+          path: "/artist/:artistID",
+          children: [
+            {
+              path: "/",
+            },
+          ],
+        },
+      ],
+    })
+    expect(findCurrentRoute(match)).toHaveProperty("path", "/")
+  })
+
+  it("should return a non-zero nested route", () => {
+    const match = getMatch({
+      routeIndices: [0, 2],
+      routes: [
+        {
+          path: "/artist/:artistID",
+          children: [
+            {
+              path: "/",
+            },
+            {
+              path: "/works-for-sale",
+            },
+            {
+              path: "/cv",
+            },
+          ],
+        },
+      ],
+    })
+    expect(findCurrentRoute(match)).toHaveProperty("path", "/cv")
+  })
+})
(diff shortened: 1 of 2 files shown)


src/Artsy/Router/Utils/__tests__/findCurrentRoute.test.ts
```diff
@@ -0,0 +1,54 @@
+import { findCurrentRoute } from "../findCurrentRoute"
+import { Match, Router, Location } from "found"
+
+describe("findCurrentRoute", () => {
```
damassi, line 4: Thanks for the test!

src/Artsy/Router/Utils/__tests__/findCurrentRoute.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    })
+    expect(findCurrentRoute(match)).toHaveProperty("path", "/")
+  })
+
+  it("should return a non-zero nested route", () => {
+    const match = getMatch({
+      routeIndices: [0, 2],
+      routes: [
+        {
+          path: "/artist/:artistID",
```
damassi, line 37: Can we add one more test that asserts against 

```
/artist/:artistID/baz
```
I'm not aware of any routes that are more than three paths deep but this would cover all of our bases.

src/Artsy/Router/Utils/findCurrentRoute.tsx
```diff
@@ -1,16 +1,16 @@
 import { Match, RouteConfig } from "found"
 
 export const findCurrentRoute = ({
-  route,
   routes,
   routeIndices,
 }: Match & { route?: RouteConfig }) => {
-  if (route) {
-    return route
+  console.log(routes, routeIndices)
```
zephraph, line 7:
```suggestion
```

damassi: Stray

src/Artsy/Router/Utils/__tests__/findCurrentRoute.test.ts
zephraph: Wait, `/artist/:artistID/cv` is tested here. Do you want one a level deeper?

damassi: baz (with children)

zephraph: Okay, I added a test for `/artist/:artistID/foo/bar` and `/bar` has children that we _don't_ want to match against.

artsyit:
:rocket: PR was released in [`v26.30.2`](https://github.com/artsy/reaction/releases/tag/v26.30.2) :rocket:

PR #3470, opened 2020-04-30
[Chore] Turns on rules-of-hooks

dzucconi:
Before merging there's one issue I've disabled inline that we can address in a few ways; will leave a comment there.

Do you think test coverage is enough to be sure here? Should I integrate with force and just try to track down each component? (I'm not terribly concerned about any of these changes, but...)


src/Apps/FeatureAKG/Components/FeaturedArticles.tsx
```diff
@@ -31,14 +31,11 @@ interface FeaturedArticlesProps {
 const FeaturedArticles: React.FC<FeaturedArticlesProps> = props => {
   const articles = [...props.articles]
 
-  if (!articles?.length) {
-    return null
-  }
-
   const firstArticle = articles.shift()
   const firstArticleImage = firstArticle?.thumbnailImage?.cropped
```
dzucconi, line 35: This is really the only line here that would be suspect but it's already guarded.

src/Utils/Hooks/useMedia.ts
```diff
@@ -26,6 +26,8 @@ export function useMedia(): { [k in Breakpoint]?: boolean } {
     (acc, [key, value]) => {
       return {
         ...acc,
+        // TODO:
+        // eslint-disable-next-line react-hooks/rules-of-hooks
```
dzucconi, line 30: OK So this guy here. It looks like it's just being used in a single component, which I can just rewrite to use the useMatchMedia hook and then remove this one (it doesn't look like it's used in force?) Otherwise we should refactor these two a bit.

src/Components/NavBar/Menus/MobileNavMenu/MobileNavMenu.tsx
```diff
@@ -126,7 +126,7 @@ const Menu: React.FC<MenuProps> = ({
 export const BackLink: React.FC = () => {
   const { trackEvent } = useTracking()
   const { pop } = useNavigation()
-  const contextModule = getTrackingContextModule()
+  const contextModule = useTrackingContextModule()
```
dzucconi, line 129: Just a naming thing.

src/Apps/FeatureAKG/Components/FeaturedArticles.tsx
```diff
@@ -47,6 +44,10 @@ const FeaturedArticles: React.FC<FeaturedArticlesProps> = props => {
     })
   }
 
+  if (articles.length === 0) {
```
damassi, line 47: We should keep the check against `null` articles, that was deliberate.

src/Utils/Hooks/useMedia.ts
damassi: If its only being used once feel free to remove 👍

src/Apps/FeatureAKG/Components/FeaturedArticles.tsx
dzucconi: It'd never be `null` though. If you look above the object is being spread into a new array (which is admittedly weird in itself).

dzucconi: I guess so that it can mutate it with shift (why do that though).
(That object itself is a non-null array according to the typing at least)

damassi: In that case, all good 👍

src/Components/NavBar/NavBar.tsx
```diff
@@ -55,7 +56,8 @@ export const NavBar: React.FC = track(
   const { trackEvent } = useTracking()
   const { mediator, user, EXPERIMENTAL_APP_SHELL } = useContext(SystemContext)
   const [showMobileMenu, toggleMobileNav] = useState(false)
-  const { xs, sm } = useMedia()
+  const xs = useMatchMedia(themeProps.mediaQueries.xs)
+  const sm = useMatchMedia(themeProps.mediaQueries.sm)
```
dzucconi, line 60: cc @damassi

dzucconi:
Looks good in Force. Merging.

artsyit:
:rocket: PR was released in [`v26.30.3`](https://github.com/artsy/reaction/releases/tag/v26.30.3) :rocket:

PR #3472, opened 2020-04-30
[mWeb] Show OTP prompt during login when 2FA-enabled

dblandin:
Add an OTP prompt to the mWeb login modal.

<details>
  <summary>📦 Published PR as canary version: <code>26.30.3-canary.3472.59408.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/reaction@26.30.3-canary.3472.59408.0
  # or 
  yarn add @artsy/reaction@26.30.3-canary.3472.59408.0
  ```
</details>

### Screen Recording

Peek 2020-04-30 18-32

### DIFF
--- a/src/Components/Authentication/Mobile/LoginForm.tsx
+++ b/src/Components/Authentication/Mobile/LoginForm.tsx
@@ -22,15 +22,81 @@ import {
 } from "../commonElements"
 import { FormProps, InputValues, ModalType } from "../Types"
 import { MobileLoginValidator } from "../Validators"
+import { StepElement } from "Components/Wizard/types"
+
+interface LoginFormState {
+  advanceWizardStep: boolean
+  error: string
+  hideBackButton: boolean
+  steps: StepElement[]
+}
 
 class MobileLoginFormWithSystemContext extends Component<
   FormProps & {
     relayEnvironment: Environment
-  }
+  },
+  LoginFormState
 > {
+  static getDerivedStateFromProps(nextProps, prevState): LoginFormState | null {
+    if (nextProps.error !== prevState.error) {
+      if (nextProps.error === "missing two-factor authentication code") {
+        return {
+          advanceWizardStep: true,
+          error: nextProps.error,
+          hideBackButton: true,
+          steps: prevState.steps.concat([
+            MobileLoginFormWithSystemContext.buildOtpStep(),
+          ]),
+        }
+      } else {
+        return {
+          ...prevState,
+          error: nextProps.error,
+        }
+      }
+    }
+
+    return null
+  }
+
+  static buildOtpStep = () => {
+    return (
+      <Step validationSchema={MobileLoginValidator.otpAttempt}>
+        {({
+          form: { errors, values, handleChange, handleBlur, setTouched },
+        }) => {
+          return (
+            <QuickInput
+              block
+              error={errors.otp_attempt}
+              name="otp_attempt"
+              placeholder="Authentication Code"
+              value={values.otp_attempt}
+              onChange={handleChange}
+              onBlur={handleBlur}
+              setTouched={setTouched}
+              touchedOnChange={false}
+            />
+          )
+        }}
+      </Step>
+    )
+  }
+
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      error: null,
+      advanceWizardStep: false,
+      hideBackButton: false,
+      steps: this.buildBaseSteps(),
+    }
+  }
+
   showError = status => {
-    const { error } = this.props
-    if (error) {
+    const { error } = this.state
+    if (error && error !== "missing two-factor authentication code") {
       return <Error show>{error}</Error>
     }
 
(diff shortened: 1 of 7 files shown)


src/Components/Authentication/Mobile/LoginForm.tsx
```diff
@@ -118,29 +176,42 @@ class MobileLoginFormWithSystemContext extends Component<
...
+            form: { handleSubmit, actions, values, status, isSubmitting },
           } = context
-          const { currentStep, isLastStep } = wizard
+
+          const { currentStep, isLastStep, next } = wizard
+
+          if (advanceWizardStep) {
+            next(values, actions)
+            this.setState({ advanceWizardStep: false })
+          }
```
dblandin, line 197: Having a state change within the `render` function doesn't feel right but I'm not sure how to trigger this step advance otherwise.

src/Components/Authentication/Mobile/LoginForm.tsx
```diff
@@ -22,15 +22,81 @@ import {
...
 class MobileLoginFormWithSystemContext extends Component<
   FormProps & {
     relayEnvironment: Environment
-  }
+  },
+  LoginFormState
 > {
+  static getDerivedStateFromProps(nextProps, prevState): LoginFormState | null {
+    if (nextProps.error !== prevState.error) {
+      if (nextProps.error === "missing two-factor authentication code") {
```
dblandin, line 42: This is brittle and it would be better for Gravity to return a stable error code in addition to the human-friendly error message. Doesn't have to be numeric. Could be `missing_otp` or `otp_required`.

src/Components/Wizard/Wizard.tsx
```diff
@@ -125,20 +125,20 @@ export class Wizard extends React.Component<WizardProps, WizardState> {
     if (this.isLastStep) {
       onComplete && onComplete(values, actions)
     } else {
-      actions && actions.setSubmitting(false)
+      actions?.setSubmitting(false)
       // If exist, call onSubmit on the current step
       const onSubmit = this.currentStep.props.onSubmit
       if (onSubmit) {
-        actions.setSubmitting(true)
+        actions?.setSubmitting(true)
```
dblandin, line 132: Ran into the following error while executing the spec without these undefined guards. Not sure how to mock this formik behavior.

```
    TypeError: Cannot read property 'setSubmitting' of undefined

      130 |       const onSubmit = this.currentStep.props.onSubmit
      131 |       if (onSubmit) {
    > 132 |         actions.setSubmitting(true)
          |                 ^
      133 |         const result = onSubmit(values, actions)
      134 | 
      135 |         if (result instanceof Boolean) {

      at setSubmitting (src/Components/Wizard/Wizard.tsx:132:17)
      at next (src/Components/Authentication/Mobile/LoginForm.tsx:195:13)
```

dblandin:
@erikdstock I noticed you had originally built the Wizard interface (:clap:). Requesting your review in case you have ideas on how to better handle this conditional step + advance behavior.

src/Components/Wizard/Wizard.tsx
anandaroop: A safe addition in any case 👍

src/Components/Authentication/Mobile/LoginForm.tsx
anandaroop: I wonder if we could replace the `:invalid_grant` error code that Gravity [currently creates](https://github.com/artsy/gravity/blob/ebe4ecfa22422ae10a4489ed7a2d11bfac67e33e/app/controllers/oauth_controller.rb#L147-L148) with `:otp_required` and `:otp_invalid` — assuming that the error code could even bubble down this far.

(The OAuth spec seems to stipulate that it's [permissible to create new codes](https://tools.ietf.org/html/rfc6749#section-8.5) subject to a few restrictions)

src/Components/Authentication/Mobile/LoginForm.tsx
anandaroop: Yeah having this "impure" render definitely feels weird, but I'm not seeing a better solution either.

src/Components/Authentication/Validators.tsx
```diff
@@ -12,6 +12,8 @@ const password = Yup.string()
 
 const loginPassword = Yup.string().required("Password required")
 
+const otpAttempt = Yup.string().required("Enter a code")
```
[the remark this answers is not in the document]
dblandin: > Not sure if the backup codes have the same format

That's the thing. Backup codes have a different format and include letters+digits. So I don't think we can have much validation beyond this.

Screenshot from 2020-05-01 13-47-54

src/Components/Authentication/Mobile/LoginForm.tsx
dblandin: For sure. Let's follow-up with this :+1:

artsyit:
:rocket: PR was released in [`v26.30.4`](https://github.com/artsy/reaction/releases/tag/v26.30.4) :rocket:

dblandin:
Looks like force actually uses the formik actions vs the error prop to pass in the server error:

https://github.com/artsy/force/blob/51733c03c8e490fa8ff705268f86c472412890c1/src/desktop/apps/authentication/helpers.ts#L90-L96

So I'll have to update this handling to consider the form state as well. I wonder where we use that `error` prop now... :thinking:

PR #3473, opened 2020-04-30
[Chore] Turns on exhaustive-deps

dzucconi:
Sets it to `warn` which is recommended.

There's 13 warnings right now, and I do think they could be auto-fixed. But I have seen missing dependencies paper over underlying bugs before so I'm reluctant to change them all at once.


artsyit:
:rocket: PR was released in [`v26.30.4`](https://github.com/artsy/reaction/releases/tag/v26.30.4) :rocket:

PR #3479, opened 2020-05-05
Fixes swallowed assertions in async tests

dzucconi:
Not 100% sure if I caught them all because it's a bit weird to search for —  but there's a few specs that were just being papered over inside of setTimeouts without actually waiting. So this fixes the specs that I came across. Notably I couldn't get one of them to pass no matter how long I waited — left a comment.


dzucconi:
I *know* this is preaching to the choir but if not: never trust a test you write that you don't see fail first! "red, green, refactor"

src/Apps/Artwork/Components/__tests__/PricingContextModal.test.tsx
```diff
@@ -27,66 +29,86 @@ describe("PricingContextModal", () => {
...
+
+    button.simulate("click")
+
+    await flushPromiseQueue()
+    component.update()
+
+    // TODO: Can't get this to update/pass
+    // expect(component.text()).not.toContain(
+    //   "This feature aims to provide insight into the range of prices for an artist's works and allow buyers to discover other available works by the artist at different price points."
+    // )
```
dzucconi, line 111: Not sure what's happening here. This works in storybooks. So... should be OK. But there you have it.

src/Apps/Artwork/Components/__tests__/PricingContext.test.tsx
```diff
@@ -13,6 +13,7 @@ import {
   PricingContext,
   PricingContextFragmentContainer,
 } from "../PricingContext"
+import { flushPromiseQueue } from "Utils/flushPromiseQueue"
```
damassi, line 16: I'm pretty sure this is only used in our tests. Mind moving to our [`DevTools`](https://github.com/artsy/reaction/tree/master/src/DevTools) folder?

dzucconi: Sure thing.

dzucconi: Moved

src/Apps/Artist/Components/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
```diff
@@ -136,6 +136,7 @@ describe("CollectionsRail", () => {
           subject: "clicked next button",
           type: "Button",
         })
+        done()
```
pepopowitz, line 139: Is there a reason this test uses the `done()` mechanism versus marking this test `async` and doing `await flushPromiseQueue()` before the assertion, which the rest do?

I wouldn't hold up a merge for this, but if it is an arbitrary choice, I'd prefer consistency with the other tests to increase the likelihood of that pattern being followed.

src/Apps/Artwork/Components/__tests__/PricingContextModal.test.tsx
```diff
@@ -14,8 +15,9 @@ jest.mock("sharify", () => ({
 }))
 
 describe("PricingContextModal", () => {
-  it("renders with the modal closed", async () => {
```
pepopowitz, line 17: good catch

src/Apps/Artist/Components/ArtistCollectionsRail/__tests__/ArtistCollectionsRail.test.tsx
dzucconi: Eh, mostly just because I left it after I noticed how many were there. There are *alot* of tests that use setTimeout (with varying amounts of time). I think the async/await pattern is a lot more legible but it'd be a fair amount more work that warrants it's own PR.

artsyit:
:rocket: PR was released in [`v26.31.4`](https://github.com/artsy/reaction/releases/tag/v26.31.4) :rocket:

PR #3484, opened 2020-05-06
Deploys Storybook to S3

dzucconi:
Would also be nice to get back PR-specific deploys — are we doing that anywhere else?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,17 @@ orbs:
   yarn: artsy/yarn@4.0.2
   codecov: codecov/codecov@1.0.5
   auto: artsy/auto@1.2.0
+  aws-s3: circleci/aws-s3@1.0.15
+
+jobs:
+  upload-storybook:
+    docker:
+      - image: "cimg/python:3.8"
+    steps:
+      - aws-s3/sync:
+          arguments: --acl public-read
+          from: storybook_build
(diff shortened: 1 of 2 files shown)


zephraph:
I don't think we're doing PR specific deploys anywhere. We could build that out, but it'd take a bit of finagling.

.circleci/config.yml
```diff
@@ -54,3 +65,24 @@ workflows:
             - lint-changed
             - yarn/type-check
             - yarn/update-cache
+      - yarn/run:
+          name: build-docs
```
zephraph, line 69:
```suggestion
          name: build-storybook
```

.circleci/config.yml
```diff
@@ -54,3 +65,24 @@ workflows:
...
+                root: .
+                paths:
+                  - storybook_build
+          filters:
+            branches:
+              only: master
+      - upload-docs:
+          context: static-sites-uploader
+          requires:
+            - build-docs
```
zephraph, line 82:
```suggestion
            - build-storybook
```

not really a necessary change, but it'd be a little more clear

.circleci/config.yml
dzucconi: 👍

artsyit:
:rocket: PR was released in [`v26.31.4`](https://github.com/artsy/reaction/releases/tag/v26.31.4) :rocket:

PR #3486, opened 2020-05-06
Fix auction results layout on mobile

zephraph:
Before

image

After

image

I don't know exactly _why_ this broke. I'm assuming some extra spacing squeezed through somewhere that caused the column layout to fail. I ripped all that out and simplified the layout.

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
@@ -207,56 +207,34 @@ const ExtraSmallAuctionItem: SFC<Props> = props => {
   )
 
   return (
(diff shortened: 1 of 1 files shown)


sweir27:
Was going to ask if you had any clues as to why the `Col` s didn't work, but this change seems good to me! #mergeongreen

zephraph:
Unfortunately I don't. I think the primary consideration here is that those cols are for building responsive layouts and they're intentionally designed to overflow/wrap to the next line if there isn't enough space. That's not really ever something we want w/ this particular view so I think they're best avoided. 

Also, just as a small aside, it feels to me like it made the overall layout harder to understand.

artsyit:
:rocket: PR was released in [`v26.31.5`](https://github.com/artsy/reaction/releases/tag/v26.31.5) :rocket:

PR #3494, opened 2020-05-07
Updates mobile artwork filters layout

dzucconi:
Re: [FX-1938](https://artsyproduct.atlassian.net/browse/FX-1938)

Makes the mobile filters fullscreen and moves the "Apply" button to the bottom. Note that here I've just renamed "Cancel" to "Close" for the time being since we don't have a real notion of cancelling. If we want a proper cancel we'll have to just hold off on this until we can refactor the filters to actually be staged.

-----

Note that I'm punting on the scroll lock, which is a separate issue — refactoring the underlying nature of the Modal is... problematic. Closed PR here: https://github.com/artsy/palette/pull/675 — I'm thinking we just pull off `ModalBase` as it's own thing and use that here (and going forward to build Modals and slowly refactor the old one but I'm not thrilled by this prospect).

### DIFF
--- a/src/Components/v2/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
+++ b/src/Components/v2/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
@@ -1,8 +1,11 @@
-import { Box, Button, Sans, Spacer } from "@artsy/palette"
-import { MobileTopBar } from "Components/MobileTopBar"
+import { Box, Button, Sans, Flex, color } from "@artsy/palette"
 import React, { SFC } from "react"
 import styled from "styled-components"
-import { useArtworkFilterContext } from "./ArtworkFilterContext"
+import {
+  useArtworkFilterContext,
+  initialArtworkFilterState,
+} from "./ArtworkFilterContext"
+import { isEqual, omit } from "lodash"
 
 export const ArtworkFilterMobileActionSheet: SFC<{
   children: JSX.Element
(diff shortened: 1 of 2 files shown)


src/Components/v2/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
```diff
@@ -1,36 +1,56 @@
...
   children: JSX.Element
   onClose: () => void
 }> = ({ children, onClose }) => {
   const filterContext = useArtworkFilterContext()
 
+  // TODO: `filterContext.hasFilters` seems to be what we want but is never actually `false`?
+  const isReset = isEqual(
+    omit(filterContext.filters, "reset", "keyword"),
+    initialArtworkFilterState
+  )
```
dzucconi, line 20: This is essentially accurate — the reset method resets to the `initialAuctionResultsFilterState` with the addition of `reset: true`, the keyword is omitted because it's not reflected in this UI, but resetting also removes the keyword.

src/Components/v2/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
```diff
@@ -1,36 +1,56 @@
...
+} from "./ArtworkFilterContext"
+import { isEqual, omit } from "lodash"
 
 export const ArtworkFilterMobileActionSheet: SFC<{
   children: JSX.Element
   onClose: () => void
 }> = ({ children, onClose }) => {
   const filterContext = useArtworkFilterContext()
 
+  // TODO: `filterContext.hasFilters` seems to be what we want but is never actually `false`?
```
sweir27, line 16: Did you dig at all? It looks like it's set up with tests and everything, so I'm surprised that it's broken 😭 .

dzucconi: I'm not 100% sure on it's meaning. Though here it's a bit more particular since we're only really talking about what's in the mobile UI. I'll dig a little bit and remove this comment.

damassi: Here's an example where `hasFilters` is false: https://www.artsy.net/search?term=hhhhhhhhhhhhhhhh

It looks like that prop only applies to search pages [and the ZeroState](https://github.com/damassi/reaction/blob/917f59f79fac9af348e4944049b077a9e6aead86/src/Apps/Search/Components/ZeroState.tsx#L30) at the moment. (This ZeroState should be moved into the `Apps/Search/Components` folder at some point.)

dzucconi: OK, makes sense. Updated comment just to be clear.

artsyit:
:rocket: PR was released in [`v26.33.0`](https://github.com/artsy/reaction/releases/tag/v26.33.0) :rocket:

PR #3497, opened 2020-05-07
[ViewingRoom] Break apart components

damassi:
This PR breaks apart our components into more concrete chunks, also fleshes out the routing behavior better and adds two new hooks:
- `useRouter`
- `useIsRouteActive`


src/Apps/ViewingRoom/Routes/Works/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <ViewingRoomCarousel />
+      <Box mt={2} mb={4}>
+        <ViewingRoomArtworkDetails />
+      </Box>
+    </Box>
+  )
+}
+
+// Top-level route needs to be exported for bundle splitting in the router
+export default WorksRoute
```
dzucconi, line 21: 👍

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomCarousel.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <Flex
+      position="absolute"
+      width={50}
+      height="100%"
+      justifyContent="center"
+      alignItems="center"
+      style={{
+        backgroundColor: "#ffffff95",
```
dzucconi, line 60: oOoO hexalpha

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
```diff
@@ -0,0 +1,69 @@
...
+const ImageArea = styled(Box).attrs({
+  width: "50%",
+})`
+  overflow: hidden;
+`
+
+const ContentArea = styled(Flex).attrs({
+  alignItems: "center",
+  justifyContent: "center",
+  width: "50%",
```
damassi, line 59: Leaving as attrs as these will be responsive in next pr

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomArtworkDetails.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react"
+import { Box, Sans, Button, Serif } from "@artsy/palette"
+
+export const ViewingRoomArtworkDetails = props => {
+  return (
+    <Box width="50%" m="auto">
+      <Box>
+        <Sans size="3">Christine Sun Kim</Sans>
```
dzucconi, line 8: Do we do like `<Sans as="h2">` or anything like that in places?

src/Apps/ViewingRoom/Routes/Statement/Components/ViewingRoomStatements.tsx
```diff
@@ -0,0 +1,63 @@
...
+          different ways of viewing and interpreting bodies and the physical and
+          social spaces that govern our everyday lives as mutable, inclusive,
+          and open-ended. Kwak's work reimagines the form, functionality, and
+          materiality of objects in order to propose different ways of viewing
+          and interpreting bodies and the physical and social spaces that govern
+          our everyday lives as mutable, inclusive, and open-ended.
+        </Serif>
+      </Box>
+
+      <Spacer my={4} />
```
dzucconi, line 35: I do still think we should implement [a Stack](https://chakra-ui.com/stack) — I find this whole spacer pattern a lil weird. But maybe it's just unfamiliar. (Though I don't like all the extra DOM)

src/Apps/ViewingRoom/Components/ViewingRoomTabBar.tsx
```diff
@@ -0,0 +1,55 @@
...
+    </Flex>
+  )
+}
+
+const Tab: React.FC<LinkProps> = ({ children, to }) => {
+  const active = useIsRouteActive(to)
+
+  const borderBottom = active
+    ? `2px solid ${color("black100")}`
+    : `1px solid ${color("black10")}`
```
dzucconi, line 27: Does the change in border width cause the content to shift up a bit?

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <HeaderContainer>
+      <ImageArea>
+        <ResponsiveImage
+          src="https://user-images.githubusercontent.com/236943/81243255-342dcc00-8fc4-11ea-9a2b-2ab96ab67c9b.png"
+          style={{
+            backgroundSize: "cover",
+            backgroundPosition: "center",
+            height: "100%",
+          }}
+        />
```
dzucconi, line 17: We gotta fix the ResponsiveImage component to not use background images so we can support proper srcSet and (and it's just annoying)

dzucconi: [Here's the one I did recently](https://github.com/auspices/eos/blob/master/src/ResponsiveImage/ResponsiveImage.tsx)

damassi: Its been super super annoying through time...

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomArtworkDetails.tsx
damassi: Ah, nice call -- we can't do `as` for "reasons" but we have an "element" prop we can use instead.

src/Apps/ViewingRoom/Routes/Statement/Components/ViewingRoomStatements.tsx
damassi: Yeah, I used something like that [here](https://github.com/damassi/reaction/blob/27418dec416d0fe6125801609c786c9affa849e0/src/Apps/ViewingRoom/Routes/Statement/index.tsx#L15) but I think we adding Stack would improve our API a bit.

src/Apps/ViewingRoom/Components/ViewingRoomTabBar.tsx
damassi: Nope, not that I can see: https://staging.artsy.net/viewing-room

artsyit:
:rocket: PR was released in [`v26.34.3`](https://github.com/artsy/reaction/releases/tag/v26.34.3) :rocket:

PR #3499, opened 2020-05-08
[ViewingRoom] Connect to data

damassi:
This PR connects everything to MP and stubs out some tests: 

data


damassi:
@dzucconi - the jest / test job is going to fail here because i'm on a fork. This is related to a change that @zephraph made earlier today (he's gonna look into it tomorrow). It should be safe to merge unless there's feedback.

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
```diff
@@ -110,10 +148,10 @@ const Metadata: React.FC = props => {
         p={2}
       >
         <Sans size={["3", "4"]} color={["white100", "black100"]}>
-          Gallery Name
+          {name}
         </Sans>
         <Sans size={["3", "4"]} color={["white100", "black100"]}>
-          Closes in 5 days
+          {endAt}
```
dzucconi, line 154: Dunno if it's of any interest or we already have a way of formatting timestamps in this project? But we could [implement a date formatting field](https://github.com/auspices/atlas/blob/master/app/graphql/shared/timestamps.rb) like there is in Metaphysics so that we could do `endAt(relative: true)` or something.

artsyit:
:rocket: PR was released in [`v26.34.3`](https://github.com/artsy/reaction/releases/tag/v26.34.3) :rocket:

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
```diff
@@ -32,7 +59,7 @@ const ViewingRoomHeaderSmall = props => {
       }}
     >
       <ResponsiveImage
-        src="https://user-images.githubusercontent.com/236943/81243255-342dcc00-8fc4-11ea-9a2b-2ab96ab67c9b.png"
+        src={heroImageURL}
```
sweir27, line 62: Since this image can be _anything_, let's pop it through gemini's `resizer` before including here to make sure it's cached behind a cdn and sized correctly. (We should also `convert_to: "jpg"` for perf reasons).

Something like: https://github.com/artsy/reaction/blob/master/src/Apps/FeatureAKG/Components/Feature.tsx#L43-L46

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
damassi: Yeah this has been pretty random across our codebases but in general we've tried to hoist this stuff to MP. I don't (think?) we have a generic formatting field, but that would be nice. I left this as raw output here to remind me to update in MP.

dzucconi: [We've got one](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/fields/date.ts) though the format option should be documented with a description and it'd be nice to throw in a relative option too.

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
damassi: Will do 👍

PR #3502, opened 2020-05-08
adding pullQuote to VR

oxaudo:
So I think this blob of text is a pullQuote on the VR level. Reshuffled things here and created a component for that.

As for the images - none of the works from this artist have more than one image. But I'll try to think of something on staging...


src/Apps/ViewingRoom/Routes/Statement/ViewingRoomStatementRoute.tsx
```diff
@@ -34,6 +35,8 @@ const StatementRoute: React.FC<ViewingRoomStatementRouteProps> = ({
           >
             View works
           </Button>
+          <ViewingRoomPullQuote viewingRoom={viewingRoom} />
```
oxaudo, line 38: @damassi - I think it's logical to add spacing on this level and not inside the `quote` component itself. This way we can shuffle things around at that level not worrying of needing to change each component.

By that logic I also think that the `Button` right above should be part of `ViewingRoomWorks` since it should always follow the works row display and if we move those - the button should follow. No?

damassi: That makes sense!

artsyit:
:rocket: PR was released in [`v26.35.0`](https://github.com/artsy/reaction/releases/tag/v26.35.0) :rocket:

PR #3508, opened 2020-05-09
[Viewing Room] Add carousel support

damassi:
This adds carousel support to desktop / mobile. Also a lot of cleanup, scroll managment, etc. 

#### Desktop 

carousel

#### Mobile 

carousel-2

### DIFF
--- a/src/Apps/Artist/routes.tsx
+++ b/src/Apps/Artist/routes.tsx
@@ -135,6 +135,7 @@ export const routes: RouteConfig[] = [
           OverviewRoute.preload()
         },
         displayNavigationTabs: true,
+        ignoreScrollBehavior: true,
         query: graphql`
           query routes_OverviewQuery($artistID: String!) @raw_response_type {
             artist(id: $artistID) {
@@ -150,6 +151,7 @@ export const routes: RouteConfig[] = [
           WorksForSaleRoute.preload()
         },
         displayNavigationTabs: true,
+        ignoreScrollBehavior: true,
         query: graphql`
           query routes_WorksQuery(
             $acquireable: Boolean
@@ -229,6 +231,7 @@ export const routes: RouteConfig[] = [
           AuctionResultsRoute.preload()
         },
         displayNavigationTabs: true,
+        ignoreScrollBehavior: true,
         query: graphql`
           query routes_AuctionResultsQuery($artistID: String!) {
             artist(id: $artistID) {
--- a/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
+++ b/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
@@ -53,7 +53,7 @@ const ViewingRoomHeaderSmall: React.FC<ViewingRoomHeaderProps> = props => {
       alignItems="center"
       height={HeaderHeight}
       style={{
-        borderBottom: `1px solid ${color("black10")};`,
+        borderBottom: `1px solid ${color("black10")}`,
         overflow: "hidden",
         position: "relative",
       }}
--- a/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
+++ b/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
@@ -56,7 +56,9 @@ const ViewingRoomHeaderLarge: React.FC<ViewingRoomHeaderProps> = props => {
     >
       <Box width="50%" style={{ overflow: "hidden" }}>
         <ResponsiveImage
+          // FIXME: Use resizer: https://github.com/artsy/reaction/pull/3499/files#r422166275
           src={heroImageURL}
+          lazyLoad={false}
           style={{
             backgroundSize: "cover",
             backgroundPosition: "center",
--- a/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
+++ b/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
@@ -39,6 +39,51 @@ export const ViewingRoomHeaderFragmentContainer = createFragmentContainer(
   }
 )
 
+/**
+ * Header used for desktop layouts
+ */
+const ViewingRoomHeaderLarge: React.FC<ViewingRoomHeaderProps> = props => {
+  const {
+    viewingRoom: { heroImageURL, title },
+  } = props
+
+  return (
+    <Flex
+      style={{
+        height: `calc(100vh - ${NavBarHeight}px)`,
+        borderBottom: `1px solid ${color("black10")}`,
+      }}
+    >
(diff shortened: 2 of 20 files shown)


src/Components/Carousel.tsx
```diff
@@ -291,14 +317,7 @@ export class BaseCarousel<T> extends React.Component<
...
             this.flickity.previous()
-
-            if (onArrowClick) {
-              onArrowClick({
-                state: this.state,
-                props: this.props,
-                flickity: this.flickity,
-              })
-            }
+            setImmediate(this.handleArrowClick)
```
damassi, line 320: Noticed this bug -- we need to wait for flickity's internal state to finish updating

src/Artsy/Router/Utils/shouldUpdateScroll.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * Handler for `found-scroll` used for determining if scroll management should
+ * be applied on route change, based on keys attached to a route.
+ *
+ * @see https://github.com/4Catalyzer/found-scroll#custom-scroll-behavior
+ */
+export function shouldUpdateScroll(_prevRenderArgs, { routes }) {
+  try {
```
damassi, line 8: Don't want to crush a 404 with an error

src/Apps/Artist/routes.tsx
```diff
@@ -135,6 +135,7 @@ export const routes: RouteConfig[] = [
           OverviewRoute.preload()
         },
         displayNavigationTabs: true,
+        ignoreScrollBehavior: true,
```
damassi, line 138: @sweir27 - added better scroll management in this PR which fixed the artist tabs behavior -- now you can click them and don't get thrown back to the top of the page.

src/Apps/ViewingRoom/Routes/Statement/ViewingRoomStatementRoute.tsx
```diff
@@ -45,14 +29,16 @@ const StatementRoute: React.FC<ViewingRoomStatementRouteProps> = ({
...
-  viewingRoom: graphql`
-    fragment ViewingRoomStatementRoute_viewingRoom on ViewingRoom {
-      ...ViewingRoomIntro_viewingRoom
-      ...ViewingRoomWorks_viewingRoom
-      ...ViewingRoomPullQuote_viewingRoom
-      ...ViewingRoomSubsections_viewingRoom
-    }
-  `,
-})
+export const ViewingRoomStatementRouteFragmentContainer = createFragmentContainer(
```
damassi, line 32: Removed bundle splitting from the sub routes. Sometimes there's diminishing returns here, in the sense that a tiny snippet of page is probably just a few kb -- however, a fetch needs to happen, a mount needs to happen, etc. A lot is going on and can lead so some layout thrashing, especially if the MP request is so fast that it outpaces the bundle request happening in parallel, leading to a white flash. 

So bundle splitting: required at the top most route (e.g., `/viewing-room`) for the new app shell architecture, and then evaluated on a case-by-case basis for `children` routes. In many cases it's just not needed and can diminish the UX.

src/Apps/ViewingRoom/routes.tsx
```diff
@@ -39,6 +37,7 @@ export const routes: RouteConfig[] = [
       {
         path: "/works",
         Component: WorksRoute,
+        ignoreScrollBehavior: true,
```
damassi, line 40: I'm going to spend time this week and get proper types into our routes. Right now things are free-form.

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomCarousel.tsx
```diff
@@ -11,29 +20,77 @@ interface ViewingRoomCarouselProps {
...
+        <Carousel
+          options={{
+            cellAlign: "center",
+            draggable: showProgressBar,
+            freeScroll: true,
+            groupCells: 1,
+            pageDots: false,
+          }}
+          data={images}
+          height="550px"
```
dzucconi, line 47: Weird having to specify the height? — We don't have proper placeholders in here?

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
```diff
@@ -39,6 +39,53 @@ export const ViewingRoomHeaderFragmentContainer = createFragmentContainer(
...
+ */
+const ViewingRoomHeaderLarge: React.FC<ViewingRoomHeaderProps> = props => {
+  const {
+    viewingRoom: { heroImageURL, title },
+  } = props
+
+  return (
+    <Flex
+      style={{
+        height: `calc(100vh - ${NavBarHeight}px)`,
```
dzucconi, line 53: I never trust this sort of thing — this is just because we can't flex: 1 out the parent?

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomCarousel.tsx
damassi: By placeholder you mean a placeholder component that shows an approximate pre-renderered layout that then fills in once the images / individual components load / stabalize?

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
damassi: Can you explain a bit more? I can try it, but my css is also really bad 😄

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomCarousel.tsx
dzucconi: Oh, yeah, I was actually assuming this would have data in it already when the page loads for SSR, in which case you'd see properly sized image placeholders or whatever. Though if this is client-side only then the explicit height makes sense.

src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
dzucconi: Like: https://codepen.io/dzuc/pen/OJyEbzN

Vs trying to calculate the height of of the header then subtract it out.

src/Apps/ViewingRoom/Routes/Works/Components/ViewingRoomCarousel.tsx
damassi: It can be both, depending on where the user enters -- direct URL = SSR pass, but then navigating from elsewhere everything is on the client.

artsyit:
:rocket: PR was released in [`v26.36.5`](https://github.com/artsy/reaction/releases/tag/v26.36.5) :rocket:

PR #3509, opened 2020-05-11
Responsive conversations layout

zephraph:
Changes:
- Redirect to first conversation when `/conversations` opened on desktop
- Added
- Refactored inbox header to be responsive
- Refactored layout to be responsive
- Added details stub

Full view (details panel not yet implemented)

image

md view

image

Still quite a bit of work to do so expect a lot of fast follows.

### DIFF
--- a/src/Apps/Conversation/Components/Conversation.tsx
+++ b/src/Apps/Conversation/Components/Conversation.tsx
@@ -1,32 +1,21 @@
 import {
-  ArrowLeftIcon,
   color,
   Flex,
   Image,
-  InfoCircleIcon,
   Link,
   Sans,
   Serif,
   Spacer,
+  Box,
 } from "@artsy/palette"
 import { Conversation_conversation } from "__generated__/Conversation_conversation.graphql"
-import { RouterLink } from "Artsy/Router/RouterLink"
 import { DateTime } from "luxon"
 import React from "react"
 import { createFragmentContainer, RelayProp } from "react-relay"
 import { graphql } from "relay-runtime"
-import styled from "styled-components"
 import { MessageFragmentContainer as Message } from "./Message"
 import { Reply } from "./Reply"
 import { fromToday, TimeSince } from "./TimeSince"
-
-const StyledHeader = styled(Flex)`
-  position: fixed;
-  background: white;
-  border-bottom: 1px solid ${color("black10")};
-  height: 55px;
-  top: 59px;
-`
 
 interface ItemProps {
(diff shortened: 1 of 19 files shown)


artsyit:
:rocket: PR was released in [`v26.36.0`](https://github.com/artsy/reaction/releases/tag/v26.36.0) :rocket:

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -124,35 +113,23 @@ export interface ConversationProps {
...
-                ? i.item.id
-                : idx
-            }
-          />
-        ))}
-        {groupMessages(conversation.messages.edges.map(edge => edge.node)).map(
-          (messageGroup, groupIndex) => {
+    <Flex flexDirection="column" width="100%">
+      <Box>
+        <Spacer mt="45px" />
```
damassi, line 118: @zephraph wouldn't this be better if we did `mt={4}` versus breaking our design system? These values shouldn't be to comp so much as they should be to the closest value in the system. 

Thoughts?

src/Apps/Conversation/ConversationApp.tsx
```diff
@@ -1,48 +1,102 @@
...
+  const conversation = me.conversationsConnection.edges[0]?.node
+
+  useEffect(() => {
+    setWidth(getViewWidth())
+    const listenForResize = debounce(() => {
+      setWidth(getViewWidth())
+    })
+    window.addEventListener("resize", listenForResize)
+    return () => window.removeEventListener("resize", listenForResize)
+  }, [])
```
damassi, line 45: Maybe https://reactjs.org/docs/hooks-reference.html#uselayouteffect ?

src/Apps/Conversation/Components/Conversation.tsx
zephraph: This is a buffer to move the conversation body far enough from under a fixed header to be visible. We're moving pretty fast on this and it's not all ideal at the moment. Will follow up.

src/Apps/Conversation/ConversationApp.tsx
zephraph: > However, we recommend starting with useEffect first and only trying useLayoutEffect if that causes a problem.

src/Apps/Conversation/Components/Conversation.tsx
damassi: We should just add a `5` to our space scale, this constantly comes up 😄 I'll PR and ping design

PR #2966, opened 2019-11-01
Update buyer's premium copy and location

bhoggard:
Ticket: https://artsyproduct.atlassian.net/browse/AUCT-676

Screen Shot 2019-11-01 at 4 01 36 PM

PR #3514, opened 2020-05-11
Move tracking code for buyers premium link click

erikdstock:
The tracking decorator for clicking on the buyers' premium link [did not move with its handler function](https://github.com/artsy/reaction/pull/2966/files#diff-c9ef16c3b323f531057622da216d4a53L24) in a recent change - h/t @sweir27 for quickly tracking it down. This fixes it. Should be fine to merge as-is but I'm working on adding a test as well.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
@@ -1,6 +1,5 @@
 import { Box, Serif } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
-import { SystemContextConsumer } from "Artsy"
 import React from "react"
 
 import { ArtworkSidebarAuctionPartnerInfo_artwork } from "__generated__/ArtworkSidebarAuctionPartnerInfo_artwork.graphql"
@@ -18,22 +17,18 @@ export class ArtworkSidebarAuctionPartnerInfo extends React.Component<
       return null
     }
     return (
-      <SystemContextConsumer>
-        {() => (
-          <Box pb={3}>
-            {partner && (
-              <Serif size="2" weight="semibold" color="black100">
-                {partner.name}
-              </Serif>
-            )}
-            {sale_artwork && sale_artwork.estimate && (
-              <Serif size="2" color="black60">
-                Estimated value: {sale_artwork.estimate}
-              </Serif>
-            )}
-          </Box>
+      <Box pb={3}>
+        {partner && (
+          <Serif size="2" weight="semibold" color="black100">
+            {partner.name}
+          </Serif>
         )}
-      </SystemContextConsumer>
+        {sale_artwork && sale_artwork.estimate && (
+          <Serif size="2" color="black60">
+            Estimated value: {sale_artwork.estimate}
+          </Serif>
+        )}
+      </Box>
     )
   }
 }
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
@@ -1,9 +1,7 @@
 import { Box, Serif } from "@artsy/palette"
-import { SystemContextConsumer } from "Artsy"
-import { track } from "Artsy/Analytics"
-import * as Schema from "Artsy/Analytics/Schema"
-import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { SystemContextConsumer } from "Artsy"
+import React from "react"
 
 import { ArtworkSidebarAuctionPartnerInfo_artwork } from "__generated__/ArtworkSidebarAuctionPartnerInfo_artwork.graphql"
 
@@ -11,16 +9,9 @@ export interface ArtworkSidebarAuctionPartnerInfoProps {
   artwork: ArtworkSidebarAuctionPartnerInfo_artwork
 }
 
-@track()
 export class ArtworkSidebarAuctionPartnerInfo extends React.Component<
   ArtworkSidebarAuctionPartnerInfoProps
 > {
-  @track(() => ({
-    context_module: Schema.ContextModule.Sidebar,
-    action_type: Schema.ActionType.Click,
-    subject: Schema.Subject.AuctionBuyerPremium,
-    type: Schema.Type.Link,
-  }))
   render() {
     const { partner, sale_artwork, sale } = this.props.artwork
     if (sale.is_closed) {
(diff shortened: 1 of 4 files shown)


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
```diff
@@ -1,34 +1,25 @@
...
-    type: Schema.Type.Link,
-  }))
   render() {
     const { partner, sale_artwork, sale } = this.props.artwork
     if (sale.is_closed) {
       return null
     }
     return (
       <SystemContextConsumer>
-        {({ mediator }) => (
```
erikdstock, line 31: This was no longer used, which makes me think we can remove the `<SystemContextConsumer/>` and probably make this a function component instead of a class.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
```diff
@@ -1,34 +1,25 @@
...
 
 export interface ArtworkSidebarAuctionPartnerInfoProps {
   artwork: ArtworkSidebarAuctionPartnerInfo_artwork
 }
 
-@track()
 export class ArtworkSidebarAuctionPartnerInfo extends React.Component<
   ArtworkSidebarAuctionPartnerInfoProps
 > {
-  @track(() => ({
```
erikdstock, line 18: This call was previously decorating the handler for clicking the BP link, but when that link moved it was suddenly decorating the render function.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
```diff
@@ -135,7 +145,11 @@ export class ArtworkSidebarCurrentBidInfo extends React.Component<
               <Serif size="2" color="black60">
                 <br />
                 This auction has a{" "}
-                <Link onClick={this.onClickBuyerPremium.bind(this, mediator)}>
```
erikdstock, line 138: In order to get tracking to fire i had to remove this binding. Is this now handled by babel?

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionPartnerInfo.tsx
```diff
@@ -1,34 +1,25 @@
...
-    subject: Schema.Subject.AuctionBuyerPremium,
-    type: Schema.Type.Link,
-  }))
   render() {
     const { partner, sale_artwork, sale } = this.props.artwork
     if (sale.is_closed) {
       return null
     }
     return (
       <SystemContextConsumer>
```
damassi, line 21: Can delete context consumer if not using

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
```diff
@@ -4,7 +4,8 @@ import { SystemContextConsumer } from "Artsy"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { get } from "Utils/get"
-
+import { track } from "Artsy/Analytics"
+import * as Schema from "Artsy/Analytics/Schema"
```
damassi, line 8: import {AnalyticsSchema} from "Artsy/Analytics"

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
```diff
@@ -20,9 +21,18 @@ export interface ArtworkSidebarCurrentBidInfoProps {
...
 > {
+  @track(() => {
+    return {
+      context_module: Schema.ContextModule.Sidebar,
+      action_type: Schema.ActionType.Click,
+      subject: Schema.Subject.AuctionBuyerPremium,
+      type: Schema.Type.Link,
+    }
+  })
   onClickBuyerPremium(mediator) {
```
damassi, line 36: `handleBuyerPremiumClick` (on = incoming props passed to component; handle = internal event handler)

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
damassi: Heh, we're in weird JS territory right now 😄 

To poorly explain: `bind` does a couple things in JS -- it allows one to attach functions to different scopes (in the most common case, `this`) and it also serves as a sort of odd form of [partial application](https://towardsdatascience.com/javascript-currying-vs-partial-application-4db5b2442be8). Meaning, by calling `foo.bind(scope, someParam)` on a function you can "preload" the function with additional arguments (someParam) that can be plucked out of the second argument when foo is invoked:

```js
function MyComponent() {
  const foo = (name, location) => {
    console.log(name, location) // chris, washington
  }

  const handleClick = foo.bind(this, "washington") 

  return <div onClick={() => handleClick("chris") />
}
```

The old code `onClick={this.onClickBuyerPremium.bind(this, mediator)}>` was using this partial application form to pass `mediator` across before the actual click occurs, which will then pass the click event arguments. 

This is confusing, can hop on call to explain!

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
```diff
@@ -135,7 +145,11 @@ export class ArtworkSidebarCurrentBidInfo extends React.Component<
               <Serif size="2" color="black60">
                 <br />
                 This auction has a{" "}
-                <Link onClick={this.onClickBuyerPremium.bind(this, mediator)}>
+                <Link
+                  onClick={() => {
+                    this.onClickBuyerPremium(mediator)
```
damassi, line 150: Since `onClickBuyerPremium` doesn't actually need any e.MouseEvent props that would be passed along by doing
```js
onClick={this.onClickBuyersPremium}
```
this is totally fine.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
erikdstock: Ah thanks! I was aware it was possible for that case but had never seen a practical application of it. I tried this in the console though and I think that the args are applied in order, so in this example "washington" would be bound to `name`:
```
>   const foo = (name, location) => {
     console.log(name, location) // chris, washington
   }

>   const handleClick = foo.bind(this, "washington")

> foo("a","b")
a b

> handleClick("A")
washington A
```

artsyit:
:rocket: PR was released in [`v26.36.9`](https://github.com/artsy/reaction/releases/tag/v26.36.9) :rocket:

PR #3517, opened 2020-05-12
Roughs in Feature app UI

dzucconi:
Nothing too wild here. Let me know if I'm missing some useful Palette components or patterns.

I need to just move on to the real data because most of what has to happen now is: improve image handling (which I want real images to build with) and build out how the different union type options display in the sets.

[desktop screen recording](http://static.damonzucconi.com/_capture/zXNgyW9ETyum.gif)
[mobile screen recording](http://static.damonzucconi.com/_capture/H01gHd0O8VKz.gif)

### DIFF
--- a/src/Apps/Feature/Components/__stories__/FeatureFeaturedLink.story.tsx
+++ b/src/Apps/Feature/Components/__stories__/FeatureFeaturedLink.story.tsx
@@ -1,0 +1,16 @@
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { FeatureFeaturedLink } from "../FeatureFeaturedLink"
+import { FEATURED_LINK } from "../__tests__/fixtures"
+import { CSSGrid } from "@artsy/palette"
+
+storiesOf("Apps/Feature", module)
+  .add("FeatureFeaturedLink", () => (
+    <CSSGrid gridTemplateColumns={`repeat(2, 1fr)`} gridGap={2}>
+      <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+      <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+    </CSSGrid>
+  ))
+  .add("FeatureFeaturedLink (wide)", () => (
+    <FeatureFeaturedLink featuredLink={FEATURED_LINK} wide />
+  ))
--- a/src/Apps/Feature/Components/__stories__/FeatureHeader.story.tsx
+++ b/src/Apps/Feature/Components/__stories__/FeatureHeader.story.tsx
@@ -1,0 +1,18 @@
+import React from "react"
+import { omit } from "lodash"
+import { storiesOf } from "storybook/storiesOf"
+import { FeatureHeader } from "../FeatureHeader"
+import { FEATURE_HEADER } from "../__tests__/fixtures"
+import { Box } from "@artsy/palette"
+
+storiesOf("Apps/Feature", module)
+  .add("FeatureHeader", () => (
+    <Box height="100vh">
+      <FeatureHeader feature={FEATURE_HEADER} />
+    </Box>
+  ))
+  .add("FeatureHeader (without image)", () => (
+    <Box height="100vh">
+      <FeatureHeader feature={omit(FEATURE_HEADER, "image")} />
+    </Box>
+  ))
--- a/src/Apps/Feature/FeatureApp.tsx
+++ b/src/Apps/Feature/FeatureApp.tsx
@@ -1,20 +1,78 @@
 import React from "react"
 import { AppContainer } from "Apps/Components/AppContainer"
-import { Box } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
-
+import { FeatureHeaderFragmentContainer as FeatureHeader } from "./Components/FeatureHeader"
+import { FeatureFeaturedLink } from "./Components/FeatureFeaturedLink"
+import { FeatureSet } from "./Components/FeatureSet"
 import { FeatureApp_feature } from "__generated__/FeatureApp_feature.graphql"
+import { Box, Sans, Spacer } from "@artsy/palette"
+import { FEATURED_LINK } from "./Components/__tests__/fixtures"
 
 interface FeatureAppProps {
-  children: React.ReactNode
   feature: FeatureApp_feature
 }
 
-const FeatureApp: React.FC<FeatureAppProps> = ({ children, feature }) => {
+const FeatureApp: React.FC<FeatureAppProps> = ({ feature }) => {
   return (
     <AppContainer maxWidth="100%">
-      <Box my={3}>{feature.name}</Box>
-      {children}
+      <Box height="100vh">
+        <FeatureHeader feature={feature} />
+      </Box>
+
+      <Box maxWidth={["100%", 460]} mx="auto" my={3} px={3}>
+        <Sans size="4">
+          {/* Feature#description */}
+          Art-world spaces have paused, museums and galleries have closed their
+          doors, exhibitions have canceled, and auctions and fairs have been
+          postponed. Still, even in times of crisis, art can bring us closer and
+          keep us connected across time and place. And in times of crises, art
+          continues to move us and keeps going, and and keeps us connected.
+        </Sans>
+
+        <Spacer my={3} />
+
+        <Sans size="6">
+          {/* Feature#callOut */}
+          Featuring: Greene Naftali, Petzel, Andrew Kreps, Skarstedt, CLEARING,
+          The Journal Gallery, CANADA, Company Gallery, Van Doren Waxter, Jack
+          Hanley, Matthew Marks, Gavin Brown, Perrotin, James Fuentes, Gagosian,
+          David Zwirnir, Galerie Lelong, Sikkema Jenkins, Two Palms, Pace,
+          Lisson, Hauser &amp; Wirth
+        </Sans>
+      </Box>
+
+      <FeatureSet mx={3}>
+        <FeatureFeaturedLink wide featuredLink={FEATURED_LINK} />
+      </FeatureSet>
+
+      <FeatureSet mx={3}>
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+      </FeatureSet>
+
+      <FeatureSet mx={3}>
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+      </FeatureSet>
+
+      <FeatureSet mx={3}>
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+      </FeatureSet>
+
+      <FeatureSet mx={3}>
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+        <FeatureFeaturedLink featuredLink={FEATURED_LINK} />
+      </FeatureSet>
     </AppContainer>
   )
 }
@@ -23,7 +81,7 @@ const FeatureApp: React.FC<FeatureAppProps> = ({ children, feature }) => {
 export default createFragmentContainer(FeatureApp, {
   feature: graphql`
     fragment FeatureApp_feature on Feature {
-      name
+      ...FeatureHeader_feature
     }
   `,
 })
--- a/src/Apps/Feature/Components/FeatureFeaturedLink.tsx
+++ b/src/Apps/Feature/Components/FeatureFeaturedLink.tsx
@@ -1,0 +1,92 @@
+import React from "react"
+import styled from "styled-components"
+import { space, SpaceProps } from "styled-system"
+import { Box, Sans, Serif, Image, Flex } from "@artsy/palette"
+import { RouterLink } from "Artsy/Router/RouterLink"
+
+const Container = styled(RouterLink)`
+  ${space}
+  display: flex;
+  text-decoration: none;
+  flex-direction: column;
+`
+
+const Figure = styled(Box)`
+  position: relative;
+`
+
+const Cover = styled(Image)`
+  display: block;
+  width: 100%;
+  height: 100%;
+`
+
+const Title = styled(Sans)`
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  background: linear-gradient(
+    rgba(255, 255, 255, 0) 0%,
+    rgba(0, 0, 0, 0.25) 100%
+  );
+`
+
+const Meta = styled(Flex).attrs({ mt: 1 })``
+
+const MetaCell = styled(Flex)`
+  flex-basis: 50%;
+`
+
+export interface FeatureFeaturedLinkProps extends SpaceProps {
+  wide?: boolean
+  featuredLink: {
+    title: string
+    href: string
+    subtitle?: string
+    description?: string
+    image: {
+      width: number
+      height: number
+      url: string
+    }
+  }
+}
+
+export const FeatureFeaturedLink: React.FC<FeatureFeaturedLinkProps> = ({
+  wide,
+  featuredLink: { title, href, subtitle, description, image },
+  ...rest
+}) => {
+  return (
+    <Container to={href} {...rest}>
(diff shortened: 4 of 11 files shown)


src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+    title: string
+    href: string
+    subtitle?: string
+    description?: string
+    image: {
+      width: number
+      height: number
+      url: string
+    }
+  }
```
dzucconi, line 52: Will just be the fragment.

src/Apps/Feature/Components/FeatureHeader/FeatureHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+`
+
+const Cover = styled(Image)`
+  object-fit: cover;
+  width: 100%;
+  height: 100%;
+`
+
+const Meta = styled(Flex)`
+  ${flexBasis}
```
dzucconi, line 28: Weird that we missed this function in the Flex component. Will be nice to just upgrade styled-system and use the `flexbox` function.

src/Apps/Feature/Components/FeatureHeader/FeatureHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export const FeatureHeader: React.FC<FeatureHeaderProps> = ({
+  feature: { name, subheadline, image },
+}) => {
+  return (
+    <Container flexDirection={["column", "row"]}>
+      {image && (
+        <Figure>
+          {/* TODO: Optimize image, support higher DPR */}
```
dzucconi, line 47: I'll be fleshing this out on Metaphysic's end of things.

src/Apps/Feature/FeatureApp.tsx
```diff
@@ -1,20 +1,80 @@
...
   feature: FeatureApp_feature
 }
 
-const FeatureApp: React.FC<FeatureAppProps> = ({ children, feature }) => {
+const FeatureApp: React.FC<FeatureAppProps> = ({ feature }) => {
   return (
     <AppContainer maxWidth="100%">
-      <Box my={3}>{feature.name}</Box>
-      {children}
+      <Box height="100vh">
```
dzucconi, line 20: I realize I need to offset by the header height here. I don't like this at all — it seems like it'd be prudent to eventually fix this in the layout so we don't have to hardcode all these calcs everywhere which is ugly.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+            {title}
+          </Title>
+        </Figure>
+      ) : (
+        <Sans size="6" color="black100" my={2}>
+          {title}
+        </Sans>
+      )}
+
+      <Meta flexDirection={wide ? ["column", "row"] : "column"}>
```
sweir27, line 76: I think we _could_ use the `Grid` here?

src/Apps/Feature/Components/FeatureSet/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./FeatureSet"
```
sweir27, line 1: I think we try to avoid these extra `index` files, but @damassi will know more about convention.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const FeatureFeaturedLink: React.FC<FeatureFeaturedLinkProps> = ({
+  wide,
+  featuredLink: { title, href, subtitle, description, image },
+  ...rest
+}) => {
+  return (
+    <Container href={href} {...rest}>
+      {image ? (
+        <Figure>
+          <Cover src={image.url} />
```
dzucconi, line 64: Will be a ResponsiveImage but I'd be interested in fixing that up a bit.

dzucconi: Need real data for placeholders blah blah.

dzucconi: I need to look at the Gemini templates for FeaturedLink images but I believe they all crop to a 4:3 rectangle. cc @sweir27  — noticed the designs are specified around portrait aspect ratios.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
```diff
@@ -0,0 +1,91 @@
...
+  );
+`
+
+const Meta = styled(Flex).attrs({ mt: 1 })``
+
+const MetaCell = styled(Flex)`
+  flex-basis: 50%;
+`
+
+export type FeatureFeaturedLinkProps = SpaceProps & {
```
sweir27, line 40: #superminor but I don't think we _need_ the `Feature` namespacing in these component names since we're already nested under the `Feature` folder.

(This is not a strong opinion, this name in particular stuck out 😄)

dzucconi:
> The other one we're considering supporting to start is Artworks

Yeah, I'm just gonna do that when I wire up with real data; we need to support quite a few sub-types, definitely artworks to start, sales would be next, the rest are much more rare.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
dzucconi: I'm of the opinion all subcomponents should be prefixed with their parent due to uniqueness constraints on relay fragment naming. This takes any thought out of it.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
sweir27: Oh really?? Ideally we'd pull the version that has the original aspect ratio and then we can pass it through the `crop` endpoint (using that helper here or via the MP call) to get the size we want.

The only caveat with that is we should make sure to do it based on fixed sizes (even via breakpoint) vs. dynamic based on exact browser size. This is just bc we'll want to benefit from caching which will break down if the request is highly variable.

src/Apps/Feature/Components/FeatureSet/index.ts
dzucconi: I personally like it so that it's simple to sprawl out into sub-components when needed and expose what you in a more explicit manner, but yeah I can see the other side of this coin. Lemme know @damassi

sweir27:
@dzucconi totally.

We can work with Nicole to spec out the rest of the treatments!

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
sweir27: Ah! That's... a good reason haha. 👍

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
dzucconi: I don't think this template has a normalized version unfortunately, we'll probably have to add it and then reprocess. But I'll double check the template.

https://github.com/artsy/quartz/blob/master/src/gravity/FeaturedLink.ts#L23-L29

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
dzucconi: But yeah, I mean we don't do it currently AFAICT 🤷

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
```diff
@@ -0,0 +1,91 @@
+import React from "react"
+import styled from "styled-components"
+import { space, SpaceProps } from "styled-system"
+import { Box, Sans, Serif, Image, Flex } from "@artsy/palette"
+
+const Container = styled.a`
```
damassi, line 6: Maybe try `StyledLink` from palette? 

But generally, if the link navigates to another page internal to Artsy.net, we should use a `RouterLink`. Can do `styled(StyledLink)` to apply additional styles, etc.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
damassi: This is also the "Relay Way", as mentioned due to fragments -- it has to be such, otherwise we have the potential for name clashes. (I'm personally not so strict for no-relay components, but then things feel inconsistent.)

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
damassi: The codebase generally uses interfaces for objects / components -- we had a lint rule in place with TSLint that we need to update (mind updating in this PR?) 

The use of `type` is mostly constrained to one-liners.

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
damassi: @sweir27 - (just my opinion) -- We should generally only use Grid where flexbox fails, especially if things are simple (reason being, we use a lib for the grid, which ideally we can rip out at some point).

src/Apps/Feature/Components/FeatureHeader/FeatureHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <Sans size="4" textAlign="center">
+            {subheadline}
+          </Sans>
+        )}
+      </Meta>
+    </Container>
+  )
+}
+
+export const FeatureHeaderContainer = createFragmentContainer(FeatureHeader, {
```
damassi, line 70: The idiom is:

```
export const FeatureHeaderFragmentContainer
```
which we then import as 
``` 
import { FeatureHeaderFragmentContainer as FeatureHeader } from ...
```

src/Apps/Feature/Components/FeatureHeader/FeatureHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+  feature: graphql`
+    fragment FeatureHeader_feature on Feature {
+      name
+      subheadline: description
```
damassi, line 74: #superminor - rather than creating divergence here, why keep the naming 1:1 across fields / components and / or update MP `subheadline` field to = `description`?

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
dzucconi: Ah interesting — I've been generally using it with styled-system props because it's much simpler to compose prop types. `type FooComponentProps = BoxProps & FlexBasisProps & { whatever: string }` — like in this instance I'm intersecting with SpaceProps, but would also probably want the StyledLinkProps or whatever since it's rest spread. This is also one of those things that doesn't matter much to me, whatever avoids bike shedding.

src/Apps/Feature/Components/FeatureHeader/FeatureHeader.tsx
dzucconi: Yah, will do.

src/Apps/Feature/Components/FeatureSet/FeatureSet.tsx
```diff
@@ -0,0 +1,54 @@
...
+  & + & {
+    margin-top: -1px;
+  }
+
+  &:last-of-type {
+    border-bottom: 0;
+  }
+`
+
+export type FeatureSetProps = Omit<BoxProps, "color"> & {}
```
damassi, line 18: Good use case for `type` over `interface` for props, often seen in codebase

src/Apps/Feature/Components/FeatureHeader/FeatureHeader.tsx
dzucconi: Oh this was just temporary because we're going to add this as a real field. I just was hooking this up to real data and wanted a placeholder for now.

sweir27: I think this is just for the sake of showing something in that field right now (we do eventually want this to be a new `subheadline` field from grav).

Since it's a little confusing now @dzucconi I'd just use a static string for now there.

sweir27: (Oops sorry for the dup comment 😅)

src/Apps/Feature/Components/FeatureSet/index.ts
damassi: `index` files make a lot of sense as a funnel in lib code, but unless the folder (in this case `FeatureSet`) will contain numerous files its best to keep things as flat as possible. (It's also nice to (not) have five tabs open in VSCode that are all named `index` 😄)

There are some exceptions to this tho:

Say you have this kind of folder structure:

```
Apps
  AppName 
     Routes
        RouteName
           Components
```
It can make sense to at the `AppName` level to have an `index` file that then coordinates various exports for the sub-app namespace. 

For individual components tho the rule is: 
- One file, flat in Components folder
- Two plus files, feel free to create folder to contain related items and then an index file is ok, but not until there's a logical grouping of items necessary.

src/Apps/Feature/Components/__stories__/FeatureFeaturedLink.story.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react"
+import { storiesOf } from "storybook/storiesOf"
+import { FeatureFeaturedLink } from "../FeatureFeaturedLink"
+import { FEATURED_LINK } from "../__tests__/fixtures"
+import { CSSGrid } from "@artsy/palette"
+
+storiesOf("Apps/Feature", module)
```
damassi, line 7: Hey, thank you!

src/Apps/Feature/Components/__stories__/FeatureHeader.story.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react"
+import { omit } from "lodash"
+import { storiesOf } from "storybook/storiesOf"
+import { FeatureHeader } from "../FeatureHeader"
+import { FEATURE_HEADER } from "../__tests__/fixtures"
+
+storiesOf("Apps/Feature", module)
+  .add("FeatureHeader", () => (
+    <div style={{ height: "100vh" }}>
```
damassi, line 9: We haven't used divs in our codebase in a long while, can we stick to box to keep things consistent?

src/Apps/Feature/Components/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./FeatureHeader"
+export * from "./FeatureFeaturedLink"
+export * from "./FeatureSet"
```
damassi, line 3: Use of an index file _could_ make sense here, but is it necessary for something that will only contain three files? I don't have a strong opinion -- just thinking editor ergonomics and how to reduce the number of matched `index` files in a search or tab bar set.

src/Apps/Feature/FeatureApp.tsx
damassi: Def down to help clean this up, I'm not even sure how its possible without a `calc`

dzucconi:
Ah yeah, @damassi — I self-assigned because I figured there would be feedback which I'd then just address and merge when I'm ready. Is this less common?

damassi:
Yah, the way it works is:

1) Dev assigns to someone, they add comments or approve
2) Dev address comments, and merges once there's an approval from assignee. OR assignee reassigns back to dev with comment saying "looks good merge ittt"

src/Apps/Feature/FeatureApp.tsx
dzucconi: You could avoid the `calc` by using `position: sticky` on the nav so that it placeholders correctly against the document. Or just keeping it fixed and using marginTop: NAV_HEIGHT a single time in the layout container (rather than having to specify it for every single page/app).

src/Apps/Feature/Components/FeatureFeaturedLink/FeatureFeaturedLink.tsx
dzucconi: OK so it looks like we can use the `wide` version (tho it's a bit weird), the rest of them are all fill crops to 4:3 or 1:1.
```
 height: 480
 width: 640
 quality: 90
 resize_to: "fill"
 key: "large_rectangle"

 height: nil
 width: 2000
 quality: 85
 resize_to: "width"
 key: "wide"

 height: 400
 width: 400
 quality: 90
 resize_to: "fill"
 key: "large_square"

 height: 200
 width: 200
 quality: 90
 resize_to: "fill"
 key: "medium_square"

 height: 100
 width: 100
 quality: 90
 resize_to: "fill"
 key: "small_square"

 height: 195
 width: 260
 quality: 90
 resize_to: "fill"
 key: "medium_rectangle"

 height: 113
 width: 148
 quality: 90
 resize_to: "fill"
 key: "small_rectangle"
```

dzucconi:
Addressed feedback and merging/continuing on.

artsyit:
:rocket: PR was released in [`v26.36.8`](https://github.com/artsy/reaction/releases/tag/v26.36.8) :rocket:

PR #3521, opened 2020-05-12
Removes placeholders to fix build

dzucconi:
Didn't realize this would actually cause the production build to fail but pass in the pull request.

*The more you know*


zephraph:
You could just move it to a non-`__test__` directory and that would be fine.

dzucconi:
Eh, I already replaced it with real data so this is fine!

artsyit:
:rocket: PR was released in [`v26.36.8`](https://github.com/artsy/reaction/releases/tag/v26.36.8) :rocket:

PR #3525, opened 2020-05-13
[Viewing Room] Add tests

damassi:
This adds some tests to the Viewing Room app. 

(Assigning again to you @dzucconi just so you can see some of our testing patterns.)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,8 +36,6 @@ module.exports = {
     node: true,
   },
   rules: {
-    "react/display-name": 0,
-    "react/prop-types": 0,
     "@typescript-eslint/ban-ts-ignore": 0,
     "@typescript-eslint/camelcase": 0,
     "@typescript-eslint/explicit-function-return-type": 0,
@@ -49,8 +47,17 @@ module.exports = {
     "@typescript-eslint/no-use-before-define": 0,
     "@typescript-eslint/no-unused-vars": 0,
     "@typescript-eslint/no-var-requires": 0,
+    "react/display-name": 0,
+    "react/prop-types": 0,
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
+    "sort-imports": [
+      "warn",
+      {
+        // Unfortunately there's no autofixer for this
+        ignoreDeclarationSort: true,
+      },
+    ],
 
     // FIXME: Investigate / reenable these rules. Disabled to introduce eslint
     // into codebase.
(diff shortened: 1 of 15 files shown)


.eslintrc.js
```diff
@@ -49,8 +47,17 @@ module.exports = {
...
     "@typescript-eslint/no-var-requires": 0,
+    "react/display-name": 0,
+    "react/prop-types": 0,
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
+    "sort-imports": [
+      "warn",
+      {
+        // Unfortunately there's no autofixer for this
+        ignoreDeclarationSort: true,
```
damassi, line 58: In which case we lean on the autofixer to sort named export keys on individual imports, but don't worry about alpha for the full import list as thats a real pain.

src/Apps/ViewingRoom/Routes/Statement/__tests__/ViewingRoomStatementRoute.test.tsx
```diff
@@ -1,5 +1,130 @@
-describe.skip("ViewingRoomStatementRoute", () => {
-  it("todo", () => {
-    //
+import React from "react"
+import { MockBoot, renderRelayTree } from "DevTools"
+import { graphql } from "react-relay"
+import { ViewingRoomStatementRoute_Test_QueryRawResponse } from "__generated__/ViewingRoomStatementRoute_Test_Query.graphql"
+import { Breakpoint } from "@artsy/palette"
+import { ViewingRoomStatementRouteFragmentContainer } from "../ViewingRoomStatementRoute"
+import { ViewingRoomStatmentRouteFixture } from "Apps/ViewingRoom/__tests__/Fixtures/ViewingRoomStatementRouteFixture"
```
sweir27, line 7: I tend to prefer putting once-used fixtures in the file in which they're used. Makes it a little longer but means we don't need to `export` and risk keeping old fixture files around that aren't used.

artsyit:
:rocket: PR was released in [`v26.36.13`](https://github.com/artsy/reaction/releases/tag/v26.36.13) :rocket:

src/Apps/ViewingRoom/Routes/Statement/__tests__/ViewingRoomStatementRoute.test.tsx
damassi: Thats a good point, yeah right now things are all spread out. Most `Apps/*` fixtures actually ended up `Apps/__tests__/Fixtures/*`, kinda far away, but it makes sense to colocate everything. Will update in follow-up.

damassi:
Yeah, unfortunately we're a long way from that world unless we start breaking up Reaction and paying overtime :/

PR #3527, opened 2020-05-13
Wires up OrderedSets and FeaturedLinks to real data

dzucconi:
The duplicated fields are just the placeholder aliases — I'm going to switch to working on Metaphysics next.

### TODO

- [ ] Expose fields for all of those yet-to-exist values
- [ ] Support markdown on FeaturedLink subtitles and OrderedSet descriptions
- [ ] Expose a `srcSet` on Metaphysics image types
- [ ] Featured link images should support different sizes depending on the media width (so using picture/source tags)
- [ ] a11y pass (alt tags, etc)

### DIFF
--- a/src/Apps/Feature/Components/FeatureFeaturedLink.tsx
+++ b/src/Apps/Feature/Components/FeatureFeaturedLink.tsx
@@ -1,8 +1,10 @@
 import React from "react"
 import styled from "styled-components"
-import { space, SpaceProps } from "styled-system"
-import { Box, Sans, Serif, Image, Flex } from "@artsy/palette"
+import { SpaceProps, space } from "styled-system"
+import { Box, Flex, ResponsiveImage, Sans, Serif, color } from "@artsy/palette"
+import { createFragmentContainer, graphql } from "react-relay"
 import { RouterLink } from "Artsy/Router/RouterLink"
+import { FeatureFeaturedLink_featuredLink } from "__generated__/FeatureFeaturedLink_featuredLink.graphql"
 
 const Container = styled(RouterLink)`
   ${space}
@@ -13,12 +15,6 @@ const Container = styled(RouterLink)`
 
 const Figure = styled(Box)`
   position: relative;
-`
-
-const Cover = styled(Image)`
-  display: block;
-  width: 100%;
-  height: 100%;
 `
 
 const Title = styled(Sans)`
@@ -40,29 +36,25 @@ const MetaCell = styled(Flex)`
 
 export interface FeatureFeaturedLinkProps extends SpaceProps {
   wide?: boolean
-  featuredLink: {
-    title: string
-    href: string
-    subtitle?: string
-    description?: string
-    image: {
-      width: number
-      height: number
-      url: string
-    }
-  }
+  featuredLink: FeatureFeaturedLink_featuredLink
 }
 
 export const FeatureFeaturedLink: React.FC<FeatureFeaturedLinkProps> = ({
   wide,
-  featuredLink: { title, href, subtitle, description, image },
+  featuredLink,
+  featuredLink: { href, title, subtitle, description, image },
   ...rest
 }) => {
   return (
     <Container to={href} {...rest}>
       {image ? (
         <Figure>
-          <Cover src={image.url} />
+          <ResponsiveImage
+            maxWidth="100%"
+            src={image.cropped.src}
+            ratio={image.cropped.height / image.cropped.width}
+            style={{ backgroundColor: color("black10") }}
+          />
 
           <Title size="6" color="white100" p={2} pt={9}>
             {title}
@@ -70,7 +62,7 @@ export const FeatureFeaturedLink: React.FC<FeatureFeaturedLinkProps> = ({
         </Figure>
       ) : (
         <Sans size="6" color="black100" my={2}>
-          {title}
+          {title || "—"}
         </Sans>
       )}
 
@@ -90,3 +82,25 @@ export const FeatureFeaturedLink: React.FC<FeatureFeaturedLinkProps> = ({
     </Container>
   )
 }
+
+export const FeatureFeaturedLinkFragmentContainer = createFragmentContainer(
+  FeatureFeaturedLink,
+  {
+    featuredLink: graphql`
+      fragment FeatureFeaturedLink_featuredLink on FeaturedLink {
+        href
+        title
+        subtitle
+        # TODO: Placeholder value
+        description: subtitle
+        image {
+          cropped(width: 800, height: 600, version: ["wide"]) {
+            src: url
+            width
+            height
+          }
+        }
+      }
+    `,
+  }
+)
--- a/src/Apps/Feature/Components/FeatureSet.tsx
+++ b/src/Apps/Feature/Components/FeatureSet.tsx
@@ -1,6 +1,9 @@
 import React from "react"
 import styled from "styled-components"
-import { Box, BoxProps, color, Sans, CSSGrid } from "@artsy/palette"
+import { Box, BoxProps, CSSGrid, Sans, color } from "@artsy/palette"
+import { createFragmentContainer, graphql } from "react-relay"
+import { FeatureFeaturedLinkFragmentContainer as FeatureFeaturedLink } from "./FeatureFeaturedLink"
+import { FeatureSet_set } from "__generated__/FeatureSet_set.graphql"
 
 const Container = styled(Box)`
   border-top: 1px solid ${color("black100")};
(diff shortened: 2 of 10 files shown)


src/Apps/Feature/Components/FeatureFeaturedLink.tsx
```diff
@@ -40,37 +36,33 @@ const MetaCell = styled(Flex)`
...
 }) => {
   return (
     <Container to={href} {...rest}>
       {image ? (
         <Figure>
-          <Cover src={image.url} />
+          <ResponsiveImage
+            maxWidth="100%"
+            src={image.cropped.src}
+            ratio={image.cropped.height / image.cropped.width}
```
dzucconi, line 55: Cropped annoyingly also doesn't expose the ratio (yet) —obv this is simpler for cropped but still: all of these helpers and fields should basically correspond to the props api for the tags. This is a bit difficult to use correctly.

src/Apps/Feature/Components/FeatureSet.tsx
```diff
@@ -47,8 +53,43 @@ export const FeatureSet: React.FC<FeatureSetProps> = ({
...
+          switch (orderedItem.__typename) {
+            case "FeaturedLink":
+              return (
+                <FeatureFeaturedLink
+                  key={orderedItem.id}
+                  featuredLink={orderedItem}
+                />
+              )
+            default:
+              // TODO: Unimplemented: Artist | Artwork | Gene | Sale | PartnerShow | Profile | OrderedSet
```
dzucconi, line 66: @sweir27 this is what's currently supported

https://github.com/artsy/metaphysics/blob/2784f200a7c75843ab898dc7b575173f8d03668c/src/schema/v2/item.ts#L15-L25

For instance Genes are used on artists index pages.

dzucconi: Which is weird weird in terms of restricting things — Sets aren't exclusively used for Feature pages, so support in this case becomes what's available in the UI rather than what's possible through the types.

src/Apps/Feature/Components/FeatureFeaturedLink.tsx
dzucconi: @damassi how prevalent is the backgroundCover abuse ResponsiveImage? To migrate this into a proper `` tag we'll have to either release a breaking change or export a fresh component.

dzucconi: Or maybe we just support a `srcs` prop then switch based on that.

src/Apps/Feature/Components/FeatureSet.tsx
sweir27: Conceptually it seems OK that we have the ability to create sets of many different types (because they are generic/can be used elsewhere), but the sets that are used _within_ Feature pages are constrained.

Agree that it should be an actual data model limitation, but for now maybe metaphysics can restrict?

dzucconi: I concur — I sort of suspect we'll have to create a different OrderedItem type that is restricted to the types we want.

dzucconi: So what do we want to restrict it to? We should compile a list of valid features that express the possibilities in staging.

sweir27: Yeah. I mean to do that ☝️ today 😄

src/Apps/Feature/Components/FeatureSet.tsx
```diff
@@ -47,8 +53,43 @@ export const FeatureSet: React.FC<FeatureSetProps> = ({
...
 }
+
+export const FeatureSetFragmentContainer = createFragmentContainer(FeatureSet, {
+  set: graphql`
+    fragment FeatureSet_set on OrderedSet {
+      name
+      description
+      itemType
+      # TODO: Handle pagination
+      orderedItems: orderedItemsConnection(first: 50) {
```
sweir27, line 82: I'd start with something lower (20?).

src/Apps/Feature/FeatureApp.tsx
```diff
@@ -3,41 +3,52 @@ import { AppContainer } from "Apps/Components/AppContainer"
...
-        </Sans>
+      <AppContainer>
+        <HorizontalPadding>
+          {(feature.description || feature.callOut) && (
+            <Box maxWidth={["100%", 460]} mx="auto" my={3} px={3}>
+              <Join separator={<Spacer my={3} />}>
+                {feature.description && (
+                  <Sans size="4">
+                    <Box
+                      dangerouslySetInnerHTML={{ __html: feature.description }}
```
sweir27, line 29: Is this our preferred way to render markdown?

src/Apps/Feature/Components/FeatureSet.tsx
dzucconi: 👍 funnily enough https://github.com/artsy/artsy-backbone-mixins/blob/6309f2a4d4e1a1458105d934b1ec89af6bb87e98/lib/feature.coffee#L113 we should make a call about how to actually handle pagination here though — just infinite scroll? Historically these were never actually paginated.

sweir27:
This looks awesome!!

We should have tests for this-- want to add in this PR or a follow-up?

src/Apps/Feature/FeatureApp.tsx
dzucconi: Would be nice if there was a more sanctioned way to do this but, we do want to render in the graph. @damassi is there a reason we don't expose this prop on Sans itself?

src/Apps/Feature/Components/FeatureSet.tsx
sweir27: Haha!

Although in the old feature pages these are rendered client-side so not as much of an issue.

I'd say infinite scroll, but there may also be some UX-reason for not supporting long features? We can ask Nicole.

dzucconi:
> We should have tests for this

Will just follow up. Though, there's no actual logic here to test. I'd be interested in knowing what we are gaining from manual render specs vs snapshots (just a 'human eye'?). Those specs make me really want to get going on a strictness migration here.

src/Apps/Feature/Components/FeatureSet.tsx
dzucconi: One of the unfortunate things about the pagination here is that we need to paginate both sets and the contents inside of them. I do wonder if this is overkill if we never plan on having set content over a specific size and can just set a reasonable limit until a use-case presents itself.

dzucconi: Or maybe one of those like truncated see-more gradient fade out buttons would be a good UX for long sets which would help make the infinite scrolling a bit more natural.

sweir27: Interesting interesting...

I'm okay with setting a reasonable default for now.

sweir27:
@dzucconi Simple render/assertion tests are a lot easier to comprehend than snapshots (which tend to just be blindly updated and mostly churn due to unrelated reasons).

I think the only logical piece is making sure we're showing the right set component based on the type, so even just a test for that would be fine I think!

And beyond that, having the tests in place makes it easy to add tests for tracking, which we should do when we get to it! The most onerous part of it all is setting up the fixture, but once that's in it's easy to add more.

src/Apps/Feature/Components/FeatureFeaturedLink.tsx
damassi: I don't think its that prevailent, in fact I (think) its just used once, in the VR app. ResponsiveImage itself isn't used much at all.

src/Apps/Feature/FeatureApp.tsx
damassi: Have never had the need (to my knowlege)

src/Apps/Feature/Components/FeatureFeaturedLink.tsx
dzucconi: Ah yeah, [it could be used in the artwork grid but it's just done in a custom fashion there](https://github.com/artsy/reaction/blob/master/src/Components/Artwork/GridItem.tsx#L115)

src/Apps/Feature/FeatureApp.tsx
dzucconi: What's weird is that it should just work there too — it's just a type error. There's a lot of machinery involved in setting the typography components up so maybe it's just getting lost.

One of the things to note here is that currently we're getting faux styles on any italics/bold that happens within Markdown rendered text. This is probably one of those things that you don't notice until you're aware of it but it. I'm gonna try to get some fixes into the font face declarations later in the week.

damassi:
Second @sweir27 on snapshots. They provide some foundational protection against unintended changes to things like Palette, but in general `-u` is blindly applied, and the output is somewhat obscure. We also run into CI issues quite frequently where snapshots fail there, but pass locally -- because the dev ran `yarn jest -u` in a linked environment, which doesn't match `yarn.lock` on CI. 

Some recent trends: The MX team recently deleted them en masse from Eigen, replacing them with render assertions. I tend to delete them when I come across them in Reaction, and do the same. Perhaps there's a workflow we can consider at some point that lives calmly in the middle 🤔.

dzucconi:
@damassi, @sweir27 I'm definitely not advocating for snapshots 😄 I guess I'm just being a bit *provocative* here and saying that these render-style no-logic specs don't offer much more value than snapshots and if we can find a path towards strict types in this repo then they would offer even less. 🤔

artsyit:
:rocket: PR was released in [`v26.37.0`](https://github.com/artsy/reaction/releases/tag/v26.37.0) :rocket:

PR #3530, opened 2020-05-13
Update artsy/palette to v9

dblandin:
Addresses one breaking change in the Modal component: https://github.com/artsy/palette/pull/680

Removes any use of the `forcedScroll` property. The component now defaults to `overflow: auto`.


dblandin:
Closing in favor of https://github.com/artsy/reaction/pull/3532

PR #3539, opened 2020-05-14
Update a conversation to be read once it's opened

lilyfromseattle:
[PURCHASE-1871]

When opening a conversation we should fire off the updateConversation mutation in metaphysics to mark any unread messages as read. 

[PURCHASE-1871]: https://artsyproduct.atlassian.net/browse/PURCHASE-1871

### DIFF
--- a/src/Apps/Conversation/Components/Conversation.tsx
+++ b/src/Apps/Conversation/Components/Conversation.tsx
@@ -1,21 +1,22 @@
 import {
-  color,
+  Box,
   Flex,
   Image,
   Link,
   Sans,
   Serif,
   Spacer,
-  Box,
+  color,
 } from "@artsy/palette"
 import { Conversation_conversation } from "__generated__/Conversation_conversation.graphql"
 import { DateTime } from "luxon"
-import React from "react"
-import { createFragmentContainer, RelayProp } from "react-relay"
+import React, { useEffect } from "react"
+import { RelayProp, createFragmentContainer } from "react-relay"
 import { graphql } from "relay-runtime"
 import { MessageFragmentContainer as Message } from "./Message"
 import { Reply } from "./Reply"
-import { fromToday, TimeSince } from "./TimeSince"
+import { TimeSince, fromToday } from "./TimeSince"
+import { UpdateConversation } from "../Mutation/UpdateConversationMutation"
 
 interface ItemProps {
   item: Conversation_conversation["items"][0]["item"]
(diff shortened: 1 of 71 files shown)


src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -112,6 +113,11 @@ export interface ConversationProps {
 
 const Conversation: React.FC<ConversationProps> = props => {
   const { conversation, relay } = props
+
+  useEffect(() => {
+    UpdateConversation(relay.environment, conversation)
```
damassi, line 118: Since this is a function being called (which matches the mutation field name, versus the type) maybe this should be `updateConversation`?

src/Apps/Conversation/Mutation/UpdateConversationMutation.ts
```diff
@@ -0,0 +1,41 @@
+import { Conversation_conversation } from "__generated__/Conversation_conversation.graphql"
+import { UpdateConversationMutation } from "__generated__/UpdateConversationMutation.graphql"
+import { Environment, commitMutation, graphql } from "relay-runtime"
+
+export const UpdateConversation = (
```
damassi, line 5: `updateConversation`

src/Apps/Conversation/Mutation/UpdateConversationMutation.ts
```diff
@@ -0,0 +1,41 @@
...
+      input: {
+        conversationId: conversation.internalID,
+        fromLastViewedMessageId: conversation.lastMessageID,
+      },
+    },
+    mutation: graphql`
+      mutation UpdateConversationMutation(
+        $input: UpdateConversationMutationInput!
+      ) {
+        updateConversation(input: $input) {
```
damassi, line 20: Matches here, which is what we're mirroring

zephraph:
@lilyfromseattle those snapshot updates don't look like things that should be included in this PR. You might need to rebase/merge master into your branch.

src/Apps/Conversation/Components/Conversation.tsx
zephraph: This pattern matches the other mutation that's used in this app. I'd be 👍 for merging this as is and updating them both in a separate PR or updating them both in this PR. @lilyfromseattle whichever you prefer would likely be fine.

damassi: Sounds good 👍 we should probably update all of them at once so that there's no more copy 🍝 but yeah, better to keep things consistent across app for now.

artsyit:
:rocket: PR was released in [`v26.40.8`](https://github.com/artsy/reaction/releases/tag/v26.40.8) :rocket:

PR #3540, opened 2020-05-14
[COM-17] Supports artworks on feature pages

dzucconi:
Re: [COM-17](https://artsyproduct.atlassian.net/browse/COM-17)

### DIFF
--- a/src/Artsy/Analytics/trackingMiddleware.ts
+++ b/src/Artsy/Analytics/trackingMiddleware.ts
@@ -12,7 +12,7 @@ import { get } from "Utils/get"
  */
 
 interface TrackingMiddlewareOptions {
-  excludePaths?: [string]
+  excludePaths?: string[]
 }
 
 export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
--- a/src/Components/Masonry.tsx
+++ b/src/Components/Masonry.tsx
@@ -1,0 +1,19 @@
+import { Box } from "@artsy/palette"
+import styled from "styled-components"
+import { GridColumnGapProps, gridColumnGap, style } from "styled-system"
+
+const columnCount = style({
+  prop: "columnCount",
+  cssProperty: "columnCount",
+})
+
+export const Masonry = styled(Box)<
+  { columnCount?: number[] | number } & GridColumnGapProps
+>`
+  ${columnCount};
+  ${gridColumnGap};
+
+  > * {
+    break-inside: avoid;
+  }
+`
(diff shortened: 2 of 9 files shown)


src/Artsy/Analytics/trackingMiddleware.ts
```diff
@@ -12,7 +12,7 @@ import { get } from "Utils/get"
  */
 
 interface TrackingMiddlewareOptions {
-  excludePaths?: [string]
+  excludePaths?: string[]
```
dzucconi, line 15: 🙃

src/Components/Masonry.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const Masonry = styled(Box)<
+  { columnCount: number[] | number } & GridColumnGapProps
+>`
+  ${columnCount};
+  ${gridColumnGap};
+
+  > * {
+    break-inside: avoid;
+  }
```
dzucconi, line 18: So I stole this from your ConsignFAQ — I didn't replace it though because of this break-inside addition. Let me know if you want me to use this there as well.

damassi: Yeah! i wanna keep exploring this for artworks, see if there's something we can hammer in.

Would you mind adding a storybooks entry for `Masonry`?

src/Artsy/Analytics/trackingMiddleware.ts
damassi: 🤦

src/Components/Masonry.tsx
dzucconi: Made the sort obvious — was thinking if we really wanted to maintain the left-to-right sorting we could partition the children into columned arrays then re-interleave the array. Probably overkill but not the worst idea. 🤷

damassi: Yeah, we actually _need_ to do that. If you scroll fast down the artwork grid, or click an artwork further down the grid while in the experiment a/b test group you ocan see that our current grid is vertically oriented -- it leads to a flash of the lazy loader, even though those items are within view, because the indexes aren't correctly left to right, and `preload` exists only for the first ten. Its a mess.

artsyit:
:rocket: PR was released in [`v26.40.3`](https://github.com/artsy/reaction/releases/tag/v26.40.3) :rocket:

src/Components/Masonry.tsx
damassi: flash

damassi: (cc @zephraph - as you have the most context around this)

PR #3543, opened 2020-05-14
Make tracking exclusion more robust

damassi:
Data team noticed that we were doing some double firing on page view tracking. This doesn't fix that, but simply makes path exclusion more robust. The same technique has been copied over to the corresponding Force PR, which does fix the problem. Uses [path-to-regex](https://github.com/pillarjs/path-to-regexp) under the hood.

Going to merge this in order to get a deploy out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
     "klaw-sync": "6.0.0",
     "lint-staged": "10.0.8",
     "patch-package": "6.2.0",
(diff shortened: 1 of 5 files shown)


artsyit:
:rocket: PR was released in [`v26.40.5`](https://github.com/artsy/reaction/releases/tag/v26.40.5) :rocket:

PR #3544, opened 2020-05-15
[COM-20] - handle missing data and include pricing info on VR artwork

oxaudo:
I assumed some data will be here since it's enforced in gravity - like VR title, partner. And Artworks can not be created without images and Title defaults to 'Untitled' in api.

Otherwise I tried to guard against blank data and adjust spaces accordingly. Figured out it's too much to write separate stories for it. Unless we can collaborate with design and maybe create one VR with absolute minimal data and see that it displays correctly. Tbd I guess 🤷‍♀️ 

I also noticed that designs display the sale message for each artwork - so added that as well.


damassi:
LTGM 👍 !

artsyit:
:rocket: PR was released in [`v26.40.5`](https://github.com/artsy/reaction/releases/tag/v26.40.5) :rocket:

PR #3545, opened 2020-05-15
Fix carousel

damassi:
Addresses weird carousel bug we see occasionally around images lacking static widths: https://github.com/metafizzy/flickity/issues/691. Added as devDep here since we use it as a dep in force.


artsyit:
:rocket: PR was released in [`v26.40.6`](https://github.com/artsy/reaction/releases/tag/v26.40.6) :rocket:

PR #3550, opened 2020-05-18
Ensure test checks don't validate actual dates

zephraph:
This caused a test failure in master due to the number of months changing from 12 to 13. 

You can see the failure here: https://app.circleci.com/jobs/github/artsy/reaction/61094

As a follow up we should figure out how to properly mock up this date data so that it's consistent between runs. Given that we're still really in flux on this work currently though, I'm just narrowing down the check.


artsyit:
:rocket: PR was released in [`v26.40.7`](https://github.com/artsy/reaction/releases/tag/v26.40.7) :rocket:

PR #3563, opened 2020-05-19
[ViewingRoom] Make sure width is auto on carousel

damassi:
One more -- rather than fixed width, we set to auto and let the height drive the shape.


artsyit:
:rocket: PR was released in [`v26.40.17`](https://github.com/artsy/reaction/releases/tag/v26.40.17) :rocket:

PR #3564, opened 2020-05-19
Fix vertical alignment of CountrySelect ::after element

dblandin:
The position of the ::after element shifts if margin is applied to the `Select` element. Seems strange but using Spacer elements instead fixes the alignment for now.

### Before

Screenshot from 2020-05-19 19-51-04

### After

Screenshot from 2020-05-19 19-50-06


artsyit:
:rocket: PR was released in [`v26.40.18`](https://github.com/artsy/reaction/releases/tag/v26.40.18) :rocket:

PR #3570, opened 2020-05-20
Feature landing page Q/A

dzucconi:
Re: [Figma board](https://www.figma.com/file/3VpNrlYhYXuSvI2xzqfVZT/Communities-Project?node-id=1%3A2600)


src/Apps/Feature/Components/FeatureHeader.tsx
```diff
@@ -1,67 +1,70 @@
...
           {subheadline && (
-            <Sans size="4" textAlign="center">
-              <Box dangerouslySetInnerHTML={{ __html: subheadline }} />
-            </Sans>
+            <HTML
+              fontFamily="sans"
+              size="4"
+              textAlign="center"
+              html={subheadline}
+            />
```
damassi, line 67: Should this be in Palette?

damassi: (Not a blocker of course)

dzucconi: Probably — it sort of seems like this doesn't come up much (?) but it'd probably be nice to have a comprehensive set of styles for all the elements we support in Markdown (h1, h2, etc). My main purpose here was just to neutralize the inner margins.

src/Apps/Feature/FeatureApp.tsx
```diff
@@ -48,6 +41,10 @@ const FeatureApp: React.FC<FeatureAppProps> = ({ feature }) => {
             feature.sets.edges.map(
               ({ node: set }) => set && <FeatureSet key={set.id} set={set} />
             )}
+
+          <Separator my={3} />
+
+          <Footer />
```
damassi, line 47: Its unfortunate that this has to be added in this way right now, but we have a bit of conflict in the A/B test, so it can't go [in the new AppShell](https://github.com/artsy/reaction/blob/master/src/Apps/Components/AppShell.tsx#L60) by default just yet.

src/Components/HTML.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react"
```
damassi, line 1: A good follow-up would be to add this to Palette

src/Components/HTML.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const HTML = ({ fontFamily = "sans", html, size, ...rest }) => {
+  const Typeface = TYPEFACES[fontFamily as Typeface]
+
+  return (
+    <Typeface size={size} {...rest}>
+      <Box dangerouslySetInnerHTML={{ __html: html }} />
+    </Typeface>
+  )
+}
```
damassi, line 41: Otherwise, this is nice 👍

sweir27:
Woo looks good to me too! Thank you!

dzucconi:
Thx all — Extracted `HTML` and bumped version in this PR.

artsyit:
:rocket: PR was released in [`v26.44.0`](https://github.com/artsy/reaction/releases/tag/v26.44.0) :rocket:

PR #3572, opened 2020-05-20
Show prices on default view of artist auction results.

artsyjian:
https://artsyproduct.atlassian.net/browse/PURCHASE-1925

Currently on Artist Auction Results, user must be logged-in to see prices. This change is to relax that restriction but only for the default view which is the view on page 1 when filters are at default. We will show prices in the default view even when no user is logged-in.

Please note, due to filter context being incorrect about the page# of the page being viewed (it always says 1), I cannot track the page# the user is on. I resort to tracking whether pagination has occurred. If pagination has occurred, prices are hidden. So if user clicks page 1 while on page 1, the page hasn't changed, but pagination has occurred, prices will be hidden again.

--
Screenshots showing effects of the change:

Not logged-in. Page 1. Default filters. Price shown.
Screenshot from 2020-05-20 16-49-44

Not logged-in. Page 1. Medium `Painting` checked. Prices hidden.
Screenshot from 2020-05-20 16-50-33

Not logged-in. Page 2. Default filters. Prices hidden.
Screenshot from 2020-05-20 16-50-57

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
@@ -26,9 +26,6 @@ import {
   ImageWithFallback,
   renderFallbackImage,
 } from "./Components/ImageWithFallback"
-import { useAuctionResultsFilterContext } from "./AuctionResultsFilterContext"
-import { isEqual } from "lodash"
-import { auctionResultsFilterResetState } from "./AuctionResultsFilterContext"
 
 export interface Props extends SystemContextProps {
   expanded?: boolean
@@ -36,7 +33,7 @@ export interface Props extends SystemContextProps {
   index: number
   mediator?: Mediator
   lastChild: boolean
-  filtersHaveUpdated: boolean
+  filtersAtDefault: boolean
   paginated: boolean
 }
 
@@ -75,13 +72,6 @@ export const ArtistAuctionResultItem: SFC<Props> = props => {
     })
   }
 
-  // Is current filter state different from the default (reset) state?
-  const filterContext = useAuctionResultsFilterContext()
-  const filtersHaveUpdated = !isEqual(
-    filterContext.filters,
-    auctionResultsFilterResetState
-  )
-
   return (
     <>
       <Media at="xs">
@@ -92,7 +82,6 @@ export const ArtistAuctionResultItem: SFC<Props> = props => {
               expanded={expanded}
               mediator={mediator}
               user={user}
-              filtersHaveUpdated={filtersHaveUpdated}
             />
           </Row>
           <Box>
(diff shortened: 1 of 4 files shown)


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -50,6 +50,9 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
     allowEmptyCreatedDates,
   } = filterContext.filters
 
+  // Count number of times user paginated.
```
jpotts244, line 53: Nitpick: Is comment relevant with your latest commit to use booleans for pagination detection?

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -298,9 +332,22 @@ const getProps = (props: Props) => {
...
+  user,
+  mediator,
+  size,
+  filtersHaveUpdated,
+  paginated
+) => {
   const textSize = size === "xs" ? "2" : "3t"
-  if (user) {
+
+  // If user is logged in we show prices. Otherwise we show prices only for the default view - on page 1 and filters not changed.
```
jpotts244, line 346: Thanks for adding a note on the functionality intention here. There is a lot of nuance around this behavior

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -298,9 +332,22 @@ const getProps = (props: Props) => {
...
+  // Ideally we get current page number from filter context 'page' property but somehow it is always '1'.
+  // So we resort to pagination detection. If user has paginated at all, prices will be hidden. even if user comes back to page 1.
+  // TODO: Fix filter context so its 'page' property has the current page number, then change this code.
```
jpotts244, line 349: Interesting discovery around the filter context page bug! Nice find!

I'm not super familiar with filter contexts but I do think we have another one working on the Artwork page. 
https://github.com/artsy/reaction/blob/master/src/Components/v2/ArtworkFilter/ArtworkFilterContext.tsx#L26

This one also handles some logic around tracking the `page` property. Would be curious to see how that one handles this and if we can use that to help us debug the page number bug you found.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
artsyjian: Oh...I can see myself getting confused by that comment in future days. Corrected. Thanks!

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
artsyjian: Oh. I will check that out. Thanks again!

zephraph:
Amazing work @artsyjian. Thanks for the detailed comments and test updates as well. 👏

artsyit:
:rocket: PR was released in [`v26.49.0`](https://github.com/artsy/reaction/releases/tag/v26.49.0) :rocket:

PR #3576, opened 2020-05-20
email confirmation message banner

erikdstock:
Related 🔒 [Jira Ticket](https://artsyproduct.atlassian.net/browse/AUCT-1001)

Low-risk code, so I will merge with one approving review.
This PR adds a simple confirmation message banner for use based on a query parameter in the url on force. Messages are currently hardcoded based on the incoming param - no reliance on the user being present.

Paired with @bhoggard with some input from @yuki24 on this.

Example linked into force (https://github.com/artsy/force/pull/5620):
image


dleve123:
@erikdstock in the future, let's always request review from at least some members of product-auctions so that there's shared visibility on our team's work. I'll review now; also requesting review from @yuki24

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+const textMessage: (msg: string) => React.FC = (message: string) => () => (
+  <Sans color="white100" size="3" lineHeight={1} weight="medium">
+    {message}
+  </Sans>
+)
+
+const messages: { [k: string]: React.FC } = {
+  confirmed: textMessage("Your email has been confirmed."),
```
dleve123, line 15: I presume these message keys need to correspond with the [URL parameters generated by Gravity](https://github.com/artsy/gravity/blob/master/app/controllers/confirmations_controller.rb#L3-L7) - is that correct? Otherwise, we'll never surface the more nuanced error cases that _should_ be presented to end-users: for example, a user already having confirmed their email.

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,29 @@
...
+  // TODO: I don't think we should show this as a fallback as any string could trigger it
+  fallback: textMessage("An error has occurred. Please try again."),
```
dleve123, line 17: agreed with this TODO – I don't think we should fallback like this. Can we remove and enumerate through the states that Gravity expresses via URL parameters (described above).

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react"
+import { Banner, Sans } from "@artsy/palette"
+
+interface Props {
+  messageCode: string
```
dleve123, line 5: Could it make sense for this to be a TS enum? We don't need to handle arbitrary strings here yet (only using for email confirmation atm), so further specifying the types feels a bit safer.

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,29 @@
...
+  fallback: textMessage("An error has occurred. Please try again."),
+}
+
+export const FlashBanner: React.FunctionComponent<Props> = ({
+  messageCode,
+}) => {
+  const Message = messages[messageCode] || messages.fallback
+  return (
+    <Banner dismissable backgroundColor="black100">
+      <Message />
```
yuki24, line 26: This looks tricker than it should be. I think it's more straight forward to inline the `<Sans color="white100" size="3" lineHeight={1} weight="medium">`:

```tsx
const MESSAGES = {
   confirmed: "Your email has been confirmed.",
  // other keys...
 }
```

```tsx
<Banner dismissable backgroundColor="black100">
  <Sans color="white100" size="3" lineHeight={1} weight="medium">
    {MESSAGES[messageCode] || "An error has occurred. Please try again."}
  </Sans>
</Banner>
```

src/Components/FlashBanner.tsx
erikdstock: should we log or send a message to sentry for unhandled messages? i think not.

src/Components/FlashBanner.tsx
erikdstock: So there are two reasons I did this, though maybe there is a better way.
1. the text size on the default palette banner does not match the mockups, so i'm adding a second `Sans` wrapper to lower it. maybe not ideal.
2. The designs for the 'email confirmation has been sent. `[send again]`' have a button, which suggests that the children might be more dynamic than just text (though I guess we could put a button inside the tag above).

Also as an aside I think we are going to drop the fallback message for unhandled flashMessage keys.

src/Components/FlashBanner.tsx
erikdstock: :+1:

src/Components/MinimalCtaBanner.tsx
```diff
@@ -58,7 +58,12 @@ export class MinimalCtaBanner extends React.Component<
               </Sans>
             </a>
             <IconContainer onClick={this.dismissCta.bind(this) as any}>
-              <Icon name="close" color={this.props.textColor} fontSize="16px" />
+              <Icon
+                name="close"
+                color={this.props.textColor}
+                fontSize="16px"
+                style={{ cursor: "pointer" }}
```
erikdstock, line 65: This and the other change to the CtaBanner story were minor bugfixes left over from early explorations. It's possible we don't even use this and/or should remove the changes from the PR but for now I'm thinking we want to prioritize upcoming reaction migration work so am not addressing unless you think it is neccessary @damassi

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+  | "expired_token"
+
+type BannerMessage = string | React.FC
+
+const messages: Record<FlashMessageKey, BannerMessage> = {
+  confirmed: "Your email has been confirmed.",
+  already_confirmed: "You have already confirmed your email.",
+  invalid_token: "Your token is invalid. Please try again.",
+  blank_token: "No token found. Please try again.",
+  expired_token: "An error has occurred. Please try again.",
```
bhoggard, line 27: Shouldn't this be something about how their confirmation token expired, not the general message? Or are we planning to come back to this once we allow people to request the email?

erikdstock: I copied these from the [eigen pr](https://github.com/artsy/eigen/pull/3325/files#diff-2996077a298cdc9e41e9bf32fc2431d1R66)- happy to change but also wonder if these are meant to be final - in most cases for example 'try again' is not going to be actionable unless there is a 'resend email' button.

dleve123: Agreed – would much rather tell people to "Contact support@artsymail.com"

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+  /* email already confirmed */
+  | "already_confirmed"
+  /* email confirmation token was invalid */
+  | "invalid_token"
+  /* email confirmation token was empty */
+  | "blank_token"
+  /* email confirmation token was expired */
+  | "expired_token"
+
+type BannerMessage = string | React.FC
```
dleve123, line 20: why the union type here? Seems like these are all strings?

src/Components/FlashBanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const FlashBanner: React.FunctionComponent<Props> = ({
+  messageCode,
+}) => {
+  const Message = messages[messageCode]
+  if (!Message) return null
+  return (
+    Message && (
+      <Banner dismissable backgroundColor="black100">
+        <Sans color="white100" size="3" lineHeight={1} weight="medium">
+          {typeof Message === "string" ? Message : <Message />}
```
dleve123, line 39: very well could be missing something, but these seem to always be strings...

dleve123:
approving to unblock merge, but still not sure what's up with the type complexity.

artsyit:
:rocket: PR was released in [`v26.49.1`](https://github.com/artsy/reaction/releases/tag/v26.49.1) :rocket:

PR #3579, opened 2020-05-21
[Viewing Room] QA

damassi:
- [x] (Header) @desktop, reduce header height to 95% viewport (so people know to scroll down)

(Artworks preview on Statement tab)
- [x]  Alignment seems to be a bit off between text, tiles, and button
- [x]  Spacing also looks off in the artwork tile type. 
- [ ]  ~Artworks should preserve their own image ratio.~ we dont adjust these images
- [ ]  [NICE TO HAVE] If possible, if a user taps a work, it should anchor the user to that artwork, in Works tab.

(Statement tab)
- [x]  @xs the spacing above statement text is a bit too tall.

(Carousel)
- [x] Curious how the progress bar width is being determined, would it be possible for the bar to match the width of the text across screen sizes?
- [x] Type needs to be tightened up

(Statement Tab) 
- [x]  Missing “View Works” button at bottom

Footer
- [x] Add responsive footer

### DIFF
--- a/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
+++ b/src/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
@@ -56,7 +56,7 @@ const ViewingRoomHeaderLarge: React.FC<ViewingRoomHeaderProps> = props => {
   return (
     <Flex
       style={{
-        height: `calc(100vh - ${NavBarHeight}px)`,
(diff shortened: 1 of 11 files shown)


artsyit:
:rocket: PR was released in [`v26.45.1`](https://github.com/artsy/reaction/releases/tag/v26.45.1) :rocket:

PR #3591, opened 2020-05-22
Remove tracking calls for now

sweir27:
We're hitting our segment quota! This PR:
- Removes publishing events that fire often and are not used.
- Comments out our "Experiment Viewed" event now that we're confident in the result of the test. I didn't remove that code entirely since I figured it would come as part of the A/B test closeout (at which point I think I bunch of code can be cleaned up).


damassi:
(Just realized since this is publishing code no worries about copying to force -- i'll move the experiment event called tho.)

artsyit:
:rocket: PR was released in [`v26.53.0`](https://github.com/artsy/reaction/releases/tag/v26.53.0) :rocket:

PR #3621, opened 2020-08-14
[HOLD] Upgrade relay

damassi:
#### ⚠️ HOLD

> We need to coordinate npm publishing / release steps with the main upgrade PR here: https://github.com/artsy/force/pull/5934

While finishing up https://github.com/artsy/force/pull/5934 I noticed that force wouldn't boot and after some digging found that it led to the article routes causing an error around the artist tooltips. Then I remembered that we have just a single Relay component in our publishing code...

Going forward, until we completely retire reaction, we'll have to perform relay upgrades in parallel.


zephraph:
Thanks for helping push this forward @damassi! 🙏

artsyit:
:rocket: PR was released in [`v28.0.0`](https://github.com/artsy/reaction/releases/tag/v28.0.0) :rocket:

PR #3622, opened 2020-08-14
Upgrades various dependencies

dzucconi:
Re: [FX-2184](https://artsyproduct.atlassian.net/browse/FX-2184) — I need to update some typography for captions, etc. But we have to get thru a few major version upgrades. I'm not sure if this has any implications for the other applications still utilizing this library?


src/__stories__/config.js
```diff
@@ -22,6 +24,8 @@ addParameters({
   sortStoriesByKind: true,
 })
 
+addDecorator(storyFn => <Theme>{storyFn()}</Theme>)
```
dzucconi, line 27: None of the stories had the theme.

src/Components/Publishing/Byline/__tests__/__snapshots__/ShareDate.test.tsx.snap
```diff
@@ -50,7 +50,7 @@ exports[`renders ShareDate properly 1`] = `
   color: black;
 }
 
-@media (max-width:768px) {
+@media (max-width:768pxpx) {
```
eessex, line 53: Looks like a typo?

dzucconi: Good catch — probably some theme that didn't get upgraded back when. Will investigate.

src/Assets/Theme.ts
```diff
@@ -13,11 +13,11 @@ export default {
...
-      xs: breakpoints.xs, // px
-      sm: breakpoints.sm,
-      md: breakpoints.md,
-      lg: breakpoints.lg,
-      xl: breakpoints.xl,
+      xs: parseInt(breakpoints.xs, 10), // px
+      sm: parseInt(breakpoints.sm, 10),
+      md: parseInt(breakpoints.md, 10),
+      lg: parseInt(breakpoints.lg, 10),
+      xl: parseInt(breakpoints.xl, 10),
```
dzucconi, line 20: @eessex ☝️

eessex: Would love to see these publishing breakpoints retired at some point so that we can use palette's theme instead!

artsyit:
:rocket: PR was released in [`v28.1.0`](https://github.com/artsy/reaction/releases/tag/v28.1.0) :rocket:

PR #3628, opened 2020-08-27
Deprecate editorial email signup

eessex:
This form was removed from use a few years back because it was not GDPR compliant. It has never been used since then, removing.


artsyit:
:rocket: PR was released in [`v28.4.0`](https://github.com/artsy/reaction/releases/tag/v28.4.0) :rocket:

PR #3632, opened 2020-09-02
Strip /works-for-sale from urls when attempting to fetch tooltop content

eessex:
Article tooltips fetch content from artsy.net URLs based on simple text parsing. Since we added a redirect on artist pages for to add `/works-for-sale` to the url, the parsing was not finding the slug where it expected it (at the end of the url). This caused fetches to use "works-for-sale" rather than the actual slug, breaking both the artist links & tooltips where the string is present. 

The editorial team has been manually removing these links from articles, that will no longer be necessary with this change.


artsyit:
:rocket: PR was released in [`v28.5.0`](https://github.com/artsy/reaction/releases/tag/v28.5.0) :rocket:

PR #3637, opened 2020-09-11
Show artwork sale message on editorial articles

williardx:
## Problem

We currently show artworks on Editorial that are for sale on Artsy. However, users do not see any pricing information which makes their for-sale status unclear.

## Solution

Show an artwork's `saleMessage` in an artwork's image caption on Editorial pages when we show an artwork. 

**Note**: The logic for `bidInfo` and `saleMessageDisplay` was lifted from Force: https://github.com/artsy/force/blob/b607b3b2bd9459981203c879ae3b3a3b04e9e587/src/v2/Components/Artwork/Details.tsx#L156. That code is thoroughly tested in Force and so tests are lighter here in comparison.

Screen Shot 2020-09-11 at 10 22 23 AM

Screen Shot 2020-09-11 at 10 22 38 AM

### DIFF
--- a/src/Components/Publishing/EditorialFeature/Components/Vanguard2019/__tests__/__snapshots__/index.test.tsx.snap
+++ b/src/Components/Publishing/EditorialFeature/Components/Vanguard2019/__tests__/__snapshots__/index.test.tsx.snap
@@ -1155,6 +1155,7 @@ exports[`Vanguard2019 Snapshots renders article series snapshot properly 1`] = `
                             }
                           />
                         </div>
+                        <div />
                       </div>
                     </div>
                   </div>
@@ -1593,6 +1594,7 @@ exports[`Vanguard2019 Snapshots renders article series snapshot properly 1`] = `
                             }
                           />
                         </div>
+                        <div />
                       </div>
                     </div>
                   </div>
@@ -2031,6 +2033,7 @@ exports[`Vanguard2019 Snapshots renders article series snapshot properly 1`] = `
                             }
                           />
                         </div>
+                        <div />
                       </div>
                     </div>
                   </div>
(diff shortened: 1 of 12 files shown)


dzucconi:
Thanks for the thorough solution here. If I'm reading correctly this will fire off individual queries for every caption within an article. Can we just lift the fragment into the parent query for these pages?

dzucconi:
(I can imagine some pathological examples here; interestingly [query batching](https://github.com/artsy/metaphysics/pull/2661) would make this less of an issue)

williardx:
@dzucconi The parent query for this is a query to Positron which doesn't store sale message data. We need the artwork slugs (which come from Positron) and once we have them can then query Gravity via Metaphysics for additional information.

dzucconi:
@williardx OK, interesting... I'm gonna get the query batching PR merged anyway which would identify and merge these distinct queries into a single network request. But I'd be marginally concerned about this before that is turned on as I don't know if there are any articles floating around with a large number of artworks which could potentially cause some issues.

cc @mzikherman take a peek at this?

I suppose we can add this to the stack of evidence that these pages need some attention. 🤔

williardx:
Yeah that makes sense -- I'm down to wait until https://github.com/artsy/metaphysics/pull/2661 is merged in.

sweir27:
@williardx we're already pulling in _some_ artwork data to generate the rest of this view-- where is that being queried for?

williardx:
As far as I can tell that data comes from Positron: We make a query using `ArticleQuery` https://github.com/artsy/force/blob/9fd6c485a217e90d40cbb0d58110e5e71bb9862f/src/desktop/apps/article/routes.tsx#L61 which can include this artwork fragment for `ImageCollection` and `ImageSet`: https://github.com/artsy/force/blob/9fd6c485a217e90d40cbb0d58110e5e71bb9862f/src/desktop/apps/article/queries/sectionFragments.js#L19.

artsyit:
:rocket: PR was released in [`v28.6.0`](https://github.com/artsy/reaction/releases/tag/v28.6.0) :rocket:

PR #3641, opened 2020-09-18
Do not display sale message when sale message is null

williardx:
https://artsyproduct.atlassian.net/browse/FX-2283

## Problem

We're currently displaying `null` when the sale message is `null`.

## Solution

Render an empty string instead of `null` when the sale message is `null`.

**Before**

Screen Shot 2020-09-18 at 12 10 46 PM

**After**

Screen Shot 2020-09-18 at 2 52 47 PM

### DIFF
--- a/src/Components/Publishing/Sections/ArtworkSaleMessage.tsx
+++ b/src/Components/Publishing/Sections/ArtworkSaleMessage.tsx
@@ -36,8 +36,12 @@ export const ArtworkSaleMessage: React.FC<ArtworkSaleMessageQueryResponse> = pro
   }
 
   const saleMessageDisplay = artwork => {
-    const { sale } = artwork
+    const { sale, saleMessage } = artwork
     const isAuction = sale && sale.isAuction
+
+    if (saleMessage === null) {
+      return ""
+    }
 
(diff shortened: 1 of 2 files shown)


src/Components/Publishing/Sections/ArtworkSaleMessage.tsx
```diff
@@ -36,9 +36,13 @@ export const ArtworkSaleMessage: React.FC<ArtworkSaleMessageQueryResponse> = pro
   }
 
   const saleMessageDisplay = artwork => {
-    const { sale } = artwork
+    const { sale, saleMessage } = artwork
     const isAuction = sale && sale.isAuction
 
+    if (saleMessage === null) {
```
sweir27, line 42: #non-blocking Slightly less verbose to do `if (!saleMessage) {` unless there's a reason to explicitly check for `null` here.

sweir27 (approved):
Seems fine to me! Merge when ready.

artsyit:
:rocket: PR was released in [`v28.7.0`](https://github.com/artsy/reaction/releases/tag/v28.7.0) :rocket:

PR #3658, opened 2020-11-02
chore: move cohesion out of dev deps

eessex:
Moves Cohesion out of dev dependencies and into regular dependencies. 

Cohesion is a real dependency for reaction, so the dev placement is inaccurate. I believe this was added when Reaction was more actively worked on to cut down on CI builds during office hours.

This change will allow us to remove Cohesion from Positron's dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
   },
   "devDependencies": {
     "@artsy/auto-config": "1.0.2",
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`v28.12.0`](https://github.com/artsy/reaction/releases/tag/v28.12.0) :rocket:

PR #3664, opened 2020-12-01
fix: Update URL parsing to more precisely extract slug

williardx:
https://artsyproduct.atlassian.net/browse/GRO-64

## Problem

`getArtsySlugsFromHTML` incorrectly parses some links in Editorial articles resulting in (1) users being unable to click on any artist link in the article and (2) tooltips for artist links not showing up. 

For example, links like `/artist-series/slug` were incorrectly interpreted as an artist slug and and links like `/artist/slug/` were incorrectly interpreted to not be an artist link.

## Solution

Update `getArtsySlugsFromHTML` to use a regex that correctly parses a greater variety of inputs. Links including trailing slashes, artist app subroutes (`/works-for-you`, `/auction-results`) and query params will now be correctly parsed.

~**DoNotMerge** until I've validated (1) I've covered all the known cases of broken links and (2) I've confirmed that showing tooltip links for links to artist app subroutes is what we want.~

### DIFF
--- a/src/Components/Publishing/Constants.ts
+++ b/src/Components/Publishing/Constants.ts
@@ -1,6 +1,5 @@
-import { compact, last, uniq } from "lodash"
+import { compact, uniq } from "lodash"
 import { DateTime } from "luxon"
-import url from "url"
 import { ArticleData, DateFormat } from "../Publishing/Typings"
 
 const APP_URL = process.env.APP_URL || "https://www.artsy.net"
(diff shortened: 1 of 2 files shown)


damassi (approved):
As long as the tests sufficiently cover the problem cases and other probable scenarios  LGTM 👍

artsyit:
:rocket: PR was released in [`v28.14.0`](https://github.com/artsy/reaction/releases/tag/v28.14.0) :rocket:

PR #3665, opened 2020-12-08
Remove unused formik dependency and legacy auth code

eessex:
Formik is no longer used in Reaction, this change removes dependency and some related legacy auth code. 

I also noticed that Force is using `formik` via Reaction's dependencies, this PR will break Force, so I will add a follow up PR there to add the lib to dependencies.


icirellik (approved):
💯 Formik has been on my radar for a while. Every now and then it causes me issues when trying to convert unit tests. There is a possibility we will see some bundle size decreases from his as well.

I'll leave this for you to merge so that you can make a follow-up PR when ready.

artsyit:
:rocket: PR was released in [`v28.15.0`](https://github.com/artsy/reaction/releases/tag/v28.15.0) :rocket:

PR #3697, opened 2021-08-12
bugfix: Fix overflow text

damassi:
Fixes https://artsyproduct.atlassian.net/browse/GRO-370

Fixes an issue where the copy text would overlap the header image on mobile. Now we just run overflow:hidden over it and be done with it. Works well enough. 

⚠️  Note: was not able to get article pages running in force without commenting out [client-side mount code here](https://github.com/artsy/force/blob/1d3a83449f2224b3c190a082f83b1b28e7a3b0e2/src/desktop/apps/article/client/article.tsx#L9-L13). For whatever reason after the SSR pass the screen blanks out on client-side mount. I couldn't figure out how to fix. 

Before:
 
Screen Shot 2021-08-12 at 2 32 00 PM

After: 

overflow


artsyit:
:rocket: PR was released in [`v29.2.1`](https://github.com/artsy/reaction/releases/tag/v29.2.1) :rocket:

PR #3710, opened 2021-10-07
fix(deps): updates @artsy deps

dzucconi:
This updates these deps to the most recent versions so that they aren't (hopefully) duplicated in the Force bundle.


dzucconi:
Oh, I'm only suggesting because this is dead code.

PR #3811, opened 2026-04-10
chore: remove Peril mention, correct label added

amonkhouse:
* Update Peril mention -> Github mention
* Correct which label is automatically added (I think the doc has been wrong for a while as we didn't change functionality there)


mc-jones (approved):

artsyit:
:rocket: PR was released in [`v29.13.0`](https://github.com/artsy/reaction/releases/tag/v29.13.0) :rocket:

